PR #2346, opened 2022-06-05
Make fallback FlatModifier slug use item slug instead of label

stwlam:


CarlosFdez:
I guess double sluggify should be harmless...

Issue #2103, opened 2022-02-22
Request: Output chat card when "Use (consumable name)" button is selected on the inventory page of character sheet.


cswendrowski:
In GitLab by @joshmvii on Feb 22, 2022, 11:26

Right now, only a message that's been used (and dice roll for healing potion, etc.) is outputted to chat when this button is selected, but it consumes the item. So in the case of items that have effects on them that need dragged, the effects have to be dragged before you use the item or else you'll lose the item (if it's your only one) and have to go find the effect. User would expect to use the item and then get the effect so it'd be a nice QoL change to get the card when it's used.

ThingsCouldGetDicey:
upvote!

But what i also suggest ontop of this is that a token should be selected before use consumable is selected and it auto applies the RE effects onto the token selected after the button is used.

Issue #1219, opened 2021-03-23
"use potion" on mutagens does not output information to chat


cswendrowski:
In GitLab by @hifillymayshere on Mar 23, 2021, 16:02

![Image1](http://gitlab.com/hooking/foundry-vtt---pathfinder-2e/uploads/02fe189dd84a3162163b86b4b3c80418/Image1.png)
When the "use potion" button is clicked in inventory, all that pops up in chat is a notification that it has been used. No useful information and the potion is immediately deleted from inventory. This causes me to have to look up the potion again to determine what it actually did. 

I suggest the "use potion" button have the same functionality as clicking the potion's icon in inventory. Output a useful description of the potion into the chat along with the drag and drop effects.

![Image2](http://gitlab.com/hooking/foundry-vtt---pathfinder-2e/uploads/5acdd40579a867d475e7c91666aeb00a/Image2.png)

PR #2361, opened 2022-06-06
Add clickable compendium link to usage chat message for used consumables

fraserstanton:
If you use the last consumable of a stack, it disappears from your inventory and you no longer have easy access to the compendium description for any draggable effects, etc.
This change adds a clickable link to the compendium item into the chat when you use an item.

Addresses issue #2103 

image


Shandyan:
Also addresses issue #1219

CarlosFdez:
I feel like we're more likely to redo consumable logic than continue to allow this workflow. Consumables are in need of a brushup and they're on my radar.

fraserstanton:
Up to you if you are going to address something soon, but it seems that it is a relatively minor change, and easy to replace/refactor in the future. Something like this will be nice to have in the future implementation anyway.

PR #2362, opened 2022-06-06
Add Adjust Modifier REs to Mighty Bulwark

stwlam:
This will make the display of the (Mighty) Bulwark modifier less confusing (before it would show both Bulwark and Dexterity until rolled)


SpartanCPA (approved):
Data changes seem fine. Did not review changes to .ts files

CarlosFdez (approved):

PR #2365, opened 2022-06-06
Add support for ItemPF2e#isOfType("physical")

stwlam:
Also moves `PHYSICAL_ITEM_TYPES` and `PhysicalItemType` to @item/physical/values and @item/physical/types, respectively


CarlosFdez:
YES

PR #2372, opened 2022-06-07
Add logic to grey Inventory, Effect and Notes tabs for NPC if relative tab is empty

fraserstanton:
Similar logic to existing Spells hiding logic.
Inventory tab title will hide if the npc has no inventory items
Effect tab title will hide if the npc has no conditions or effects
Notes tab title will hide if the npc has neither public or private notes

image

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -123,6 +123,7 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
             totalWealth,
             totalWealthGold,
             inventory: this.prepareInventory(),
+            hasInventory: this.actor.inventory.size === 0 ? false : true,
         };
 
         this.prepareItems(sheetData);
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -123,7 +123,6 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
             totalWealth,
             totalWealthGold,
             inventory: this.prepareInventory(),
-            hasInventory: this.actor.inventory.size === 0 ? false : true,
         };
 
         this.prepareItems(sheetData);
--- a/static/templates/actors/npc/sheet.html
+++ b/static/templates/actors/npc/sheet.html
@@ -6,7 +6,7 @@
         {{> "systems/pf2e/templates/actors/npc/partials/header.html"}}
         <nav class="sheet-tabs tabs" data-group="primary">
             <a class="item" data-tab="main">{{localize "PF2E.NPC.MainTab"}}</a>
-            <a class="item {{#if document.inventory.size{{/if}}" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
+            <a class="item {{#if (not document.inventory.size)}}disabled{{/if}}" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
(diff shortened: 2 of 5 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -123,6 +123,7 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
             totalWealth,
             totalWealthGold,
             inventory: this.prepareInventory(),
+            hasInventory: this.actor.inventory.size === 0 ? false : true,
```
CarlosFdez, line 126:
```suggestion
            hasInventory: !!this.actor.inventory.size,
```

CarlosFdez: TBH you don't even need this, in the sheet if you do `{{#if document.inventory.size}}` that's the same result.

fraserstanton: Updated to use suggested simplification. Thanks.

static/templates/actors/npc/sheet.html
```diff
@@ -6,9 +6,9 @@
         {{> "systems/pf2e/templates/actors/npc/partials/header.html"}}
         <nav class="sheet-tabs tabs" data-group="primary">
             <a class="item" data-tab="main">{{localize "PF2E.NPC.MainTab"}}</a>
-            <a class="item" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
+            <a class="item {{#if document.inventory.size{{/if}}" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
```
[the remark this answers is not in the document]
fraserstanton: Rushed the fix. Forgot to refresh before I tested and it still worked on the old code. Updated now to the correct logic and this time I remembered to refresh before testing :)
Thanks again for the reviews!

static/templates/actors/npc/sheet.html
```diff
@@ -6,10 +6,10 @@
         {{> "systems/pf2e/templates/actors/npc/partials/header.html"}}
         <nav class="sheet-tabs tabs" data-group="primary">
             <a class="item" data-tab="main">{{localize "PF2E.NPC.MainTab"}}</a>
-            <a class="item" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
+            <a class="item {{#if (not document.inventory.size)}}disabled{{/if}}" data-tab="inventory">{{localize "PF2E.NPC.InventoryTab"}}</a>
             <a class="item {{#if (not isSpellcaster)}}disabled{{/if}}" data-tab="spells">{{localize "PF2E.NPC.SpellsTab"}}</a>
-            <a class="item" data-tab="effects">{{localize "PF2E.NPC.EffectsTab"}}</a>
-            <a class="item" data-tab="notes">{{localize "PF2E.NPC.NotesTab"}}</a>
+            <a class="item {{#if (not hasConditionOrEffect)}}disabled{{/if}}" data-tab="effects">{{localize "PF2E.NPC.EffectsTab"}}</a>
+            <a class="item {{#with (not data.details.publicNotes data.details.privateNotes)}}disabled{{/with}}" data-tab="notes">{{localize "PF2E.NPC.NotesTab"}}</a>
```
CarlosFdez, line 12: use unless and or instead. `with` is superfluous to if/unless if all you need is evaluation. not only applies to a single argument, ignoring the second one.

fraserstanton: Updated with to unless and replaced not with or to get required logic to check if either field has data. Thanks again.

PR #2393, opened 2022-06-08
Add support for damage dice downgrades

stwlam:
Also cleanup unused code and useless test


avagdumortesin:
This does not appear to be working properly. Adding the following DamageDice RE to a Falchion makes the Falchion roll 1d12 instead of 1d8:
```
{
  "key": "DamageDice",
  "selector": "damage",
  "override": {
    "downgrade": true
  }
}
```

src/module/system/damage/weapon.ts
```diff
@@ -567,8 +567,14 @@ export class WeaponDamagePF2e {
 
         // First, increase the damage die. This can only be done once, so we
```
CarlosFdez, line 568:
```suggestion
        // First, increase or decrease the damage die. This can only be done once, so we
```

src/module/system/damage/damage.ts
```diff
@@ -1,29 +1,21 @@
...
-    }
+const dieFacesTuple = ["d4", "d6", "d8", "d10", "d12"] as const;
+const DAMAGE_DIE_FACES = new Set(dieFacesTuple);
+type DamageDieSize = SetElement<typeof DAMAGE_DIE_FACES>;
+
+function nextDamageDieSize(next: { upgrade: DamageDieSize }): DamageDieSize;
+function nextDamageDieSize(next: { downgrade: DamageDieSize }): DamageDieSize;
+function nextDamageDieSize(next: { upgrade: DamageDieSize } | { downgrade: DamageDieSize }): DamageDieSize {
+    const [faces, direction] = "upgrade" in next ? [next.upgrade, 1] : [next.downgrade, -1];
+    return dieFacesTuple[dieFacesTuple.indexOf(faces) + direction] ?? faces;
```
fryguy1013, line 14: this is really more clever than it needs to be imo. `faces` should probably be called `dieSize`

PR #2381, opened 2022-06-08
Traits updates for da

avagdumortesin:
Adding traits for upcoming book into the main branch.

PR #2401, opened 2022-06-10
Fix typo in trait slug/name and add new traits

avagdumortesin:
Previously submitted as https://github.com/foundryvtt/pf2e/pull/2381, but the branch was deleted when I deleted my fork when trying to troubleshoot an issue with getting the changes in https://github.com/foundryvtt/pf2e/pull/2393 to work.


TMunsie:
This cannot be merged for the time being as we do not have licensing information in the system for the book this comes from.

Issue #2317, opened 2022-05-24
[Feature Request] New Attribute hp.recoveryMod for increasing HP gained from resting by more than just multiplying the rate.


cswendrowski:
In GitLab by @squirrelkiller on May 24, 2022, 24:01

When looking into ways to add a value to the recovery rate from a night's rest, I realized that there was only a multiplier, and no way to add, for example, +4 HP per level from a night's rest. While currently there is nothing in the officially released content that needs this, it isn't uncommon in homebrew and could be useful to have for the future.

PR #2407, opened 2022-06-10
Add additive modifier to the recovery rate from a night's rest 

kolontaev:
Add additive bonus to the recovery rate from a night's rest and add it to the Hillock Halfling heritage.
Tried to resolve https://github.com/foundryvtt/pf2e/issues/2317


stwlam:
Note that this isn't a bonus--it doesn't follow any rules of bonuses and penalties like type stacking rules.

kolontaev:
Renamed recoveryBonus to recoveryAddend. New name is consistent with recoveryMultiplier and better convey its function.

PR #2431, opened 2022-06-12
Update brimstone rat and bugbear marauder

avagdumortesin:
NPC updates for Troubles in Otari Bestiary


SpartanCPA:
What exactly is updated here? All that is changing are the IDs of items on the actors, which means no actual data was wrong...

avagdumortesin:
Per Tikael's instructions in https://discord.com/channels/880968862240239708/880969253765935174/985311728529178764, I re-added the items from the compendia and ensured that the data matches the stat blocks in the adventure.

SpartanCPA (approved):

PR #2434, opened 2022-06-12
Prevent buttons from retaining focus when clicked

stwlam:
Currently, whenever a button is clicked, it retains focus--even when the button is one that closes a Foundry application. When the document body doesn't have focus, canvas hotkeys stop functioning until the canvas is clicked again. Removing button focus after any click successfully preserves hotkey functionality with (hopefully) no significant side effects.

Upstream issue: https://github.com/foundryvtt/foundryvtt/issues/7270


CarlosFdez:
lets see what happens!

PR #2436, opened 2022-06-12
Audit Ralldar (Age of Ashes)

SpartanCPA:
Replaced Ralldar with the version of the Greater Barghest from bestiary 1 and set the correct mutation


SpartanCPA:
It might even be better to remove the NPC entirely, as it's just the normal Greater Barghest with the mutation pre-picked by the adventure

Issue #2291, opened 2022-05-15
Marshal Dedication: Has no effect


cswendrowski:
In GitLab by @Peenicks on May 15, 2022, 13:52

**Problem:**
In addition, you’re surrounded by a marshal’s aura in a 10-foot emanation. Your aura has the emotion, mental, and visual traits and grants you and allies within the aura a +1 status bonus to saving throws against fear.

**Expected Behavior:**
It should have it's own custom effect

![image](http://gitlab.com/hooking/foundry-vtt---pathfinder-2e/uploads/efae6a3890e52b3971fa2f49b0002559/image.png)

PR #2442, opened 2022-06-12
Add effect for Marshal Dedication

kolontaev:
Fix https://github.com/foundryvtt/pf2e/issues/2291

### DIFF
--- a/packs/data/feat-effects.db/effect-marshal-dedication.json
+++ b/packs/data/feat-effects.db/effect-marshal-dedication.json
@@ -1,0 +1,51 @@
+{
+    "_id": "tDKxQ0JaNvzvKM5E",
+    "data": {
+        "badge": null,
+        "description": {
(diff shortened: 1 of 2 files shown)


stwlam:
This isn't correct, exactly. The initial effect belongs to an aura, which doesn't yet have system support. Additionally, the same aura has other benefits given other feats, making "Effect: Martial Dedication" a very confusing name.

kolontaev:
Do you suggest to rename effect or do you mean that it must be implemented other way?

According to Archives of Nethys only two feats give additional benefits to marchal's aura: Dread Marshal Stance and Inspiring Marshal Stance. Both have their own effects in compendium.

So we can rename the effect to `Effect: Marchal's Aura`. Players could apply it and additionally apply and change stances' effects.

stwlam:
Superseded by #2869

PR #2455, opened 2022-06-12
Automate human feats

Shandyan:


packs/data/feat-effects.db/effect-saoc-astrology.json
```diff
@@ -0,0 +1,141 @@
...
+            "value": 0
+        },
+        "level": {
+            "value": 1
+        },
+        "rules": [
+            {
+                "choices": [
+                    {
+                        "label": "-1 penalty",
```
TikaelSol, line 21: I think this is way overcomplicated. You can just have the -1,1,2,3,4 with 1, 3,  and 4 predicated for the heir feat. No need for labels here, just value with the number and reference that number with `{item|flags.pf2e.rulesSelections.bonusChoice}` or something.

packs/data/feats.db/devils-advocate.json
```diff
@@ -11,19 +11,42 @@
...
             "value": []
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "predicate": {
+                    "all": [
+                        "target:trait:devil"
+                    ]
```
TikaelSol, line 32:
```suggestion
                    "any": [
                        "target:trait:devil",
                        "origin:trait:devil"
                    ]
```
This will work if a devil feints for example.

packs/data/feats.db/shory-aerialist.json
```diff
@@ -29,7 +30,25 @@
...
+                },
+                "selector": "acrobatics",
+                "type": "circumstance",
+                "value": 2
+            },
+            {
+                "key": "FlatModifier",
+                "selector": "fly-speed",
+                "type": "status",
+                "value": 5
```
TikaelSol, line 49: This is technically wrong. We don't have a way to denote that you are flying via magic or because you have something else like an inventor item or some versatile heritage nonsense.

packs/data/feats.db/viking-shieldbearer.json
```diff
@@ -19,14 +19,51 @@
...
+                "key": "ActiveEffectLike",
+                "mode": "upgrade",
+                "path": "data.martial.weapon-base-battle-axe.rank",
+                "predicate": {
+                    "all": [
+                        "viking:battle-axe"
+                    ]
+                },
+                "value": 1
             }
```
TikaelSol, line 67:
```suggestion
            {
                "key": "ActiveEffectLike",
                "mode": "upgrade",
                "path": "data.martial.weapon-base-{item|flags.pf2e.rulesSelections.weaponChoice}.rank",
                "value": 1
            }
```
Path can just be resolved directly and we can add a flag to give it a nice name to refer to instead of spelling out the feat name.

packs/data/feats.db/viking-shieldbearer.json
```diff
@@ -19,14 +19,51 @@
...
+                        "value": "longsword"
+                    },
+                    {
+                        "label": "PF2E.Weapon.Base.battle-axe",
+                        "value": "battle-axe"
+                    }
+                ],
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Human.VikingShieldbearer.Prompt",
+                "rollOption": "viking"
```
TikaelSol, line 44:
```suggestion
                "flag": "weaponChoice"
```

packs/data/feats.db/witch-warden.json
```diff
@@ -19,11 +19,49 @@
...
+            {
+                "key": "Note",
+                "predicate": {
+                    "any": [
+                        "curse",
+                        "origin:trait:hag"
+                    ]
+                },
+                "selector": "saving-throw",
+                "text": "{item|name} @Localize[PF2E.SpecificRule.Human.WitchWarden.SaveNote]"
```
TikaelSol, line 62:
```suggestion
                "text": "<p class=\"compact-text\"><strong>{item|name}</strong> @Localize[PF2E.SpecificRule.Human.WitchWarden.SaveNote]</p>"
```

packs/data/feats.db/devils-advocate.json
TikaelSol: Hmm. This suggestion looks like it's going to break the json format though. So maybe manually do this.

packs/data/feats.db/shory-aerialist.json
Shandyan: I struggled with this one. Is there a non-magical way to fly? I think this bonus will be correct in 99% of cases, but can remove it if we need to go for 100%

packs/data/feat-effects.db/effect-saoc-astrology.json
Shandyan: I've simplified this massively - thanks for the advice (at some point, I'll 'get' choicesets properly, but still not quite there!).
I've left the labels in, as when I tested without them, it says I "have no valid options for Effect: blah".

TikaelSol: Hmm, should be able to just have a value? I'd like to fix that if we can't, instead of having more labels on things.

packs/data/feats.db/shory-aerialist.json
TikaelSol: Move speeds are a mess, but yeah there's non magical flight. The Soaring Dynamo feat or Blast Boots for example. But yeah it would be right for most characters and the ones it isn't right for are unlikely to take the feat so it's maybe fine? Maybe we could have some clever way to parse the traits of the thing granting a move speed in the future? Though that seems a bit excessive for tackling one feat like this.

PR #2476, opened 2022-06-13
Handle null price values for Coins

CarlosFdez:
Updates the old migration to remove null as well if it finds it.

### DIFF
--- a/src/module/item/physical/helpers.ts
+++ b/src/module/item/physical/helpers.ts
@@ -10,7 +10,8 @@ class CoinsPF2e implements Coins {
     gp: number;
     pp: number;
 
-    constructor(data: Coins = {}) {
+    constructor(data?: Coins | null) {
+        data ??= {};
         this.cp = data.cp ?? 0;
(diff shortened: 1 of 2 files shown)


src/module/item/physical/helpers.ts
```diff
@@ -10,11 +10,11 @@ class CoinsPF2e implements Coins {
     gp: number;
     pp: number;
 
-    constructor(data: Coins = {}) {
-        this.cp = data.cp ?? 0;
-        this.sp = data.sp ?? 0;
-        this.gp = data.gp ?? 0;
-        this.pp = data.pp ?? 0;
+    constructor(data?: Coins | null) {
```
stwlam, line 13: I'm suspecting that was an aberrant case from a single user--does this need to be a thing the helper?

CarlosFdez: new CoinsPF2e(null) leading to 0 coins kinda makes sense for the API to do anyways.

PR #2507, opened 2022-06-14
Add support for spell variants of spells with a variable cast time

In3luki:
This enables the creation of spell variants based on cast time (like Heal) and adds a prompt to select which variant to cast.

https://user-images.githubusercontent.com/41452412/173586017-499d679d-8195-4f75-88c7-70bd819c9553.mp4

Example setup for the Heal spell with variants:
- The description of the base spell can stay the same:
image
  
- Add a new variant in the details tab:
image
  
- Variants have their own spell sheet and can be edited independently of the base spell:
image
image
  
- Add additional variants as needed:
image
<br>
<br>
The variants can be sorted via drag and drop on the spell sheet and the prompt buttons are displayed in the same order.

### DIFF
--- a/src/module/actor/sheet/item-summary-renderer.ts
+++ b/src/module/actor/sheet/item-summary-renderer.ts
@@ -195,11 +195,17 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                 }
 
                 if (actor instanceof CharacterPF2e) {
-                    if (chatData.isAttack) {
+                    if (chatData.variants) {
+                        const label = game.i18n.localize("PF2E.Item.Spell.Variants.SelectVariantLabel");
                         buttons.append(
-                            `<span><button class="spell_attack" data-action="spellAttack">${game.i18n.localize(
-                                "PF2E.AttackLabel"
-                            )}</button></span>`
+                            `<span><button class="spell_attack" data-action="selectVariant">${label}</button></span>`
+                        );
+                        break;
+                    }
+                    if (chatData.isAttack) {
+                        const label = game.i18n.localize("PF2E.AttackLabel");
+                        buttons.append(
+                            `<span><button class="spell_attack" data-action="spellAttack">${label}</button></span>`
                         );
                     }
                     if (chatData.hasDamage) {
@@ -211,12 +217,12 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
 
                 break;
             case "consumable":
-                if (item instanceof ConsumablePF2e && item.charges.max > 0 && item.isIdentified)
+                if (item instanceof ConsumablePF2e && item.charges.max > 0 && item.isIdentified) {
+                    const label = game.i18n.localize("PF2E.ConsumableUseLabel");
                     buttons.append(
-                        `<span><button class="consume" data-action="consume">${game.i18n.localize(
-                            "PF2E.ConsumableUseLabel"
-                        )} ${item.name}</button></span>`
+                        `<span><button class="consume" data-action="consume">${label} ${item.name}</button></span>`
                     );
+                }
                 break;
             default:
         }
@@ -239,6 +245,9 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                     break;
                 case "consume":
                     if (item instanceof ConsumablePF2e) item.consume();
+                    break;
+                case "selectVariant":
+                    spell?.variantPrompt().then((variant) => variant?.toMessage());
                     break;
             }
         });
--- a/src/module/chat-message/data.ts
+++ b/src/module/chat-message/data.ts
@@ -26,6 +26,7 @@ type ChatMessageFlagsPF2e = foundry.data.ChatMessageFlags & {
         preformatted?: "flavor" | "content" | "both";
         isFromConsumable?: boolean;
         journalEntry?: DocumentUUID;
+        spellVariant?: { overlayIds: string[] };
         [key: string]: unknown;
     };
     core: NonNullable<foundry.data.ChatMessageFlags["core"]>;
--- a/src/module/chat-message/index.ts
+++ b/src/module/chat-message/index.ts
@@ -111,6 +111,13 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
             item.trickMagicEntry = trick;
         }
 
+        const spellVariant = this.data.flags.pf2e.spellVariant;
+        if (spellVariant && item?.isOfType("spell")) {
+            return item.loadVariant({
+                overlayIds: spellVariant.overlayIds,
+            });
+        }
+
         return item;
     }
 
--- a/src/module/item/spell/data/types.ts
+++ b/src/module/item/spell/data/types.ts
@@ -48,6 +48,19 @@ export interface SpellHeightenLayer {
     data: Partial<SpellSystemData>;
 }
 
+interface SpellOverlayOverride extends Partial<SpellSource> {
+    overlayType: "override";
+}
+
+/** Not implemented */
+interface SpellOverlayDamage {
+    overlayType: "damage";
+    choices: DamageType[];
+}
+
+type SpellOverlay = SpellOverlayOverride | SpellOverlayDamage;
+type SpellOverlayType = SpellOverlay["overlayType"];
+
 interface SpellSystemSource extends ItemSystemSource, ItemLevelData {
     traits: SpellTraits;
     level: {
@@ -87,6 +100,7 @@ interface SpellSystemSource extends ItemSystemSource, ItemLevelData {
         value: Record<string, SpellDamage>;
     };
     heightening?: SpellHeighteningFixed | SpellHeighteningInterval;
+    overlays?: Record<string, SpellOverlay>;
     save: {
         basic: string;
         value: SaveType | "";
@@ -122,4 +136,4 @@ interface SpellSystemData extends SpellSystemSource, ItemSystemData {
     traits: SpellTraits;
 }
 
-export { SpellData, SpellSource, SpellSystemData, SpellSystemSource };
+export { SpellData, SpellSource, SpellSystemData, SpellSystemSource, SpellOverlay, SpellOverlayType };
--- a/src/module/item/spell/index.ts
+++ b/src/module/item/spell/index.ts
@@ -21,9 +21,11 @@ import { DamageCategorization, DamageType } from "@system/damage";
 import { CheckPF2e } from "@system/rolls";
 import { StatisticRollParameters } from "@system/statistic";
 import { EnrichHTMLOptionsPF2e } from "@system/text-editor";
-import { ErrorPF2e, objectHasKey, ordinal } from "@util";
-import { SpellData, SpellHeightenLayer, SpellSource } from "./data";
+import { ErrorPF2e, getActionIcon, objectHasKey, ordinal } from "@util";
+import { SpellData, SpellHeightenLayer, SpellOverlay, SpellOverlayType, SpellSource } from "./data";
 import { MagicSchool, MagicTradition, SpellComponent, SpellTrait } from "./types";
+import { SpellOverlayCollection, SpellVariantPrompt } from "./overlay";
+import { LocalizePF2e } from "@system/localize";
 
 interface SpellConstructionContext extends ItemConstructionContextPF2e {
     fromConsumable?: boolean;
@@ -34,6 +36,8 @@ class SpellPF2e extends ItemPF2e {
 
     /** The original spell. Only exists if this is a variant */
     original?: SpellPF2e;
+    /** The overlays that were applied to create this variant */
+    appliedOverlays?: Map<SpellOverlayType, string>;
 
     /** Set if casted with trick magic item. Will be replaced via overriding spellcasting on cast later. */
     trickMagicEntry: TrickMagicItemEntry | null = null;
@@ -104,6 +108,14 @@ class SpellPF2e extends ItemPF2e {
         return !!this.original;
     }
 
+    get hasVariants(): boolean {
+        return this.overlays.size > 0;
+    }
+
+    override get uuid(): ItemUUID {
+        return this.isVariant ? this.original!.uuid : super.uuid;
+    }
+
     constructor(data: PreCreate<ItemSourcePF2e>, context: SpellConstructionContext = {}) {
         super(data, mergeObject(context, { pf2e: { ready: true } }));
         this.isFromConsumable = context.fromConsumable ?? false;
@@ -156,7 +168,7 @@ class SpellPF2e extends ItemPF2e {
 
         // If this isn't a variant, it probably needs to be heightened via overlays
         if (!this.isVariant) {
-            const variant = this.loadVariant(castLevel);
+            const variant = this.loadVariant({ castLevel });
             if (variant) return variant.getDamageFormula(castLevel, rollData);
         }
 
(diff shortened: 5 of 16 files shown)


src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -183,6 +183,14 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                 }
 
                 if (actor instanceof CharacterPF2e) {
+                    if (chatData.variants) {
+                        buttons.append(
+                            `<span><button class="spell_attack" data-action="selectVariant">${game.i18n.localize(
+                                "PF2E.SpellVariantSelectVariantLabel"
+                            )}</button></span>`
+                        );
```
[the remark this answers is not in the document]
In3luki: That would require a partial refactor of the item summary renderer as the other buttons are created in the same way.

src/module/item/spell/index.ts
```diff
@@ -231,16 +233,23 @@ class SpellPF2e extends ItemPF2e {
...
-        if (heightenEntries.length === 0) return null;
+        if (heightenEntries.length === 0 && isObjectEmpty(overlayData)) return null;
 
         const override = this.toObject();
+        // Sanitize base data
+        delete override.data.variants;
+        override.data.rules = [];
+
+        // Apply overlay data
+        mergeObject(override, overlayData, { overwrite: true });
```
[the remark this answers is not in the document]
In3luki: 👍

src/module/item/spell/index.ts
```diff
@@ -231,16 +233,23 @@ class SpellPF2e extends ItemPF2e {
...
+    loadVariant(castLevel: number, overlayData: Partial<SpellSource> = {}): SpellPF2e | null {
         if (this.original) {
-            return this.original.loadVariant(castLevel);
+            return this.original.loadVariant(castLevel, overlayData);
         }
 
         // Retrieve and apply variant overlays to override data
         const heightenEntries = this.getHeightenLayers(castLevel);
-        if (heightenEntries.length === 0) return null;
+        if (heightenEntries.length === 0 && isObjectEmpty(overlayData)) return null;
```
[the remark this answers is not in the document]
In3luki: I can pull the new `isEmpty` from v10 and add it to our utils if you want. Otherwise there  isn't much of an alternative in v9.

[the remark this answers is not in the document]
stwlam: just county the keys :stuck_out_tongue:

src/module/item/spell/index.ts
```diff
@@ -231,16 +233,23 @@ class SpellPF2e extends ItemPF2e {
...
+        if (heightenEntries.length === 0 && Object.keys(overlayData).length === 0) return null;
 
         const override = this.toObject();
+        // Sanitize base data
+        delete override.data.variants;
+        override.data.rules = [];
+
+        // Apply overlay data
+        mergeObject<Partial<SpellSource>>(override, overlayData, { overwrite: true });
+
```
stwlam, line 252: Assign

CarlosFdez:
Is this done? I can take a look at it soonish and see how it is. Any final things you wanna look at before I start deep diving?

In3luki:
> Is this done? I can take a look at it soonish and see how it is. Any final things you wanna look at before I start deep diving?

I still need to add a way for migrations in foundry and the packs to handle the partial variant data.  
I'm also planning to add an easier way to cast a specific variant, probably by making the dialog buttons draggable to the macro bar.  
Also the UI in the spell sheet could be better.
But I was waiting to see if this gets approved at all.

CarlosFdez:
For the variants source data, is it possible to add a variant type in there? I'm thinking for a complete solution, we'll need to allow more than one variant type, so not locking ourselves down to only spell source data might be ideal.

So maybe something like
```
interface SpellOverlayOverride {
    type: "override";
    // anything else you need (like img? name?)
    data: Partial<SpellSystemData>;
}

interface SpellOverlayDamage {
    type: "damage";
    choices: DamageType[];
}

type SpellOverlay = SpellVariantOverride | SpellVariantDamage;
```

and then later on:
```ts
interface SpellSystemSource {
    ...
    variants: Record<string, SpellVariant>;
}
```

I also don't know if it'll help, but what I did when I was working on this was keep a distinction between "overlay" and "variant". A variant is the spell you are casting. Overlay is data you are using to create it, and a spell variant can be composed of one or more overlays.

src/module/item/spell/index.ts
```diff
@@ -231,16 +233,21 @@ class SpellPF2e extends ItemPF2e {
      * This handles heightening as well as alternative cast modes of spells.
      * If there's nothing to apply, returns null.
      */
-    loadVariant(castLevel: number): SpellPF2e | null {
+    loadVariant(castLevel: number, overlayData: Partial<SpellSource> = {}): SpellPF2e | null {
```
CarlosFdez, line 236: The goal behind this function is that it should be able to *resolve* what variant you're trying to load. Instead it should give it a variant id or list of variant ids and then it should fetch it from there.

In3luki:
I've added support for different types of spell overlay data and made it possible to merge multiple overlays in `loadVariants`.
The `override` overlay is the only one that is implemented for now, but I've added a type stub for a `damage` overlay.
While working on this I also realized that embedding the prepared variants is not necessary so I've removed that.
Updating the spell with with `data.update()` and re-rendering the sheet works fine.

src/module/item/spell/data/types.ts
```diff
@@ -87,6 +100,7 @@ interface SpellSystemSource extends ItemSystemSource, ItemLevelData {
         value: Record<string, SpellDamage>;
     };
     heightening?: SpellHeighteningFixed | SpellHeighteningInterval;
+    overlay?: Record<string, SpellOverlay>;
```
CarlosFdez, line 103: This should be overlays?: to be plural.

src/module/item/spell/index.ts
```diff
@@ -231,22 +237,57 @@ class SpellPF2e extends ItemPF2e {
...
+                            // Sanitize data
+                            delete source.data.overlay;
+                            source.data.rules = [];
+
+                            source = mergeObject(source, overlayData, { overwrite: true });
+                            break;
+                        }
+                    }
+                    // Store overlay type as a flag on the variant spell
+                    source.flags = mergeObject(source.flags, { pf2e: { overlayType: overlayData.overlayType } });
```
CarlosFdez, line 271: Not sure if this is the right call. If there is more than one overlay spell, the one prioritized is the one applied second. Is there a reason to store in the flags?

src/module/item/spell/index.ts
```diff
@@ -231,22 +237,57 @@ class SpellPF2e extends ItemPF2e {
...
+
+                    // Override excludes other overlay types
+                    if (overlayData.overlayType === "override" && overlayIds.length > 1) {
```
CarlosFdez, line 274: Instead of doing this check here, it might be more ideal to do an `overlayIds.map(id => this.data.data.overlays?.[overlayId])` or the like to create a list of overlay data to apply, and *then* perform the checks before looping. 

Also, the goal of separate overlay types is to allow the stacking. The only thing that shouldn't be allowed is more than one override. However an override with a separate damage type selection should be fine (and might handle some nasty use cases).

CarlosFdez:
So here's something that *really* came in handy for me when I was implementing it, which might come in handy for you.

You can try to create a new Collection subclass called SpellOverlayCollection. In the same vein that ActorInventory and ActorSpellcasting work, you can add special behavior and methods to it, which you can access via spell.overlays.create() or what-have-you.

This will allow you to reduce the number of overlay specific methods in the spell that are more "internal" to the implementation, and also allow you to create special overlay methods.

src/module/item/spell/data/types.ts
In3luki: True.

src/module/item/spell/index.ts
```diff
@@ -261,6 +304,33 @@ class SpellPF2e extends ItemPF2e {
             .sort((first, second) => first.level - second.level);
     }
 
+    /** Returns a specific variant of this spell or a shows a prompt to select one. */
+    async getVariantChoices(this: Embedded<SpellPF2e>, overlayId?: string): Promise<Embedded<SpellPF2e> | null> {
```
CarlosFdez, line 308: Might need a better name, this doesn't return the variant choices.

@stwlam can you think of something? I'm thinking of getOrPromptVariant().

src/module/chat-message/data.ts
```diff
@@ -26,6 +26,7 @@ type ChatMessageFlagsPF2e = foundry.data.ChatMessageFlags & {
         preformatted?: "flavor" | "content" | "both";
         isFromConsumable?: boolean;
         journalEntry?: DocumentUUID;
+        spellVariant?: { original: ItemUUID; id: string };
```
CarlosFdez, line 29: Perhaps we should rename id to overrideId or override? We can later expand this to add the other types that involve a secondary choice.

CarlosFdez: Actually, now I'm wondering what original is actually for here. Shouldn't origin.uuid be pointing to the original spell, not the variant?

In3luki: `original` is the UUID of the base spell (based on `SpellPF2e#original`) and `id` is the overlay id. The id field should actually be a string array now.
`originalSpell: ItemUUID` and `overlayIds: string[]` might be more descriptive.

src/module/item/spell/index.ts
In3luki: I'll strip out the `overlayId` param as that would now only work with an override overlay and rename it to `promptVariant()`.
The `overlayIds[] -> spellVariant` concept needs a way to specifiy which overlayIds represent a single spell variant.
When that is done the method could be called with the `variantId` again.

src/module/chat-message/data.ts
CarlosFdez: Is there a reason why the uuid of the variant spell, when cast, can't be the same uuid as the original spell? That way you won't need to do regex to handle that.

In3luki: I'm using the override overlay `id` as the document `_id` of the variant spell. That might no longer be necessary... I'll think about it.

In3luki: Ok, sheet editing breaks down when I omit the id from the override variants so I opted to override `SpellPF2e#uuid` and return the original uuid for variant spells. This way only the overlayIds need to be stored in the chat message flags.

CarlosFdez: Ah, that's a shame. The only alternative I can think of is a keepId field that defaults to true that you can use when creating the variants for sheet usage by setting to false. But I'm fine with overriding uuid on the spell itself.

CarlosFdez: Another possible idea might have been to override the sheet's id getter, if the problem with sheet editing is that you can't tell what's unique or not.

src/module/chat-message/index.ts
```diff
@@ -111,6 +111,15 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
             item.trickMagicEntry = trick;
         }
 
+        const spellVariant = this.data.flags.pf2e.spellVariant;
+        if (spellVariant && item instanceof SpellPF2e) {
+            return (
+                (item.loadVariant({
+                    overlayIds: spellVariant.overlayIds,
+                }) as Embedded<ItemPF2e> | null) ?? null
```
CarlosFdez, line 119: I don't think you need to cast here.

src/module/item/spell/index.ts
```diff
@@ -231,23 +243,58 @@ class SpellPF2e extends ItemPF2e {
...
+                        `Error loading variant of Spell ${this.name} (${this.uuid}). Cannot apply multiple override overlays.`
+                    );
+                }
+                for (const overlayId of overlayIds) {
+                    const overlayData = this.overlays.get(overlayId);
+                    if (!overlayData) {
+                        throw ErrorPF2e(
+                            `Spell ${this.name} (${this.uuid}) does not have an overlay with id: ${overlayId}`
+                        );
+                    }
```
CarlosFdez, line 272:
```suggestion
                for (const overlayData of overlays) {
```

src/module/item/spell/overlay.ts
```diff
@@ -0,0 +1,137 @@
...
+                    const variantSpell = this.spell.loadVariant({ overlayIds: [id] });
+                    if (variantSpell) {
+                        variantSpell.sheet.render(true);
+                    }
+                }
+                break;
+        }
+    }
+
+    async update(overlayId: string, _data: Record<string, unknown>): Promise<void> {
```
CarlosFdez, line 51: What is this method for? It doesn't seem to do anything.

src/module/chat-message/data.ts
CarlosFdez: Other than a keepId flag so that variants created for item sheets have a different id, I think this is the only other way. Honestly I'm fine with it. Overriding get uuid() is probably a solid call.

src/module/item/spell/overlay.ts
In3luki: That will be the method to update the other types of overlay data. I was contemplating to make the first parameter `string | SpellPF2e` and integrating `updateOverrideData` into it but wasn't sure if that was a good idea.

CarlosFdez: Might be better to not have this here at all for now, or if it needs to be stubbed for whatever reason, leave a comment.

src/module/item/spell/sheet.ts
```diff
@@ -196,6 +219,103 @@ export class SpellSheetPF2e extends ItemSheetPF2e<SpellPF2e> {
...
+        });
+
+        $html.find(`[data-action="variant-edit"]`).on("click", (event) => {
+            const id = $(event.target).closest("[data-action='variant-edit']").attr("data-id");
+            if (id) {
+                this.item.loadVariant({ overlayIds: [id] })?.sheet.render(true);
+            }
+        });
+
+        $html.find(`[data-action="variant-delete"]`).on("click", (event) => {
```
stwlam, line 234:
```suggestion
        $html.find("[data-action=variant-delete]").on("click", (event) => {
```

static/lang/en.json
```diff
@@ -2990,6 +2991,11 @@
         "SpellCategorySpell": "Spell",
         "SpellUsesLabel": "Uses",
         "SpellUnlimitedLabel": "Unlimited",
+        "SpellVariantDeleteDialogTitle": "Delete Spell Variant",
+        "SpellVariantDeleteDialogText": "Are you sure you want to delete '{variantName}'?",
+        "SpellVariantSelectVariantLabel": "Select Variant",
+        "SpellVariantsLabel": "Spell Variants",
+        "SpellVariantSheetTitle": "{originalName} (Variant)",
```
stwlam, line 2998: We're avoiding bloating the top-level namespace now: move all this to `PF2E.Item.Spell.Variant`

src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -195,6 +195,14 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                 }
 
                 if (actor instanceof CharacterPF2e) {
+                    if (chatData.variants) {
+                        buttons.append(
+                            `<span><button class="spell_attack" data-action="selectVariant">${game.i18n.localize(
+                                "PF2E.SpellVariantSelectVariantLabel"
+                            )}</button></span>`
```
stwlam, line 202: Avoid doing so much in a string interpolation that it requires multiple lines: localize the label above it and interpolate that.

PR #2516, opened 2022-06-15
Fix players not being able to see save DCs from their own spells

stwlam:


src/module/chat-message/index.ts
```diff
@@ -166,7 +166,7 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
     override prepareData(): void {
         super.prepareData();
 
-        const rollData = this.actor?.getRollData();
+        const rollData = { ...this.actor?.getRollData(), actor: this.actor, item: this.item };
```
CarlosFdez, line 169: actor.getRollData() doesn't do anything, funny enough. Its still the default foundry implementation. Its item.getRollData() that does something.

I think actor.getRollData() should return `{ actor: this }` and then this should be `this.item?.getRollData() ?? this.actor?.getRollData()`. That might work the best out of all of them.

stwlam: `actor.getRollData()` dumps the system data properties in there, doesn't it?

CarlosFdez: currently yeah. item.getRollData() does that baseline as well but we overwrote it. Making actor.getRollData() into something useful might be a good idea.

I'm digging into the copy of V10 I have, and it seems ChatMessage has a getRollData() method too, and it defaults to doing actor.getRollData() on the speaker actor. Might also be good to put this there?

stwlam: Well, I'll leave it to you to figure that out. This PR is unrelated.

PR #2533, opened 2022-06-15
Cover two gaps in modifier adjustments getting applied

stwlam:
Adjustments weren't getting picked up for (1) custom modifiers and (2) modifiers with the "all" selector.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -460,13 +460,16 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
     }
 
     getModifierAdjustments(selectors: string[], slug: string): ModifierAdjustment[] {
-        return Array.from(
+        const adjustmentsRecord = this.synthetics.modifierAdjustments;
+        const selectorMatches = Array.from(
             new Set(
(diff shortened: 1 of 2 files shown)


src/module/actor/base.ts
```diff
@@ -460,13 +460,16 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
...
-        return Array.from(
+        const adjustmentsRecord = this.synthetics.modifierAdjustments;
+        const selectorMatches = Array.from(
             new Set(
-                selectors
-                    .flatMap((s) => this.synthetics.modifierAdjustments[s] ?? [])
-                    .filter((a) => [slug, null].includes(a.slug))
+                selectors.includes("all")
+                    ? Object.values(adjustmentsRecord).flat()
+                    : selectors.flatMap((s) => this.synthetics.modifierAdjustments[s] ?? [])
```
CarlosFdez, line 468:
```suggestion
                    : selectors.flatMap((s) => adjustmentsRecord[s] ?? [])
```

if you're going through the trouble of aliasing, might as well use it?

PR #2542, opened 2022-06-16
Appropriately nest spellcasting roll options

stwlam:


CarlosFdez:
The concern is that now we're adding the spellcasting entry's slug, and its possible for the user to trick predicates into thinking its primal or something. So now we might have to nest everything here, even though there's only ever one possible value per nest (spellcasting: options only show the current casted entry).

This might affect @TikaelSol so he needs to be the one to review it.

PR #2561, opened 2022-06-17
Separate NPC system data and source

stwlam:


CarlosFdez:
Also adding the ability for NPCs to be controlled by rules based vision here?

stwlam:
> Also adding the ability for NPCs to be controlled by rules based vision here?

nope

PR #2563, opened 2022-06-17
Automate Kobold Features

leerhund:
Add ChoiceSet to Draconic Exemplar
Remove multiple Dragonscaled Kobold heritages in favor of one with Resistance based on Draconic Exemplar choice.


Shandyan (commented):
I don't know if the prompt text in Draconic Exemplar should be more specific. Flagging for Tikael/Spartan to make the call. 

Otherwise, everything else looks fine!

SpartanCPA:
I'm gonna leave this one for Tikael, since it appears very close to what he planned on doing for Battlezoo Dragons

TikaelSol:
I've been waiting for the fabled "Actions that Act" to do these so we don't need multiple copies of the same item. In the meantime we can merge at least the choices and whatnot but I don't want to introduce more items we just need to consolidate in the near future.

leerhund:
> I've been waiting for the fabled "Actions that Act" to do these so we don't need multiple copies of the same item. In the meantime we can merge at least the choices and whatnot but I don't want to introduce more items we just need to consolidate in the near future.

Should be good to go.

PR #2572, opened 2022-06-19
Warn about mode before priority for AELike

CarlosFdez:
priority is derived from mode, so missing mode and priority will warn about priority currently, when it should warn about mode first.


stwlam:
The priority is set in the constructor if it's not defined.

CarlosFdez:
The priority is set in the constructor *if* there is a valid mode, otherwise its set to NaN. If you make an AE-Like that is missing both, it will tell you that priority is missing.

Issue #2334, opened 2022-06-03
Mark for death (Assassins Dedication)


cswendrowski:
In GitLab by @AGulyaev12 on Jun 3, 2022, 01:24

The action right now has only the description but neither the GM nor a player can drag'n'drop the effect onto a target. does the effect actauly exists and work ?

TMunsie:
Presently this is the same as hunted shot. We can't use an effect that needs to know who two actors are. We could, similar to hunted shot, create toggles.

PR #2575, opened 2022-06-19
Automate Mark for Death Action

avagdumortesin:
Fixes https://github.com/foundryvtt/pf2e/issues/2334

### DIFF
--- a/packs/data/actions.db/mark-for-death.json
+++ b/packs/data/actions.db/mark-for-death.json
@@ -77,7 +77,7 @@
                 "definition": {
                     "any": [
                         "weapon:trait:finesse",
-                        "weapon:trait:agile",
+                        "weapon:trait:agile"
                     ]
                 },
                 "key": "AdjustStrike",
--- a/packs/data/actions.db/mark-for-death.json
+++ b/packs/data/actions.db/mark-for-death.json
@@ -16,7 +16,146 @@
         "requirements": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.MarkForDeath.TargetMark",
+                "option": "marked-for-death",
+                "toggleable": true
+            },
+            {
+                "key": "FlatModifier",
+                "label": "PF2E.SpecificRule.MarkForDeath.TargetMark",
+                "predicate": {
+                    "all": [
+                        "action:seek",
+                        "marked-for-death"
+                    ]
+                },
+                "selector": "perception",
+                "type": "circumstance",
+                "value": 2
+            },
+            {
+                "key": "FlatModifier",
+                "label": "PF2E.SpecificRule.MarkForDeath.TargetMark",
+                "predicate": {
+                    "all": [
+                        "action:feint",
+                        "marked-for-death"
+                    ]
+                },
+                "selector": "deception",
+                "type": "circumstance",
+                "value": 2
+            },
+            {
+                "definition": {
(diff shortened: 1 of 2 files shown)


packs/data/actions.db/mark-for-death.json
```diff
@@ -16,7 +16,146 @@
         "requirements": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.MarkForDeath.TargetMark",
```
Shandyan, line 23: You can leave the label out. If there's no label, it defaults to the name of the source item.

If you are adding a specific label, you also need to update re-en.json to specify what will be displayed.

packs/data/actions.db/mark-for-death.json
```diff
@@ -16,7 +16,146 @@
...
+                        "marked-for-death"
+                    ]
+                },
+                "selector": "deception",
+                "type": "circumstance",
+                "value": 2
+            },
+            {
+                "definition": {
+                    "any": [
```
Shandyan, line 55: I think this ability should only work on unarmed attacks that are also agile or finesse. As worded, it will work on any unarmed attack.

packs/data/actions.db/mark-for-death.json
avagdumortesin: Added the labels in the new commit.

packs/data/actions.db/mark-for-death.json
avagdumortesin: Removed the unarmed predicate in the new commit.

leerhund:
@avagdumortesin You've left a stray comma on line 79 of the Mark for Death json

avagdumortesin:
> @avagdumortesin You've left a stray comma on line 79 of the Mark for Death json

Yup; noticed it when the CI check failed. Fixed it in commit b797784.

packs/data/actions.db/mark-for-death.json
```diff
@@ -16,7 +16,141 @@
...
+                "key": "AdjustStrike",
+                "mode": "add",
+                "predicate": {
+                    "all": [
+                        "marked-for-death"
+                    ]
+                },
+                "property": "weapon-traits",
+                "value": "deadly-d8"
+            },
```
SpartanCPA, line 112: Can you verify in Foundry that this will not result in the weapon having _both_ deadly d6 and deadly d8, and thus rolling too many dice?

leerhund: Can confirm.
First roll without Toggle, 2nd with.
image

avagdumortesin: It looks like the portions that upgrade the deadly die just cascade, and everything loads up to Deadly D12.

Would brackets perhaps work, or should we remove the automation for the upgrade, and leave notes on how to upgrade it in the RE?

avagdumortesin: Alternatively, I can leave in the Deadly D6 > Deadly D8, and we can add notes to change that RE if Deadly D8 or higher is already on, thus only having 2 `AdjustStrike` REs.

SpartanCPA (approved):

packs/data/actions.db/mark-for-death.json
leerhund: Reverse the order of the Adjust Strikes. Have the one for advancing to d12 first, then go down from there...
Check the Discord, I'll @ you with a modified action file to import and test.

PR #2586, opened 2022-06-20
Dynamically prevent invalid items from being dragged onto actors

CarlosFdez:


CarlosFdez:
I may have missed some item types, bring them up if you notice any.

PR #2620, opened 2022-06-21
Fixed: Vivacious speed class feature.

joshua-baller:
Lack of 'slug' was causing second rule to be ignored. Slug was added.


stwlam:
Superceded by #2619

PR #2621, opened 2022-06-21
Support increment/decrement/source for conditions in the familiar sheet

CarlosFdez:
As a side effect, this also removes the wonky looking divider border from NPC sheet, and fixes condition and effect name alignment.

This also removes character sheet condition capping for doomed/dying/wounded, but that is not only bugged but also didn't work from token hud or effect panel.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -1,7 +1,7 @@
 import { SkillAbbreviation } from "@actor/creature/data";
 import { MODIFIER_TYPE, ProficiencyModifier } from "@actor/modifiers";
 import { ActorSheetDataPF2e } from "@actor/sheet/data-types";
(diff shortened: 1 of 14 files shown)


stwlam:
So now you can increase dying, doomed, and wounded, with no limit, from all interfaces?

CarlosFdez:
Yup, as opposed to being able to increase dying, doomed, and wounded, with no limit, from most interfaces (that are the common use ones).

stwlam:
So then this introduces a regression? On purpose?

CarlosFdez:
Its more that the original feature never worked correctly. We need to implement it properly, separately, at either the actor level or condition manager level, rather than at the sheet level.

(also if doomed updates dying max, then its impossible to cap doomed anyways, and it seemed you were willing to let that one go away).

PR #2647, opened 2022-06-22
Remove ranged attacks benefiting from flat-footed due to flanking

stwlam:


CarlosFdez:
This works for combination weapons as well here? Not sure if this is the contextual item or not.

stwlam:
> This works for combination weapons as well here? Not sure if this is the contextual item or not.

Yes

PR #2658, opened 2022-06-23
Refine roll-option toggle disabling with disabledValue property

stwlam:
If a roll option toggle is disabled due to its `disabledIf` predicate failing, an additional `disabledValue` property (defaults to false) can control whether the roll option is still active when toggled on but disabled

### DIFF
--- a/packs/data/feats.db/game-hunter-dedication.json
+++ b/packs/data/feats.db/game-hunter-dedication.json
@@ -42,8 +42,8 @@
             },
             {
                 "domain": "all",
-                "enabledIf": {
-                    "all": [
+                "disabledIf": {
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I'm guessing the goal is to have predicates that cannot be detected by the user but are on anyways? What for? Why do it this way instead of a hideFromUser option?

stwlam:
Visibility doesn't play into this, no.

CarlosFdez:
TBH, I don't understand enabledIf in the first place, so its making it harder to understand what this is doing. I thought it was a visibility toggle from the comment, is it just something preventing user input?

PR #2684, opened 2022-06-24
Set vehicles' fortitude save Statistic

stwlam:


src/module/actor/vehicle/data.ts
```diff
@@ -52,17 +53,10 @@ interface VehicleSystemData extends ActorSystemData {
         speed: number;
     };
     saves: {
-        fortitude: {
-            rank: number;
-            value: number;
-            saveDetail: string;
-        };
+        fortitude: StatisticModifier;
```
CarlosFdez, line 56: This should probably be SaveData instead. Or something equivalent that extends StatisticCompatData.

PR #2686, opened 2022-06-24
Allow thrown weapons to have reload 0

JDCalvert:
Currently, a weapon is defined as being thrown if it is ranged and has reload of `-` or `null`. Also, thrown weapons' reload times are forcibly set to `-` as this is the value that _almost_ all thrown weapons use. However, the [shuriken](https://2e.aonprd.com/Weapons.aspx?ID=79) has a reload of 0. This is currently impossible to set in the system, as even if one tries to manually alter it, it is set back to `-`.

I've changed four things:
- The reload value is set to `-` only if it is not already set to something else. This allows setting thrown weapons to have a different reload value.
- The definition of a thrown weapon is now a ranged weapon with the "thrown" trait. This removes the dependency on the reload value.
- The definition of a weapon requiring ammunition is now a ranged weapon that _is not thrown_ and has a non-`-` reload time.
- The shuriken now has a reload of 0.

### DIFF
--- a/packs/data/equipment.db/shuriken.json
+++ b/packs/data/equipment.db/shuriken.json
@@ -39,13 +39,13 @@
             "value": "0"
         },
         "potencyRune": {
-            "value": null
+            "value": 0
         },
         "preciousMaterial": {
-            "value": ""
+            "value": null
         },
         "preciousMaterialGrade": {
-            "value": ""
+            "value": null
         },
         "price": {
             "value": {
@@ -67,26 +67,29 @@
             "value": ""
         },
         "propertyRune1": {
-            "value": ""
+            "value": null
         },
         "propertyRune2": {
-            "value": ""
+            "value": null
         },
         "propertyRune3": {
-            "value": ""
+            "value": null
         },
         "propertyRune4": {
-            "value": ""
+            "value": null
         },
         "quantity": 1,
         "range": 20,
         "reload": {
-            "value": "-"
+            "value": "0"
         },
         "rules": [],
         "size": "med",
         "source": {
             "value": "Pathfinder Core Rulebook"
+        },
+        "specific": {
+            "value": false
         },
         "splashDamage": {
             "value": null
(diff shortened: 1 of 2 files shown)


stwlam:
I'm not seeing a reason to consider the shuriken case as anything other than an obvious error.

JDCalvert:
I'm not sure what you mean exactly.

If you mean that a user will easily spot it as an error, and should know that it should be reload 0, then I don't agree since the Foundry system should give them the correct information.

If you mean that the shuriken is a data entry error and should be corrected, that's impossible without the rest of this change because the reload value is current reset to `-` during data preparation because it's a thrown weapon.

CarlosFdez:
I think shark means that shuriken should be -

JDCalvert:
Ah, I see. I don't see any evidence that it _is_ an error. After three printings of the CRB, the shuriken still has reload 0 and I've seen no discussions around it being an error.

There is also precedence for this from 1st edition, where the shuriken was "...treated as ammunition for the purposes of drawing them...", which I think is what Paizo was going for by making it reload-0.

stwlam:
> Ah, I see. I don't see any evidence that it _is_ an error. After three printings of the CRB, the shuriken still has reload 0 and I've seen no discussions around it being an error.
> 
> There is also precedence for this from 1st edition, where the shuriken was "...treated as ammunition for the purposes of drawing them...", which I think is what Paizo was going for by making it reload-0.

A _lot_ of errors persist in the CRB despite errata. Some even get worse. Actually giving it a reload value of zero opens a massive can of worms that have no clear solutions. I'd like to get official clarification on it and all its ramifications before moving forward on treating it differently than any other thrown weapon.

JDCalvert:
I've tried to have a look and I can't find any discussions around this rule being an error, apart from one off-hand comment back in November on the Discord server about it possibly being a typo (there was no further discussion I don't think). Usually errors like this are fairly well-documented on forums.

I doubt it'll be possible to get anyone official to state that this isn't a typo, since that claim could be made about any ruling someone disagrees with. However, I've asked the question on the Paizo forum so maybe something will come out of it.

PR #2711, opened 2022-06-26
Clean up the Ability Score builder, again and by a different person

fryguy1013:
Several things are broken by the changes that have been made to "improve" it.

Also, improve behavior of key ability score being set manually.

### DIFF
--- a/src/module/actor/character/index.ts
+++ b/src/module/actor/character/index.ts
@@ -434,6 +434,7 @@ class CharacterPF2e extends CreaturePF2e {
     override prepareEmbeddedDocuments(): void {
         super.prepareEmbeddedDocuments();
 
+        this.setAbilityModifiers();
         this.setNumericRollOptions();
         this.deity?.setFavoredWeaponRank();
     }
@@ -856,7 +857,9 @@ class CharacterPF2e extends CreaturePF2e {
             // Record base values: same as stored value if in manual mode, and prior to RE modifications otherwise
             ability.base = ability.value;
         }
+    }
 
+    private setAbilityModifiers(): void {
         // Set modifiers
         for (const ability of Object.values(this.data.data.abilities)) {
             ability.mod = Math.floor((ability.value - 10) / 2);
--- a/src/module/actor/sheet/popups/ability-builder.ts
+++ b/src/module/actor/sheet/popups/ability-builder.ts
@@ -24,19 +24,6 @@ export class AbilityBuilderPopup extends Application {
         return `ability-builder-${this.actor.id}`;
     }
 
-    private get boostFlawState(): BoostFlawState {
-        return {
-            lockedFlaw: false,
-            lockedBoost: false,
-            boosted: false,
-            available: false,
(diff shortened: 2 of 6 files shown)


static/templates/actors/ability-builder.html
```diff
@@ -135,7 +135,7 @@ <h2>{{localize "PF2E.Actor.Character.AbilityBuilder.Flaws"}}</h2>
                 <div class="full-row">{{localize "PF2E.Actor.Character.AbilityBuilder.BackgroundMissingHelp"}}</div>
             {{/if}}
 
-            <div class="hover-content" class="background-tooltip">
+            <div class="hover-content" id="background-tooltip">
```
CarlosFdez, line 138: The value of id must be unique within the entire page. This shouldn't use it.

CarlosFdez: That said, it seems that data-tooltip-content seems to use ids everywhere, even in the parts that should be unique. I guess without using tooltipster document side, it has to use ids unfortunately.

Might be better to use .tooltipster() for this like most of the codebase uses?

fryguy1013: It is using tooltipster like most of the rest of the codebase uses.

CarlosFdez: I meant feeding in the content directly via the ability builder application. But making the ids unique is fine enough.

src/module/actor/character/index.ts
```diff
@@ -102,6 +102,9 @@ class CharacterPF2e extends CreaturePF2e {
     class!: Embedded<ClassPF2e> | null;
     deity!: Embedded<DeityPF2e> | null;
 
+    // base ability scores, before any rule element breaks things
+    baseAbilityScores!: Abilities | null;
+
```
stwlam, line 107: This is unnecessary and can get acquired in the builder application

src/module/actor/sheet/popups/ability-builder.ts
```diff
@@ -3,14 +3,15 @@ import { Abilities } from "@actor/creature/data";
 import { AbilityString } from "@actor/types";
 import { ABILITY_ABBREVIATIONS } from "@actor/values";
 import { AncestryPF2e, BackgroundPF2e, ClassPF2e } from "@item";
+import { ActorPF2e } from "@actor/index";
 
-export class AbilityBuilderPopup extends Application {
+export class AbilityBuilderPopup extends FormApplication<ActorPF2e> {
```
stwlam, line 8: There's no point in making this a `FormApplication` only to later prevent it from being submitted.

src/module/actor/character/index.ts
fryguy1013: There is no place that the after calculated ability scores are available. This is for builder mode too.

stwlam: How about the bases be set for each score in the abilities object?

fryguy1013: ok i'll do that

PR #2714, opened 2022-06-26
Display "base" ability scores from ability builder

stwlam:


fryguy1013:
I think the solution in my PR is probably better for this, since it's super confusing in builder mode if there's an apex item too

stwlam:
> I think the solution in my PR is probably better for this, since it's super confusing in builder mode if there's an apex item too

Updated

PR #2721, opened 2022-06-26
Have migration 760 handle one P element and one Text node

stwlam:


SpartanCPA (approved):
Data is good. Didn't look at the migration

CarlosFdez:
CONFLICTS

PR #2742, opened 2022-06-27
Match strike by id and slug when creating card damage buttons

stwlam:


CarlosFdez:
I'm hoping someone doesn't attempt to optimize this later, this looks wasteful, but I more or less understand the why.

PR #2760, opened 2022-06-29
Remove TreatWounds with alt skills setting, the answer is yes

CarlosFdez:
Converts treat wounds macro to typescript, we should probably include a migration later (gated by source id maybe)


stwlam:
Conflicts!

PR #2761, opened 2022-06-29
Fix rage damage of animal instinct barbarians.

TikaelSol:
Adding back in the load bearing jank.


stwlam:
Doesn't this do the same thing as a selector targeting that specific attack?

TikaelSol:
It does the same thing, sort of. You would need to reference the item id and not the specific attack since some items have multiple  strikes generated. I'm indifferent to how we solve it but we should do something before the next update.

PR #2762, opened 2022-06-29
Move crit formula setting to variant rules

CarlosFdez:
image

### DIFF
--- a/src/module/system/settings/index.ts
+++ b/src/module/system/settings/index.ts
@@ -36,19 +36,6 @@ export function registerSettings() {
         config: true,
     });
 
-    game.settings.register("pf2e", "critRule", {
-        name: "PF2E.SETTINGS.CritRule.Name",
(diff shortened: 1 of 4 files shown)


stwlam:
not a variant rule

CarlosFdez:
Its effectively a variant rule in both form and function (de-facto per-say), the only part where it isn't a variant rule is where its located. And also, it'd be nice to reduce the options in the main section and shove this thing that is a variant rule in look and feel somewhere else.

If the only thing that matters to be in that section is that its explicitly called out as a variant rule, then sure. But it basically is one and discussion pointed it out as such.

But whatever. If we're keeping it in the main section then there it shall stay.

CarlosFdez:
That said, were we to move it, we'd have all the "paizo says you can run this way instead" rules in a single place and that'd be sweet.

PR #2766, opened 2022-06-29
Show action to chat and edit/delete in view mode for beta hazard sheet

CarlosFdez:


static/templates/actors/hazard/sheet-new.html
```diff
@@ -99,14 +99,13 @@ <h3>Actions</h3>
                                     <div class="action-name rollable">
                                         <h4><a>{{action.name}} {{{actionGlyph action.actionCost}}}</a></h4>
                                     </div>
-                                    {{#if ../../owner}}
                                     <div class="item-controls">
-                                        {{#if @root.editing}}
+                                        <a class="item-chat chat" title="{{ localize 'PF2E.NPC.SendToChat'}}"><i class="fas fa-comment-alt"></i></a>
```
stwlam, line 103: data-action? :)

src/module/actor/npc/sheet.ts
```diff
@@ -497,15 +495,6 @@ export class NPCSheetPF2e<TActor extends NPCPF2e> extends CreatureSheetPF2e<TAct
...
 
-    private async onClickToChat(event: JQuery.ClickEvent): Promise<void> {
-        event.preventDefault();
-
-        const itemId = $(event.currentTarget).parents(".item").attr("data-item-id") ?? "";
-        const item = this.actor.items.get(itemId, { strict: true });
-        if (item instanceof PhysicalItemPF2e && !item.isIdentified) return;
-        await item.toChat(event);
-    }
-
```
stwlam, line 508: This has some unique behavior to it that'd be lost: not sure of the wisdom of that behavior, but removing it doesn't seem germane to the PR.

CarlosFdez: I added that behavior to the new one, other than the throwing an error if it doesn't exist bit.

static/templates/actors/hazard/sheet-new.html
CarlosFdez: wouldn't be bad, would also be a change to NPC but I'd be for it.

src/module/actor/npc/sheet.ts
stwlam: I think it's fine that it throw an error: if the item ID is bogus, a much bigger problem is present.

static/templates/actors/hazard/sheet-new.html
CarlosFdez: Actually this is more invasive than we think, the sibling elements key using "edit" and "delete" css classes. IMO might be best to leave them all the same for now.

PR #2770, opened 2022-06-29
Review B3 - sea dragons

MSAbaddon:


SpartanCPA (commented):
Why are you moving the AC bonus to an effect as opposed to the toggle? It can't affect any creature besides the dragon itself.

MSAbaddon:
Until now I always did things with durations as effects for easier tracking. Liquify is a reaction, the AC bonus lasts until the beginning of the dragon's next turn, which might include a whole round.

SpartanCPA (approved):

PR #2776, opened 2022-06-30
Update Adventurer's Pack with the correct amount of chalk

MrVauxs:
It should be 10 pieces of chalk, not just 1. This also causes a pricing discrepancy. A miniscule one, but a discrepancy nonetheless.


stwlam:
It's a single item with ten uses

MrVauxs:
> It's a single item with ten uses
> 
> Or maybe that was changed back too ...

It was changed to being 1 item with a price for 10

PR #2777, opened 2022-06-30
Correct spellheart usage and add toggles for resistances/damage

dmarcuse:
Spellhearts work like talismans in that they're affixed to armor or weapons, rather than being held. This PR updates the data to accordingly, and also adds toggles to automate the bonus strike damage and damage resistances they grant when activated.

There already seem to be effects for the five-feather wreath spellhearts to add the acrobatics bonus, so I didn't add toggles for those - though I'm happy to convert those to toggles as well for consistency.

### DIFF
--- a/packs/data/equipment.db/five-feather-wreath-greater.json
+++ b/packs/data/equipment.db/five-feather-wreath-greater.json
@@ -50,7 +50,7 @@
             ]
         },
         "usage": {
-            "value": "held-in-one-hand"
+            "value": "affixed-to-armor-or-a-weapon"
         },
         "weight": {
             "value": "-"
--- a/packs/data/equipment.db/five-feather-wreath-major.json
+++ b/packs/data/equipment.db/five-feather-wreath-major.json
@@ -50,7 +50,7 @@
             ]
         },
         "usage": {
-            "value": "held-in-one-hand"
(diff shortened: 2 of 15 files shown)


TikaelSol:
Has anyone looked at this? Kind of hit at an awkward time for me work wise so I'm just now getting to it.

Would we want spellhearts as toggles? I think most of the weapons and some of the armor things are best done as effects since they have a set duration of `until the end of your next turn` which is best tracked by effects. For the constant armor benefits I'd like a way to just ask where it's affixed, which is probably best done waiting for a more general solution to affixing. We can use choiceset, ask if attaching it to armor or weapon, but the choice won't proc on equip only on creation and there may be some strangeness if you tried to equip the same type of spellheart to a weapon and armor forcing the roll options to proc both bonuses on both items.

dmarcuse:
Now that you mention it, effects probably make more sense. I used toggles because that's how I have it set up for my PFS characters so that I can import the actor to any world and have it work without also copying effects over, but once this is implemented in the system itself that's moot.

The effects when affixed to armor only apply after casting a spell rather than constantly, so I don't think the spellhearts should do anything automatically when equipped (but a way to indicate that something is affixed would still be nice in general).

TikaelSol:
Closing this PR, it will conflict with v10 changes anyways.

PR #2780, opened 2022-06-30
Standardize Ability Scores Header and use for Feat section

CarlosFdez:
image


CarlosFdez:
Willing to have a better selector than h3.header, but <header> might need some refactoring since global styles are involved. Is it worth it.

PR #2783, opened 2022-06-30
Add Backgrounds from Pathfinder Society Year 4 Blog

avagdumortesin:
Also updated OGL to include the blog post.


SpartanCPA:
Consolidate the backgrounds that give a choice into one background with a ChoiceSet. Refer to the Blood Lords backgrounds for an example.

avagdumortesin:
@SpartanCPA done; if there's a way to take the response from the `ChoiceSet` to a `GrantItem` to another `ChoiceSet`, I can fix `Friend of Greensleeves` to automatically provide Assurance with that skill. I'm not seeing any documentation on how to use `GrantItem` to set the choice on Assurance, though.

PR #2789, opened 2022-07-01
Fix roll twice api option for character strikes, add for skills/saves

CarlosFdez:
It doesn't seem like NPC's types include it. Could be added, but that's not a fix I think

### DIFF
--- a/src/module/actor/character/index.ts
+++ b/src/module/actor/character/index.ts
@@ -1895,7 +1895,8 @@ class CharacterPF2e extends CreaturePF2e {
                     }
 
                     const item = context.self.item;
-                    const rollTwice = extractRollTwice(synthetics.rollTwice, selectors, finalRollOptions);
+                    const rollTwice =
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Unlike most options, I figured it'd be better if roll twice REs are prioritized over the option. I could reverse it if need be.

src/module/actor/character/index.ts
```diff
@@ -1895,7 +1895,8 @@ class CharacterPF2e extends CreaturePF2e {
                     }
 
                     const item = context.self.item;
-                    const rollTwice = extractRollTwice(synthetics.rollTwice, selectors, finalRollOptions);
+                    const rollTwice =
+                        extractRollTwice(synthetics.rollTwice, selectors, finalRollOptions) || args.rollTwice;
```
rsimoes, line 1899:
```suggestion
                        extractRollTwice(synthetics.rollTwice, selectors, finalRollOptions) ?? args.rollTwice;
```

PR #2799, opened 2022-07-01
Errata of the Blast lance (greater)

rectulo:
Level 4 -> 8
Price 100 -> 500 gp
fire damage 2d8 -> 3d8
page reference deleted


SpartanCPA (approved):

stwlam:
Let's not call PRs "errata" except to indicate that they come from errata. An error hopefully isn't being introduced with this PR, after all.

PR #2814, opened 2022-07-02
Add an effect for Aid bonus

SpartanCPA:


SpartanCPA:
We don't really have a place to put "Basic Action Effects" so I stuck it in Feature Effects

PR #2824, opened 2022-07-03
B3 - Review forest dragon spellcasters

MSAbaddon:
image

Read as written only the prepared spells are cumulative, but not the rituals. That's why I removed plant growth from the adult dragon and didn't give any rituals to the ancient dragon at all.
That said I'm also fine re-adding them, just leave a comment here.


SpartanCPA:
Re-add the rituals

MSAbaddon:
On the other hand if you take a look at the sea dragon, this might actually be inteded. There, the rituals for each dragon are explicitly mentioned - even if a younger version also has that ritual.

MSAbaddon:
Done

SpartanCPA (approved):

PR #2841, opened 2022-07-04
Fix sheet breakage with long deity names

CarlosFdez:
image


stwlam:
Can it be wise to whether the class depends on it (granted it)?

stwlam:
And this might as well be a context menu for all four of the thingies there.

CarlosFdez:
Removing deities blocked until proper context menus exist, replaced with only the sheet fixes.

PR #2845, opened 2022-07-04
Wipe URL sources from items

stwlam:


TMunsie (commented):
I don't have an issue with this, although the whole spell should probably be replaced and PFS season 1 should be avoided while I work on it.

stwlam:
> I don't have an issue with this, although the whole spell should probably be replaced and PFS season 1 should be avoided while I work on it.

I'll exclude season 1

avagdumortesin:
I can go through the list of NPCs here and change the spells over.

TikaelSol:
In a dumb twist this is actually how I've been tracking which are the oldest bits of data entry to do. All the NPCs with these URLs still on them need brushed up.

avagdumortesin:
> In a dumb twist this is actually how I've been tracking which are the oldest bits of data entry to do. All the NPCs with these URLs still on them need brushed up.

I can do the full audit of these; stick to max of 5/PR?

TikaelSol:
B2 is being tackled by Abaddon, Spartan was working on some AoE stuff but I don't think we've really done that in earnest. We should brushup all of AoE and these NPCs will get fixed along the way. If you want to start at book 6 and move back, 4-5 per PR to keep the diffs reasonable and keep potential merge conflicts down to a minimum that would work.

SpartanCPA:
> B2 is being tackled by Abaddon

Bestiary 3 actually

stwlam:
> In a dumb twist this is actually how I've been tracking which are the oldest bits of data entry to do. All the NPCs with these URLs still on them need brushed up.

You'll have this commit as a reference. :o)

CarlosFdez:
Should probably be kept to compendiums that start with `Compendium.pf2e.` I think it'd be ok to make that change after merging this though if you'd prefer?

PR #2847, opened 2022-07-04
Fix feat double summary rendering

CarlosFdez:
Only find pre-existing summaries among direct children


src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -49,23 +49,23 @@ export class ItemSummaryRendererPF2e<AType extends ActorPF2e> {
...
+            const $summary = $element.children(".item-summary");
             if (options.instant) {
                 $summary.hide().empty();
             } else {
                 $summary.slideUp(200, () => $summary.hide().empty());
             }
         } else {
             const $summary = (() => {
-                const $existing = $li.find(".item-summary");
+                const $existing = $element.children(".item-summary");
```
CarlosFdez, line 61: This is the only line with the fix, everything else is renaming a variable

PR #2854, opened 2022-07-05
Condense effects styling and use non-svg header for PC effects

CarlosFdez:
chrome_2022-07-05_06-06-23


CarlosFdez:
I'm considering doing the alternating color thing feats do. I'm also tempted on making new CSS variables only for the character sheet for things like `--font-item-name` so we don't have so many differing font families depending on where we are in the sheet.

PR #2855, opened 2022-07-05
Fix deity and society boon creation

CarlosFdez:
I suspect this broke when the ability to add normal feats was removed.

I'd like to save rewriting this section for at least V10.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -1025,7 +1025,7 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
     }
 
     /** Handle creating a new Owned Item for the actor using initial data defined in the HTML dataset */
-    private onClickCreateItem(event: JQuery.ClickEvent) {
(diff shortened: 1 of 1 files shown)


stwlam:
Can we just remove deity boon creation from sheets?

CarlosFdez:
I'd have to know what happens in actual play, but I do like to give custom ones, and custom ones don't really make any sense to add to world items to then add to the sheet.

If we're gonna be removing the ability to add deity boons and society boons, pardon me for the slippery slope, but I think that might apply to everything. We might as well remove the ability to create anything (spells/physical items/etc) from the sheet. With regular feats there was an argument to be made that adding a basic feat the normal way was useless and it kept being added to bonus feats, but for this not really.

...and actually I'm tempted on reintroducing create feat if we can solve the bleed into bonus feats problem, but I'm not rushing to do that.

stwlam:
Good point, let's remove the button for boons, too

CarlosFdez:
likely will just be a new item type or re-use effects

PR #2871, opened 2022-07-06
Update original spell chat card after selecting a variant of that spell

In3luki:
Replace the chat message content with the variant spell content instead of adding a new variant spell chat card.

Recording 2022-07-06 at 21 11 17

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -70,7 +70,19 @@ export const ChatCards = {
                 else if (action === "spellDamage") spell?.rollDamage(event);
                 else if (action === "spellCounteract") spell?.rollCounteract(event);
                 else if (action === "spellTemplate") spell?.placeTemplate();
-                else if (action === "selectVariant") (await spell?.variantPrompt())?.toMessage();
+                else if (action === "selectVariant") {
+                    const variantSpell = await spell?.variantPrompt();
+                    if (variantSpell) {
+                        const spellLvl = $html.find<HTMLDivElement>("div.chat-card").attr("data-spell-lvl");
+                        const variantMessage = await variantSpell.toMessage(undefined, {
+                            create: false,
+                            data: { spellLvl },
+                        });
+                        if (variantMessage) {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I'd like input from @stwlam @SpartanCPA and @TikaelSol . Especially the latter two once they actually start trying out spell variants.

stwlam:
The menu is tiny. Can the buttons get labels? What happens if there are variants with no change in number of actions?

In3luki:
> The menu is tiny. Can the buttons get labels? What happens if there are variants with no change in number of actions?

It kind of only works for variants with a different number of actions right now. In the short term I can change the buttons to only show an action icon if the number is different and default to the normal icon. The label can be the name of the variant spell but then each variant has to have a different name to tell them apart.
Long term this need its own FormApplication to deal with other types of variants and maybe show a preview of the variant spell chat card.

CarlosFdez:
data entry isn't really commenting it, so I guess we gotta decide what sort of content editing they're going to be doing. Talked it out a bit and it seems instead of a dialog, maybe do this instead:

1) Remove [Select] button, put it back to the [Cast] button. Make it spit out the same chat thing the others do.
2) Put selection choices for the variant inline the chat itself. We can handle more complex multi-selection cases later.
3) Lock down description editing for the variant. Read only on the override variant. No data curation or separate translations needed.

SpartanCPA:
> Since data entry can't really comment, we gotta decide what they're going to be doing.

I haven't gotten a chance to really test it out yet. It's been a busy work week. First comment I have from the gif is "can the variant spell card remove the text for the action costs that wasn't selected?" Which is more a DE thing than a comment on the implementation.

I agree with shark that the buttons could use being a bit bigger/more than just the action icons, but I don't have much more comment from a UI standpoint.

CarlosFdez:
as for why it might not be desirable, localization and editorializing. For heal it'd be easy to remove the other versions of the spell, but look at scorching ray's description.

In3luki:
See https://github.com/foundryvtt/pf2e/pull/2890

PR #2807, opened 2022-07-02
Add ChoiceSet Rules Elements for Backgrounds that have multiple skill options

avagdumortesin:

PR #2877, opened 2022-07-07
Trim down ChoiceSet flags

TikaelSol:
To make shark happy.

### DIFF
--- a/packs/data/bestiary-effects.db/effect-adaptive-strike.json
+++ b/packs/data/bestiary-effects.db/effect-adaptive-strike.json
@@ -30,7 +30,7 @@
                         "value": "silver"
                     }
                 ],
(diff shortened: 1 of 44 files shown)


packs/data/bestiary-effects.db/effect-adaptive-strike.json
```diff
@@ -30,7 +30,7 @@
                         "value": "silver"
                     }
                 ],
-                "flag": "strikeChoice",
```
stwlam, line 33: What's this supposed to do?

avagdumortesin:
I'm surprised that it's not showing any conflicts; I already updated all of the `skillChoice` options, and it was merged earlier. https://github.com/foundryvtt/pf2e/pull/2807

PR #2890, opened 2022-07-07
Update spell variant implementation

In3luki:
- Remove the SpellVariantPrompt class.
- Show spell variants as inline buttons on the spell chat card.
- Update the existing chat card with the selected variant data.
- Add a button to variant spell chat cards that allows the selection of another variant.
- Disable editing the description of spell variants.
- Remove spell chat buttons for users that are not the owner of that spell.

Recording 2022-07-07 at 23 15 14

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -70,7 +70,34 @@ export const ChatCards = {
                 else if (action === "spellDamage") spell?.rollDamage(event);
                 else if (action === "spellCounteract") spell?.rollCounteract(event);
                 else if (action === "spellTemplate") spell?.placeTemplate();
-                else if (action === "selectVariant") (await spell?.variantPrompt())?.toMessage();
+                else if (action === "selectVariant") {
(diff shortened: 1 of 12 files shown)


src/module/chat-message/listeners/cards.ts
```diff
@@ -70,7 +70,45 @@ export const ChatCards = {
...
+                        const overlayIds = overlayIdString.split(",").map((id) => id.trim());
+                        const variantSpell = spell?.loadVariant({ overlayIds, castLevel: spellLvl });
+                        if (variantSpell) {
+                            const variantMessage = await variantSpell.toMessage(undefined, {
+                                create: false,
+                                data: { spellLvl },
+                            });
+                            if (variantMessage) {
+                                const messageSource = variantMessage.toObject();
+                                messageSource.flags = mergeObject(
```
CarlosFdez, line 87: Is this not already handled by Spell#toMessage()? It injects some flags if its a variant.

In3luki: It wasn't working without it and I forgot to investigate further. I'll take another look.

PR #2901, opened 2022-07-08
Show carried bulk for loot actors

CarlosFdez:
Foundry_Virtual_Tabletop_2022-07-08_17-46-34

### DIFF
--- a/src/styles/actor/loot/_index.scss
+++ b/src/styles/actor/loot/_index.scss
@@ -63,6 +63,13 @@
             width: calc(100% - 6px);
             background-color: var(--tertiary);
         }
+
+        .bulk {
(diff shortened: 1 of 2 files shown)


CarlosFdez:
If we add it to the bottom of the inventory list, it'd be with the contingent that max bulk will be configurable in the future, and that it'd show unlimited max bulk for now

stwlam:
Why show the capacity bar at the bottom, though?

CarlosFdez:
Cause its where PCs show it. If it has the concept of a max bulk, but it defaults to "unlimited", then we can do a progress bar similar to what PCs have if its a non-merchant actor. If not, then where I have it now is best I think.

PR #2902, opened 2022-07-09
Add Deity Boons/Curses from the Gods and Magic Web Supplement 

LebombJames:
Made as effects, as @stwlam told me

### DIFF
--- a/packs/data/boons-and-curses.db/achaekek-major-boon.json
+++ b/packs/data/boons-and-curses.db/achaekek-major-boon.json
@@ -1,0 +1,38 @@
+{
+    "_id": "eFwCopyvVam6GiCT",
+    "data": {
+        "badge": null,
+        "description": {
+            "value": "<p><strong>Major Boon:</strong> Death lives in your very blade. You gain the @Compendium[pf2e.classfeatures.Sneak Attack]{Sneak Attack} rogue class feature, dealing 3d6 precision damage. if you already have that class feature, you increase your sneak attack damage by 3d6.</p>"
+        },
+        "duration": {
+            "expiry": "turn-start",
+            "sustained": false,
+            "unit": "unlimited",
+            "value": -1
+        },
+        "level": {
+            "value": 0
+        },
+        "rules": [],
+        "source": {
+            "value": ""
+        },
+        "start": {
+            "initiative": null,
+            "value": 0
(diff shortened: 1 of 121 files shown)


packs/data/boons-and-curses.db/achaekek-major-boon.json
```diff
@@ -0,0 +1,38 @@
...
+            "sustained": false,
+            "unit": "unlimited",
+            "value": -1
+        },
+        "level": {
+            "value": 0
+        },
+        "rules": [],
+        "source": {
+            "value": ""
```
SpartanCPA, line 19:
```suggestion
            "value": "Pathfinder Loast Omens: Gods and Magic"
```
Make this change on every item in this MR

packs/data/boons-and-curses.db/achaekek-major-curse.json
```diff
@@ -0,0 +1,38 @@
+{
+    "_id": "GJdHcY56q2c3kSiA",
+    "data": {
+        "badge": null,
+        "description": {
+            "value": "<p><strong>Major Curse:</strong> The Mantis God doesn't waste his time toying with or tormenting those who truly anger him, or those who dare to consider themselves divine. He rips open a portal to your location, kills you, drags your soul to judgment in a way that prevents resurrection magic, and then leaves.</p>"
```
SpartanCPA, line 6: Does this even need an effect?

SpartanCPA:
None of these effects actually *do* anything? They all seem to be made just to act as player markers?

packs/data/boons-and-curses.db/achaekek-major-curse.json
LebombJames: Realistically, no, but I added it just for the sake of GM reference and completeness.

LebombJames:
> None of these effects actually _do_ anything? They all seem to be made just to act as player markers?

Ones that give bonuses/penalties have REs. Ones that let you take a different degree of success all the time are done, but not ones that apply once or once per day, as there's no way to let the user choose to use it or not, as far as I'm aware.

There are also some messy ones to automate, like being unable to benefit from circumstance/status bonuses or fortune effects, which I think technically may be possible, but I'm not exactly sure how

PR #2903, opened 2022-07-09
Handle ranged attacks with only max range, abort strikes if out of range

stwlam:


src/module/actor/npc/index.ts
```diff
@@ -607,6 +607,15 @@ class NPCPF2e extends CreaturePF2e {
                                 rollOptions.push("melee");
                             }
 
+                            // Check whether target is out of maximum range; abort early if so
+                            if (context.self.item.isRanged && typeof context.target?.distance === "number") {
+                                const maxRange = item.maxRange ?? 10;
```
CarlosFdez, line 612: Why 10? Just to have a number?

stwlam: Just to have one! Current data validation is loose, so one can say an attack is ranged but then not provide a range

CarlosFdez: there can probably be better handling, but the two that come to mind are extremes of each other so might as well give it a try and see what people say.

my thoughts are to either not restrict if not set, or extra restrict and block it entirely. Not sure which direction to lean.

stwlam: A default of 10 is already in place and has done a decent job of highlighting data problems

PR #2905, opened 2022-07-09
Remove edit/delete/add buttons for hazard actions/strikes in view mode

CarlosFdez:


static/templates/actors/hazard/sheet-new.html
```diff
@@ -80,7 +80,7 @@ <h3>{{localize "PF2E.HazardResetLabel"}}</h3>
             <div class="section-container">
                 <div class="section-header">
                     <h3>Actions</h3>
-                    {{#if owner}}
```
stwlam, line 83: If these `owner` checks are elsewhere, they should be replaced with checking `editable`

static/templates/actors/hazard/sheet-new.html
```diff
@@ -101,7 +101,7 @@ <h4><a>{{action.name}} {{{actionGlyph action.actionCost}}}</a></h4>
                                     </div>
                                     <div class="item-controls">
                                         <a class="item-chat chat" title="{{ localize 'PF2E.NPC.SendToChat'}}"><i class="fas fa-comment-alt"></i></a>
```
stwlam, line 103:
```suggestion
                                        <a class="item-chat chat" title="{{localize "PF2E.NPC.SendToChat"}}"><i class="fas fa-comment-alt"></i></a>
```

CarlosFdez: copy paste laziness strikes again

static/templates/actors/hazard/sheet-new.html
CarlosFdez: these are the last ones I think unless the editor params count

stwlam: It should just be ensured that nothing is editable when viewed from a locked compendium, on pain of server errors.

PR #2915, opened 2022-07-09
Make actions sortable in new hazard sheet, reactions go first

CarlosFdez:
image

### DIFF
--- a/src/module/actor/hazard/sheet-new.ts
+++ b/src/module/actor/hazard/sheet-new.ts
@@ -3,7 +3,7 @@ import { ActorSheetDataPF2e } from "@actor/sheet/data-types";
 import { SAVE_TYPES } from "@actor/values";
 import { ConsumablePF2e, SpellPF2e } from "@item";
 import { ItemDataPF2e } from "@item/data";
-import { ErrorPF2e, objectHasKey } from "@util";
+import { ErrorPF2e } from "@util";
 import { HazardPF2e } from ".";
 import { HazardSystemData } from "./data";
 import { HazardActionSheetData, HazardSaveSheetData, HazardSheetData } from "./types";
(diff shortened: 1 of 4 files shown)


CarlosFdez:
not 100% sure about this. Would we prefer to force an ordering? Reactions above actions? I'm about to do the same for the NPC sheet depending.

TMunsie:
> Not 100% sure about this. Would we prefer to force an ordering? Reactions above actions? I'm about to do the same for the NPC sheet depending.

Reaction and free action above other actions for hazards. Hazards only use those to start combat/activate for complex/simple hazards

CarlosFdez:
The main concern is, should we enforce the order in code, or do we just encourage that as a convention? Allowing sorting means that data entry can control what goes where.

We can separate reaction specifically, but I'm not sure about free actions. Can you give some examples?

CarlosFdez:
To clarify, the reason for the MR is actually to get rid of action buckets as a system, or minimize them. They're kinda in the way of us doing item activations or refactoring action costs.

CarlosFdez:
I'll update it to reactions, then "the rest", then strikes.

SpartanCPA:
> To clarify, the reason for the MR is actually to start getting rid of action buckets as a system, or minimize them. They're kinda in the way of us doing item activations or refactoring action costs.

Action Buckets as in the interaction/defensive/offensive divide? That might break our sorting script on extract that Graven made

CarlosFdez:
Action Buckets right now are actions, reactions, free, and passive, in that order.  Hazards enforce that order. NPCs instead enforce passive > free > reaction > action, however passive is listed separately and explicitly.

CarlosFdez:
We could do this if we must keep reactions separate 
image

PR #2919, opened 2022-07-10
Auto-adjust current hit points when changing PC level or adding ABC item

stwlam:
[auto-adjust-hp.webm](https://user-images.githubusercontent.com/106829671/178130896-a10e63f0-0c9d-412b-8d51-052f11abef9b.webm)

### DIFF
--- a/src/module/actor/character/index.ts
+++ b/src/module/actor/character/index.ts
@@ -1,6 +1,12 @@
 import { CreaturePF2e, FamiliarPF2e } from "@actor";
 import { Abilities, CreatureSpeeds, LabeledSpeed, MovementType, SkillAbbreviation } from "@actor/creature/data";
-import { AttackItem, AttackRollContext, StrikeRollContext, StrikeRollContextParams } from "@actor/creature/types";
+import {
+    AttackItem,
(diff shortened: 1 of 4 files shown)


src/module/actor/character/index.ts
```diff
@@ -2166,15 +2172,24 @@ class CharacterPF2e extends CreaturePF2e {
...
-        if (level?.value !== undefined) {
+        if (typeof level?.value === "number") {
             level.value = Math.clamped(Number(level.value) || 0, 0, 30);
+
+            // Adjust hit points if level is changing
+            const clone = this.clone(changed);
+            const hpMaxDifference = clone.hitPoints.max - this.hitPoints.max;
+            if (hpMaxDifference !== 0) {
+                options.allowHPOverage = true;
+                const newHitPoints = this.hitPoints.value + hpMaxDifference;
```
CarlosFdez, line 2190: this should probably be clamped to 1 hp if current hp is positive, and stay at 0 hp if its currently 0 hp.

CarlosFdez (changes requested):

CarlosFdez (approved):

PR #2920, opened 2022-07-10
Refactor feat group assignment into feat categories

CarlosFdez:
Should be easier to add new custom feat groups after this. However the way feats are assigned still bothers me, but I don't know a good way to handle that yet. If actor.items is sorted, I could do it in prepareActorData() and it might be cleaner?

After this, it should be easy to store bonus feat categories in actor flags or homebrew settings or whatever and construct them.

### DIFF
--- a/src/module/actor/character/index.ts
+++ b/src/module/actor/character/index.ts
@@ -45,7 +45,6 @@ import { AncestryBackgroundClassManager } from "@item/abc/manager";
 import { ActionTrait } from "@item/action/data";
 import { ARMOR_CATEGORIES } from "@item/armor/data";
 import { ItemSourcePF2e, ItemType, PhysicalItemSource } from "@item/data";
-import { ItemGrantData } from "@item/data/base";
 import { ItemCarryType } from "@item/physical/data";
 import { getPropertyRunes, getPropertySlots, getResiliencyBonus } from "@item/runes";
 import { MAGIC_TRADITIONS } from "@item/spell/values";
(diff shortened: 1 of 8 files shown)


CarlosFdez:
This should be ready now. I can keep building on top of this, but as is this should be a halfway decent refactor.

The only API gotcha is that now its character.feats.get("class").feats and that double feats is kinda silly. I'm thinking of renaming the second instance of feats to contents.

src/module/actor/character/index.ts
```diff
@@ -1242,190 +1200,104 @@ class CharacterPF2e extends CreaturePF2e {
...
+        // Add campaign feats if enabled
+        if (game.settings.get("pf2e", "campaignFeats")) {
+            this.feats.createCategory({ id: "campaign", label: "PF2E.FeatCampaignHeader" });
+        }
+
+        // Add background skill feat slot
         const background = this.background;
         if (background && Object.keys(background.data.data.items).length > 0) {
-            this.featGroups.skill?.feats.unshift({
+            this.feats.get("skill")?.feats.unshift({
```
stwlam, line 1281: Can the `get` method signature get overloaded to make ones like these guaranteed to return an array?

CarlosFdez: We could, but if we do it might be wise to move the creation of the base categories to the constructor to guarantee this is the case. That would completely get rid of CharacterPF2e#prepareFeats() but that might be fine and probably a better result anyways.

stwlam: Creating the core categories in the constructor would definitely be a necessary step.

CarlosFdez: Should be good to go, prepareFeats() is still there of course cause there's other things that do still need doing there right now.

PR #2927, opened 2022-07-10
Fix spell variants check in item summary renderer

In3luki:


src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -195,7 +195,7 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                 }
 
                 if (actor instanceof CharacterPF2e) {
-                    if (chatData.variants) {
+                    if ((chatData.variants as Array<unknown>).length) {
```
stwlam, line 198: Is that always going to be present for all items?

In3luki: Currently yes. Alternatively I can set it to undefined if no variants exist and leave the check as it is.

stwlam: That sounds better

In3luki: Done

PR #2933, opened 2022-07-10
Add target item traits as roll options for the Repair action

nikolaj-a:
For the Repair basic action, add all traits of the target item as roll options with the format "target:trait:xyz". This will allow situational modifiers to predicate on those traits.

### DIFF
--- a/src/module/system/action-macros/crafting/repair.ts
+++ b/src/module/system/action-macros/crafting/repair.ts
@@ -37,6 +37,8 @@ async function repair(options: RepairActionOptions) {
             return;
         })();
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This should be done instead via feeding the item to the roll function. That will automatically spawn item:trait:x flags. You'll see them for example if you do a spell attack roll with a spell and check the roll inspector.
image

nikolaj-a:
It gives the wrong namespace for the traits. The `item:` namespace should refer to the repair kit instead of the target item.

CarlosFdez:
That's a good point.

PR #2940, opened 2022-07-11
Update Sorching Ray with spell variants

In3luki:


packs/data/spells.db/scorching-ray.json
```diff
@@ -58,6 +42,76 @@
...
+                                "type": {
+                                    "categories": [],
+                                    "subtype": "",
+                                    "value": "fire"
+                                },
+                                "value": "4d6"
+                            }
+                        }
+                    },
+                    "heightenedLevel": {},
```
CarlosFdez, line 62: What is this doing here? That shouldn't exist anymore in top level spell data. Is something slipping it in?

In3luki: I've just added a new heightened entry on the variants. Seems like it was added there. I'll take a look.

PR #2942, opened 2022-07-11
Fix SpellPF2e#loadVariants adding unused heightenedLevel data

In3luki:


CarlosFdez:
Why is that being set? heightenedLevel is a property in spell.data.data.location, not in spell.data.data

In3luki:
That was already there so I assumed it was needed. Let me check what happens if I remove it completely.

In3luki:
Everything still works fine without it.

CarlosFdez:
Might have been something that was accidentally left in. Might be worth investigating that fixed heightening spells work correctly right now.

PR #2952, opened 2022-07-11
Fix spell variant buttons falling back to myster man image with invalid actions

In3luki:
Also a minor CSS fix for buttons with no image


CarlosFdez:
Eventually we need to consider fixing actionCost for spells and then making a getActionImage handlebars helper like we a have a getActionGlyph one.

PR #2953, opened 2022-07-11
Expand choice set owned items to non-weapons

CarlosFdez:
Existing ones should still pick weapons, but the types array will allow non-spells. There is a migration, but it doesn't affect our compendium at all.


CarlosFdez:
Main gotcha right now is that failed validation doesn't block effect creation. Having no options to select does block creation, but having options and not picking doesn't. I have no idea how to block creation outside of throwing an exception, and I couldn't quite find how having no options to select is blocking creation. At this point, I can only imagine its because the async promise is never resolved.

PR #2969, opened 2022-07-13
Expand filter parameters for game.pf2e.compendiumBrowser.openTab()

chepseskhaf:
I would like to use the game.pf2e.compendiumBrowser.openTab function in macros to open the compendium browser with feats or spells of a particular level but it does not currently take a "minLevel" parameter.

To keep the function backwards compatible, I added the parameter in the last position with a default value. I would be happy to change it to something like levelFilter: { min: number; max: number; } if need be.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -736,10 +736,27 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
     private async onClickBrowseFeatCompendia(event: JQuery.ClickEvent): Promise<void> {
         const maxLevel = Number($(event.currentTarget).attr("data-level")) || this.actor.level;
         const button: HTMLElement = event.currentTarget;
-        const filter = button.dataset.filter?.split(",").filter((f) => !!f) ?? [];
-        if (filter.includes("feattype-general")) filter.push("feattype-skill");
+        const checkboxesFilterCodes = button.dataset.filter?.split(",").filter((f) => !!f) ?? [];
+        if (checkboxesFilterCodes.includes("feattype-general")) checkboxesFilterCodes.push("feattype-skill");
 
-        await game.pf2e.compendiumBrowser.openTab("feat", filter, maxLevel);
+        const filter: Record<string, string[]> = {};
+        for (const filterCode of checkboxesFilterCodes) {
+            const splitValues = filterCode.split("-");
+            if (splitValues.length !== 2) {
+                console.error(
+                    `Invalid filter value for opening the compendium browser:\n'${JSON.stringify(
+                        checkboxesFilterCodes
+                    )}'`
+                );
+                return;
+            }
+
+            const [filterType, value] = splitValues;
+            const filterCategory = filter[filterType] ?? (filter[filterType] = []);
+            filterCategory.push(value);
+        }
+
+        await game.pf2e.compendiumBrowser.openTab("feat", Object.assign(filter, { levelRange: { max: maxLevel } }));
     }
 
     /** Handle changing of proficiency-rank via dropdown */
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -566,12 +566,29 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
         }
     }
 
-    private onClickBrowseEquipmentCompendia(event: JQuery.ClickEvent<HTMLElement>) {
-        const filter: string[] = [$(event.currentTarget).attr("data-filter")].filter(
+    private async onClickBrowseEquipmentCompendia(event: JQuery.ClickEvent<HTMLElement>): Promise<void> {
+        const checkboxesFilterCodes: string[] = [$(event.currentTarget).attr("data-filter")].filter(
             (element): element is string => !!element
         );
-        console.debug(`Filtering on: ${filter}`);
-        game.pf2e.compendiumBrowser.openTab("equipment", filter);
+
+        const filter: Record<string, string[]> = {};
+        for (const filterCode of checkboxesFilterCodes) {
+            const splitValues = filterCode.split("-");
+            if (splitValues.length !== 2) {
+                console.error(
+                    `Invalid filter value for opening the compendium browser:\n'${JSON.stringify(
+                        checkboxesFilterCodes
+                    )}'`
+                );
+                return;
+            }
+
+            const [filterType, value] = splitValues;
+            const filterCategory = filter[filterType] ?? (filter[filterType] = []);
+            filterCategory.push(value);
+        }
+
+        await game.pf2e.compendiumBrowser.openTab("equipment", filter);
     }
 
     protected override _canDragStart(selector: string): boolean {
--- a/src/module/apps/compendium-browser/index.ts
+++ b/src/module/apps/compendium-browser/index.ts
@@ -5,11 +5,23 @@ import { ErrorPF2e, isObject, objectHasKey } from "@util";
 import { LocalizePF2e } from "@system/localize";
 import { BrowserTab } from "./tabs";
 import { TabData, PackInfo, TabName, TabType, SortDirection } from "./data";
-import { CheckboxData, RangesData } from "./tabs/data";
+import {
+    BaseFilterData,
+    CheckboxData,
+    InitialActionFilters,
+    InitialBestiaryFilters,
+    InitialEquipmentFilters,
+    InitialFeatFilters,
+    InitialFilters,
+    InitialHazardFilters,
+    InitialSpellFilters,
+    RangesData,
+} from "./tabs/data";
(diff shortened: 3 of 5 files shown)


CarlosFdez:
How much would expect to break if we change it to be an options parameter? If possible, might be better to break sooner rather than later if we want to toss in configuration options.

chepseskhaf:
I don't think that there is an easy way to know since changing the parameters would mainly impact other Foundry modules that try to open the compendium browser with arbitrary filters and (possibly) user macros.

CarlosFdez:
Personally, I'm of the mind we should wait till V10 and then break all compatibility and use object params. Use the fresh start to make this method much nicer.

CarlosFdez:
Compatibility can now be broken, we start the v10 train. Go nuts!

chepseskhaf:
Ok. I did go nuts.

I tried to cover every filter currently available in the compendium browser while trying to derive most data types from the existing ones.

**Usage example in macros :**

Show every uncommon level 3 spell taking 3 actions, sorted by descending name:
```
game.pf2e.compendiumBrowser.openTab("spell", {
  selects: {
    timefilter: "3"
  },
  checkboxes: {
    level: ["3"],
    rarity: ["uncommon"]
  },
  orderBy: "name",
  orderDirection: "desc"
});
```

Show every common alchemical item of level 5 or more costing 5000 gp or less, sorted by price:
```
game.pf2e.compendiumBrowser.openTab("equipment", {
  checkboxes: {
    rarity: ["common"]
  },
  multiselects: {
    traits: ["alchemical"]
  },
  ranges: {
    price: { max: "3000gp" }
  },
  sliders: {
    level: { min: 5 }
  },
  orderBy: "price"
});
```

CarlosFdez:
Would need @stwlam's input, but personal preference, not a fan of splitting by control type. Splitting by just actual field might be better.

So like, maybe just
```ts
game.pf2e.compendiumBrowser.openTab("equipment", {
  rarity: ["common"],
  traits: ["alchemical"],
  price: { max: "3000gp" },
  level: { min: 5 },
  orderBy: "price",
});
```

I'd go one step further and try to have rarity as part of traits, since rarity is a trait and we can sort out the difference in the browser, but that's personal preference.

chepseskhaf:
I started doing just that but then realized it would have been much harder to present a common and strongly typed method signature for openTab : the equipment tab uses a multiselect for traits whereas others use checkboxes and the spells tab uses checkboxes for levels whereas others use a slider. I also tried to stay close to the existing data type to facilitate future changes.

I would keep the new openTab() function as is but add some wrapper functions (like OpenSpellTab that already exists) to convert from the "user-friendly" data types to the "technical" one.

Also, rarity might be a trait in the rulebooks but it is currently handled separately in all data entries. I think it makes sense to keep it separate than to artificially merge it with an other parameter then having to extract it with a special case.

CarlosFdez:
I'll have to double check, but give it was able to handle a list of string properties based on {key}-{value} joins, it doesn't seem like the type of the form element should affect anything (it was able to resolve the param regardless of the form element). A few might be tricky (probably level), but it doesn't seem like traits and a few others were at all.

I will need to double check to be 100% certain though.

chepseskhaf:
Sure, representing all filters with a string code is a possible solution.

Previously, filters could only apply to checkboxes and were built like this : [category]-[checkboxValue].

We could have a similar system and check the category name in each control type instead of just checkboxes but that would mean having almost completely type-less parameter : everything would be of type "Record<string, string | number | { min?: number | string; max?: number | string }>".

Since this is primarily aimed to be used outside of a TypeScript environment, having a strong typing may be overkill.

Up to you.

CarlosFdez:
I think those that are a range (like level) can be separate from those that are "checkboxes", but given that the tagify components have a whitelist, they're functionally no different in the data model compared to checkboxes. They're just another way to represent the same data.

This is an API after all, and we don't want to leak implementation details to the outside world. I don't really mind if traits are a different array than categories (that sounds like a great idea) nor that rarity is separate from traits (can go either way), but leaking to the outside world whether something is a range or a slider or a checkbox is really a no go.

I guess also make sure you don't break the feat browse filtering. Clicking browse on ancestry feats does filtering based on your ancestry for example.

Strangley-Browne:
In the compendium browser would it be possible to get a new filter for Feats For  [Trained in...], [Expert in...], [Master in...], [... Dedication], [Basic..], [Initial...], etc
Just so with can filter on Prerequisites?

TMunsie:
Prerequisites are presently text so this is very unlikely until we have a better backend and a lot of data entry work

chepseskhaf:
Now, the names of the filter parameters are the same as the ones in the interfaces based on BaseFilterData, except for sliders and ranges. These parameters are names "levelRange" and "priceRange". This is necessary because there is already a "level" parameter for a checkboxes filter in the "spells" tab.

**Same examples :**

Show every uncommon level 3 spell taking 3 actions, sorted by descending name:

```
game.pf2e.compendiumBrowser.openTab("spell", {
  timefilter: "3",
  level: ["3"],
  rarity: ["uncommon"],
  orderBy: "name",
  orderDirection: "desc"
});
```

Show every common alchemical item of level 5 or more costing 3000 gp or less, sorted by price:

```
game.pf2e.compendiumBrowser.openTab("equipment", {
  rarity: ["common"],
  traits: ["alchemical"],
  priceRange: { max: "30000sp" },
  levelRange: { min: 5 },
  orderBy: "price"
});
```

src/module/apps/compendium-browser/index.ts
```diff
@@ -272,44 +296,210 @@ class CompendiumBrowser extends Application {
...
+            else {
+                console.warn(`'${filterType}' is not a valid filter for tab '${currentTab.tabName}'.`);
+            }
         }
 
-        this.render(true);
+        this.initialFilter = {};
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
CarlosFdez, line 488: avoid disabling the linter for any. Instead leave value as type unknown and take advantage of typeof and the isObject() helper function.

In3luki (changes requested):

src/module/apps/compendium-browser/tabs/data.ts
```diff
@@ -98,6 +98,81 @@ interface SpellFilters extends BaseFilterData {
...
 
+// Models used for initializing filters
+interface BaseInitialFilters {
+    searchText?: string;
+    orderBy?: SortByOption;
+    orderDirection?: SortDirection;
+}
+
+interface InitialActionFilters extends BaseInitialFilters {
+    traits?: Array<string>;
```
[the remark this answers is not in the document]
chepseskhaf: You can add an eslint rule to enforce this.

CarlosFdez:
ESLint rule has been merged in

src/module/actor/character/sheet.ts
```diff
@@ -736,10 +736,27 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
     private async onClickBrowseFeatCompendia(event: JQuery.ClickEvent): Promise<void> {
         const maxLevel = Number($(event.currentTarget).attr("data-level")) || this.actor.level;
         const button: HTMLElement = event.currentTarget;
-        const filter = button.dataset.filter?.split(",").filter((f) => !!f) ?? [];
-        if (filter.includes("feattype-general")) filter.push("feattype-skill");
+        const checkboxesFilterCodes = button.dataset.filter?.split(",").filter((f) => !!f) ?? [];
+        if (checkboxesFilterCodes.includes("feattype-general")) checkboxesFilterCodes.push("feattype-skill");
+
+        const filter: Record<string, Array<string>> = {};
```
In3luki, line 742:
```suggestion
        const filter: Record<string, string[]> = {};
```

src/module/actor/sheet/base.ts
```diff
@@ -578,12 +578,29 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
...
-    private onClickBrowseEquipmentCompendia(event: JQuery.ClickEvent<HTMLElement>) {
-        const filter: string[] = [$(event.currentTarget).attr("data-filter")].filter(
+    private async onClickBrowseEquipmentCompendia(event: JQuery.ClickEvent<HTMLElement>): Promise<void> {
+        const checkboxesFilterCodes: string[] = [$(event.currentTarget).attr("data-filter")].filter(
             (element): element is string => !!element
         );
-        console.debug(`Filtering on: ${filter}`);
-        game.pf2e.compendiumBrowser.openTab("equipment", filter);
+
+        const filter: Record<string, Array<string>> = {};
```
In3luki, line 586:
```suggestion
        const filter: Record<string, string[]> = {};
```

src/module/actor/character/sheet.ts
In3luki: We've added the eslint rule so these need to be changed too.

In3luki (changes requested):

CarlosFdez (approved):

PR #2972, opened 2022-07-14
Report bulk differences in the inventory

CarlosFdez:
image

### DIFF
--- a/src/module/actor/data/size.ts
+++ b/src/module/actor/data/size.ts
@@ -133,4 +133,8 @@ export class ActorSizePF2e {
         this.length = newSpace.length;
         this.width = newSpace.width;
     }
+
+    toString() {
+        return game.i18n.localize(CONFIG.PF2E.actorSizes[this.value]);
+    }
 }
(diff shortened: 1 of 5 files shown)


stwlam:
What do you think about adjusting the name to make it more immediately clear? E.g., "Large Mace" when on a medium creature (or always?).

CarlosFdez:
Is that possible to do as part of the current actor/item autonaming thing? That might be best to append (large) after the name in greyed out text with a smaller font.

stwlam:
> Is that possible to do as part of the current actor/item autonaming thing? That might be best to append (large) after the name in greyed out text with a smaller font.

Yeah, it's just a matter of changing the name during data preparation. But if you prefer to append a parathetical to the name separately and formatted differently, it may be better to just perform that in the item-line template.

CarlosFdez:
What if instead of the asterisk and a parenthetical after the name, we add something like (L) or (T) or (H) next to the bulk value?

Parenthetical might look weird for like consumables with charges. Though there's also a part of me that thinks that consumables and treasure shouldn't receive bulk adjustments at all.

ebuckle:
I think a parenthetical would look odd. Especially for an object of light bulk: `L (L)`

PR #2979, opened 2022-07-15
Ensure unique actors are checked for aura effects

stwlam:
When multiple tokens are within an aura, it may be the case the more than one refer back to a single actor. This won't solve every issue with auras hitting an actor twice, but it eliminates one source of the problem.


CarlosFdez:
Does this guarantee that if there's one in the aura and one that isn't, its always treated as they're in it? It'll be relevant for mirror thaumaturge implement I believe.

stwlam:
> Does this guarantee that if there's one in the aura and one that isn't, its always treated as they're in it? It'll be relevant for mirror thaumaturge implement I believe.

It doesn't guarantee anything of the sort, but that isn't relevant to the PR.

PR #2980, opened 2022-07-15
Allow multiple school traits to be selected for spells

avagdumortesin:
This will fail due to errors in the following locations:
```
src/module/item/spellcasting-entry/trick.ts:26:14
src/module/migration/migrations/621-remove-config-spell-schools.ts:33:19
src/module/migration/migrations/621-remove-config-spell-schools.ts:37:19
src/module/system/settings/homebrew/cleanup-migration.ts:98:25
src/module/system/settings/homebrew/cleanup-migration.ts:98:59
```

I could probably fumble around and find a way to fix `trick.ts`, but I'm not sure how I would handle the migrations.

### DIFF
--- a/src/module/item/sheet/data-types.ts
+++ b/src/module/item/sheet/data-types.ts
@@ -101,7 +101,7 @@ export interface SpellSheetData extends ItemSheetDataPF2e<SpellPF2e> {
         sort: number;
         actions: string;
     }[];
-    magicSchools: SheetOptions;
+    magicSchools: ConfigPF2e["PF2E"]["magicSchools"];
     spellCategories: ConfigPF2e["PF2E"]["spellCategories"];
     spellLevels: ConfigPF2e["PF2E"]["spellLevels"];
     spellTypes: ConfigPF2e["PF2E"]["spellTypes"];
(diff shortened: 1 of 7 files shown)


avagdumortesin:
Looks like the last commit did fix the error with `trick.ts`; still not sure how to go about messing with the migrations, however. Any pointers, @stwlam or @CarlosFdez?

CarlosFdez:
What is this for? If this is for a ritual, might be best to not bother. Otherwise, it might be best to derive schools from traits during data preparation. Because otherwise, you're going to need to make some changes in the spell sheet as well, which it doesn't seem like you did either.

But even then, instead of ValuesList<MagicSchool>, it should just be MagicSchool[].

avagdumortesin:
There are multiple spells now that have two schools as their trait, and there is no way to select multiple schools. The goal is to make the Schools into a checkbox system on the sheet like Traits and Traditions.

Maybe I misunderstood what line 58 in `src/module/item/spell/sheet.ts` does; is this not the spell sheet?

CarlosFdez:
spell-details.html as well. What spells are these? Send them my way privately.

avagdumortesin:
Superseded by https://github.com/foundryvtt/pf2e/pull/2986

CarlosFdez:
Its not superceded, shark backed out of the downgrade them all to traits and my time was wasted.

It'll still need some code in compendium browser (more or less what I did) to work, but past that nothing really.

PR #2987, opened 2022-07-16
Add Visual trait to Hazard actors and Apocryphal Subdomains

avagdumortesin:
Many of the Hazards in DA have the Visual trait.


stwlam:
Try not to add unrelated commits onto open PRs like this. We could have merged the visual trait earlier without the apocryphal domains.

avagdumortesin:
Fair enough; they're both background info for Dark Archive data entry, and the Apocryphal stuff I already added in with https://github.com/foundryvtt/pf2e/pull/2959, I just did it wrong, which I didn't realize until last night.

PR #2993, opened 2022-07-16
Add magical school traits to general spell traits

avagdumortesin:
Also removes Schools as a checkbox in the Compendium Browser, making it show in Traits only.

### DIFF
--- a/src/scripts/config/traits.ts
+++ b/src/scripts/config/traits.ts
@@ -344,6 +344,7 @@ const spellTraits = {
     ...damageTraits,
     ...elementalTraits,
     ...spellOtherTraits,
+    ...magicSchools,
 };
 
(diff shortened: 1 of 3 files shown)


stwlam:
Pull back to the single update to the spell traits record. We can get everything else in line after.

avagdumortesin:
Done

src/scripts/config/traits.ts
```diff
@@ -344,6 +344,7 @@ const spellTraits = {
     ...damageTraits,
     ...elementalTraits,
     ...spellOtherTraits,
+    ...magicSchools,
```
stwlam, line 347:
```suggestion
    ...magicSchools,
    ...spellOtherTraits,
```

PR #3001, opened 2022-07-17
Fixed source value and separator for the Atheism deity

chepseskhaf:
There is currently only one data entry with more than one source: the "Atheism" deity.

```
"source": {
    "value": "Pathfindier Core Rulebook, Pathfinder Lost Omens: Gods & Magic"
},
```

In addition to a spelling error ("Pathfindier" instead of "Pathfinder"), it uses "," as a separator. However, some sources such as "Pathfinder Lost Omens: Absalom, City of Lost Omens" use a "," in their name, invalidating the use of "," as a separator.

The future-proof way to solve this would be to change the data type to an array but that would mean changing every data entry. This is not ideal, especially for only one problematic entry.

As a compromise, I would only change the "," to ";" in this specific data entry (and all future data entries with mixed sources), hoping that Paizo will never release something with a ";" in its name...

### DIFF
--- a/packs/data/deities.db/atheism.json
+++ b/packs/data/deities.db/atheism.json
@@ -17,7 +17,7 @@
         "rules": [],
         "skill": null,
         "source": {
-            "value": "Pathfindier Core Rulebook, Pathfinder Lost Omens: Gods & Magic"
+            "value": "Pathfinder Core Rulebook; Pathfinder Lost Omens: Gods & Magic"
         },
         "spells": {},
         "weapons": []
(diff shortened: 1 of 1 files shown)


packs/data/deities.db/atheism.json
```diff
@@ -17,7 +17,7 @@
         "rules": [],
         "skill": null,
         "source": {
-            "value": "Pathfindier Core Rulebook, Pathfinder Lost Omens: Gods & Magic"
+            "value": "Pathfinder Core Rulebook; Pathfinder Lost Omens: Gods & Magic"
```
SpartanCPA, line 20:
```suggestion
            "value": "Pathfinder Lost Omens: Gods & Magic"
```

The versions in the CRB and Gods and Magic are different, this one if from Gods and Magic.

PR #3007, opened 2022-07-17
Improve feat drag/drop and handle invalid feat drops

### LINKED ISSUE #2918: Feature Request: Lock Button/Toggle for Items/Feats in character sheet
Trying to highlight text within feats (specifically) often times leads me to "grabbing" the feat and "throwing" it off of my character sheet. CTRL+Z does not undo this action. The best way that I can find to fix is to reopen the compendium the drag the feat back onto the sheet.
Highlighting text really helps me as I'm reading across all platforms. Having the option to lock "item" boxes that I definitely don't want to remove would be immensely helpful.

CarlosFdez:
The empty slots no longer interact with mouse drags, and the browse button is gone from the unorganized feats. Resolves #2918

image

### DIFF
--- a/src/module/actor/character/feats.ts
+++ b/src/module/actor/character/feats.ts
@@ -15,6 +15,7 @@ interface FeatCategoryOptions {
 }
 
 class CharacterFeats extends Collection<FeatCategory> {
+    /** Feats with no actual category (referred to by paizo as bonus feats)  */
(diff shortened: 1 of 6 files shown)


CarlosFdez:
A nice side-effect now is that the drag-drop preview for the spell list should look better.

stwlam:
I don't think the section should be renamed

Issue #2860, opened 2022-07-05
Effect for Sniping Duo dedication


OnkOnk:
An effect with a toggable box that says something like "partner hit target with a strike this round" would be perfect for this effect which is otherwise hard to keep track of.

PR #3013, opened 2022-07-18
add re for sniping duo

ebuckle:
For: https://github.com/foundryvtt/pf2e/issues/2860

Add FlatModifier and RollOption for [Sniping Duo Dedication](https://2e.aonprd.com/Feats.aspx?ID=3258).


LebombJames:
> Obviously this will only affect the player with the dedication feat. I could add an effect to be given to the spotter so they also have access to the re?

An effect is what I did for my game, think that'll work well.

TikaelSol:
Let's do an effect for this one instead.

ebuckle:
Done!

PR #3034, opened 2022-07-19
Add configuration for NPC auto heighten

CarlosFdez:
Later on we should show the heighten level on the sheet for all actors in the cantrip section, but separate PR.


stwlam:
So is this an auto-heighten max?

CarlosFdez:
This exposes the cantrip/focus spell heighten level, which sets what that cantrip heightens to. Some NPCs have cantrips at a level separate from their actual caster level.

stwlam:
So then yes? Increase level, cantrips and focus spells heighten no further?

CarlosFdez:
Or vice versa I think.

PR #3042, opened 2022-07-20
Add Implement and Innovation as "other tags"

stwlam:
Also expands other-tagging to armor and equipment item types


src/module/item/weapon/sheet.ts
```diff
@@ -132,6 +132,13 @@ export class WeaponSheetPF2e extends PhysicalItemSheetPF2e<WeaponPF2e> {
             rangedOnlyTraits.some((trait) => traitSet.has(trait));
         const mandatoryMelee = sheetData.data.traits.value.some((trait) => /^thrown-\d+$/.test(trait));
 
+        // Restrict the Implement tag to one-handed weapons
+        const otherTags = ((): SheetOptions => {
+            const otherWeaponTags: Record<string, string> = deepClone(CONFIG.PF2E.otherWeaponTags);
+            if (this.item.hands !== "1") delete otherWeaponTags.implement;
```
CarlosFdez, line 138: I wish they were clear on if a 1+ was a one handed weapon, but for now I suppose we can assume they're not.

PR #3048, opened 2022-07-20
Add frequency and display to actions and feats in the character sheet

CarlosFdez:
There's a part of me that wants get frequency() to return a special frequency class that can make such comparisons. But I'd like to have eyes on this first.

image

image


CarlosFdez:
Actually, tbh, I'd rather leave the Frequency class and built in use of Duration saved for when we actually start do things with it other than just rest for the night.

PR #3049, opened 2022-07-20
Pass inline check traits to Statistic and CheckPF2e

### LINKED ISSUE #3047: Feature Request: show the traits used in DC Rollable Buttons in the roll outcome
It would be nice if you could see the traits used in the dc rollable buttons e.g. @Check[type:diplomacy|dc:20|traits:auditory,concentrate,emotion,exploration,linguistic,mental|showDC:gm|name:Coerce] in the outcome of the roll as well (like the skill action macros):
skillissue

In3luki:
`@Check[type:diplomacy|dc:20|traits:auditory,concentrate,emotion,exploration,linguistic,mental|showDC:gm|name:Coerce] `
image

Closes #3047

### DIFF
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -350,9 +350,7 @@ class StatisticCheck {
         }
 
         // Add additional traits
-        const traitObjects = args.traits
-            ? args.traits.map((trait) => traitSlugToObject(trait, CONFIG.PF2E.actionTraits))
-            : [];
+        const traitObjects = args.traits?.map((trait) => traitSlugToObject(trait, CONFIG.PF2E.actionTraits)) ?? [];
 
         // Create parameters for the check roll function
         const context: CheckRollContext = {
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -16,7 +16,7 @@ import { eventToRollParams } from "@scripts/sheet-util";
 import { CheckRoll } from "@system/check/roll";
 import { CheckDC } from "@system/degree-of-success";
 import { CheckPF2e, CheckRollCallback, CheckRollContext, CheckType, RollTwiceOption } from "@system/rolls";
-import { isObject } from "@util";
+import { isObject, objectHasKey } from "@util";
 import {
     BaseStatisticData,
     StatisticChatData,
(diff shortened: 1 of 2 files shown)


src/scripts/ui/inline-roll-links.ts
```diff
@@ -78,97 +76,89 @@ export const InlineRollLinks = {
...
                 return;
             }
-
+            const creatureActors = actors.filter((actor): actor is CreaturePF2e => actor.isOfType("creature"));
+            const parsedTraits = pf2Traits
+                ? pf2Traits
+                      .split(",")
+                      .map((trait) => trait.trim())
+                      .filter((trait) => !!trait)
+                : undefined;
```
stwlam, line 85:
```suggestion
            const parsedTraits = pf2Traits
                ?.split(",")
                .map((trait) => trait.trim())
                .filter((trait) => !!trait);
```

src/scripts/ui/inline-roll-links.ts
```diff
@@ -78,97 +76,89 @@ export const InlineRollLinks = {
...
-                                        .map((trait) => trait.trim())
-                                        .filter((trait) => !!trait);
-                                    options.push(...traits);
-                                }
-                                flatCheck.check.roll({ ...eventRollParams, extraRollOptions: options, dc });
-                            } else {
-                                console.warn(`PF2e System | Skip rolling flat check for '${actor}'`);
-                            }
+                        } else {
+                            console.warn(`PF2e System | Skip rolling flat check for '${actor}'`);
```
stwlam, line 130:
```suggestion
                            console.warn(`PF2e System | Skip rolling flat check for "${actor}"`);
```

src/module/system/statistic/index.ts
```diff
@@ -342,6 +349,11 @@ class StatisticCheck {
             }
         }
 
+        // Add additional traits
+        const traitObjects = args.traits
+            ? args.traits.map((trait) => traitSlugToObject(trait, CONFIG.PF2E.actionTraits))
+            : [];
```
stwlam, line 355:
```suggestion
                const traitObjects = args.traits?.map((trait) => traitSlugToObject(trait, CONFIG.PF2E.actionTraits)) ?? [];
```

stwlam:
Actually, I don't think we can merge this. The "traits" from inline thingies aren't actually traits but roll options, some of which happen to correspond with traits. They're not traits for the check.

In3luki:
True, there would have to be separate option for real traits but then the current `traits` would have to be renamed to `rollOptions` which would require a migration. Probably not worth it for a minimal gain.

I'll open a new PR with the cleaned up `inline-roll-links.ts` later.

PR #3055, opened 2022-07-21
Tagify spell traits and traditions

stwlam:
image


CarlosFdez:
How well does it work when all 4 traditions are up?

stwlam:
I assume it doubles up like with the traits

CarlosFdez:
I solved it for note REs, so I'll look into it for this as well after I suppose.

PR #3057, opened 2022-07-21
Begin split of spell list from spellcasting ability

CarlosFdez:


CarlosFdez:
My first concern is naming. its SpellCollection and collections, but also its SpellcastingEntryPF2e#spells. Collection also has a loaded meaning, on second though. However actor.spellcasting.lists is probably even worse?

src/module/actor/sheet/base.ts
```diff
@@ -694,23 +694,23 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
         // or moving it from one spellcastingEntry to another
         if (item.isOfType("spell")) {
             const targetLocation = $dropContainerEl.attr("data-container-id") ?? "";
-            const entry = this.actor.spellcasting.get(targetLocation);
-            if (!entry) {
+            const collection = this.actor.spellcasting.collections.get(targetLocation);
+            if (!collection) {
                 console.warn("PF2E System | Failed to load spellcasting entry");
                 return [];
             }
```
stwlam, line 701:
```suggestion
            const collection = this.actor.spellcasting.collections.get(targetLocation, { strict: true });
```

src/module/actor/sheet/base.ts
```diff
@@ -845,15 +845,15 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
         if (item.isOfType("spell") && itemSource.type === "spell") {
             if (dropContainerType === "spellcastingEntry") {
                 const entryId = $containerEl.attr("data-container-id") ?? "";
-                const entry = this.actor.spellcasting.get(entryId);
-                if (!entry) {
+                const collection = this.actor.spellcasting.collections.get(entryId);
+                if (!collection) {
                     console.warn("PF2E System | Failed to load spellcasting entry");
                     return [];
                 }
```
stwlam, line 852:
```suggestion
                const collection = this.actor.spellcasting.collections.get(entryId, { strict: true });
```

stwlam (approved):

PR #3066, opened 2022-07-21
Automate Signifier Armor Expertise feat

TikaelSol:


stwlam (dismissed):
These need predicates

TikaelSol:
Frankly I forgot I added those roll options already

PR #3078, opened 2022-07-22
Review Veshumirix

manfred7168:
Changed "vs" to "vs." for standardization

### DIFF
--- a/packs/data/age-of-ashes-bestiary.db/veshumirix.json
+++ b/packs/data/age-of-ashes-bestiary.db/veshumirix.json
@@ -1299,7 +1299,7 @@
                 }
             },
             "img": "systems/pf2e/icons/actions/Passive.webp",
-            "name": "+1 Status to All Saves vs Magic",
+            "name": "+1 Status to All Saves vs. Magic",
             "sort": 1500000,
             "type": "action"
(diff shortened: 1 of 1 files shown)


TikaelSol:
The current ability in the glossary does not have the `.`, and the compendiums are split evenly on which one is used which tells me we changed the name of the ability at some point. We can easily fix all of them at once, just need to decide on a way to go. @SpartanCPA should we add the `.` or remove it from everything?

TikaelSol:
It has `.` on the statblock for at least this one. So let's add them then. Easy enough regex, I'll close this since it will get caught in the big sweep.

manfred7168:
Abaddon and me Need this to reduce the different Translation items in the german translation.
Abaddon did this to other Monsters too, so that all are using the same Syntax for easier translation.

Gesendet von Mail für Windows

Von: SpartanCPA
Gesendet: Freitag, 22. Juli 2022 19:17
An: foundryvtt/pf2e
Cc: Manni; Author
Betreff: Re: [foundryvtt/pf2e] Review Veshumirix (PR #3078)

I would add them personally, as it's more grammatically correct. How is it formatted on an official statblock?

Issue #2962, opened 2022-07-13
Adding the drained condition past drained 1 does not damage a character correctly.


sandersontavares:
As per the title. Easy to reproduce. Add drained 1, you take damage and lose max HP. Add more drained after that and you lose the max HP but do not take any damage.

EDIT: To reproduce, first get a character and damage them to a point where the gap between their current hp and their max is higher than twice their level, then apply drained 2.

avagdumortesin:
I cannot reproduce this behavior.

Dev:

https://user-images.githubusercontent.com/55938242/178631777-c0ba2eca-a308-44f0-a384-3791be448805.mov

3.12.2.10:

https://user-images.githubusercontent.com/55938242/178632060-376ffab2-d253-4866-be49-b39a9a03aa2e.mov

leerhund:
> I cannot reproduce this behavior.
> 
> Dev:
> 
>  Screen.Recording.2022-07-12.at.6.41.02.PM.mov 
> 3.12.2.10:
> 
>  Screen.Recording.2022-07-12.at.6.45.03.PM.mov

The problem is that the Lose Hitpoints RE is only triggering on Drained 1. It should trigger all the way down. So in the first example when you lower the Drained, they should remain as 53 Current HP.

avagdumortesin:
Ah, if I mark the HP at below max, I see the behavior.

sandersontavares:
Yes, sorry, I'll edit that to make it easier to reproduce for other people.

PR #3079, opened 2022-07-22
Fix Drained Condition Health Removal

ebuckle:
Potential fix for https://github.com/foundryvtt/pf2e/issues/2962

I hope there can be a neater solution to this but I struggled to find something that didn't resort to hard coding. The problems with drain are:
- Hooking into preUpdateActor won't work since the condition is being updated, not the actor
- Making a new preUpdateItem method could be promising but...
- Since we're incrementing a condition we need to know the previous value and new value, to avoid removing too much health

There were other issues but essentially I struggled to find a way to do this in a clean, generic way. This solution works but it most likely would be better somewhere else instead of on an RE. Happy to hear suggestions.

### DIFF
--- a/src/module/item/base.ts
+++ b/src/module/item/base.ts
@@ -590,8 +590,6 @@ class ItemPF2e extends Item<ActorPF2e> {
         type WithPreUpdate = RuleElementPF2e & { preUpdate: NonNullable<RuleElementPF2e["preUpdate"]> };
         const rules = this.rules.filter((r): r is WithPreUpdate => !!r.preUpdate);
         if (rules.length > 0) {
-            const clone = this.clone(changed, { keepId: true });
-            this.data.flags.pf2e.rollOptions = clone.data.flags.pf2e.rollOptions;
             for (const rule of rules) {
                 await rule.preUpdate(changed);
             }
--- a/src/module/item/base.ts
+++ b/src/module/item/base.ts
@@ -587,12 +587,8 @@ class ItemPF2e extends Item<ActorPF2e> {
         }
 
         // Run preUpdateItem rule element callbacks
-        type WithPreUpdate = RuleElementPF2e & { preUpdate: NonNullable<RuleElementPF2e["preUpdate"]> };
-        const rules = this.rules.filter((r): r is WithPreUpdate => !!r.preUpdate);
-        if (rules.length > 0) {
-            for (const rule of rules) {
-                await rule.preUpdate(changed);
-            }
+        for (const rule of this.rules) {
+            await rule.preUpdate?.(changed);
         }
 
         await super._preUpdate(changed, options, user);
--- a/src/module/item/base.ts
+++ b/src/module/item/base.ts
@@ -585,6 +585,18 @@ class ItemPF2e extends Item<ActorPF2e> {
         if (changed.data?.description?.value === null) {
             changed.data.description.value = "";
         }
+
+        // Run preUpdateItem rule element callbacks
+        type WithPreUpdate = RuleElementPF2e & { preUpdate: NonNullable<RuleElementPF2e["preUpdate"]> };
+        const rules = this.rules.filter((r): r is WithPreUpdate => !!r.preUpdate);
+        if (rules.length > 0) {
+            const clone = this.clone(changed, { keepId: true });
+            this.data.flags.pf2e.rollOptions = clone.data.flags.pf2e.rollOptions;
+            for (const rule of rules) {
+                await rule.preUpdate(changed);
+            }
+        }
+
         await super._preUpdate(changed, options, user);
     }
 
--- a/src/module/rules/rule-element/base.ts
+++ b/src/module/rules/rule-element/base.ts
@@ -385,6 +385,10 @@ interface RuleElementPF2e {
     preDelete?({ pendingItems, context }: RuleElementPF2e.PreDeleteParams): Promise<void>;
 
     /**
+     * Runs before this rules element's parent item is updated */
+    preUpdate?(changes: DeepPartial<ItemSourcePF2e>): Promise<void>;
+
+    /**
      * Runs after an item holding this rule is added to an actor. If you modify or add the rule after the item
      * is already present on the actor, nothing will happen. Rules that add toggles won't work here since this method is
      * only called on item add.
--- a/src/module/rules/rule-element/lose-hit-points.ts
+++ b/src/module/rules/rule-element/lose-hit-points.ts
@@ -1,25 +1,48 @@
 import { CreaturePF2e } from "@actor";
 import { ItemPF2e } from "@item";
+import { ItemSourcePF2e } from "@item/data";
 import { RuleElementPF2e, RuleElementSource } from "./";
 import { RuleElementOptions } from "./base";
 
 /** Reduce current hit points without applying damage */
 export class LoseHitPointsRuleElement extends RuleElementPF2e {
-    constructor(data: RuleElementSource, item: Embedded<ItemPF2e>, options?: RuleElementOptions) {
+    /** Lost hitpoints should reevaluate on item update, with the parent actor losing the difference in HP between the new and old values */
+    protected reevaluateOnUpdate: boolean;
+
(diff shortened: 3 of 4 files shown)


ebuckle:
Worth noting that remove hitpoints re is currently only being used for Drained. But if we do decide to use a specialised RE just for drained we should probably make a new one instead of changing it.

stwlam:
Go for a pre update

ebuckle:
New pre update hook for ItemPF2e?

stwlam:
> New pre update hook for ItemPF2e?

For `RuleElementPF2e`. Conditionally update the actor without rendering, following a similar pattern as these:

```ts
    /**
     * Runs before this rules element's parent item is created. The item is temporarilly constructed. A rule element can
     * alter itself before its parent item is stored on an actor; it can also alter the item source itself in the same
     * manner.
     */
    preCreate?({ ruleSource, itemSource, pendingItems, context }: RuleElementPF2e.PreCreateParams): Promise<void>;

    /**
     * Runs before this rules element's parent item is created. The item is temporarilly constructed. A rule element can
     * alter itself before its parent item is stored on an actor; it can also alter the item source itself in the same
     * manner.
     */
    preDelete?({ pendingItems, context }: RuleElementPF2e.PreDeleteParams): Promise<void>;
```

ebuckle:
Something like this?

src/module/rules/rule-element/base.ts
```diff
@@ -384,6 +384,10 @@ interface RuleElementPF2e {
      */
     preDelete?({ pendingItems, context }: RuleElementPF2e.PreDeleteParams): Promise<void>;
 
+    /**
+     * Runs before this rules element's parent item is updated */
+    preUpdate?(changed: DeepPartial<ItemPF2e["data"]["_source"]>): Promise<void>;
```
stwlam, line 389: use `ItemSourcePF2e`

ebuckle:
Now with far less hard coding 😄

ebuckle:
Characters with existing drained conditions will be unnaffected, since I included `reevaluateOnUpdate`. Characters will only see the new changes when adding a new instance of Drained.

Also realizing now I can do this without needing flags at all, will make a quick change

src/module/rules/rule-element/lose-hit-points.ts
```diff
@@ -1,15 +1,20 @@
...
-    constructor(data: RuleElementSource, item: Embedded<ItemPF2e>, options?: RuleElementOptions) {
+    /** Lost hitpoints should reevaluate on item update, with the parent actor losing the difference in HP between the new and old values */
+    protected reevaluateOnUpdate: boolean;
+
+    constructor(data: LoseHitPointsSource, item: Embedded<ItemPF2e>, options?: RuleElementOptions) {
         super(data, item, options);
         const actorIsCreature = this.actor instanceof CreaturePF2e;
         const valueIsValid = typeof data.value === "number" || typeof data.value === "string";
         if (!(actorIsCreature && valueIsValid)) this.ignored = true;
+        this.reevaluateOnUpdate = Boolean(data.reevaluateOnUpdate ?? false);
```
stwlam, line 17:
```suggestion
        this.reevaluateOnUpdate = !!data.reevaluateOnUpdate;
```

src/module/rules/rule-element/lose-hit-points.ts
```diff
@@ -20,6 +25,27 @@ export class LoseHitPointsRuleElement extends RuleElementPF2e {
             actorUpdates["data.attributes.hp.value"] = Math.max(currentHP - value, 0);
         }
     }
+
+    override async preUpdate(changed: DeepPartial<ItemSourcePF2e>): Promise<void> {
+        if (!this.reevaluateOnUpdate || this.ignored) return;
+        const previousValue = Math.abs(Number(this.resolveValue()) || 0);
+        mergeObject(this.item.data, changed);
+        const newValue = Math.abs(Number(this.resolveValue()) || 0);
```
stwlam, line 33: Instead of mutating the current item, clone the source (`this.item.toObject()`), merge that, and fish out the same RE.

src/module/rules/rule-element/lose-hit-points.ts
```diff
@@ -20,6 +25,27 @@ export class LoseHitPointsRuleElement extends RuleElementPF2e {
...
+        if (typeof previousValue === "number" && typeof newValue === "number") {
```
stwlam, line 34: These are already guaranteed to be numbers

stwlam (changes requested):

src/module/rules/rule-element/lose-hit-points.ts
ebuckle: Problem is I can't resolve the value of the new RE with the context of the preUpdated item since it doesn't exist.

stwlam: You can clone the item with the changes.

ebuckle: Before I commit, I'd be doing something like:
```
const newItem = this.item.clone();
mergeObject(newItem.data, changed);
const rule = newItem.data.data.rules.find((r) => r.key === this.key);
const newValue = Math.abs(Number(this.resolveValue(rule?.value, 0, { resolvables: { item: newItem } })));
```
?

ebuckle: This works I just want to make sure it's what you had in mind

stwlam: include the change in the clone method

ebuckle: Nice, thank you, that's all in now

.betterer.results
```diff
@@ -26,7 +26,7 @@ exports[`too-much-lint`] = {
       [23, 25, 3, "Unexpected any. Specify a different type.", "193409811"],
       [36, 38, 3, "Unexpected any. Specify a different type.", "193409811"]
     ],
-    "src/module/item/base.ts:2799968150": [
+    "src/module/item/base.ts:1027049593": [
```
stwlam, line 29: This is getting reverted; exclude file from PR unless you're intentionally updating it.

stwlam (changes requested):

.betterer.results
ebuckle: Gone 👍

stwlam (approved):

src/module/item/base.ts
```diff
@@ -585,6 +585,18 @@ class ItemPF2e extends Item<ActorPF2e> {
         if (changed.data?.description?.value === null) {
             changed.data.description.value = "";
         }
+
+        // Run preUpdateItem rule element callbacks
+        type WithPreUpdate = RuleElementPF2e & { preUpdate: NonNullable<RuleElementPF2e["preUpdate"]> };
+        const rules = this.rules.filter((r): r is WithPreUpdate => !!r.preUpdate);
+        if (rules.length > 0) {
+            const clone = this.clone(changed, { keepId: true });
+            this.data.flags.pf2e.rollOptions = clone.data.flags.pf2e.rollOptions;
```
stwlam, line 594: What's this part for?

ebuckle: Ah true, don't need to clone the rollOptions for items, that was lifted from the implementation in `ActorPF2e _preUpdate`

src/module/item/base.ts
```diff
@@ -585,6 +585,16 @@ class ItemPF2e extends Item<ActorPF2e> {
...
         }
+
+        // Run preUpdateItem rule element callbacks
+        type WithPreUpdate = RuleElementPF2e & { preUpdate: NonNullable<RuleElementPF2e["preUpdate"]> };
+        const rules = this.rules.filter((r): r is WithPreUpdate => !!r.preUpdate);
+        if (rules.length > 0) {
+            for (const rule of rules) {
+                await rule.preUpdate(changed);
+            }
+        }
```
stwlam, line 596:
```suggestion
        for (const rule of this.rules) {
            await rule.preUpdate?.(changed);
        }
```

PR #3095, opened 2022-07-24
Distinguish between cast and slot level in spell chat data

stwlam:


static/templates/chat/spell-card.html
```diff
@@ -1,4 +1,12 @@
-<div class="pf2e chat-card item-card" {{#if data.item}}data-embedded-item="{{data.item}}"{{/if}} data-spell-lvl="{{data.spellLvl}}" data-actor-id="{{actor.id}}" data-item-id="{{item.id}}" {{#if tokenId}}data-token-id="{{tokenId}}"{{/if}}>
+<div class="pf2e chat-card item-card"
+    {{#if data.item}}data-embedded-item="{{data.item}}"{{/if}}
+    data-spell-lvl="{{data.slotLevel}}"
```
CarlosFdez, line 3:
```suggestion
    data-spell-lvl="{{data.castLevel}}"
```
If this is sticking around for backwards compatibility, for the spellcard this was always the cast level.

src/module/chat-message/listeners/cards.ts
```diff
@@ -71,16 +71,17 @@ export const ChatCards = {
...
                 else if (action === "selectVariant") {
-                    const spellLvl = Number($html.find<HTMLDivElement>("div.chat-card").attr("data-spell-lvl")) || 1;
+                    const slotLevel =
+                        Number($html[0].querySelector<HTMLElement>("div.chat-card")?.dataset.slotLvl) || 1;
                     const overlayIdString = $button.attr("data-overlay-ids");
                     const originalId = $button.attr("data-original-id") ?? "";
                     if (overlayIdString) {
                         const overlayIds = overlayIdString.split(",").map((id) => id.trim());
-                        const variantSpell = spell?.loadVariant({ overlayIds, castLevel: spellLvl });
+                        const variantSpell = spell?.loadVariant({ overlayIds, castLevel: slotLevel });
```
CarlosFdez, line 80: Only 90% sure on this but shouldn't the chat message listeners for loading a spell always use the cast level?

src/module/item/spell/index.ts
```diff
@@ -596,14 +603,13 @@ class SpellPF2e extends ItemPF2e {
...
-            const cardData = card ? card.dataset : {};
-            return Number(cardData.spellLvl) || 1;
-        })();
-
+    async rollDamage(
+        this: Embedded<SpellPF2e>,
+        event: JQuery.ClickEvent<unknown, unknown, HTMLElement>
+    ): Promise<void> {
+        const slotLevel =
+            Number(event.currentTarget.closest<HTMLElement>("*[data-slot-level]")?.dataset.slotLevel) || 1;
```
CarlosFdez, line 611: After the current changes, this should use cast level, not slot level I think. That way cantrip damage button from the spell summary can scale.

static/templates/chat/spell-card.html
stwlam: But there was no cast level sent to the template

src/module/chat-message/listeners/cards.ts
stwlam: @In3luki Do you remember what this was expecting?

In3luki: That should be the cast level. `loadVariant` applies fixed heigthening based on that value.

static/templates/chat/spell-card.html
CarlosFdez: image

spell level is set to 4 for a cantrip. That means its the cast level, not the slot level.

CarlosFdez: In general, what spell-lvl refers to in live changed depending on what phase you were in. In the actor sheet it was slot level, but in the chat card it was cast level.

CarlosFdez: Ah, because I forgot to clarify, this is why I said it was "just cantrips" and that an alternative is to adjust drag/drop code. Spell level being "what it will be casted as" vs "what it got casted as" wasn't really broken as a paradigm until you got to cantrips. An alternative was to alter the sibling checks to prioritize cantrips over spell level, and make spell level the "will cast at" level always.

Having slot level and cast level separate might be easier to wrap someone's head around though.

CarlosFdez:
Looks great to me. Though I think it might be best to save this for next major release.

Issue #3064, opened 2022-07-21
Feature Request: Update Bloodline:Wyrmblessed with Bloodline:Draconic automation


Zakor11:
Just a short request to copy the automation features for damage and resistance type from Bloodline:Draconic to Bloodline:Wyrmblessed.

PR #3096, opened 2022-07-24
Localize Wyrmblessed and Draconic Bloodlines ChoiceSet RE and Add Rules Elements to Wyrmblessed

avagdumortesin:
Closes https://github.com/foundryvtt/pf2e/issues/3064.


packs/data/classfeatures.db/bloodline-wyrmblessed.json
```diff
@@ -54,6 +54,209 @@
                 "path": "data.proficiencies.aliases.sorcerer",
                 "priority": 39,
                 "value": "divine"
+            },
+            {
+                "adjustName": true,
+                "choices": [
+                    {
+                        "label": "Black Dragon",
```
TikaelSol, line 62:
```suggestion
                        "label": "PF2E.SpecificRule.DragonDisciple.Black",
```
We should localize all these, Dragon disciple already has a good chunk of them in re-en.json so let's just expand that list to include all of them.

PR #3116, opened 2022-07-24
Run migration to remove duplicated image

avagdumortesin:
If necessary to prevent any issues from https://github.com/foundryvtt/pf2e/pull/3115. Shamelessly copied from Migration 766. If not necessary, just kill it.


avagdumortesin:
@stwlam or @CarlosFdez if either of you think this is needed, I'll mark it as ready for review.

PR #3134, opened 2022-07-25
Add @check to locks

InfamousSky:


packs/data/equipment.db/lock-average.json
```diff
@@ -4,7 +4,7 @@
         "baseItem": null,
         "containerId": null,
         "description": {
-            "value": "<p>Picking an average lock requires four successful Thievery checks at DC 25.</p>"
+            "value": "<p>Picking an average lock requires four successful Thievery checks at DC 25. (@Check[type:thievery|dc:25|traits:action:pick-a-lock|immutable:true])</p>"
```
TikaelSol, line 7:
```suggestion
            "value": "<p>Picking an average lock requires four successful @Check[type:thievery|dc:25|traits:action:pick-a-lock|immutable:true]) checks.</p>"
```
Little wording change but this will read better I think.

Also I should look at just making all inlines on equipment default to immutable I think.

InfamousSky: changed the wording

PR #3135, opened 2022-07-25
AoA-GhastlyBear-RuleLabel

manfred7168:
Changed RuleLabel of GhastlyBear to use re-en.json


SpartanCPA:
Change the ability to use the same rule elements as the Bestiary bears. Mauler is not specific to just this NPC.

TikaelSol:
I'll do a sweep of unlocalized labels after v10 changes get rolling, closing this for now.

PR #3143, opened 2022-07-26
Locked RE textarea scaling to vertical

CarlosFdez:


CarlosFdez:
Just in case its ambiguous, all it affects is this little thing. Resizing the window itself will scale the textarea as normal 
image

PR #3147, opened 2022-07-26
Fix notification when feat is dropped into bonus feats category

stwlam:


CarlosFdez:
I think category id should probably err closer to null for bonus, not away from null, but its not a very particularly important hill.

PR #3150, opened 2022-07-26
Adjust modifiers instead of adding new ones for elite/weak adjustments

stwlam:
Normal:
Screenshot from 2022-07-26 16-20-41

Elite:
Screenshot from 2022-07-26 16-20-34

Weak:
Screenshot from 2022-07-26 16-20-49


CarlosFdez:
Looks fine to me (for as much as I can understand without poking around in an IDE). Do we merge this now or wait for V10? I'm erring towards the latter but I'm only a coward when it comes to my code.

stwlam:
Let's sit on this and merge it in just prior to V10 work.

PR #3153, opened 2022-07-26
Localize NPC resistance exceptions

MSAbaddon:


stwlam:
Not crazy about this one. We shouldn't be exposing localization keys in sheet inputs.

MSAbaddon:
> Not crazy about this one. We shouldn't be exposing localization keys in sheet inputs.

This is rather in order to localize the resistance exceptions within resistance rule elements. Should the localization happen in the rule element instead?

MSAbaddon:
@stwlam Just realized I still have this pull request open. I switched localization to the rule element. This way the localization key isn't exposed to the sheet input.

stwlam (approved):

PR #3190, opened 2022-07-28
Handle Token Auras making Perfect Vision act even naughtier

stwlam:


CarlosFdez:
I'll merge this, but would it not be simpler to make AuraRenderers functions perform no-ops if naughty module is naughty?

stwlam:
> I'll merge this, but would it not be simpler to make AuraRenderers functions perform no-ops if naughty module is naughty?

The functions aren't there. Token Auras monkey-patches an `auras` property over it.

CarlosFdez:
That is the most terrifying thing I've heard.

PR #3200, opened 2022-07-28
Fix NPC item rarity colors

### LINKED ISSUE #3199: Rarity Colors on NPC Sheets don't work
There are two problems. The first is that the data-item-rarity does not get set:
image
This can be fixed by changing item-line.html line 2 from:
```<div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.data.traits.rarity}}">```
to
```<div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.data.data.traits.rarity}}">```

However, that is not a complete fix as for some reason the styles in _inventory.sccs like ```&[data-item-rarity="common"] {``` do not apply.

So, I'm hoping someone that actually understands sass can figure it out and fix it. :)

### PR DESCRIPTION
**Pull request by CarlosFdez**
Closes #3199

### DIFF
--- a/static/templates/actors/partials/item-line.html
+++ b/static/templates/actors/partials/item-line.html
@@ -1,5 +1,5 @@
 <li class="list-row">
-    <div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.data.data.traits.rarity}}">
+    <div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.system.traits.rarity}}">
         <div class="item-name rollable{{#if (and (ne @root.actor.type "loot") (not owner))}} long{{/if}}">
             <div class="item-image" style="background-image: url('{{item.img}}')"></div>
             <h4>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: static/templates/actors/partials/item-line.html [hunk 1]
```diff
@@ -1,5 +1,5 @@
 <li class="list-row">
-    <div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.data.traits.rarity}}">
+    <div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.data.data.traits.rarity}}">
```
stwlam, line 2:
```suggestion
    <div class="item {{#if item.isTemporary}}temporary-item{{/if}}" data-item-id="{{item.id}}" data-item-is-container="{{isContainer}}" data-item-rarity="{{item.system.traits.rarity}}">
```
:o)

CarlosFdez: I forgot that because its a document, we can do this without repercussions pre v10.

PR #3201, opened 2022-07-28
Convert Actor classes and sheets from data.data to system (first pass)

In3luki:
Also changes most `instanceof` checks to use `isOfType` instead.
I think the sheet `getData` method needs a complete rewrite so I left it for now.
This is not compatible with v9.


In3luki:
I've rebased this and am left with only the breaking changes so I'll close this for now until it's time for full v10 conversion.

PR #3226, opened 2022-07-29
Add automation to Implement's Empowerment feature

### LINKED ISSUE #3219: Feature Request: Toggleable "Implement Empowerment" checkbox
The features page does allow to click to roll this additional damage, but user experience would be greatly improved by having this a toggle able field similar to sneak attack, as it is almost always on, though there will be some cases where it would not be, and should simply be added in to the damage the thaumaturge makes rather than so many extra steps.

TikaelSol:
Closes #3219

### DIFF
--- a/packs/data/classfeatures.db/implements-empowerment.json
+++ b/packs/data/classfeatures.db/implements-empowerment.json
@@ -11,7 +11,7 @@
             "value": null
         },
         "description": {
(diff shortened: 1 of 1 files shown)


packs/data/classfeatures.db/implements-empowerment.json
```diff
@@ -23,7 +23,25 @@
...
+            },
+            {
+                "key": "FlatModifier",
+                "predicate": {
+                    "all": [
+                        "implements-empowerment"
+                    ]
+                },
+                "selector": "strike-damage",
+                "slug": "gravity-weapon",
```
CarlosFdez, line 41: hmmm

TikaelSol, line 41:
```suggestion
```

PR #3257, opened 2022-07-30
Review spells against Nethys

SoldierC4:
Review spells by pulling data from Archives of Nethys' Elasticsearch backend. All changes were manually reviewed against the original PDFs; except Shadow Zombie, which I do not own.

### DIFF
--- a/packs/data/spells.db/shadow-zombie.json
+++ b/packs/data/spells.db/shadow-zombie.json
@@ -88,7 +88,9 @@
         "traits": {
             "custom": "",
             "rarity": "uncommon",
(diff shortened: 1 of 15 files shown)


packs/data/spells.db/shadow-zombie.json
```diff
@@ -88,7 +88,9 @@
         "traits": {
             "custom": "",
             "rarity": "uncommon",
-            "value": []
+            "value": [
+                "shadow"
```
SoldierC4, line 92: Need someone who owns Blood Lords to verify this

PR #3273, opened 2022-07-31
Remove source IDs pointing to world items

TikaelSol:
Might be sensible to add this to extractpacks


stwlam:
These should have been pulled from compendium items so that they can be refreshed

PR #3276, opened 2022-07-31
Make aura effects have unlimited duration

stwlam:


kyles-ep:
Is there a strong reason to force the effect duration to be unlimited? It seems like there's a great deal of benefit to allowing the applied effect have its own duration. For example, it would allow an aura to apply an effect that has a duration like Inspire Defense (4 rounds). We would not lose the ability for the applied effect to have unlimited duration if the responsibility was shifted to whoever is using the aura to point to an unlimited version of the buff.

A low-effort approach could be a boolean flag to toggle "coerce unlimited duration" to force the duration.

Even better would be to accept an object of settings overrides that get applied to the "template" effect as part of the effect data.

KSops:
Is there a strong reason to force the effect duration to be unlimited? It seems like there's a great deal of benefit to allowing the applied effect have its own duration. For example, it would allow an aura to apply an effect that has a duration like Inspire Defense (4 rounds). We would not lose the ability for the applied effect to have unlimited duration if the responsibility was shifted to whoever is using the aura to point to an unlimited version of the buff.

A low-effort approach could be a boolean flag to toggle "coerce unlimited duration" to force the duration.

Even better would be to accept an object of settings overrides that get applied to the "template" effect as part of the effect data.

PR #3317, opened 2022-08-01
Add basic automation to Fang Sharpener feat

TikaelSol:
We can't do the removal of the finesse trait just yet, there's a way we could cheat that by predicating the strike on the base goblin heritage on not having this feat then just add a new strike here but I'd rather just do an AdjustStrike when we have a remove mode.


stwlam:
Were the strikes slugged you could replace them via that

TikaelSol:
They weren't. I think merge or close and I'll fix the rest before the next release.

PR #3337, opened 2022-08-03
Refactor CompendiumBrowser listeners

In3luki:
This technically isn't a v10 change but I noticed the infinite scrolling in the compendium browser result list was getting stuck after a re-render if the mouse wasn't moved.
I've changed the scroll event to append the new results to the DOM instead of re-rendering the whole application.
I've also removed JQuery from the other listeners while I was at it.
The browser should work  exactly the same after this PR.


src/module/apps/compendium-browser/tabs/base.ts
```diff
@@ -20,6 +20,8 @@ export abstract class CompendiumBrowserTab {
     scrollLimit = 100;
     /** The name of this tab */
     tabName: Exclude<TabName, "settings">;
+    /** The path to the result list template of this tab */
+    templatePath = "";
```
stwlam, line 24:
```suggestion
    abstract templatePath: string;
```

PR #3353, opened 2022-08-04
Add support for modifying Infused crafting batch sizes

Gronex:
Generally intended to help with support for the 10 batch size for Munitions Crafter (https://2e.aonprd.com/Feats.aspx?ID=3158).
It is implemented as a rule element to be put on the feat.

To make it work i also had to add `type:<consumableType>` to the `getRollOptions` list, to help filter down to the ammunition.

```json
{
  "key": "CraftingInfusion",
  "selector": "munitionsCrafter",
  "predicate": {
    "all": [
      "consumable:trait:alchemical",
      "consumable:type:ammo",
      "consumable:level:0"
    ]
  },
  "batchSize": 10
}
```


CarlosFdez:
Is there anything else this is good for? For example, can this handle the witches cauldron feat?

A rule element needs to be as general as possible, otherwise hardcoding (such as what handwraps do currently) is preferred.

stwlam:
Is a separate rule element for this needed? What made expanding `CraftingEntry` undesirable?

Gronex:
Regarding the witches cauldron feat, the rule could be added to that one as well. It depends on how the feats are supposed to interact, because advanced alchemy explicitly states a batch of 2.

Making it a rule element may well have been overkill.
The rationale of doing it that way was to make it able to be independent of the `CraftingEntry`, so later feats, although i don't know that any exists currently, could modify further.

CarlosFdez:
The concern would be that if it can't be CraftingEntry and must be its own entry, making it a rule element that can alter the quantity of a crafting attempt, whether its infusion or regular crafting, might make it more generally useful. That way you don't have a rule element for only one thing.

In general we need to take another look at the CraftingEntry rule element though, because the batch size is definitely supposed to be handled by that. We might need to consider giving it tweaks somehow.

ebuckle:
So especially for Witches Cauldron I didn't bother adding any additional automation as there was no need. There should be nothing stopping you from Crafting 6 potions at a time and crafting them has nothing to do with Crafting Entries

CarlosFdez:
Forgetting that the same argument applies to munitions machinist....

If all you want is mundane ammo crafting, and want to take the easiest route to it, you could add "munitions-machinist" as a field discovery and assign to it, like the field discovery feats do. Then you can add hardcode hande the batch size.

Then later in the future we can have a real solution and migrate all of them to the real solution.

ebuckle:
Munitions crafter this definitly applies to, since it uses the Crafting Entry system and "spends" infused reagents to craft batches of ammo. So I can definitely understand the need for this.

Field discovery could definitely be a workaround for now, pending some review of the crafting code in future yes 🙂

CarlosFdez:
More so that you don't technically *need* automation for munitions crafter either. You can add 10 bullets to your inventory, give them the infused trait, and manually deduct the reagent. Granted its more annoying to do so than cauldron, where you only need to tick an arrow twice. 

Field discovery is probably the least intrusive way to do it right now. An RE to change batch sizes would need to be applied to all the alchemist subclasses, not just munitions crafter.

Gronex:
That definitional makes, the only thing i wonder is if we will get some wonky behavior for an alchemist with gunslinger dedication, who can effectively have multiple field discoveries then.

CarlosFdez:
Oh don't worry, that already has wonky behavior. But field discovery is assigned via a path, so you'd assign the field discovery to the munitions crafter entry, not the alchemist one, and it'll be fine.

Gronex:
I updated the code in the direction discussed. One thing that caused some issues was the summation.
```js
Math.floor(fieldDiscoveryQuantity / fieldDiscoveryBatchSize) +
Math.ceil(((fieldDiscoveryQuantity % fieldDiscoveryBatchSize) + otherQuantity) / batchSize)
```
Gave 1 more than expected, so I changed it to be simpler, although I worry that I broke some edge case I have not thought of by that simplification.

ebuckle:
Please post a screenshot of it giving the wrong value

Gronex:
image
Interestingly it is only when the batch size does not match exactly to the field discovery batch size.

ebuckle:
What was the batch size for the field discovery in that case? Two?

CarlosFdez:
you'll probably need to special case ammo to default to the batch size. But also I guess make sure only mundane ammo is affected. What you did seems to affect all ammo.

packs/data/feats.db/munitions-crafter.json
```diff
@@ -51,6 +51,30 @@
             {
                 "key": "GrantItem",
                 "uuid": "Compendium.pf2e.feats-srd.Alchemical Crafting"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "system.crafting.entries.munitionsCrafter.fieldDiscovery",
```
ebuckle, line 58: I don't think there's a need to have these all be AEs.

The ones for alchemist only do that because the choice of Field Discovery will update the entry. You could set all of these values in the Crafty Entry RE and just update the Crafting Entry RE to use them if needed

src/module/item/base.ts
```diff
@@ -100,6 +100,10 @@ class ItemPF2e extends Item<ActorPF2e> {
             options.unshift(`${delimitedPrefix}type:${itemType}`);
         }
 
+        if (this.isOfType("consumable")) {
+            options.push(`${delimitedPrefix}type:${this.consumableType}`);
```
ebuckle, line 104: Ahh very nice, I was missing knowledge of how to do this when I first did the crafting code. 🙂

packs/data/feats.db/munitions-crafter.json
Gronex: That is a good point, makes it easier to get an overview as well

CarlosFdez:
Looks fine to me mostly, only gotcha is that if you're going to add min level and mundane restrictions in the RE instead of in code, it might better to make that a predicate. Then you'd predicate on "item:level:0" and also probably `not: { ["item:trait:magical"] }`.

A magical check is probably not required if you're doing a level 0 check, that was my mistake.

CarlosFdez:
Hmm, on second though, a predicate could also predicate on the item type. If we're going that far, we might as well add a migration for the other field crafts and convert them all to "predicate + batch size" things. And then we'd need a migration for all the existing ones. And perhaps a new CraftingBatchSize RE with selectors for either mundane or infused crafting.

Since a migration would be required anyways, I'm down to merge it, and we can revisit this later if they ever make more feats of this type in the future. Since this is pertaining to REs, are you ok with that @stwlam ?

src/module/item/base.ts
stwlam: Move to consumable class and call it a subtype.

src/module/rules/rule-element/crafting/entry.ts
```diff
@@ -47,8 +48,12 @@ class CraftingEntryRuleElement extends RuleElementPF2e {
             requiredTraits: this.data.requiredTraits,
             maxItemLevel: this.data.maxItemLevel,
             maxSlots: this.data.maxSlots,
+            batchSize: this.data.batchSize,
+            fieldDiscovery: this.data.fieldDiscovery,
+            fieldDiscoveryBatchSize: this.data.fieldDiscoveryBatchSize,
+            fieldDiscoveryRequiresMundane: this.data.fieldDiscoveryRequiresMundane,
+            maxFieldDiscoveryItemLevel: this.data.maxFieldDiscoveryItemLevel,
```
stwlam, line 55: This is way too hard-coded for alchemy

CarlosFdez:
Alright, got something more decisive now. Remove the two new properties you added, `maxFieldDiscoveryItemLevel` and `fieldDiscoveryRequiresMundane`. Instead hardcode the ammo field discovery to be mundane only in the code itself (which was my original suggestion). Zero future proofing. There's no need to add configurability on something that will get the cattlegun later.

Also, if possible, post a video of it at work once you're done. Some things are harder to gauge from images. Play around with it if you can. If you can't do so easily nor have the time l'll check it out locally and try to break it myself.

For someone else (not necessarily you) that is sufficiently motivated in the future, my thought right now is to introduce a CraftingQuantity RE that uses a predicate to match a valid item, and has a field to determine if the crafting is regular or advanced. This will handle alchemist field reagents, munitions crafting, and cauldron, and any other "you craft X instead of Y" or "you make a batch of X per reagent" feature in the future. That'll require a migration, so I'd rather not force you to do that just to get munitions crafter working for your game.

src/module/rules/rule-element/crafting/entry.ts
Gronex: I am not sure if you wanted all of it gone, or just the mundane and maxItemlevel ones.

The others seemed to have just been missed, as they were properties on the entry before, and could make sense to set on add

ebuckle:
You'll need to work with: https://github.com/foundryvtt/pf2e/pull/3424 once it's in. Should make things easier. 

I skipped the migration on my PR to do it here instead once everything is sorted for munitions crafter

CarlosFdez:
I'm busy currently, but from a glance it looks fine and its more or less what I suggested. I can dig more later.

Far2Casual:
No more move on this ? Ammunition crafting is still a chore without it :innocent:

ebuckle:
@Gronex Are you able to rebase this with the newest crafting changes and continue?

If not I can pick this issue up

Gronex:
I won't be able to take a look until Tuesday at the earliest, so you are welcome to pick it up

ebuckle:
The potential solution for this is to turn Field Discovery into a Predicate, allowing Machinist to function as expected with a field discovery batch size of 10. 

You hit another problem when you get to Precious Munitions, as there's currently no way to support 3 different batch sizes in a single entry. And there's currently no way to set precious materials on ammunition.

Since munitions crafter and it's feat tree seem rather unique (at least for now), Precious Muntions could make another crafting entry that only accepts precious material ammo, with a batch size of 1. 

Either that or a more sophisticated way of calculating batch sizes for crafting entries, which feels unnecessary to support a single feat.

Gronex:
I suppose with #3787 there is no longer a need for this PR, as that one handles the same in a much nicer way, as far as i can tell.

DawidIzydor:
@Gronex since #3787 was dropped can we reopen/merge this? Ammunition crafting is still not working correctly as far as I'm aware

Gronex:
@DawidIzydor I am not really sure why the other one was closed, so to have any real hope of getting it merged that may need to be clarified. Regardless I think we might be better of starting over, since there has been a fairly large amount of restructuring in the crafting area since this PR was abandoned

PR #3368, opened 2022-08-05
Convert embedded spell variants from `data` to `system`

In3luki:


stwlam:
Could this cover all overlays?

In3luki:
I can't find any other `data` references in the packs apart from spells embeddd in consumables but those are already converted. Did I miss something?

stwlam:
```ts
            for (const overlay of heightenEntries) {
                mergeObject(source.system, overlay.data);
            }
```

In3luki:
That `data` is added in `SpellPF2e#getHeightenLayers`. No migration needed. I can rename it for consistency though.

In3luki:
I'll do it in a separate PR, it's a more invlolved than I expected.

PR #3383, opened 2022-08-06
Add a way to catch up embedded item source data to the current DataModel

In3luki:
`Item.migratedData` removes `data` and adds `system` but does not update the database so the item has both after the migration but I believe that shouldn't be a problem.


stwlam:
Doesn't the server do this for world stuff?

In3luki:
This is for spells embedded in consumable items.

PR #3384, opened 2022-08-06
Add support for spell variants from consumable items

### LINKED ISSUE #3382: Spell variants are treated as orphaned spells when cast through consumable
Reproduction steps:
1. Drag an Acid Splash spell (e.g. from quick insert) onto a spellcaster's **inventory**
2. Keep selection (heightened level 1, scroll) and click Create
3. Expand the newly created item
4. Click "Use scroll of ..."
5. Observe chat log

Expected behavior:  chat card should have attack roll and damage buttons

Actual behavior:  it doesn't have those, plus there's a console warning about "Orphaned spell Acid Splash"

After investigating a bit, it seems that the problem may be due to spell variants temporarily created during calculated and lacking the "isFromConsumable" field of their parents?

In3luki:
The `Select Other Variant` button on chat cards from consumable items deletes the original message instead of updating it because I didn't want make the whole thing even more janky. I probably should've renamed `isFromConsumable` but the name is technically still correct because it points at the consumable the spell is from. 😄 

Closes #3382

### DIFF
--- a/src/module/actor/sheet/item-summary-renderer.ts
+++ b/src/module/actor/sheet/item-summary-renderer.ts
@@ -1,5 +1,5 @@
-import { ActorPF2e, CharacterPF2e, CreaturePF2e } from "@actor";
-import { ConsumablePF2e, ItemPF2e, PhysicalItemPF2e, SpellPF2e } from "@item";
+import { ActorPF2e, CreaturePF2e } from "@actor";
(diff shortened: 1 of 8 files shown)


In3luki:
I just realized this can't handle Trick Magic Item...  Back to the drawing board.

PR #3394, opened 2022-08-07
Use newly available wand-magic-sparkles icon for PC spellcasting tab

stwlam:
Before:
image

After:
image


CarlosFdez:
Technically a foundryv10 thing, since wand with sparkles is the v9 look, so added the tag.

PR #3395, opened 2022-08-07
Remove `ActorPF2e#physicalItems` getter

stwlam:


CarlosFdez:
Its about that time isn't it

PR #3402, opened 2022-08-08
Convert a few item sheets to use getBaseData

CarlosFdez:


CarlosFdez:
This breaks a feature in persistent damage, so right now during v10 is the perfect time to do it!

PR #3413, opened 2022-08-08
Refactor wands and scrolls to not require embedded spell data

### LINKED ISSUE #3382: Spell variants are treated as orphaned spells when cast through consumable

In3luki:
Newly created wands and spells will only have a UUID and some information about the embedded spell.
Old items with embedded spell data are still fully supported but will show a warning in the console that that will change in a future system version. Spell variants also work with the old items now. No migration needed. 

Only old spell cards with the spell data embedded in the DOM will no longer work.

Closes #3382

### DIFF
--- a/src/module/actor/sheet/item-summary-renderer.ts
+++ b/src/module/actor/sheet/item-summary-renderer.ts
@@ -1,5 +1,5 @@
-import { ActorPF2e, CharacterPF2e, CreaturePF2e } from "@actor";
-import { ConsumablePF2e, ItemPF2e, PhysicalItemPF2e, SpellPF2e } from "@item";
+import { ActorPF2e, CreaturePF2e } from "@actor";
+import { ItemPF2e, PhysicalItemPF2e } from "@item";
 import { ItemSummaryData } from "@item/data";
 import { isItemSystemData } from "@item/data/helpers";
 import { InlineRollLinks } from "@scripts/ui/inline-roll-links";
@@ -198,7 +198,7 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
                     buttons.append(`<span>${save.label}</span>`);
                 }
 
-                if (actor instanceof CharacterPF2e) {
+                if (actor.isOfType("character")) {
                     if (Array.isArray(chatData.variants) && chatData.variants.length) {
                         const label = game.i18n.localize("PF2E.Item.Spell.Variants.SelectVariantLabel");
                         buttons.append(
@@ -221,7 +221,7 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
 
                 break;
             case "consumable":
-                if (item instanceof ConsumablePF2e && item.uses.max > 0 && item.isIdentified) {
+                if (item.isOfType("consumable") && item.uses.max > 0 && item.isIdentified) {
                     const label = game.i18n.localize("PF2E.ConsumableUseLabel");
                     buttons.append(
                         `<span><button class="consume" data-action="consume">${label} ${item.name}</button></span>`
@@ -233,11 +233,11 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
 
         $div.append(buttons);
 
-        buttons.find("button").on("click", (event) => {
+        buttons.find("button").on("click", async (event) => {
             event.preventDefault();
             event.stopPropagation();
 
-            const spell = item instanceof SpellPF2e ? item : item instanceof ConsumablePF2e ? item.embeddedSpell : null;
+            const spell = item.isOfType("spell") ? item : null;
 
             // which function gets called depends on the type of button stored in the dataset attribute action
             switch (event.target.dataset.action) {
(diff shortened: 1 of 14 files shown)


src/module/chat-message/document.ts
```diff
@@ -97,13 +97,12 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
 
     /** Get the owned item associated with this chat message */
     get item(): Embedded<ItemPF2e> | null {
```
CarlosFdez, line 99: get item() is used by modules and should always return a result. If you want to not embed spells in chat messages, you'll need to make this async. I don't quite mind if items that aren't in the actor are embedded, but it can work either way so long as this method continues to work.

src/module/actor/spellcasting.ts
```diff
@@ -24,8 +24,7 @@ export class ActorSpellcasting extends Collection<SpellcastingEntryPF2e> {
         return this.regular.filter((entry) => entry.isPrepared || entry.isSpontaneous);
     }
 
-    canCastConsumable(item: ConsumablePF2e): boolean {
-        const spell = item.embeddedSpell;
+    canCastConsumable(spell?: SpellPF2e): boolean {
```
CarlosFdez, line 27: canCastConsumable(spell) feels off to me lmao. That said, this method was here to also in the future be able to handle things like scroll thaumaturgy. I'd continue to give it the consumable, make the method async, but cache the embedded spell (cache wiped during data prep).

src/module/chat-message/document.ts
CarlosFdez: Basically, if we make this async, an update can be made to automated animations to fix their stuff hopefully. But if this returns null for async stuff, then we can't fix their module afterwards.

Continuing to store the spell source data in flags completely avoids any updates, but I know you wanna reduce the update burden.

In3luki:
Making `ChatMessagePF2e#item` async lead to breaking enrichment in `prepareData` because that is sync.
Core does the enrichment in `ChatMessage#getHTML` now, so I removed the `super` call from `ChatMessagePF2#getHTML` and copied the necessary code to our class.

CarlosFdez:
isn't enrichment also async now? They're starting making that async anyways.

In3luki:
> isn't enrichment also async now? They're starting making that async anyways.

Yeah, but the types are not updated yet so I kept it synchronous for now.

src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -233,11 +233,11 @@ export class CreatureSheetItemRenderer<AType extends CreaturePF2e> extends ItemS
 
         $div.append(buttons);
 
-        buttons.find("button").on("click", (event) => {
+        buttons.find("button").on("click", async (event) => {
```
CarlosFdez, line 236: async is unnecessary if there's no await

src/module/item/consumable/index.ts
```diff
@@ -7,8 +7,13 @@ import { TrickMagicItemPopup } from "@actor/sheet/trick-magic-item-popup";
 import { TrickMagicItemEntry } from "@item/spellcasting-entry/trick";
 import { ValueAndMax } from "@module/data";
 import { ItemSummaryData } from "@item/data";
+import { createConsumableFromSpell } from "./spell-consumables";
+import { SpellSource } from "@item/data";
 
 class ConsumablePF2e extends PhysicalItemPF2e {
+    /** The cached spell embedded in this item */
+    #spell?: Embedded<SpellPF2e>;
```
CarlosFdez, line 15: Should probably wipe this in prepareBaseData()

src/module/item/consumable/index.ts
```diff
@@ -190,12 +299,30 @@ class ConsumablePF2e extends PhysicalItemPF2e {
...
-            if (entry instanceof SpellcastingEntryPF2e) {
-                systemData.location.value = entry.id;
+        return entry ? entry : null;
+    }
+
+    async castEmbeddedSpell(
+        this: Embedded<ConsumablePF2e>,
+        spell?: Embedded<SpellPF2e>,
+        trickMagicItemData?: TrickMagicItemEntry
+    ): Promise<void> {
```
CarlosFdez, line 309: I think you can omit the spell parameter since embedded spell is cache. There's no longer a performance gain from doing this and it'll make the API less awkward.

stwlam:
V10 may actually make storing embedded data more viable. :)

src/module/item/consumable/index.ts
In3luki: I'm only caching spells from embedded data and from a compendium currently. Caching world items has the problem that changes to the source items are not reflected until the consumable data is prepared again.

In3luki:
> V10 may actually make storing embedded data more viable. :)

I can still trim this down to just the changes needed to make spell variants work and drop the uuid approach for now.

PR #3427, opened 2022-08-09
Implement Brutal trait for PCs

CarlosFdez:
image


src/module/item/weapon/index.ts
```diff
@@ -549,6 +549,8 @@ class WeaponPF2e extends PhysicalItemPF2e {
                         !(t.startsWith("thrown") && !this.isThrown) &&
                         // Finesse trait on thrown attacks with thrown melee weapons
                         !(t === "finesse" && this.isRanged) &&
+                        // Brutal trait on melee attacks
+                        !(t === "brutal" && this.isMelee) &&
```
CarlosFdez, line 553: not sure what this is for tbh, it seems to not do anything for finesse either, but put it here because maybe there is wisdom in this.

stwlam: It makes it so the ranged attack generated for a thrown melee weapon won't have the finesse trait.

CarlosFdez: image

still seems to have it

stwlam: yeah but what method is that in :slightly_frowning_face:

PR #3439, opened 2022-08-10
Add async overloads to `TextEditor.enrichHTML` and `TextEditorPF2e.enrichHTML`

In3luki:
Based on `Roll#evaluate`.
The first overload in `TextEditorPF2e` was necessary to pass type checks, not sure why.


stwlam:
There's an overload for when:
1. you know `async` is omitted/undefined or false (default): will return a string
2. you know `async` is true: will return a promise
3. you don't know the value of `async`: will return either a promise or string

In system code, there will then be the implementation, which will have the same signature as the last (the implementation doesn't know)

PR #3450, opened 2022-08-11
Explicitly set the `async` option for all references to `enrichHTML`

In3luki:
Setting the default option in the super call makes the async handling of the override pointless.


CarlosFdez:
We need to convert to all being async if we want custom enrichers don't we?

In3luki:
I shouldn't read code after just waking up. I didn't realize the spread overrides the default `async: false` option so it should still work as expected currently.
I think the synchronous version of enrichtHTML will be removed in v12 so we do have some time to solve that.

CarlosFdez:
We kinda do have to solve it sooner rather than later. @Localize is bugged rn. Though I don't know if a custom enricher will fix that bug.

I'm busy currently so I can't really start looking into more complex stuff until the weekend.

PR #3463, opened 2022-08-12
Rename `ConsumablePF2e#charges` to `#uses`, remove `uses.current`

stwlam:


CarlosFdez:
Ah, didn't notice till after, but remember to update template.json

Issue #3403, opened 2022-08-08
Level 0 characters don't get the initial four free ability boosts


VaingloriousReptile:
Level 0 characters, as described in page 195 of the Gamemastery Guide, are supposed to get the same initial four free ability boosts that all player characters get, but these ability boosts are only given to characters of level 1 or higher, thus requiring the player to manually calculate ability scores for level 0 characters.

PR #3475, opened 2022-08-13
Allow ability score boosts at level 0

ebuckle:
Fix for #3403 

Level 0 characters are [still allowed to take level 1 boosts](https://2e.aonprd.com/Rules.aspx?ID=1344)

Could also rename `Level 1` to `Initial`? But I think level 0 characters are niche enough to not warrant a change.

### DIFF
--- a/src/module/actor/character/index.ts
+++ b/src/module/actor/character/index.ts
@@ -275,7 +275,7 @@ class CharacterPF2e extends CreaturePF2e {
         const boostLevels = [1, 5, 10, 15, 20] as const;
         const allowedBoosts = boostLevels.reduce((result, level) => {
             const allowed = (() => {
-                if (this.level < 1) return 0;
+                if (this.level === 0 && level === 1) return 4;
                 if (isGradual) return 4 - Math.clamped(level - this.level, 0, 4);
                 return this.level >= level ? 4 : 0;
             })();


stwlam:
But they can't take class boosts

ebuckle:
Doesn't conflict thankfully. 

Class pages only mention "Boosts" in the context of key ability scores and free boost at level 5 and above.

stwlam:
> A level 0 character still gets the four free ability boosts from Step 6 of the normal character creation process, but not the class ability boost.

ebuckle:
I'm not exactly sure where we're disagreeing.

This is how a level 0 character should look on choosing boosts right?

image

stwlam:
> I'm not exactly sure where we're disagreeing.
> 
> This is how a level 0 character should look on choosing boosts right?
> 
> image

At level zero, you'll have picked a class

ebuckle:
> ...but you stop after choosing your ancestry and background.

Picking a class comes after these steps, so a level 0 character should have no class:
image

CarlosFdez:
I think Eddie is right here, though we'll definitely need to look into (non-jquery) context menus to make classes deletable.

stwlam:
See https://2e.aonprd.com/Rules.aspx?ID=1349

ebuckle:
Ah, so basically needs an additional check to ignore class boosts when level 0

CarlosFdez:
Even though we still need class, going to merge it anyways since its still a step forward. Class w/o boost can be a separate PR.

PR #3491, opened 2022-08-13
Remove `TextEditorPF2e` class and move its methods to `EnrichContentPF2e`

In3luki:
I've kept `game.pf2e.TextEditor` for compatibility.

- Chat message visibility is handled in `ChatMessagePF2e#getHTML`.
- Item description visibility is handled in the base item sheet and the item summary renderer.

Alternatively we could also add a handlebars helper to handle visibility in the templates.


stwlam:
'tis the season to break: :axe: `TextEditorPF2e`

CarlosFdez:
go for it, if I add persistent damage to core, I'll have to do it with a bunch of hooks, but if you're ok with that its fine.

CarlosFdez (dismissed):

PR #3494, opened 2022-08-13
Remove unnecessary partial-template paths from ItemSheetPF2e#getData

stwlam:


CarlosFdez:
This does break persistent damage, I do rely on it in order to inject a details tab. I can be a bit more forceful though I guess, but then I'll also have to override tab logic.

stwlam:
Can you add a tab on render?

CarlosFdez (approved):

stwlam:
On hold until we can get the persistent damage module's sticky fingers out of there

PR #3509, opened 2022-08-14
Fix regression in status effect HUD on Windows

In3luki:


src/module/canvas/status-effects.ts
```diff
@@ -274,6 +274,7 @@ export class StatusEffects {
      * @param event The window click event
      */
     static async #setStatusValue(token: TokenPF2e, event: JQuery.TriggeredEvent): Promise<void> {
+        event.preventDefault();
```
stwlam, line 277: There is no default default action here.

src/module/canvas/status-effects.ts
```diff
@@ -323,6 +324,7 @@ export class StatusEffects {
     }
 
     static async #toggleStatus(token: TokenPF2e, event: JQuery.TriggeredEvent): Promise<void> {
+        event.preventDefault();
```
stwlam, line 327: Same

src/module/canvas/token/object.ts
```diff
@@ -309,7 +309,7 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
             const { width, height, texture } = source;
             this.document.reset();
             // If any of the following changed, a full redraw should be performed
-            const postChange = pick(this.document, ["width", "height", "texture"]);
+            const postChange = pick(this.document.toObject(), ["width", "height", "texture"]);
```
stwlam, line 312: The current code here has a bug, but prepared data is what needs to be compared against.

src/module/canvas/status-effects.ts
In3luki: It could be moved inside the `contextmenu` check that happens a bit later in the method, but the default behaviour needs to be prevented for right clicks.

src/module/canvas/token/object.ts
In3luki: I was just happy I found the root cause and this seemed to fix it. 😄

src/module/canvas/status-effects.ts
stwlam: That's the thing, though: there is no default action in this case.

In3luki: Recording 2022-08-14 at 17 43 07

stwlam: I'm not getting that here. :\

In3luki: Chrome 104.0.5112.81 on Windows 10

src/module/canvas/status-effects.ts
```diff
@@ -342,6 +343,7 @@ export class StatusEffects {
 
         const conditionIds: string[] = [];
         if (event.type === "contextmenu") {
+            event.preventDefault();
```
stwlam, line 346: The oriiiiginal handler :)

In3luki:
Third time's the charm? 😁

Issue #1903, opened 2021-11-29
Munitions Crafter, Ammunition Crafting


cswendrowski:
In GitLab by @Velara on Nov 29, 2021, 17:25

At present when a Gunslinger with the munitions crafter feat adds an ammunition formula such as Rounds (Jazail) to their advanced alchemy section. An error message is displayed "Item is not compatible with crafting entry trait requirements, must match one of [["alchemical, bomb"], ["alchemical"]

Moreover, when crafting ammunition or Black Powder doses, they are treated like normal alchemical items and cost 1 reagent per 2, while level 0 ammunition should instead cost 1 reagent per 10 rounds.  Per the Munititions Crafter feat. "If you use a batch of infused reagents to create basic level-0 ammunition such as black powder cartridges or black powder doses, you produce 10 rounds of ammunition."

Awesome job on the crafting window, by the way. Love it.

Far2Casual:
And just for info, there has been a regression of some sort as the warning for incompatible item when adding rounds is back. 
![image](https://user-images.githubusercontent.com/63196064/194763997-77843871-9fe5-4d37-9bd8-5a0f974fde1e.png)

shemetz:
Reproduced again, [link to discussion on Discord](https://discord.com/channels/880968862240239708/880969253765935174/1069627739591688263).


![image](https://user-images.githubusercontent.com/6516621/215522046-293982f8-7a8b-4067-aab9-ce39d893d77f.png)

There's no `item:subtype:ammo`, even though the Consumable Type is "Ammunition".

PR #3510, opened 2022-08-14
Fix reagant cost computation for alchemical ammunition crafting

Far2Casual:
Quantity for reagant cost was not computed right for alchemical ammunition

### DIFF
--- a/src/module/actor/character/crafting/entry.ts
+++ b/src/module/actor/character/crafting/entry.ts
@@ -69,11 +69,17 @@ export class CraftingEntry implements CraftingEntryData {
     get reagentCost(): number {
         if (!this.isAlchemical) return 0;
 
+        const ammunitionQuantity = this.preparedFormulas
+            .filter((f) => f.isAmmo)
+            .reduce((sum, current) => sum + (current.quantity || 1), 0);
+
         const fieldDiscoveryQuantity = this.preparedFormulas
+            .filter((f) => !f.isAmmo)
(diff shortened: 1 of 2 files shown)


TikaelSol:
For ease of lookup https://github.com/foundryvtt/pf2e/issues/1903

Please retitle this to fit the standard for PRs "If merged this PR will <PR title>", so something like "Enable crafting of basic ammunition"

stwlam:
An alchemist can also use infused reagents to craft alchemical ammunition but doesn't have the ability to craft 10 per reagent.

Far2Casual:
> 



> An alchemist can also use infused reagents to craft alchemical ammunition but doesn't have the ability to craft 10 per reagent.

Re-reading this it seems that you are right, though I would consider that a very RAW interpretation (and possibly an oversight).

What would be the best way to proceed ? Should I try to make that behaviour only if the actor has specifically the Ammunition Crafting feat ?

stwlam:
I think it easily could be intentional: crafting 10 rounds of ammunition with a single reagent is a special perk of using advancd alchemy as a Munitions Crafter. I want to say the CraftingEntry RE has some way to specify batch size, but I'm not sure.

Far2Casual:
I could test for the "munitionsCrafter" selector instead !
image

ebuckle:
At this point I think you're overalpping with https://github.com/foundryvtt/pf2e/pull/3353

Far2Casual:
> 

My bad, wasn't aware. I'll close this PR then.

PR #3511, opened 2022-08-14
Unpersist Crafting Entries and Refactor Predicates

ebuckle:
Functional but messy, will be cleaning things up when I get time and investigating potential issues.


ebuckle:
Removing from Draft, currently missing migration changes to remove Crafting Entry data stored on actors and to transfer entry data to formuals.

PR #3524, opened 2022-08-15
Clean up tag selector classes and make functional in V10

stwlam:
The resistances and weakness selectors need to be merged: will happen in a followup PR

### DIFF
--- a/src/module/system/tag-selector/speeds.ts
+++ b/src/module/system/tag-selector/speeds.ts
@@ -1,10 +1,10 @@
-import { ActorPF2e, NPCPF2e } from "@actor";
-import { LabeledString } from "@module/data";
+import { ActorPF2e } from "@actor";
+import { ErrorPF2e } from "@util";
 import { SelectableTagField, TagSelectorOptions } from ".";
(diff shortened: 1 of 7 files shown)


src/module/system/tag-selector/speeds.ts
```diff
@@ -17,27 +17,30 @@ export class SpeedSelector extends TagSelectorBase<ActorPF2e> {
...
     }
 
-    override getData() {
-        const data: any = super.getData();
-
-        if (this.object instanceof NPCPF2e) {
-            data.hasExceptions = true;
+    override async getData(): Promise<SpeedSelectorData<TActor>> {
+        if (!this.object.isOfType("creature")) {
+            throw ErrorPF2e("The Speed selector is usable only with creature-type actors");
```
CarlosFdez, line 22: Looking at the code, this particular one seems to be non-npcs

CarlosFdez: Ah, I might be misunderstanding hasExceptions. Nevermind.

PR #3525, opened 2022-08-15
Remove split between item sheet getBaseData() and getData()

CarlosFdez:
Removed action sidebar all sneaky like. I checked every sheet type I saw, but we will need some vigilance.


src/module/item/physical/sheet.ts
```diff
@@ -26,22 +26,30 @@ export class PhysicalItemSheetPF2e<TItem extends PhysicalItemPF2e = PhysicalItem
...
             isPhysical: true,
             activations: this.item.activations.map((action) => ({
                 action,
                 id: action.id,
                 base: `system.activations.${action.id}`,
                 traits: createSheetTags(actionTraits, action.traits ?? { value: [] }),
             })),
+
+            // Do not let user set bulk if in a stack group because the group determines bulk
+            bulkDisabled: stackGroup !== undefined && stackGroup !== null && stackGroup.trim() !== "",
```
stwlam, line 52:
```suggestion
            bulkDisabled: !stackGroup?.trim(),
```

CarlosFdez: I really should have tried to understand what it was doing rather than trusting it 🙃

CarlosFdez: Should be !!, I'll force push it in.

PR #3526, opened 2022-08-15
Derive trait whitelist from CONFIG

CarlosFdez:
I'd like to find a way to link validTraits with the type itself, either changing the data types to be the return type of the class, or setting the return type of the getter to match the type (this one might be a bit tougher). If you got an idea I'm all ears though, but I can't really be satisfied with either solution.


src/module/item/kit/sheet.ts
```diff
@@ -17,19 +16,15 @@ export class KitSheetPF2e extends ItemSheetPF2e<KitPF2e> {
     }
 
     override async getData(options?: Partial<DocumentSheetOptions>) {
-        const sheetData = await super.getData(options);
-        const traits = createSheetTags(CONFIG.PF2E.classTraits, sheetData.data.traits);
-
         return {
-            ...sheetData,
+            ...(await super.getData(options)),
```
stwlam, line 20: :open_mouth:

CarlosFdez: That should be correct, but I looked again and I forgot to remove the unnecessary template paths from there, so uh, updated!

PR #3527, opened 2022-08-15
Fix running migrations from console, unnest embedded spells by a level

stwlam:
This reverts commit e17cf6878df11afeff84a5e6071d34544a2e1fea.

Includes sample run on NPC with a wand


CarlosFdez (approved):

packs/scripts/run-migration.ts
```diff
@@ -219,6 +222,9 @@ async function migrate() {
                     }
                     for (const updatedItem of updatedActor.items) {
                         delete (updatedItem.system as { schema?: unknown }).schema;
+                        if (updatedItem.type === "consumable" && !updatedItem.system.spell) {
+                            delete (updatedItem.system as { spell?: unknown }).spell;
```
In3luki, line 226: This only deletes `updateItem.system.spell` if it doesn't exists on the item which seems redundant or am I reading that wrong?

src/module/migration/runner/base.ts
```diff
@@ -79,9 +78,11 @@ export class MigrationRunnerBase {
         for await (const migration of migrations) {
             await migration.updateItem?.(current);
             // Handle embedded spells
-            if (current.type === "consumable" && current.system.spell.data) {
-                this.preMigrateEmbeddedSource(current.system.spell.data);
-                await migration.updateItem?.(current.system.spell.data);
+            if (current.type === "consumable") {
+                await this.preMigrateEmbeddedSpell(current);
```
In3luki, line 82: This runs once for every migrations which could be a lot depending on the current world system schema version

packs/scripts/run-migration.ts
stwlam: It deletes it if it's `null`

PR #3532, opened 2022-08-15
Fix `@Localize` generating malformed HTML

In3luki:
Before:
before

After:
after


src/module/system/text-editor.ts
```diff
@@ -33,7 +37,7 @@ class TextEditorPF2e extends TextEditor {
     static async enrichString(
         data: RegExpMatchArray,
         options: EnrichHTMLOptionsPF2e = {}
-    ): Promise<HTMLSpanElement | void> {
+    ): Promise<HTMLElement | void> {
```
stwlam, line 40: Return `null` instead of `void` or `undefined`

In3luki:
I've rechecked foundry.js and `Promise<HTMLElement | null>` is the expected return type for a custom enricher. Fixed.

PR #3543, opened 2022-08-16
Add traits to the top of item sheets as tagify

CarlosFdez:
image

On an item that doesn't have a valid traits list
image


stwlam:
Wicked

PR #3555, opened 2022-08-17
Apply journal theme to TinyMCE, make styling more reliable

CarlosFdez:
This allows you to target the text content of text page types using `.journal-entry-page.{theme} .text-content` as the selector.  This also fixes support for tiny mce. Targetting `.journal-sheet.{theme}` is how you target the window chrome and it'll work whether or not the current sheet is the main one or a popout page.

Unfortunately the default html vary widely between page in entry, popout prosemirror, and popout tinymce styles. This is an attempt at consolidating them.

image


CarlosFdez:
The typing isn't exactly correct, the return type is either the tinymce editor or the prosemirror one, but since you're rewriting the types, its not important I get them 100% right.

src/module/journal-entry/sheet.ts
```diff
@@ -51,4 +29,24 @@ class JournalSheetStyledPF2e extends JournalSheetPF2e {
...
 
-export { JournalSheetPF2e, JournalSheetStyledPF2e };
+class JournalTextPageSheetPF2e extends JournalTextPageSheet {
+    override async activateEditor(
+        name: string,
+        options: Partial<TinyMCE.EditorOptions> = {},
+        initialContent = ""
+    ): Promise<TinyMCE.Editor> {
+        const editor = await super.activateEditor(name, options, initialContent);
+        console.log(options.target);
```
stwlam, line 39: :mega:

PR #3576, opened 2022-08-20
Move level capping to feat category

CarlosFdez:


src/module/actor/character/feats.ts
```diff
@@ -12,6 +12,7 @@ interface FeatCategoryOptions {
     featFilter?: string;
     supported?: FeatType[];
     levels?: FeatSlotLevel[];
+    level?: number;
```
stwlam, line 15: levels and level?

CarlosFdez: WTB better names. levels are granted slots. level is the cap.

Could rename levels to slots and that might be better. But maybe level should be maxLevel.

stwlam: `level: ValueAndMax`? :slightly_smiling_face:

CarlosFdez: Doesn't quite work. level is basically an override for the actor level, since SoT uses things like that. Its either unspecified or a specific value.

Things that grant levels (such as class) probably shouldn't be responsible for filtering the levels. That's what this change is about before I start doing more....controversial things.

I renamed levels to slots and I think now its good enough.

PR #3577, opened 2022-08-20
Use tagified traits input for compendium browser feat search

stwlam:
Bonus: Fix compendium browser feat search launched from PC sheet

### DIFF
--- a/src/module/actor/character/feats.ts
+++ b/src/module/actor/character/feats.ts
@@ -1,6 +1,7 @@
 import { FeatPF2e, ItemPF2e } from "@item";
 import { ItemGrantData } from "@item/data/base";
 import { FeatType } from "@item/feat/data";
+import { sluggify } from "@util";
 import { CharacterPF2e } from ".";
 import { BonusFeat, GrantedFeat, SlottedFeat } from "./data";
 
@@ -9,7 +10,7 @@ type FeatSlotLevel = number | { id: string; label: string };
 interface FeatCategoryOptions {
     id: string;
     label: string;
-    featFilter?: string;
+    featFilter?: string | null;
     supported?: FeatType[];
     slots?: FeatSlotLevel[];
     level?: number;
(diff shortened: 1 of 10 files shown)


src/module/actor/character/feats.ts
```diff
@@ -42,15 +43,27 @@ class CharacterFeats extends Collection<FeatCategory> {
...
+                ? `hb_${classSlug}`
+                : null;
+
         this.createCategory({
             id: "class",
             label: "PF2E.FeatClassHeader",
-            featFilter: "classes-" + actor.class?.slug,
-            supported: ["class"],
+            featFilter: classTrait ? `traits-${classTrait},traits-archetype` : null,
+            supported: ["class", "archetype"],
```
CarlosFdez, line 66: Interestingly enough, this doesn't do anything. archetype resolves into either skill or class depending on whether it has the skill trait, according to isFeatValid(). You can change that behavior if you want, but we'd end up allowing archetype skill feats in class/archetype feat slots.

src/module/actor/character/feats.ts
```diff
@@ -42,15 +43,27 @@ class CharacterFeats extends Collection<FeatCategory> {
         this.createCategory({
             id: "ancestry",
             label: "PF2E.FeatAncestryHeader",
-            featFilter: "ancestry-" + actor.ancestry?.slug,
+            featFilter: actor.system.details.ancestry?.trait
+                ? `ancestry-${actor.system.details.ancestry?.trait}`
+                : null,
```
CarlosFdez, line 48: Later on it'd be nice to do versatile heritages, but I guess there's no need to rush that yet.

stwlam: Yeah, I thought about that, but it was a detour while on the detour.

src/module/actor/character/feats.ts
stwlam: archetype skill feats are of the "skill" feat type. Incidentally, we don't really need to have a feat type for anything except features.

src/module/actor/character/sheet.ts
```diff
@@ -746,37 +748,37 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
...
         $formulas.find(".daily-crafting").on("click", async () => await this.actor.performDailyCrafting());
 
-        PCSheetTabManager.initialize(this.actor, $html.find<HTMLAnchorElement>('a[data-action="manage-tabs"]')[0]);
+        PCSheetTabManager.initialize(this.actor, $html.find<HTMLAnchorElement>("a[data-action=manage-tabs]")[0]);
 
         // Feat Browser shortcut links
-        $html.find(".feat-browse").on("click", (event) => this.onClickBrowseFeatCompendia(event));
+        for (const link of Array.from(html.querySelectorAll<HTMLAnchorElement>(".feat-browse"))) {
+            link.addEventListener("click", this.onClickBrowseFeatCompendia);
+        }
```
CarlosFdez, line 756: I'm starting to see why jquery still exists. DOM api is really tedious, and node list in general are kinda painful. Is there any lightweight lib you're aware of to not deal with all this garbage?

stwlam: we could acquiesce to the `NodeList#forEach`

src/module/actor/character/feats.ts
CarlosFdez: archetype class feats are of the "class" feat type :)

stwlam: naw

stwlam: cf.,
- https://github.com/foundryvtt/pf2e/blob/master/packs/data/feats.db/sentinel-dedication.json
- https://github.com/foundryvtt/pf2e/blob/master/packs/data/feats.db/steel-skin.json

PR #3601, opened 2022-08-23
Exclude invalid options from cleric/champion deity choices

stwlam:
The deities compendium had to be updated to fill in default categories so that the nedb query would work


SpartanCPA (approved):
Works for me. @CarlosFdez pls look at the document.ts change

PR #3603, opened 2022-08-23
Add option groups to the item create dialog

CarlosFdez:
image

We can always add more later, but physical items is the biggest category.

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -1568,6 +1568,12 @@
             "Condition": {
                 "Flanked": "Flat-Footed (Flanked)"
             },
+            "CreationDialog": {
+                "Categories": {
+                    "Physical": "Physical Items",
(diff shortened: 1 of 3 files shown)


static/lang/en.json
```diff
@@ -1565,6 +1565,10 @@
                 "StowingLabel": "Is Stowing Container?"
             },
             "CannotAddType": "{type} items cannot be added to this actor.",
+            "Categories": {
+                "Physical": "Physical Items",
+                "Other": "Other Items"
+            },
```
stwlam, line 1571: Move to `PF2E.Item.CreationDialog` or the like--top-level entries should be rare

CarlosFdez: I saw there were a few there already, and categories weren't just creation dialog as a concept (if subtypes occur, categories are the super types). But sure.

PR #3606, opened 2022-08-23
Accept roll-option sets in more functions

stwlam:
In the near future everything should be working entirely from sets


CarlosFdez:
I suppose the extra code to convert some arrays to strings will be less than the code to make sure arrays don't have duplicates 🙃

PR #3617, opened 2022-08-24
Make cleanup pass on `CheckModifiersDialog` application

stwlam:


src/module/system/check-modifiers-dialog.ts
```diff
@@ -72,10 +81,19 @@ export class CheckModifiersDialog extends Application {
...
+                if (options instanceof Set) {
+                    options.delete(option);
+                } else {
+                    options?.findSplice((o) => o === option);
+                }
+            } else if (options instanceof Set) {
+                options.add(option);
             } else {
-                this.context.options?.push(`substitute:${substitution.slug}`);
+                options?.push(`substitute:${substitution.slug}`);
```
CarlosFdez, line 96:
```suggestion
                options?.push(option);
```

CarlosFdez:
I think for a proper cleanup path you should probably lock in context.options's type earlier (like in the constructor?), that way you don't need to resolve set vs array in getData()

stwlam:
It's temporary until the next PR.

CarlosFdez:
Then sure, a little regression is ok then. Was wondering because as is this isn't actually a cleanup.

PR #3627, opened 2022-08-24
Relax melee/ranged coercion of darts

### LINKED ISSUE #3626: Chakram has a variant strike, and Stiletto Pen and Tamchal Chakram are missing a variant strike
Chakram is a ranged weapon, but has both a melee and thrown variant.
Stiletto Pen and Tamchal Chakram are melee weapons with the thrown trait, but are missing the thrown variant.
There may be more, but these are the ones I noticed.
image

stwlam:
Closes #3626

### DIFF
--- a/src/module/item/weapon/values.ts
+++ b/src/module/item/weapon/values.ts
@@ -1,12 +1,14 @@
 const WEAPON_CATEGORIES = new Set(["unarmed", "simple", "martial", "advanced"] as const);
 
+const MELEE_OR_RANGED_GROUPS = new Set(["dart", "knife"] as const);
+
 const MELEE_WEAPON_GROUPS = new Set([
+    ...MELEE_OR_RANGED_GROUPS,
     "axe",
     "brawling",
     "club",
     "flail",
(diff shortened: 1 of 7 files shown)


src/module/item/weapon/values.ts
```diff
@@ -1,22 +1,25 @@
 const WEAPON_CATEGORIES = new Set(["unarmed", "simple", "martial", "advanced"] as const);
 
+const MELEE_OR_RANGED_GROUPS = new Set(["dart", "knife"] as const);
```
CarlosFdez, line 3: There's pure ranged knives? Was wondering what this separate property was for and why it can't be in melee weapon groups.

stwlam: Chakram

stwlam: And it is in melee groups

CarlosFdez: Its more of I don't know why there's a separate set if its not used anywhere, but perhaps its for readability. Sure.

PR #3628, opened 2022-08-24
Implement Grant Item RE Form

CarlosFdez:
image

### DIFF
--- a/src/module/item/sheet/base.ts
+++ b/src/module/item/sheet/base.ts
@@ -82,6 +82,13 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
         const validTraits = this.validTraits;
         const hasRarity = !this.item.isOfType("action", "condition", "deity", "effect", "lore", "melee");
 
+        // Activate rule element sub forms
+        this.ruleElementForms = {};
+        for (const [idx, rule] of rules.entries()) {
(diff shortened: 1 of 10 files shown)


stwlam:
What's it look like when an injectable?

CarlosFdez:
Like this, at least for the barbarian
image

CarlosFdez:
I'm double checking, and I think to get anything more reliable, we either need to reconstruct the rule element from the base item sheet or within the form, or have a more robust connected between rule elements objects and rule element sources in the item list. The rules we edit in the item sheet come from item.toObject().system.rules which include the "invalid" ones, which I think the actual rule element list skips over?

src/module/item/sheet/base.ts
```diff
@@ -82,6 +82,13 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
         const validTraits = this.validTraits;
         const hasRarity = !this.item.isOfType("action", "condition", "deity", "effect", "lore", "melee");
 
+        // Activate rule element sub forms
+        this.ruleElementForms = {};
+        for (const [idx, rule] of rules.entries()) {
+            const FormClass = RULE_ELEMENT_FORMS[rule.key ?? ""] || RuleElementForm;
```
stwlam, line 88:
```suggestion
            const FormClass = RULE_ELEMENT_FORMS[rule.key ?? ""] ?? RuleElementForm;
```

PR #3631, opened 2022-08-24
Add extreme ChoiceSet option to oracle curses, update Battle curse predicate terminology

LebombJames:
-Update curse ChoiceSets to return the value `extreme` when the extreme curse is chosen, as opposed to `major`
-Updated all features predicated off major curse to also predicate off extreme curse, as you still suffer your major curse effect when under your extreme curse.
-Renamed rollOption naming scheme, `self:curse` to `oracular-curse:stage`


stwlam:
Refrain from using "self:" on any of these.

stwlam:
> > Refrain from using "self:" on any of these.
> 
> So just `curse:minor`, for example? These had `self:` on already, so should I get rid of that on all of them?

Yep. It might be worth distinguishing it further; e.g., "oracular-curse:stage:minor"

PR #3635, opened 2022-08-25
Defer to core `Hotbar#_onDrop` unless `data.type` is handled by system

stwlam:


CarlosFdez (approved):

CarlosFdez:
Lmao we both did the same thing at the same time. Merge whichever one you prefer. I'm going for a walk.

PR #3637, opened 2022-08-25
Add badges to roll options

CarlosFdez:


src/module/item/condition/index.ts
```diff
@@ -51,11 +50,10 @@ class ConditionPF2e extends AbstractEffectPF2e {
...
 
     /** Set a self roll option for this condition */
-    override prepareActorData(this: Embedded<ConditionPF2e>): void {
-        const slug = this.slug ?? sluggify(this.name);
-        this.actor.rollOptions.all[`self:condition:${slug}`] = true;
+    override prepareActorData(): void {
+        super.prepareActorData();
         if (this.slug === "flat-footed") {
-            this.actor.rollOptions.all["self:flatFooted"] = true;
+            this.actor!.rollOptions.all["self:flatFooted"] = true;
```
stwlam, line 56: Let's get rid of this: it's a backward compatibility holdover for data that has since been migrated.

CarlosFdez: Fair, good to know.

src/module/item/condition/index.ts
```diff
@@ -50,15 +49,6 @@ class ConditionPF2e extends AbstractEffectPF2e {
         systemData.value.value = systemData.value.isValued ? Number(systemData.value.value) || 1 : null;
     }
 
-    /** Set a self roll option for this condition */
-    override prepareActorData(this: Embedded<ConditionPF2e>): void {
-        const slug = this.slug ?? sluggify(this.name);
-        this.actor.rollOptions.all[`self:condition:${slug}`] = true;
-        if (this.slug === "flat-footed") {
-            this.actor.rollOptions.all["self:flatFooted"] = true;
-        }
```
stwlam, line 59: Sorry, I meant just get rid of this part :sweat_smile:

CarlosFdez: But its no longer needed, the superclass handles it. The rest of the method only existed to add this part.

src/module/item/abstract-effect/document.ts
```diff
@@ -1,10 +1,40 @@
 import { ItemPF2e } from "@item";
+import { sluggify } from "@util";
 import { EffectBadge } from "./data";
 
 /** Base effect type for all PF2e effects including conditions and afflictions */
 export abstract class AbstractEffectPF2e extends ItemPF2e {
+    slugNormalized!: string;
```
stwlam, line 7: Maybe call this `rollOptionSlug` or the like and document as such

CarlosFdez: I think the name is fine but its such a small thing so changed. Documentation was needed though.

PR #3654, opened 2022-08-26
Refactor melee sheet events to melee sheet

CarlosFdez:
I did this before I found it was a core issue, but its still an improvement so here we go.

### DIFF
--- a/src/module/item/melee/sheet.ts
+++ b/src/module/item/melee/sheet.ts
@@ -13,6 +13,29 @@ export class MeleeSheetPF2e extends ItemSheetPF2e<MeleePF2e> {
             attackEffects: createSheetOptions(this.getAttackEffectOptions(), item.system.attackEffects),
         };
     }
+
+    override activateListeners($html: JQuery<HTMLElement>): void {
+        super.activateListeners($html);
+
+        // Add Damage Roll
+        $html.find(".add-damage").on("click", (event) => {
+            event.preventDefault();
+            const newKey = randomID();
+            const newDamageRoll = { damage: "", damageType: "" };
+            this.item.update({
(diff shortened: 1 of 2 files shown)


src/module/item/melee/sheet.ts
```diff
@@ -13,6 +13,29 @@ export class MeleeSheetPF2e extends ItemSheetPF2e<MeleePF2e> {
...
     }
+
+    override activateListeners($html: JQuery<HTMLElement>): void {
+        super.activateListeners($html);
+
+        // Add Damage Roll
+        $html.find(".add-damage").on("click", (event) => {
+            event.preventDefault();
+            const newKey = randomID(20);
+            const newDamageRoll = { damage: "", damageType: "" };
```
stwlam, line 24: Maybe give this a default, like 1d4 bludgeoning

src/module/item/melee/sheet.ts
```diff
@@ -13,6 +13,29 @@ export class MeleeSheetPF2e extends ItemSheetPF2e<MeleePF2e> {
...
         };
     }
+
+    override activateListeners($html: JQuery<HTMLElement>): void {
+        super.activateListeners($html);
+
+        // Add Damage Roll
+        $html.find(".add-damage").on("click", (event) => {
+            event.preventDefault();
+            const newKey = randomID(20);
```
stwlam, line 23:
```suggestion
            const newKey = randomID();
```

src/module/item/melee/sheet.ts
CarlosFdez: I think that'll make it more work to enter data, cause now they'll have to delete it right?

stwlam: It'll be valid, at least :shrug:

PR #3670, opened 2022-08-28
Refactor status-effects.ts for V10 TokenHUD API and personal sanity

stwlam:
Bonus: lock badge!

### DIFF
--- a/src/module/system/conditions/manager.ts
+++ b/src/module/system/conditions/manager.ts
@@ -1,4 +1,3 @@
-import { StatusEffects } from "@module/canvas/status-effects";
 import { ConditionSlug, ConditionSource } from "@item/condition/data";
 import { ConditionPF2e } from "@item";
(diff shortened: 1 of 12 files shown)


src/module/system/conditions/manager.ts
```diff
@@ -198,73 +197,69 @@ export class ConditionManager {
...
-        });
+        }
 
         // Iterate the overriding bases.
-        for (const base of overriding) {
+        for (const slug of overriding) {
             // Make sure to get the most recent version of a condition.
-            const overrider = updates.get(appliedConditions.get(base)?._id ?? "") ?? appliedConditions.get(base);
+            const overrider = updates.get(appliedConditions.get(slug)?._id ?? "") ?? appliedConditions.get(slug);
+            if (!overrider) return;
```
CarlosFdez, line 238: perhaps this should be continue instead of return?

PR #3671, opened 2022-08-28
Render temp hp above health bar

CarlosFdez:
image

### DIFF
--- a/src/module/canvas/token/object.ts
+++ b/src/module/canvas/token/object.ts
@@ -123,6 +123,61 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
         return flankingBuddies.some((b) => onOppositeSides(this, b, flankee));
     }
 
+    /** Overrides _drawBar() to also draw pf2e variants of normal resource bars (such as temp health) */
+    protected override _drawBar(number: number, bar: PIXI.Graphics, data: TokenResourceData): void {
+        if (!canvas.dimensions) return;
+
+        const actor = this.document.actor;
+        const isHealth = data.attribute === "attributes.hp" && actor?.attributes.hp;
+
+        if (!isHealth) {
+            return super._drawBar(number, bar, data);
+        }
+
+        const { value, max, temp } = actor.attributes.hp;
+        const healthPercent = Math.clamped(value, 0, max) / max;
+
+        // Compute the color based on health percentage, this formula is the one core foundry uses
+        const black = 0x000000;
+        const color = number
+            ? PIXI.utils.rgb2hex([0.5 * healthPercent, 0.7 * healthPercent, 0.5 + healthPercent / 2])
+            : PIXI.utils.rgb2hex([1 - healthPercent / 2, healthPercent, 0]);
+
+        // Bar size logic stolen from core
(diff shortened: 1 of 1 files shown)


src/module/canvas/token/object.ts
```diff
@@ -1,4 +1,5 @@
 import { TokenDocumentPF2e } from "@module/scene";
+import { Graphics } from "@pixi/graphics";
```
stwlam, line 2:
```suggestion
```

src/module/canvas/token/object.ts
```diff
@@ -123,6 +124,56 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
         return flankingBuddies.some((b) => onOppositeSides(this, b, flankee));
     }
 
+    protected override _drawBar(number: number, bar: Graphics, data: TokenResourceData): void {
```
stwlam, line 127:
```suggestion
    protected override _drawBar(number: number, bar: PIXI.Graphics, data: TokenResourceData): void {
```

stwlam (changes requested):

src/module/canvas/token/object.ts
```diff
@@ -123,6 +124,58 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
...
+            bar.beginFill(color, 1.0).drawRoundedRect(0, healthBarY, healthPercent * this.w, barHeight, 2);
+
+            // Draw the container (outermost border)
+            bar.beginFill(black, 0).lineStyle(bs, black, 1.0).drawRoundedRect(0, 0, this.w, h, 3);
+
+            // Set position
+            bar.position.set(0, number === 0 ? this.h - h : 0);
+        } else {
+            super._drawBar(number, bar, data);
+        }
```
stwlam, line 176: Can this be checked up top and returned early?

CarlosFdez: That'll make it trickier to extend later, but I guess it can be reorganized when it is.

CarlosFdez: I'll push it, but I think it made it trickier to read :(

src/module/canvas/token/object.ts
```diff
@@ -123,6 +123,62 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
...
+    /** Overrides _drawBar() to also draw pf2e variants of normal resource bars (such as temp health) */
+    protected override _drawBar(number: number, bar: PIXI.Graphics, data: TokenResourceData): void {
+        if (!canvas.dimensions) return;
+
+        const actor = this.document.actor;
+        const isHealth = data.attribute === "attributes.hp" && actor?.attributes.hp;
+
+        if (!isHealth) {
+            super._drawBar(number, bar, data);
+            return;
```
stwlam, line 135:
```suggestion
            return super._drawBar(number, bar, data);
```

CarlosFdez: Surprised you're ok with return void. I know core does it, but it always felt like a bad idea. That might be my previous programming exp talking.

stwlam (approved):

PR #3674, opened 2022-08-28
Implement option to disable sorting on ChoiceSet

LebombJames:
Currently, ChoiceSets always sort choices, which is fine most of the time, but sometimes you want an arbitrary order. A good example of this is Oracle Curses - it would make more sense if they were sorted Minor -> Extreme.

This merge adds a `"sort"` boolean to the ChoiceSet RE, which defaults to true. When set to false, the prompt will follow the order declared in the ChoiceSet RE itself.

Left is `"sort": true` (or omitted), right is `"sort": false`

image

(First real code merge, hope I haven't broken anything 🙏)

### DIFF
--- a/src/module/rules/rule-element/choice-set/rule-element.ts
+++ b/src/module/rules/rule-element/choice-set/rule-element.ts
@@ -32,6 +32,9 @@ class ChoiceSetRuleElement extends RuleElementPF2e {
     /** An optional roll option to be set from the selection */
     private rollOption: string | null;
 
+    /** Should the options on the prompt be sorted alphabetically? */
+    private sort: boolean;
+
     constructor(data: ChoiceSetSource, item: Embedded<ItemPF2e>, options?: RuleElementOptions) {
         super(data, item, options);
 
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Personal opinion is that explicit choice sets should not be sorted, and queried ones should be. But this is shark's baby, so up to him to decide.

src/module/rules/rule-element/choice-set/rule-element.ts
```diff
@@ -42,6 +45,7 @@ class ChoiceSetRuleElement extends RuleElementPF2e {
         this.allowedDrops = isObject(data.allowedDrops) ? new PredicatePF2e(data.allowedDrops) : new PredicatePF2e();
         this.allowNoSelection = !!data.allowNoSelection;
         this.rollOption = typeof data.rollOption === "string" && data.rollOption ? data.rollOption : null;
+        this.sort = !!(data.sort ?? true);
```
stwlam, line 48:
```suggestion
        this.sort = !!data.sort;
```

stwlam:
> Personal opinion is that explicit choice sets should not be sorted, and queried ones should be, rather than introducing a sort parameter. But this is shark's baby, so up to him to decide.

Yeah, I think that'd make for the best results.

CarlosFdez:
#3678 was chosen instead, using the given arbitrary order whenever there is a choice list. While this should solve your issue, thank you for making this PR and bringing up the issue.

PR #3677, opened 2022-08-29
Fix rules json editor vertically overflowing with more lines

CarlosFdez:
This bug also exists in the previous version


stwlam:
What's this fixing, exactly?

CarlosFdez:
image

Close/Apply are supposed to remain visible, not get bumped off the page when it grows too big.

PR #3681, opened 2022-08-29
Fix use of Earn Income macro

xdy:
With this change the earnIncome macro works on master.
Expected result is a popup with options, followed by a chat message with the income earned.
Without this fix the popup is displayed, but no chat message.
I have to admit that I don't understand why this change is required, .has on a Set *should* work. There's probably a better solution.
To test it I used:
```
let pack = game.packs.get("pf2e.pf2e-macros");
await pack.getIndex().then((index) => {
	let id = index.find((e) => e.name === "Earn Income")?._id;
	if (id) {
		pack.getDocument(id).then((e) => e.execute());
	}
});
```

### DIFF
--- a/src/module/system/rolls.ts
+++ b/src/module/system/rolls.ts
@@ -196,6 +196,7 @@ class CheckPF2e {
         const substitutions = context.substitutions ?? [];
 
         // Acquire the d20 roll expression and resolve fortune/misfortune effects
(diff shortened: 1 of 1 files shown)


stwlam:
This looks like bad data is being passed to the roll method.

stwlam:
This doesn't fix the macro but changes code so that the outdated macro js will work

PR #3683, opened 2022-08-30
Fix rule forms failing to receive DOM listeners for rule at index zero

stwlam:


src/module/item/sheet/base.ts
```diff
@@ -328,7 +328,7 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
         const html = $html.get(0)!;
         const ruleSections = html.querySelectorAll<HTMLElement>(".rules .rule-form");
         for (const ruleSection of Array.from(ruleSections)) {
-            const idx = Number(ruleSection.dataset.idx) || -1;
+            const idx = ruleSection.dataset.idx ? Number(ruleSection.dataset.idx) : NaN;
```
CarlosFdez, line 331:
```suggestion
            const idx = Number(ruleSection.dataset.idx);
```

☀️

PR #3685, opened 2022-08-30
Fade out name in encounter if mystified

CarlosFdez:
image


stwlam:
I like the idea of it but wonder if some are not going to appreciate the name being harder to read.

CarlosFdez:
We could try a different value for the transparency, though we'd have to define the variable ourselves. And it might look weird next to the other ones, but maybe that's fine?

That said if the name is harder to read for mystified ones, its harder to read for hidden ones. If we need to change one, ideally we'd change the variable itself so that both are more visible.

PR #3689, opened 2022-08-30
Localize coin abbreviations in `CoinsPF2e#toString`

stwlam:


CarlosFdez:
I think might need to change fromString() as well otherwise editing will break.

stwlam:
That makes this pretty rough.

PR #3698, opened 2022-08-31
Restore "str-damage" selector to NPC damage rolls when appropriate

### LINKED ISSUE #3590: Enfeebled condition does not decrease NPC strike damage.
See image below. The attack roll suffers the status penalty from the enfeebled condition, but the damage roll does not. This problem only applied to NPCs, PCs work correctly.

image

stwlam:
Closes #3590

### DIFF
--- a/src/module/system/damage/weapon.ts
+++ b/src/module/system/damage/weapon.ts
@@ -700,11 +700,18 @@ class WeaponDamagePF2e {
             "strike-damage",
             "damage",
         ];
-        if (weapon.isOfType("melee")) return selectors;
+
+        if (weapon.isOfType("melee")) {
+            if (this.strengthBasedDamage(weapon)) {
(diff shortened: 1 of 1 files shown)


src/module/system/damage/weapon.ts
```diff
@@ -700,18 +700,26 @@ class WeaponDamagePF2e {
             "strike-damage",
             "damage",
         ];
-        if (weapon.isOfType("melee")) return selectors;
+
+        if (weapon.isOfType("melee")) {
+            if (this.strengthBasedDamage(weapon)) {
+                selectors.push("str-damage");
```
CarlosFdez, line 706: you're restoring str-based, so should this be str-based, not str-damage?

stwlam: fixed

PR #3706, opened 2022-08-31
Add shortforms to skill selectors

TikaelSol:
I'm sure I missed some selectors somewhere since these are scattered around and not made using consistent code.  This is needed to simplify REs for certain effects.


stwlam:
We're actually globally moving away from short forms.

PR #3713, opened 2022-08-31
Add icon for overridden conditions in token hud

stwlam:
image


CarlosFdez:
That looks like a remove button.

PR #3722, opened 2022-09-01
Change default of `showRollDialogs` user setting to false

stwlam:


CarlosFdez:
This is for check dialogs right? Isn't that better to be on, given its the only way certain traits and effects (guidance) can work?

stwlam:
> This is for check dialogs right? Isn't that better to be on, given its the only way certain traits and effects (guidance) can work?

I'm going through and adding tooltips to roll buttons. But it's definitely better off.

CarlosFdez:
With it on, its a lot more discoverable that it exists. Some people end up turning it off permanently, but via that process they should at least remember it exists for when they need it on again. With it off, new users will not realize it will exist, will not have any inkling on how to add situational bonuses, until they start digging through wiki guides or asking around on discords.

If anything, the default should be a world setting. I'd rather not have this be yet another thing I have to teach players.

CarlosFdez:
What sort of tooltips are you adding?

CarlosFdez:
Sorry for the third message in a row. If the Roll button on the dialog had focus, so that the enter key would roll it immediately, would that help with its usability?

PR #3738, opened 2022-09-02
Remove Lore Items

ebuckle:
Current implementation of removing lore items to replace with system data:
- Replace lore items with a ShortForm/LoreData object in system
- The ShortForm key is the sluggified lore name
- LoreRE to add new lores to a character
    - These can be given a non-Int ability
    - These are immutable, so the player cannot delete or change the name
    - Added option of locking proficiency, so the play cannot change the proficiency value
    - Ability to have lore name/key to be different values, meaning we can localize known lore names and still have a consistent short form for roll options

Current question for me is how to handle Backgrounds:
- We already have the `trainedLore` field on backgrounds, so the Lore RE could pull from that?
- Some of these fields are in a form like `<Non-Magical> Lore`, should these ask for user input first?

Aside from that, currently missing migration of old Lore items.


CarlosFdez:
I think ideally there wouldn't be a lore re, and it should be baked in as part of whatever gives skill boosts.  We're going to want something smarter for skill boosts that allows alternative skills if you already have it depending on the source anyways.

ebuckle:
True, part of my thinking was to not include the RE in this PR.

But currently that RE is the only way to include Lore skills that don't use Int

ebuckle:
image

^ Added an ability select to address that last comment

ebuckle:
Taking off draft.

Added migration and migration run to hit existing iconics.

Understand this will probably sit for a while due to release issues being more pressing 👍

stwlam:
This isn't going to do anything to address the problem of automating lore skills. Transferring all the problems of lore items to actor data is absolutely not a goal.

CarlosFdez:
How would you handle it? I thought transferring it to actor data (removing an item type and allowing res to affect them) specifically was the goal and this is pretty similar to how I thought it'd end up being (though I'd key with ids instead of slugs honestly for the source data and may not have it as top level data). The only alternative I can really imagine is forcing everything to be via REs and skill boosts, thus preventing someone from adding new lores unless you're clever with the skill boost screen.

ebuckle:
This seemed like the obvious solution but I'm open to implement alternatives if this is a complete no go. 

Key by slug may not be the best solution, exactly. I wanted to have some way of seperating short form and label to allow localization of "known" lores granted by backgrounds, and to keep the paths for REs consistent. But I guess the same can be achieves with id keys.

ebuckle:
Closing, pending some further talk about how best to do this

PR #3744, opened 2022-09-02
Clear search bar in compendium browser when filters are cleared

LebombJames:
The params on `mergeObject` seemed to be backwards: the current `filterData` should be the object being overwritten by the default values. Plus there doesn't seem to be a reason to declare `filterData` as `{ search }` beforehand.

Issue #2448, opened 2022-06-12
Feature Request: Clear all Filters should not remove the search entry


TemplarsKnight:
That in the compendium "Clear all Filters" always also removes the search entry is not workflow optimal. 

I enter a search, see that a filter is still active...want to remove the restriction and bang re-enter my search. With the search you could offer a separate X/Clear to remove that quickly.

**filter != search**

Ty

PR #3747, opened 2022-09-02
Set compendium search box to type "search"

LebombJames:
If we can't have #3744 then can we get this instead 😜


stwlam:
I actually wouldn't mind reversing course on #2448--#3744's implementation was just silly. But there's no inherent difference between input of type "search" and "text"--are you wanting to take advantage of specific browser behavior?

LebombJames:
Type `search` adds an "X" clear button to the search box on Chrome, and assumably on most other browsers. Plus V10 changed all search boxes from `text` to `search`, so if only for consistency's sake.

(For #3744 I somehow convinced myself I had to use `mergeObject`, even if there's, in retrospect, a much easier implementation)

PR #3750, opened 2022-09-03
No longer save Elite/Weak in traits

CarlosFdez:
Fixes their behavior now that traits have been tagified

### DIFF
--- a/src/module/migration/migrations/779-elite-weak.ts
+++ b/src/module/migration/migrations/779-elite-weak.ts
@@ -1,0 +1,17 @@
+import { ActorSourcePF2e } from "@actor/data";
+import { MigrationBase } from "../base";
+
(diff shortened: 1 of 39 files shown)


src/module/migration/migrations/779-elite-weak.ts
```diff
@@ -0,0 +1,16 @@
+import { ActorSourcePF2e } from "@actor/data";
+import { MigrationBase } from "../base";
+
+export class Migration779EliteWeak extends MigrationBase {
```
stwlam, line 4:
```suggestion
/** Store indication of elite/weak adjustment on boolean property instead of as trait */
export class Migration779EliteWeak extends MigrationBase {
```

stwlam:
Title :)

PR #3755, opened 2022-09-03
Add See Invisibility sense, move sense localization to `Actor.Creature`

stwlam:


CarlosFdez:
Looks good to me. Requested someone else for the data part (since I don't remember what standards they're using).

PR #3762, opened 2022-09-04
Automate Crowbar

JeremyHouser:
Added crowbar item penalty rollOption into Force Open action in order to automate -2 item penalty

Added crowbar possession rollOption to Forced Entry feat as this feat allows for ignoring the item penalty from not having a crowbar

Added crowbar possession rollOption into both crowbar and crowbar(levered) items

Has been tested as functional via github actions which I copied from this repo and on a foundry test instance!

### DIFF
--- a/src/module/actor/character/helpers.ts
+++ b/src/module/actor/character/helpers.ts
@@ -107,4 +107,16 @@ class StrikeWeaponTraits {
     }
 }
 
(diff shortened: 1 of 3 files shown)


stwlam (changes requested):
This is better done with an `AdjustModifier` RE that suppresses the penalty.

PR #3768, opened 2022-09-04
Numify creature speeds, defer BaseSpeed REs

stwlam:
I would have converted the whole thing over to a `Record<MovementType, SpeedData>` but will instead do that after the deletion-key bug is fixed.


stwlam:
@CarlosFdez would you mind pinging me before merging this? I'd like to push up the migration result before merging.

CarlosFdez (approved):

Issue #3771, opened 2022-09-04
Request: Make DC module available to macros


tyrielv:
I have some custom macros that I want to make checks with level-based DC. It would be nice to have the DC module available at runtime to custom macros.

tyrielv:
It looks like I may be able to meet my scenario for skill checks using the inline chat option "dc:@self.level" - but that doesn't work for saving throws, which is what I'm actually aiming for.

stwlam:
Copy the code if you'd like, but this isn't public API.

PR #3773, opened 2022-09-04
Expose DC module on game.pf2e

tyrielv:
This enables custom macros to access the functions for calculating DCs (eg level-based, modified by difficulty and/or rarity) in the module at runtime.
#3771 

Please let me know if there's a way to do this already.


CarlosFdez:
What is it you're trying to do specifically? If its for a level based dc as part of a roll formula, I've been considering something like a levelBased() function for rolls that you can do something like `levelBased(@actor.level)+2` with or something.

stwlam:
This isn't API we're ready to make public.

tyrielv:
@CarlosFdez that would work

PR #3787, opened 2022-09-05
Predicate Crafting Entry Batch Sizes

### LINKED ISSUE #1903: Munitions Crafter, Ammunition Crafting

ebuckle:
Alternate to: https://github.com/foundryvtt/pf2e/pull/3353

Resolves #1903

As title, instead of having fieldDiscovery being a trait, instead use a predicate fieldDiscoveryItems in the same way as craftableItems, allowing the problem of munitions crafter to be more cleanly solved.

The only pain with this is the field discovery ae-likes, since they're adding a RawPredicate postDerived and have to be validated in the crafting entry. Thoughts?

As I stated in the other PR, this still leaves a problem of Precious Munitions but that needs a wider fix to actually allow materials to be applied to ammunition.


stwlam:
Rather than refactoring this, we should remove `fieldDiscovery` (and anything else alchemist-specific) from the rule element.

ebuckle:
So yeah, my thinking to make this even more generic was to have an array of of predicate/batch sizes with each formula using the greatest valid batch size or the default.

ebuckle:
Experiment with a batchSize RE to allow any predicate/batch size pairings to be adding to crafting entries.

Not too happy with this as an implementation. Ideally a crafting entry would have all the predicates by default, with new feats adding roll options to an actor rather than adding new predicates later.

Thoughts?

ebuckle:
Crafting entries now have no mention of field discoveries, Alchemist/Munition crafter entries now have all the required predicate knowledge by default

ebuckle:
Updated batch size field to use new predicates

RobertBeilich:
Hi,
so what is the status on this?
#3510 was closed in favor of #3353, which in turn was closed in favor of this pr, but now this one is stale for a month.?

stwlam:
At a glance, all the predicates are malformed.

Far2Casual:
> Hi, so what is the status on this? #3510 was closed in favor of #3353, which in turn was closed in favor of this pr, but now this one is stale for a month.?

Thanks for bumping it up, after a year it gets boring for me to patch it up locally after every version :smile:

src/module/item/physical/document.ts
```diff
@@ -161,6 +161,7 @@ abstract class PhysicalItemPF2e extends ItemPF2e {
             magical: this.isMagical,
             uninvested: this.isInvested === false,
             [`material:${this.material.precious?.type}`]: !!this.material.precious,
+            [`material-grade:${this.material.precious?.grade}`]: !!this.material.precious,
```
stwlam, line 164: Use `material:grade`

src/module/actor/character/crafting/entry.ts
```diff
@@ -14,11 +14,11 @@ export class CraftingEntry implements Omit<CraftingEntryData, "parentItem"> {
...
     craftableItems: PredicatePF2e;
     maxSlots: number;
-    fieldDiscovery?: "bomb" | "elixir" | "mutagen" | "poison";
-    batchSize?: number;
-    fieldDiscoveryBatchSize?: number;
+    defaultBatchSize: number;
     maxItemLevel: number;
     parentItem: Embedded<ItemPF2e>;
+    batchSizes: BatchSizeData[];
+    batchSizeModifier: number;
```
stwlam, line 21: Document all these

ebuckle:
Good for another review.

I'll make a seperate contribution to the wiki detailing the re when I get some time

ebuckle:
Is there anything else missing for this to be merged?

RobertBeilich:
> Is there anything else missing for this to be merged?

@stwlam, any comment on this?

Far2Casual:
@stwlam I'm sad to see this dropped. Ammunition Crafting (lvl 1 feat for Gunslinger) is still broken and requires me to patch my game locally every version for about a year and a half :(

PR #3801, opened 2022-09-06
Fix Localize only returning first paragraph

TikaelSol:
The empty line is back regardless but by only returning the first child we lose the rest of the text.


stwlam:
This would cause fragments to not return at all.

PR #3802, opened 2022-09-06
Fix empty gap created by @Localize and use all sub-elements

CarlosFdez:
Turns out those empty `<p>` tags have always existed, but in foundry v10 they all have a minimum height now.


stwlam:
Isn't it primarily a problem with chat messages?

CarlosFdez:
NPC actions primarily.

PR #3811, opened 2022-09-06
Fix hazard updates when traits are empty

CarlosFdez:


src/module/sheet/helpers.ts
```diff
@@ -32,6 +32,23 @@ function createSheetTags(options: Record<string, string>, selections: SheetSelec
...
+    for (const inputEl of tagifyInputElements.values()) {
+        const path = inputEl.name;
+        const inputValue = data[path];
+        const selections = inputValue && typeof inputValue === "string" ? JSON.parse(inputValue) : inputValue;
+        if (Array.isArray(selections)) {
+            data[path] = selections.map((w: { id?: string; value?: string }) => w.id ?? w.value);
+        }
+    }
+}
+
```
stwlam, line 51: Can this be used in homebrew settings as well?

CarlosFdez: Not sure, I can't find any tagify code other than the onSubmit override. I think homebrew settings use string list tagify's, which do not need any post-processing.

stwlam: It doesn't

CarlosFdez: When I was taking a crack at a campaign feats editor before I underwent burnout, I did add this to the settings menu. It'll probably come back around then.

The one other place this will come in handy is journals.

PR #3815, opened 2022-09-07
Show Granted Item on same line in form

CarlosFdez:
image


CarlosFdez:
There's some housekeeping stuff in there from the flat modifiers progress because I keep having to stash the changes and bring them back, so figured I might as well push up the safer stuff.

PR #3826, opened 2022-09-07
Move RE forms to folder

CarlosFdez:
Another piece of the ongoing flat modifier re stash that keeps having to be restashed


src/module/item/sheet/rule-elements/base.ts
```diff
@@ -0,0 +1,43 @@
...
+    template = "systems/pf2e/templates/items/rules/default.html";
+    constructor(protected item: ItemPF2e, protected index: number, protected rule: TSource) {}
+    async getData() {
+        return {
+            index: this.index,
+            rule: this.rule,
+        };
+    }
+
+    async render() {
```
stwlam, line 24:
```suggestion
    async render(): Promise<string> {
```

PR #3827, opened 2022-09-07
Fix removing Elite/Weak adjustments

CarlosFdez:
As a side effect this also makes the adjustment itself easier to click

### DIFF
--- a/src/module/actor/npc/sheet.ts
+++ b/src/module/actor/npc/sheet.ts
@@ -209,8 +209,8 @@ export class NPCSheetPF2e<TActor extends NPCPF2e> extends CreatureSheetPF2e<TAct
         });
 
         // Adjustments
-        $html.find(".adjustment").on("click", (evt) => {
-            const adjustment = String(evt.target.dataset.adjustment);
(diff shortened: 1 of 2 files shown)


src/module/actor/npc/sheet.ts
```diff
@@ -209,8 +209,13 @@ export class NPCSheetPF2e<TActor extends NPCPF2e> extends CreatureSheetPF2e<TAct
         });
 
         // Adjustments
-        $html.find(".npc-elite-adjustment").on("click", () => this.actor.applyAdjustment("elite"));
-        $html.find(".npc-weak-adjustment").on("click", () => this.actor.applyAdjustment("weak"));
+        $html.find(".adjustment").on("click", (evt) => {
+            const adjustment = String(evt.target.dataset.adjustment);
```
stwlam, line 213:
```suggestion
        $html.find(".adjustment").on("click", (event) => {
            const adjustment = String(event.target.dataset.adjustment);
```

CarlosFdez: Whoops. Was in a rush so I fell back to old habits!

PR #3840, opened 2022-09-08
Change default order-by for most compendium browser tabs to level

stwlam:


src/module/apps/compendium-browser/index.ts
```diff
@@ -310,10 +310,8 @@ class CompendiumBrowser extends Application {
         }
 
         // Sorting
-        if (this.initialFilter.orderBy) {
-            currentTab.filterData.order.by = this.initialFilter.orderBy;
-            currentTab.filterData.order.direction = this.initialFilter.orderDirection ?? "asc";
-        }
+        currentTab.filterData.order.by = this.initialFilter.orderBy ?? "name";
```
CarlosFdez, line 313: is this meant to be name?

stwlam: It's a safe default if not set since not all item types have levels

PR #3849, opened 2022-09-08
Add strike getter to chat message for module use

CarlosFdez:
Seems to continue working for spells and strikes. Interestingly enough, the original logic resolved spells to unarmed strikes, so I had to add an extra check.


stwlam:
Not too crazy about having all chat messages have a getter just for strikes, but ok :disappointed:

CarlosFdez:
The reason is for AA. They have a rough time grabbing RE strikes from the chat message, or alt usage strikes. Hacks abound.

PR #3851, opened 2022-09-09
Add 'send to chat' to action items

### LINKED ISSUE #1418: Feature Request: Add "Send to Chat" button to item sheets.
In GitLab by @SpartanCPA on May 12, 2021, 22:59

image

If possible, add a "Send Description to Chat" button in an open area of an item card. I've underlined 2 areas that might work in the header.

image

At the moment, the only way to send an item's description is to put the item on a sheet and then click the image to send to chat. That makes it difficult to send one-off items or features that I a player may not always want on their actions tab, as they are infrequently applicable (like Grab an Edge, or Fly)

GravenImageRD:
image

Closes #1418

Had to add some plumbing for posting to chat from the compendium entries rather than a copy added to a sheet. If there's a better way of handling that, lemme know.

### DIFF
--- a/src/module/item/action/sheet.ts
+++ b/src/module/item/action/sheet.ts
@@ -57,7 +57,7 @@ export class ActionSheetPF2e extends ItemSheetPF2e<ActionItemPF2e> {
     override _getHeaderButtons(): ApplicationHeaderButton[] {
         const buttons = super._getHeaderButtons();
         buttons.unshift({
-            label: "",
+            label: game.i18n.format("PF2E.NPC.SendToChat"),
             class: "send-to-chat",
             icon: "fas fa-comment-alt",
             onclick: () => {},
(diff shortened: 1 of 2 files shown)


TikaelSol:
Please rename according to the project guidelines, PR titles should follow the pattern `If merge this PR will <title>`, so in this case something like `Add 'send to chat' button to action items`. Also you'd want to say the magic words: Closes #1418

src/module/item/action/sheet.ts
```diff
@@ -33,6 +34,35 @@ export class ActionSheetPF2e extends ItemSheetPF2e<ActionItemPF2e> {
...
+                };
+                ChatMessagePF2e.create(chatData, { renderSheet: false });
+            });
+        });
+    }
+
+    override _getHeaderButtons(): ApplicationHeaderButton[] {
+        const buttons = super._getHeaderButtons();
+        buttons.unshift({
+            label: "",
```
link3333, line 60: Would `game.i18n.format(PF2E.NPC.SendToChat)` be a good label?  I'm assuming based on the type, Foundry doesn't support a tooltip here.

GravenImageRD:
Preview of what it looks like with the label
image

stwlam:
I wouldn't say action items among others have anything unique to them that warrant this treatment

stwlam:
I see the use of a send-to-chat button in items, but there's nothing special about actions that make the feature more useful than other types.

PR #3853, opened 2022-09-09
Sort item traits in sheet header

CarlosFdez:
There's a small delay where the traits visibly shift around after adding it, since tagify will add the new tag then foundry will update. I think that should be fine 🤷 .


stwlam:
I think the solution here is to send tagify pre-sorted objects

CarlosFdez:
That's what the PR does though, it replaces what tagify receives with the keys of the value sorted traits.

stwlam:
What introduced the visible shift?

CarlosFdez:
The delay between tagify adding the new entry and updating the input, and the sheet re-render. Tagify adds the newly added tag to the end, then foundry picks it up and starts to re-render process. On the next sheet re-render, its sorted again.

PR #3857, opened 2022-09-09
Show the correct item name in parentheses when mystified

GravenImageRD:
Before:
image

After:
image

### DIFF
--- a/src/module/item/physical/index.ts
+++ b/src/module/item/physical/index.ts
@@ -234,15 +234,13 @@ abstract class PhysicalItemPF2e extends ItemPF2e {
         super.prepareDerivedData();
 
         const systemData = this.system;
-        if (!systemData.identification.identified) {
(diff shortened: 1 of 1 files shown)


src/module/item/physical/index.ts
```diff
@@ -234,13 +234,15 @@ abstract class PhysicalItemPF2e extends ItemPF2e {
...
-        };
+        if (!systemData.identification.identified) {
+            systemData.identification.identified = {
+                name: this.name,
+                img: this.img,
+                data: {
+                    description: { value: this.description },
+                },
+            };
+        }
```
stwlam, line 245: This is due to a core bug in which data preparation is done twice.

```suggestion
        systemData.identification.identified ??= {
            name: this.name,
            img: this.img,
            data: {
                description: { value: this.description },
            },
        };
```

PR #3859, opened 2022-09-09
Add levels to actors in actor directory

### LINKED ISSUE #3688: Feature Req: Show N/PC Level in Actors list
It would be useful if an Actor's Level was displayed in the list of Actors. 

Specifically, I'd love to have it when I'm designing an encounter with a set of known NPCs/monsters, and adding extra lackeys to help fill it out.

Something like this mockup, maybe?

image

GravenImageRD:
Closes #3688 

This is how it looks right now..
image

Doesn't work when popped out (looking into that) but wanted to start getting feedback on presentation. Thinking I should try to get it to where the name+level is vertically aligned to the center. Right now the name is, and the level is under, but that forces a pretty small font size. Maybe fine? Let me know how y'all feel.

### DIFF
--- a/src/styles/_globals.scss
+++ b/src/styles/_globals.scss
@@ -321,3 +321,14 @@ i[data-pf2-repost] {
 a[href]:hover {
     text-shadow: 0 0 8px var(--color-text-hyperlink);
 }
+
+/* ----------------------------------------- */
+/* Actor Directory                           */
+/* ----------------------------------------- */
+
+.directory-item.actor .actor-level {
+    display: flex;
+    line-height: normal;
+    margin-top: -16px;
+    font-size: x-small;
+}
--- a/src/module/apps/compendium-browser/index.ts
+++ b/src/module/apps/compendium-browser/index.ts
@@ -949,6 +949,14 @@ class CompendiumBrowser extends Application {
             ev.preventDefault();
             this.openTab("bestiary");
         });
+
+        // Add levels to actors.
+        $html.find(".actor").filter(":not([level-added])").each((_,actor) => {
+            const documentId = $(actor).attr("data-document-id");
+            $(actor).attr("level-added", "true");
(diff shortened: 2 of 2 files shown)


GravenImageRD:
Updated sample
image

stwlam:
We subclass `Actors`

GravenImageRD:
Not sure what that comment is supposed to mean in this context, may just be being thick headed though. For what it is worth, I created an actor of each type to test it and any of the ones that had levels worked, and those without looked fine as well.
image

stwlam:
You're using a hook to add HTML to the actor directory, when the actor directory is a system subclass. You're also, for some reason I can't make out, having this code live in the compendium browser.

GravenImageRD:
Ah, well, I was just following the example of how the Bestiary Browser injection was being handled. Looking at a module that added permission markers to entries they were also doing injection via a hook similar to this, so I didn't really think much past that. I'll take it back to the drawing board.

src/styles/_globals.scss
```diff
@@ -321,3 +321,20 @@ i[data-pf2-repost] {
...
+        line-height: normal;
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+    }
+
+    .actor-level {
+        font-size: x-small;
+    }
+}
```
stwlam, line 340: Don't add non-global styling to the global sheet

static/templates/actor-directory/actor-document-partial.html
```diff
@@ -0,0 +1,6 @@
+<li class="directory-item document {{@root.documentCls}} flexrow" data-document-id="{{this.id}}">
```
stwlam, line 1: put this in static/templates/sidebar

stwlam:
:thinking:

stwlam:
Reuse the grayer color seen in the compendium directory :slightly_smiling_face: 

image

PR #3863, opened 2022-09-09
Fix Fix Compendium Browser pack selection not being saved

In3luki:


src/module/apps/compendium-browser/index.ts
```diff
@@ -222,11 +222,7 @@ class CompendiumBrowser extends Application {
     }
 
     loadSettings(): void {
-        const settings = game.settings.get("pf2e", "compendiumBrowserPacks");
-        // Only override the generated list if the setting is not empty
-        if (Object.keys(settings).length) {
-            this.settings = settings;
-        }
+        this.settings = game.settings.get("pf2e", "compendiumBrowserPacks");
```
stwlam, line 225: I think we can kill this method now. :grin:

PR #3868, opened 2022-09-09
Use `progress` element for XP bar

stwlam:
Before:
xp-before

After:
xp-after


CarlosFdez:
Does this break in firefox? I checked MDN and it seems those pseudo elements are unsupported.

stwlam:
Forgot to add its selector, but here it is with it:
image

PR #3870, opened 2022-09-09
Keep darkvision mode always on

stwlam:


stwlam:
It's mostly inconsequential for the system, and dev8675309 requested it for PV compatibility.

Issue #2450, opened 2022-06-12
Feature Request: Alchemical Sciences Methodologies


TemplarsKnight:
Currently, the Investigator lacks automation for the Alchemy subclass. It's ok like it is, but it takes effort to track all the elixirs during the day. With higher levels, the effort increases accordingly, because you have more vials. Would be great if there could be a support here similar to the alchemist. Thanks

![image](https://user-images.githubusercontent.com/107361938/173243210-cad35399-314c-4311-bb9e-c07998d94e58.png)

PR #3872, opened 2022-09-10
Add support for Investigator's versatile vials

ebuckle:
Resolves https://github.com/foundryvtt/pf2e/issues/2450

image

Add tracking for versatile vials resource with REs on relevant investigator features, as well as a toggle for the Quick Tincture ability, simiar to existing Quick Alchemy.


ebuckle:
Just pending a fix to the migration, probably tomorrow since it's late

ebuckle:
All good 👍

src/module/actor/character/data/sheet.ts
```diff
@@ -34,18 +34,25 @@ export interface CraftingEntriesSheetData {
...
+    resources: {
+        infusedReagents: {
+            value: number;
+            max: number;
+        };
+        versatileVials: {
+            value: number;
+            max: number;
+        };
+    };
```
stwlam, line 55: This is isn't going to work. We're not hard-coding every advanced-alchemy-like feature ever published and to be published.

ebuckle: Attempt made to make this more generic.

Should I also extend this to crafting resources not being strictly typed?

PR #3883, opened 2022-09-10
Fix error when enriching null strings

### LINKED ISSUE #3873: Select NPCs Won't Open
image

Any way to recover these or am I going to have to rebuild affected NPCs from scratch? Most NPCs are not affected, but I cannot find what might have been in common between characters that are having this issue.

CarlosFdez:
The built in text editor enrich is able to handle null. Our override was not. Fixes #3873 

Though I'm uncertain if we even need this bit of code here.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -16,10 +16,10 @@ class TextEditorPF2e extends TextEditor {
     static override enrichHTML(content?: string, options?: EnrichHTMLOptionsPF2e): string;
     static override enrichHTML(
         this: typeof TextEditor,
-        content = "",
+        content?: string,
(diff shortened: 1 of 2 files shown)


src/module/system/text-editor.ts
```diff
@@ -16,10 +16,10 @@ class TextEditorPF2e extends TextEditor {
     static override enrichHTML(content?: string, options?: EnrichHTMLOptionsPF2e): string;
     static override enrichHTML(
         this: typeof TextEditor,
-        content = "",
+        content?: string,
```
stwlam, line 19:
```suggestion
        content?: string | null,
```

stwlam:
:smiley:

PR #3890, opened 2022-09-10
Automate tiefling feats

Shandyan:
Plus a missed tengu feat, an aasimar feat, adding the missed strikes onto a conrasu feat (oops), and adding icons to an automaton strike feat.

### DIFF
--- a/packs/data/feats.db/form-of-the-fiend.json
+++ b/packs/data/feats.db/form-of-the-fiend.json
@@ -1,0 +1,158 @@
+{
+    "_id": "eUyZBi8vV5QDxOXD",
+    "img": "systems/pf2e/icons/features/feats/feats.webp",
+    "name": "Form Of The Fiend",
+    "system": {
+        "actionCategory": {
+            "value": ""
+        },
+        "actionType": {
+            "value": "passive"
+        },
+        "actions": {
+            "value": null
+        },
+        "description": {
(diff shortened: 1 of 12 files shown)


packs/data/feats.db/form-of-the-fiend.json
```diff
@@ -0,0 +1,158 @@
...
+                        "value": "jaws"
+                    },
+                    {
+                        "label": "PF2E.BattleForm.Attack.Tail",
+                        "value": "tail"
+                    }
+                ],
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Tiefling.FormOfTheFiend.Prompt",
+                "rollOption": "formfiend"
```
TikaelSol, line 49: Should probably make this `form-fiend`.

static/lang/re-en.json
```diff
@@ -1530,6 +1530,11 @@
                     "Prompt": "Select an element."
                 }
             },
+            "Tengu": {
+                "TenguWeaponFamiliarity": {
+                    "Prompt": "Select a weapon."
```
TikaelSol, line 1535: Do we really not have these prompts elsewhere? I think there's a general prompt section.

Shandyan: I can only find 'Select a weapon group' in re-en.json, which isn't right for this feat

packs/data/feats.db/form-of-the-fiend.json
Shandyan: Oh yes. Done

static/lang/re-en.json
TikaelSol: Let's move them under SpecificRule.Prompt so everything can use them.

Shandyan: Looking again this morning, there's a specific weapon prompt right above the weapon group one. So the tengu feat now points to that, and I've deleted the tengu weapon prompt options.

static/lang/re-en.json
```diff
@@ -1544,6 +1544,11 @@
                 },
                 "TalismanEsoterica": "Talisman Esoterica"
             },
+            "Tiefling": {
+                "FormOfTheFiend": {
+                    "Prompt": "Select an unarmed attack."
```
TikaelSol, line 1549: I think this one can move to the generic section too.

PR #3896, opened 2022-09-10
Fix strike drops on hotbar

stwlam:


CarlosFdez:
This one might need a rebase? It has compendium stuff in there.

PR #3914, opened 2022-09-11
Add missing Dread runes

TikaelSol:


TikaelSol:
I missed one of the things I need to change I think, the material can't be set on items despite appearing in the menu.

stwlam:
Is the material part of the rune?

PR #3920, opened 2022-09-11
Fix distribute coin popup

In3luki:
This just popped up in my game. The fix is adding the `value` back in the template and the rest is a bit of clean up.


stwlam:
Is this going to be an array if there is only one actor?

In3luki:
Good catch. It's a string if there is only on viable actor.

In3luki:
I've restored the `_onSubmit` override.

PR #3921, opened 2022-09-11
Update Advanced Synergy to be taken multiple times

avagdumortesin:


stwlam:
Bring source in line with other items from the same book

PR #3928, opened 2022-09-12
Remove stored identified data on physical items

stwlam:
No identification data is in compendium JSON, so no internal migration needed.


src/module/migration/migrations/786-remove-identified-data.ts
```diff
@@ -0,0 +1,24 @@
+import { ItemSourcePF2e } from "@item/data";
+import { isPhysicalData } from "@item/data/helpers";
+import { IdentificationStatus } from "@item/physical/data";
+import { MigrationBase } from "../base";
+
+/** Convert kit and ABC item pack/id references to UUIDs */
```
CarlosFdez, line 6: 🌞

PR #3930, opened 2022-09-12
Add Flat Modifier Form (dev only)

CarlosFdez:
Dev only so I can get more feedback (and so urgency drops a tad if any issues arise). Once data entry has used it enough with incident and its had maybe or two more revisions, we can publish it for real (hopefully for 4.1).

image


CarlosFdez:
So the height difference between value and selector is actually because details-container-flex-row imposes a font size, which makes no sense for it to do that. Removing that might require some careful analysis to not break any other styles that use it.

PR #3934, opened 2022-09-12
Add optional `force` property to `FlatModifier` RE and `ModifierPF2e`

stwlam:
Alternative to a suppressing `AdjustModifier` RE


CarlosFdez:
Does it make sense for force to be used for anything other than ability type flat modifiers?

stwlam:
> Does it make sense for force to be used for anything other than ability type flat modifiers?

Listen, I know you hate homebrew, but we have to open things sometimes

PR #3955, opened 2022-09-13
Remove data entry world fixes line from change log

stwlam:


VestOfHolding:
Looks like more changes than the change log ended up in here.

PR #3959, opened 2022-09-13
Add support for perpetual infusions

### LINKED ISSUE #2043: Perpetual Infusion choices on crafting tab, and symbols for it and signature items
In GitLab by @danielagardeazabal on Jan 29, 2022, 08:35

Using Quick Alchemy to create Perpetual Infusions / Potency choices, it should not spend Infused Reagents.

In addition, there should be some visual on that tab to track which items are Perpetuals, similar to the star that marks signature items. Since signature items, don't matter beyond level 4 (thanks to Field Discovery), it makes more sense for the star to represent Perpetual items, which matter for most levels (7 and up). Although it would be better if both abilities has symbols.

Also, both Signatures and Perpetuals matter at the level of the formulas themselves, so the visual symbols indicating them should be in the formulas section. For example, a signature item's star is currently shown in the Advanced Alchemy part. Thus, if the PC decides to not prepare that item for even one day, the information about which item is signature is lost.

ebuckle:
Resolves #2043

Changes:
- Fix Herbalist Dedication incorrectly granting Alchemical Crafting
- Send item link to chat when formula image is clicked
- Add support for Perpetual Infusions
	- Added RollOption formatting to ChoiceSet RE to maintain consistency with RollOption RE
	- Highlights perpetual items on formula list w/ tooltip to explain perpetual infusions
image


enumag:
Can this be merged? I'd like to use Perpetual Infusions.

enumag:
@stwlam Why was this closed without merging?

CarlosFdez:
Its 8 months out of date

PR #3967, opened 2022-09-14
Student of the staff deadly scaling

dmarcuse:
This fixes some oversights in my initial implementation of Student of the Staff automation, avoiding console errors when not in arcane cascade and correctly scaling up the extra critical damage dice with the weapon striking runes as the deadly trait is meant to do.

### DIFF
--- a/packs/data/classfeatures.db/arcane-cascade.json
+++ b/packs/data/classfeatures.db/arcane-cascade.json
@@ -28,6 +28,13 @@
             {
                 "key": "GrantItem",
                 "uuid": "Compendium.pf2e.actionspf2e.Arcane Cascade"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.arcaneCascade.damageType",
+                "priority": -100,
+                "value": ""
             }
         ],
         "source": {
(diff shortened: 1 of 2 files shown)


stwlam:
I think we can do this better by making strike adjustments work for damage rolls.

TikaelSol:
What's the status on this? The conflicts will at least need to be cleared up (just need migrations ran on it for the new predicate structure).

@stwlam how hard would it be to get this working with AdjustStrike on damage rolls? Do you mean something like this to override the damage type added by deadly?

```json
{
    "key": "AdjustStrike",
    "mode": "override",
    "property": "damageType",
    "value": "{actor|flags.pf2e.arcaneCascade.damageType}",
    "slug": "deadly-d6",
    "predicate": [
        "whatever"
    ]
}
```

dmarcuse:
I'm happy to update this PR to use whichever approach is best, as long as someone can point me at some documentation, an example, or some code I can use as reference

TikaelSol:
Yeah, just want some clarification on what shark proposed really. For now rebase and clear those conflicts.

stwlam:
I'll fix the bug preventing the weapon trait from having an effect.

stwlam:
Actually, that it changes the damage type of the deadly die is an additional complication we're not setup to handle. Go ahead and merge this.

TikaelSol:
Alright @dmarcuse sorry about the long delay here. Fix the conflicts and I'll merge it.

dmarcuse:
@TikaelSol rebased and fixed! While working on this I also noticed that the migration from `Note` to `CriticalSpecialization` rule elements seems to have accidentally applied critical specialization for all clubs rather than just staves, so I've added a commit fixing that as well.

PR #3969, opened 2022-09-14
Prune prototype tokens on exctract except to preserve differing names

stwlam:


CarlosFdez:
no more non-default compendium images?

stwlam:
We have no non-default compendium images.

CarlosFdez:
I assume they're gone then. I remember seeing some silly images that some creatures used.

PR #3979, opened 2022-09-14
Fix Cover effect

TikaelSol:


stwlam:
`coverType`? :slightly_smiling_face: 

image

stwlam:
The main problem with equating each cover type with its bonus is it's true except when it ain't.

TikaelSol:
It's just the flag name, it can be whatever we like.

stwlam:
Yeah but the values are also numeric :slightly_frowning_face:

PR #3980, opened 2022-09-14
Move spell events and spell preparation form to creature

CarlosFdez:


CarlosFdez:
I've converted some of the jquery but not all, and that's why line count has increased.

PR #3986, opened 2022-09-15
Optional-ize Spell collection

CarlosFdez:
Everything should still work the same since nothing actually sets the spell collection to null currently. The structure of the result of getSpellData() is starting to get creaky at the seams, but I'd rather not break compatibility with TAH for no good reason.


stwlam:
No it ain't!

CarlosFdez:
One might say all of pf2e is optional, even the game itself.

PR #3993, opened 2022-09-15
Make tagify styles global

CarlosFdez:


stwlam:
Seems to warrant a directory if there are two related files sitting at the top level

CarlosFdez:
That's a directory for so little code. Merged the two files instead. Needed a comment to clarify that tags in tagify is not necessarily tags in pf2e 🙃

PR #4002, opened 2022-09-16
Fix all Bestiary instances of Talking Corpse with missing save

Far2Casual:


SpartanCPA:
Please fully replace the spell with a fresh compendium copy, instead of just editing the one line item. More things may be missing we don't know about.

PR #4010, opened 2022-09-16
Register system fonts with Foundry font management

stwlam:


stwlam:
Leaving action icons alone for now

PR #4017, opened 2022-09-17
Add support for tremorsense in RBV

stwlam:
image

### DIFF
--- a/src/module/canvas/perception/modes.ts
+++ b/src/module/canvas/perception/modes.ts
@@ -47,6 +47,32 @@ class HearingDetectionMode extends DetectionMode {
     }
 }
 
+class DetectionModeTremorPF2e extends DetectionModeTremor {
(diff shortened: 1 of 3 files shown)


CarlosFdez (approved):
I think the implementation is fine, so feel free to merge as is, however I suspect that you might need to handle precise tremorsense differently maybe (if that somehow exists).

Shandyan:
> I think the implementation is fine, so feel free to merge as is, however I suspect that you might need to handle precise tremorsense differently maybe (if that somehow exists).

Beastkin can get precise tremorsense via [https://2e.aonprd.com/Feats.aspx?ID=2493](https://2e.aonprd.com/Feats.aspx?ID=2493)

stwlam:
The handling of it is in hidden/concealed checks, which aren't in yet.

Issue #4021, opened 2022-09-17
Effect:Blaze uses wrong predicate


shemetz:
It has "not: fire" instead of "all: fire".

PR #4023, opened 2022-09-17
Fix Blaze effect

shemetz:
Fixed Blaze effect by replacing "not" with "all", in the predicate.


shemetz:
This is my first PR here, let me know if I'm doing anything wrong.

Fixes: #4021

PR #4024, opened 2022-09-17
Allow Herbalists to use Nature on Crafting Checks

### LINKED ISSUE #2373: Crafting items for PC with Herbalist Dedication does not use nature to craft herbal items
Not sure on the best way to automate this, perhaps it should use the higher of nature or crafting?
image

"You can create remedies and other herbal products. You gain the basic alchemy benefits, though they apply only for **herbal items: alchemical items with the healing trait, plus antidote and antiplague**. You gain batches of infused reagents per day equal to your level or half your level if you didn't make your daily preparations in the wilderness. Your advanced alchemy level for creating these is 1 and doesn't increase on its own. You become an expert in Nature and **can use Nature instead of Crafting to Craft herbal items**. You don't need to be trained in Crafting or have the Alchemical Crafting feat to do so, and you can use healer's tools instead of alchemist's tools."

ebuckle:
Resolves #2373

Small hard-coded change to allow Herbalists to use Nature in place of Crafting for certain items.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -647,6 +647,20 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
                 scope: "check",
             };
 
+            const skill = () => {
+                const rollOptions = this.actor.getRollOptions(["all"]);
+                if (rollOptions.includes("feat:herbalist-dedication")) {
+                    if (
+                        formula.item.isAlchemical &&
+                        formula.item.traits.has("healing") &&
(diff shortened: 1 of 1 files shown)


TMunsie:
I'd prefer that we find another way to do this. Hard coding is the last, worst option.

ebuckle:
Same here, this was more intended as a temporary measure until we have some better way of allowing alternate skills for checks. 

We're not currently handling those in any way right?

ebuckle:
Closing for now, pending a better solution

PR #4030, opened 2022-09-17
Clean up and defer damage dice synthetics

stwlam:


src/module/rules/rule-element/damage-dice.ts
```diff
@@ -1,83 +1,174 @@
...
+
+            if (this.override) {
+                this.override.damageType &&= this.resolveInjectedProperties(this.override.damageType);
+                this.override.dieSize &&= this.resolveInjectedProperties(this.override.dieSize);
+            }
+
+            // damageType?: string;
+            // override?: DamageDiceOverride;
+            // diceNumber?: number;
+
```
CarlosFdez, line 116:
```suggestion
```

PR #4043, opened 2022-09-18
Add finer validation to ActorTraits RE and resolve injectables

stwlam:


src/module/rules/rule-element/actor-traits.ts
```diff
@@ -1,46 +1,60 @@
...
+
         data.add ??= [];
         data.remove ??= [];
-        super(data, item, options);
-        if (!data.add?.length && !data.remove?.length) {
-            this.failValidation("Actor traits rule element requires at least a non-empty add or remove field");
+
+        if (
+            !(Array.isArray(data.add) && Array.isArray(data.remove) && (data.add.length > 0 || data.remove.length > 0))
+        ) {
```
CarlosFdez, line 17: De-morgans actually made it take a while for me to wrap my head around it. Not a good candidate for it methinks. 

But wrapping my head around it, I think this might be wrong. if you have add *and* remove, it'll fail validation, because !(true && true && true) is false.

I think doing some self documenting code with something like the below might make it simpler.
```
const hasAdd = Array.isArray(data.add) && data.add.length > 0;
const hasRemove = Array.isArray(data.remove) && data.remove.length > 0;
if (!hasAdd && !hasRemove) {
```

PR #4051, opened 2022-09-18
Release FlatModifier form and restrict damage category

CarlosFdez:


stwlam:
Checkbox with value and data-dtype?

CarlosFdez:
Where are you finding that?

stwlam:
The category field would be better as that.

CarlosFdez:
Where would splash/persistent be you think later on?

Issue #2323, opened 2022-05-29
Inconsistency in obfuscation of DCs posted to chat


cswendrowski:
In GitLab by @SchnozzleCat on May 29, 2022, 07:52

There seem to be inconsistencies in DCs posted to the chat window.
In all examples below my "Show DCs on attacks and saves" value is set to GM.

If I post a `@Check[type:fortitude|dc:19]` from an item, no information about it is obfuscated in the chat (meaning the item name and DC are shown), even if the permissions for that item are set to `None` for all players. This seems to conflict with the fact that the default `showDC` value is set to `owner`, and the players aren't owning the item in this case.

If I move the same item to an NPC actor sheet not owned by a player, posting the check from the item in the NPCs inventory correctly obfuscates information.

Ideally the information would be hidden unless the player has the permission to see the check DC in the first place on the item's description.

Here is a short video demonstrating this behavior: https://transfer.sh/wgCr5W/brave_4ZXnUQgZ2P.mp4

PR #4056, opened 2022-09-18
fix DC visibility in reposts from journals

scapegoat57:
This PR fixes DCs from reposts from journlas being visible by players. #2323

To work around this GMs could have a token selected before sending the post or they can give themselves a character in their user configuration. 

Implementing the first commit in this PR introduces a slight bug that is fixed by the second commit. After the first commit journals owned by a player will no longer show the DC of inline checks. Before the commit journals always showed the DC, which is also incorrect behavior. After the second commit DCs are only showed in journals if the player is an owner. 

I just started using this system recently with the beginner box so there might be systems that my changes affect that I'm unaware of. I tried to change as little code as possible to get it working. My limited testing shows no regressions with reposts from journals, items, or npcs.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -37,7 +37,8 @@ class TextEditorPF2e extends TextEditor {
     static processUserVisibility(content: string, options: EnrichHTMLOptionsPF2e): string {
         const $html = $("<div>").html(content);
         const actor = options.rollData?.actor ?? null;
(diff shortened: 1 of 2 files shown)


stwlam (changes requested):
Instead of adding another property, just pass the single relevant `document`.

stwlam:
We need to do a release today, sorry. Superseded by #4073

PR #4077, opened 2022-09-20
Remove JQuery from base item sheet

In3luki:
I've removed two listeners that were targeting the lore sheet and should no longer be needed.
I've also removed a listener from the rules editing block that was targeting a `textarea` in the `ProseMirror` editor that doesn't exists as far as I can tell.
Additionally I've moved one listeners to the feat sheet that was only used on that sheet.

### DIFF
--- a/src/module/item/sheet/base.ts
+++ b/src/module/item/sheet/base.ts
@@ -210,11 +210,9 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
         }
 
         const ruleElementSelect = html.querySelector<HTMLSelectElement>("select[data-action=select-rule-element]");
-        if (ruleElementSelect) {
-            ruleElementSelect.addEventListener("change", () => {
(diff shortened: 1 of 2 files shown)


src/module/item/sheet/base.ts
```diff
@@ -206,76 +203,56 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
...
-                rules.splice(index, 1);
-                this.item.update({ "system.rules": rules });
-            }
-        });
+        const ruleElementSelect = html.querySelector<HTMLSelectElement>("select[data-action=select-rule-element]");
+        if (ruleElementSelect) {
+            ruleElementSelect.addEventListener("change", () => {
+                this.selectedRuleElementType = ruleElementSelect.value;
+            });
+        }
```
stwlam, line 217:
```suggestion
        ruleElementSelect?.addEventListener("change", () => {
            this.selectedRuleElementType = ruleElementSelect.value;
        });
```

PR #4097, opened 2022-09-21
Allow multiple actors to rest for the night via macro

stwlam:


CarlosFdez:
I actually have no idea if character over selected tokens is correct. At first I thought not but thinking deeper, yeah players are going to screw up and try to rest while their familiar is selected.

PR #4117, opened 2022-09-21
Reduce width of spell attack/dc data

CarlosFdez:
Bonus: Spell Preparation alignment is a bit cleaner, Focus Pool is easier to interact with.

image


CarlosFdez:
We can also reduce the button size for spell preparation, but this is actually enough space to work with I think for my purposes.

stwlam:
DOES IT DO THE SPINNY

CarlosFdez:
IT DO NOW

stwlam:
I'MA SMASH THAT MERGER SO HARD

PR #4122, opened 2022-09-22
Support class dc for spellcasting features

CarlosFdez:
Requires #4019. It also changes class dc to use statistic (all the inline rolls continue to work, as do automatic prof bumps). Modifiers are also automatically retrieved from the synthetics rather than injected.

This changes spellcasting to stem from a base tradition proficiency. This base tradition proficiency could also be extended with alternative domains to create a counteract statistic in the future (that would not gain a +1 from inspire courage). 

The extend's function signature is *gnarly*. Typescript doesn't seem capable of understanding that StatisticCheckData is an object for the purposes of DeepPartial. Perhaps a typescript update would allow a simpler type. Or maybe you have another idea.


stwlam:
Conflicts!

PR #4127, opened 2022-09-22
Disable editing speeds set by REs

stwlam:


CarlosFdez:
What's the benefit to this one? Have we been getting reports of people tripping over this?

stwlam:
It mirrors senses, preventing ephemeral changes from being stored.

PR #4128, opened 2022-09-22
Default spellcasting entries to invalid statistic

CarlosFdez:
One source of sheet explody when data prep fails (also a better way to handle how the class dc pr does invalid statistics).


CarlosFdez:
I'll include this as part of the class dc one

PR #4137, opened 2022-09-22
Fix overflow for long tag elements

### LINKED ISSUE #2950: Long Prerequisites do not Stay within Bounding box in Feat Sidebar
This should be considered for fix prior to v10.
Screen Shot 2022-07-11 at 11 31 23 AM

tyrielv:
Closes #2950
If a tag is longer than the width allows, currently the text overflows on the bottom but the border doesn't expand with it. This change makes the border expand so it looks right, only if the text requires it.
image

### DIFF
--- a/src/styles/_tags.scss
+++ b/src/styles/_tags.scss
@@ -244,3 +244,13 @@ tags.pf2e-tagify, tags.tags.paizo-style {
         background: rgba(0, 0, 0, 0.05);
     }
 }
+
+// Ensure shadow fills the whole tag even if it wraps lines
+.tagify {
+    --tag-inset-shadow-size: 10em
+}
+
+// Ensure that remove button doesn't get covered up in when window is small
+.tagify__tag__removeBtn {
+    overflow: visible;
+}
--- a/src/styles/_tags.scss
+++ b/src/styles/_tags.scss
@@ -111,6 +111,10 @@ li.chat-message .tags .tag {
     padding: 0.05em;
     gap: 0;
 
+    select.tag {
+        height: 22px;
+    }
+
     tag, .tag, select.tag {
         border: 2px solid var(--color-border-trait);
         margin: 0;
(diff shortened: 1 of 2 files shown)


CarlosFdez:
how does it look like for tagify lists? Can you try making a mix? Do item headers (which use the same styling) still look ok?

tyrielv:
Tagify lists don't seem to be affected - I don't think they use the same styling.

image

For item headers, do you mean like the list of traits? It also expands with this change:
image

stwlam:
Tagify lists will be at the top of most item sheets in V10.

CarlosFdez:
I think its only the tagify lists at the top of every sheet that uses those stylings yeah, if I recall correctly. The paizo-style ones.

CarlosFdez:
Tested locally with the one line change, I think you need to also change the tagify styling a bit. Same file.

image

tyrielv:
I see, I was still testing on foundry 9. 
Testing on latest for tagify tags,
Before: 
before
Set min-height, but it makes the Select tag taller and the color doesn't fill the whole tag:
middle
After setting height for select tag and expanding the shadow:
after

CarlosFdez:
Does it still look good for normal traits? Is the select still the same height as the traits?

tyrielv:
Looks like I got caught up in before/after screenshots and forgot to commit some of the changes. They're back now.
Also fixed the cut-out X, and the shadow for multi-line on the homebrew settings page.
fixed
homebrew

CarlosFdez:
Seems the alignment has gotten off and there's a white gap before the yellow border for some reason. Looks like this in master, whereas its properly aligned

image

If might be due to css specificity, but maybe something is making the select bigger than its supposed to.

tyrielv:
Wow, good eye. I had to zoom way in to see that. I'll take another look later today.

tyrielv:
It looks like the gaps are an artifact of using the browser zoom - they are present in master if browser zoom is 125%
image

and absent in patch-1 with zoom 100%:
image

Dragging the item window around also tends to make it come out more.

Also, I discovered that master sometimes has the issue with the shadow not completely filling in - see the thin grey line when hovering on Android here: 
image

CarlosFdez:
Ah, if its a product of zoom then that's fine. Thanks for putting up with my nitpicking!

tyrielv:
I appreciate the attention to detail.

PR #4139, opened 2022-09-23
Track whether actors emit sound and use in hearing detection tests

stwlam:


CarlosFdez:
I think it might be better to default emits sound to false, then flip the default for creature. Because I think vehicles and siege weapons currently emit sound.

PR #4157, opened 2022-09-24
Fix rule elements on Celestial Armor

TikaelSol:


packs/data/equipment.db/celestial-armor.json
```diff
@@ -68,41 +68,23 @@
...
                 "predicate": {
                     "all": [
                         "origin:trait:fiend"
                     ]
                 },
-                "selector": "ac",
+                "selector": [
+                    "ac",
+                    "saving-throw"
+                ],
```
SpartanCPA, line 79: Can we add two selectors at once like this?

TikaelSol: Yup, it's a relatively new thing that you can make the selector an array. see my other PR for condensing spell effect REs.

PR #4163, opened 2022-09-24
Fix loss of actor context in item card @Checks

### LINKED ISSUE #4162: Class DC-based save button within Action shows as DC 0 when posted to chat
Class DC buttons are not displaying properly when the item card containing them is posted to chat.

Hello!

On Foundry V10 build 286, no modules installed and system version 4.1.3, using a freshly created Fighter actor and a Basic Reflex Save button based on its class DC (`@Check[type:reflex|dc:resolve(@actor.attributes.classDC.value)|basic:true]`)

stwlam:
Item context is still absent: using the core enrichers API means we're never getting an item in the roll data.

Closes #4162

### DIFF
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -239,14 +239,14 @@ export class Statistic {
 
     /** Returns data intended to be merged back into actor data */
     getTraceData(this: Statistic, options: RollOptionParameters = {}): StatisticTraceData {
-        const { check } = this.withRollOptions(options);
+        const { check, dc } = this.withRollOptions(options);
 
         return {
             slug: this.slug,
             label: this.label,
-            value: check?.mod ?? 0,
-            totalModifier: check?.mod ?? 0,
-            breakdown: check?.breakdown ?? "",
+            value: dc.value,
+            totalModifier: check.mod ?? 0,
+            breakdown: check.breakdown ?? "",
             _modifiers: check.modifiers.map((m) => m.toObject()),
         };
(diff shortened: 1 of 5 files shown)


src/module/system/statistic/index.ts
```diff
@@ -239,14 +239,14 @@ export class Statistic {
...
-        const { check } = this.withRollOptions(options);
+        const { check, dc } = this.withRollOptions(options);
 
         return {
             slug: this.slug,
             label: this.label,
-            value: check?.mod ?? 0,
-            totalModifier: check?.mod ?? 0,
-            breakdown: check?.breakdown ?? "",
+            value: dc.value,
```
CarlosFdez, line 247: Is this correct? Most uses of value are StatisticModifier are totalModifier rather than the dc. For example, perception.

stwlam: But not AC or (formerly) Class DC. Note how current PCs are showing their "class modifier" because of that change.

CarlosFdez: Got it, its just an inconsistent mess then. At the very skills/perception use the built statistic rather than trace data.

stwlam: Maybe we should have totalModifier and dc in the trace data?

CarlosFdez: I think this change is fine. It includes both now. I just wasn't aware how inconsistent of a mess statistic modifier usage was.

Issue #4126, opened 2022-09-22
Better Hexagon Map Measument Templates


FolkvangrForgent:
The hex grid measurement templates are lagging quite behind when compared to the current square grid measurements. I would love to see improvement to the current hex grid measurements to better facilitate play on hex grids (hex grids being a variant rule briefly presented on page 14 of the GMG). The particularly problematic templates that I often run into are the cone and the circle which do not properly highlight what would be considered within a hex template. I have also found that the cone snaps to every 45° even on hex grid where it should snap every 30° and the internal angle should also be 60° instead of 90°. I would also love to see the wall collision feature brought to the hex measurement templates. I have attached a picture that I often reference for hex measurement templates and believe exhibits the de-facto measurement rules for hex grids.
![image](https://user-images.githubusercontent.com/25625121/191661388-7230672c-1100-4c2c-ba42-f0de1745cd8a.png)

stwlam:
Unfortunately, there aren't fully defined hex-play rules, and the amount of unanswered questions left by the GMG are such that it isn't something we're able to build and maintain in the system implementation.

CarlosFdez:
That said if someone wants to make a module for it and certain changes can be done to core to better facilitate it somehow (maybe exposing certain methods or restructuring code or something) feel free to hit us up.

FolkvangrForgent:
Just want to note that I opened a commission for a module (in the league of foundry developers) as I currently am too busy and don't like js/ts so hopefully someone will take that up. If I or someone else was to have related core questions would the discord be the best place to reach?

stwlam:
Making some small accommodations as specified here.

PR #4173, opened 2022-09-25
Add logic to snap templates to 30 degree instead of 45 on a hex grid when initially placing a template.

FolkvangrForgent:
Changes:
- Added logic to check if the current scene hasHexGrid and to snap the template to every 30 degrees instead of every 45 in the case that it does. This should complement pr #4143 which addressed issue #4126 .

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -5,8 +5,7 @@ export class TemplateLayerPF2e<
 > extends TemplateLayer<TTemplate> {
     /** Originally by Furyspark for the PF1e system */
     protected override _onDragLeftMove(event: PlaceablesLayerEvent<TTemplate>): void {
-        if (!canvas.scene) return;
-        if (!canvas.dimensions) return;
+        if (!canvas.scene || !canvas.dimensions) return;
 
         // From PlaceablesLayer#_onDragLeftMove
         const preview = event.data.preview;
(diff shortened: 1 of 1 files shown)


src/module/canvas/layer/template.ts
```diff
@@ -33,8 +33,9 @@ export class TemplateLayerPF2e<
             // Update the shape data
             if (["cone", "circle"].includes(template.type)) {
                 const direction = ray.angle;
+                const snapAngle = (canvas.scene.hasHexGrid ? (Math.PI * 1/6) : (Math.PI * 0.25));
```
stwlam, line 36:
```suggestion
                const snapAngle = canvas.scene.hasHexGrid ? Math.PI * (1 / 6) : Math.PI * 0.25;
```

src/module/canvas/layer/template.ts
```diff
@@ -33,8 +33,9 @@ export class TemplateLayerPF2e<
             // Update the shape data
             if (["cone", "circle"].includes(template.type)) {
                 const direction = ray.angle;
+                const snapAngle = Math.PI / (canvas.scene.hasHexGrid ? 6 : 4);
                 template.document.direction = Math.toDegrees(
-                    Math.floor((direction + Math.PI * 0.125) / (Math.PI * 0.25)) * (Math.PI * 0.25)
+                    Math.floor((direction + Math.PI * 0.125) / snapAngle) * snapAngle
```
stwlam, line 38: Make sure this is an integer

Issue #3657, opened 2022-08-26
HTML of resistances and immunities uses "data-weakness"


itamarcu:
I found this in the debugger of HTML elements when opening an NPC actor sheet:

![image](https://user-images.githubusercontent.com/6516621/186880561-36f40bb3-ed28-49fe-9aa9-85a503a6c6cf.png)

Apparently, all three of IWR use "data-weakness" - a misleading name, when only one is about weakness while the other two are about the opposites.

This bug doesn't affect users, but it confused me as a modder.

stwlam:
Nothing is depending on it. If and when it's addressed, the entire section will be refactored.

itamarcu:
Sure thing, when this refactor happens I'll just need to remember to patch https://github.com/itamarcu/pf2e-see-simple-scale-statistics

stwlam:
Feel free to make a PR if you need it fixed for your module.

shemetz:
Seems like this was fixed anyway! :]

PR #4175, opened 2022-09-25
Show resistance exceptions in hazard IWR

CarlosFdez:
image

### DIFF
--- a/static/templates/actors/hazard/partials/sidebar.html
+++ b/static/templates/actors/hazard/partials/sidebar.html
@@ -89,7 +89,7 @@
             </div>
             <div class="tags">
                 {{#each immunities as |immunity slug|}}
-                    <span class="tag" data-slug="{{slug}}">{{localize immunity.label}}</span>
+                    <span class="tag" data-immunity="{{slug}}">{{localize immunity.label}}</span>
                 {{/each}}
             </div>
         </section>
@@ -101,7 +101,7 @@
             </div>
             <div class="tags">
                 {{#each data.traits.dv as |weakness|}}
-                    <span class="tag" data-slug="weakness.type">{{localize weakness.label}} {{weakness.value}}</span>
+                    <span class="tag" data-weakness="{{weakness.type}}">{{localize weakness.label}} {{weakness.value}}</span>
                 {{/each}}
             </div>
         </section>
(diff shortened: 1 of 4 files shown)


static/templates/actors/hazard/partials/sidebar.html
```diff
@@ -89,7 +89,7 @@ <h2>{{localize "PF2E.ImmunitiesLabel"}}</h2>
             </div>
             <div class="tags">
                 {{#each immunities as |immunity slug|}}
-                    <span class="tag" data-slug="{{slug}}">{{localize immunity.label}}</span>
```
stwlam, line 92: I think workshop uses that?

CarlosFdez: Does it? The weakness/resistance ones for hazard were not right at all, and NPC does data-weakness (for all of them....). Should probably make them all the same then. @xdy do you use these slugs, and if so, which one do you use? data-weakness?

stwlam: Maybe you can fix 🙂 
https://github.com/foundryvtt/pf2e/issues/3657

CarlosFdez: I'll wait to see what the response is, but I'm leaning towards data-slug and data-type.

xdy: No, I do it in a createChatMessage hook on a damageroll and then check message.flags?.pf2e?.damageRoll?.types for the attack (or if that's empty, get the original item and check damage.value and get the types) and then compare to traits?.di/dr/dv.value on the target. (In a very messy way. Total rewrite needed.)

CarlosFdez: Sorry, we got you mixed up. Up to @shemetz then, but I'm likely going to at least be adding data-slug. Renaming data-weakness might be too annoying to do right now though?

CarlosFdez: Changed it to data-weakness="slug", data-resistance="slug", and data-immunity="slug".

shemetz: Thanks for notifying me!

PR #4177, opened 2022-09-25
Added several missing action macros to the system

fraserstanton:
Added several missing action macros to the system

Evangelize
Reveal Machinations
No Cause For Alarm
Reveal True Name
Scare To Death
Disturbing Knowledge
Battle Prayer
Sacred Defense


stwlam:
We're not making anymore macros, sorry

PR #4187, opened 2022-09-25
Fix NPC saves displaying dc instead of modifier

CarlosFdez:


static/templates/actors/character/sidebar/class-dc.html
```diff
@@ -8,7 +8,7 @@ <h2>{{#if (gt classDCs.dcs.length 1)}}{{localize "PF2E.Actor.Character.ClassDC.P
     {{#each classDCs.dcs as |classDC|}}
         <div class="roll-data difficulty-checks class-dc{{#unless classDC.primary}} secondary{{/unless}}">
             <h3>
-                {{classDC.value}}
+                {{classDC.dc}}
```
stwlam, line 11: What about the modifiers tooltip?

CarlosFdez: Still seems to work the same
image

PR #4199, opened 2022-09-26
Remove last explicit any type

In3luki:
The removed `varaints` are `Lore` variants that are not used anywhere else in the code as far is I can tell. Also the `New Skill Variant` button on the lore item sheet is non-functional.

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -22,7 +22,6 @@ import {
 } from "@util";
 import { ActorSheetPF2e } from "../sheet/base";
 import { CreatureConfig } from "./config";
-import { SkillData } from "./data";
 import { SpellPreparationSheet } from "./spell-preparation-sheet";
 import { CreatureSheetData, SpellcastingSheetData } from "./types";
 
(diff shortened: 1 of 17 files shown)


src/module/actor/modifiers.ts
```diff
@@ -409,8 +411,32 @@ class StatisticModifier {
...
+    shortform?: string;
+
+    // NPC
+    adjustedHigher?: boolean;
+    adjustedLower?: boolean;
+
+    isLore?: boolean;
+
+    // Hazard
+    base?: number;
```
stwlam, line 439: I think most of these can be defined in the extended interface of `StatisticModifier`. Many already are.

In3luki: There is an interface for `RawModifier` that is implemented by `ModifierPF2e` but I can't find an interface for `StatisticModifier`.

stwlam: The properties you added are on the `StatisticModifier` class.

In3luki:
Updated

src/module/actor/creature/sheet.ts
```diff
@@ -68,13 +67,13 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
 
         // Update skill labels
         if (sheetData.data.skills) {
-            const skills: Record<string, SkillData & Record<string, string>> = sheetData.data.skills;
-            const mainSkills: Record<string, string> = CONFIG.PF2E.skills;
-            for (const key in skills) {
-                const skill = skills[key];
+            const skills = sheetData.data.skills;
+            for (const [key, skill] of Object.entries(skills)) {
+                if (!skill.rank) continue;
```
stwlam, line 72: Some skills have no rank?

In3luki: NPC skill are supposed to have no rank but the prepared skill data still has it with a value of 1. I wasn't sure if every `Creature` subclass defines a skill rank so I thought it would be better to add a check. 
Now that I look at the old code again it shouldn't ever be undefined or it would have blown up already.

PR #4218, opened 2022-09-28
Fix display of uses in inventory

CarlosFdez:
image


stwlam:
Can we not display if there is only 1 maximum use?

CarlosFdez:
1 maximum use is relevant for wands right now unfortunately.

PR #4219, opened 2022-09-28
Utilize `ChatMessage#getRollData`

stwlam:
Once upstream uses it themselves next release, we'll benefit from it immediately.

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -171,12 +171,17 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
         return game.scenes.get(sceneId)?.tokens.get(tokenId) ?? null;
     }
 
+    override getRollData(): Record<string, unknown> {
+        const { actor, item } = this;
+        return { ...actor?.getRollData(), ...item?.getRollData() };
+    }
+
     override async getHTML(): Promise<JQuery> {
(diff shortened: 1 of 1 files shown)


src/module/chat-message/document.ts
```diff
@@ -171,12 +171,17 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
         return game.scenes.get(sceneId)?.tokens.get(tokenId) ?? null;
     }
 
+    override getRollData(): Record<string, unknown> {
+        const { actor, item } = this;
+        return { ...actor?.getRollData(), ...(item?.getRollData() ?? { actor }) };
```
CarlosFdez, line 176:
```suggestion
        return { ...actor?.getRollData(), ...item?.getRollData() };
```

if I recall, actor.getRollData() already has actor in the roll data now as the document itself, as of the v10 release.

stwlam:
```js
  /**
   * Prepare a data object which defines the data schema used by dice roll commands against this Actor
   * @returns {object}
   */
  getRollData() {
    return this.system;
  }
```

CarlosFdez: Sorry, ActorPF2e has the override. I should have been more specific.

PR #4228, opened 2022-09-28
Allow FlatModifier REs to only (or never) apply to critical damage

### LINKED ISSUE #1271: Flat modifier rule element not respecting critical call
In GitLab by @snake444 on Apr 3, 2021, 18:57

When using a flat modifier for the critical specialization for a pick the flat modifier is not respecting the ****"critical": true**** call. This ends up rolling flat modifier in regular and critical damage, doubling the flat on a critical. {"key":"PF2E.RuleElement.FlatModifier","selector":"greatpick-damage","label":"Critical","predicate":{"all":["rage"]},"critical":true,"value":2} (This is for a raging barabrian

Code used to apply flat modifier: 

Screenshot_from_2021-04-03_16-58-06

stwlam:
Closes #1271

### DIFF
--- a/src/module/rules/rule-element/flat-modifier.ts
+++ b/src/module/rules/rule-element/flat-modifier.ts
@@ -30,6 +30,9 @@ class FlatModifierRuleElement extends RuleElementPF2e {
     /** Whether this modifier comes from equipment or an equipment effect */
     fromEquipment: boolean;
 
+    /** If a damage modifier, whether it applies given the presence or absence of a critically successful attack roll */
+    critical: boolean | null;
+
     constructor(data: FlatModifierSource, item: Embedded<ItemPF2e>, options?: RuleElementOptions) {
         super(data, item, options);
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Are there situations where we want it to show for crit and success, but not double on crit? I remember people asking about that before, but I don't recall.

stwlam:
That'd be something else to handle. I can look around for requests.

CarlosFdez:
I suspect it'd make critical a string value instead of a boolean, so definitely would be worth looking into. If there's no need or the design would be completely different though I think this looks fine.

stwlam:
It'd be easy enough to add a fourth value to indicate including in both but refrain from doubling. Or we could make the `false` value serve that purpose.

CarlosFdez:
I see. Regardless it seems the critical boolean thing is deep seated into the damage dice the more I look at it, so it doesn't hurt to handle it separately anyways.

PR #4232, opened 2022-09-28
Allow hearing detection mode to function when RBV is disabled

stwlam:


CarlosFdez:
I think leaving as is works fine, but it should probably occur after the emitsSound check (since that's an explicit actor type / actor setting thing).

PR #4243, opened 2022-09-29
Make GrantItem onDelete actions configurable

stwlam:


src/module/rules/rule-element/grant-item/base.ts
```diff
@@ -42,6 +45,20 @@ class GrantItemRuleElement extends RuleElementPF2e {
                 : {};
     }
 
+    static ON_DELETE_ACTIONS = ["cascade", "detach", "restrict"] as const;
+
+    #getOnDeleteActions(data: GrantItemSource): Partial<OnDeleteActions> | null {
```
CarlosFdez, line 50: Does Pick<OnDeleteActions, "granter" | "grantee"> work here?

stwlam: That would be equivalent to `OnDeleteActions`

src/module/rules/rule-element/grant-item/helpers.ts
```diff
@@ -0,0 +1,68 @@
...
+                if (!pendingItems.includes(grantee)) {
+                    pendingItems.push(grantee);
+                    await processGrantDeletions(grantee, pendingItems);
+                }
+                break;
+            }
+            case "detach": {
+                // Unset the grant flag and leave the granted item on the actor
+                await grantee.update({ "flags.pf2e.-=grantedBy": null }, { render: false });
+                break;
```
CarlosFdez, line 32: Should this have a default case? I'd assume detach is the default case.

PR #4245, opened 2022-09-29
Set correct ability when class dcs are picked

CarlosFdez:
This hides the ability selection as well when the proficiency is set to anything other than class tradition. Base tradition statistics do not have set abilities, and thus its safe to prioritize the base statistic's ability over the entry's.

### DIFF
--- a/src/module/actor/sheet/spellcasting-dialog.ts
+++ b/src/module/actor/sheet/spellcasting-dialog.ts
@@ -47,7 +47,7 @@ class SpellcastingCreateAndEditDialog extends FormApplication<Embedded<Spellcast
         return {
             ...(await super.getData()),
             actor: this.actor,
-            data: this.object.toObject(true).system,
+            data: this.object.toObject().system,
             magicTraditions: CONFIG.PF2E.magicTraditions,
             spellcastingTypes: CONFIG.PF2E.preparationType,
             abilities: CONFIG.PF2E.abilities,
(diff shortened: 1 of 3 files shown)


src/module/actor/sheet/spellcasting-dialog.ts
```diff
@@ -47,36 +47,51 @@ class SpellcastingCreateAndEditDialog extends FormApplication<Embedded<Spellcast
         return {
             ...(await super.getData()),
             actor: this.actor,
-            data: this.object.toObject(false).system,
+            data: this.object.toObject(true).system,
```
stwlam, line 50:
```suggestion
            data: this.object.toObject().system,
```

src/module/actor/sheet/spellcasting-dialog.ts
```diff
@@ -47,36 +47,51 @@ class SpellcastingCreateAndEditDialog extends FormApplication<Embedded<Spellcast
...
 
         // We may disable certain form data, so reinject it
-        inputData.system = mergeObject(
+        const system = mergeObject(
             inputData.system ?? {},
             {
                 prepared: {
                     value: this.object.system.prepared.value,
                 },
+                ability: { value: "cha" },
```
stwlam, line 78: Ability is cha?

CarlosFdez: Charisma is the default ability as defined by template.json. This is needed since ability can disappear.

It'd be better to make a best guess at what ability to use, but that's a separate feature and a separate PR.

PR #4261, opened 2022-09-30
Always show invest item, and equip the item to the correct slot when investing

CarlosFdez:


src/module/actor/character/index.ts
```diff
@@ -1992,7 +1992,14 @@ class CharacterPF2e extends CreaturePF2e {
             throw ErrorPF2e("Unexpected error toggling item investment");
         }
 
-        return !!(await item.update({ "system.equipped.invested": !item.isInvested }));
+        const invested = item.isInvested;
+
+        // If investing and unequipped, equip first
+        if (!invested && !item.isEquipped) {
+            await this.adjustCarryType(item, item.system.usage.type, item.system.usage.hands, true);
```
stwlam, line 1999: Can this not render?

CarlosFdez: Unfortunately it does render. We could possibly add a parameter to stop it from doing so, but it'll have to cascade. adjustCarryType is a bit messy in its implementation, but I don't have any inspirations on how to start cleaning that function up.

CarlosFdez: I could either add an options to all the functions to not render, or somehow try to return the updates to perform. I think the former might be enough and is significantly either, but wouldn't be atomic. But that might be good enough in this case?

PR #4262, opened 2022-09-30
Remove most of condition manager class

stwlam:


src/module/actor/base.ts
```diff
@@ -1112,17 +1112,21 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
                     ? Math.min(existing.value + 1, max)
                     : existing.value + 1;
             })();
-            if (conditionValue === null || conditionValue > (max ?? 0)) return;
+            if (conditionValue === null || conditionValue > (max ?? 0)) return null;
             await game.pf2e.ConditionManager.updateConditionValue(existing.id, this, conditionValue);
+            return existing;
```
CarlosFdez, line 1117: Does this existing include the updated condition value?

stwlam: It's the same item, so I hope so

PR #4264, opened 2022-09-30
Convert Earn Income to skill statistic and fix check dialog title

CarlosFdez:


CarlosFdez:
I think this is ok to leave till after the next release (maybe during the hotfix stages). Its just more preliminary work for converting skill/perception/initiative fully to statistic.

PR #4269, opened 2022-09-30
Allow Choice Set homebrew drop when there are no valid default choices

LebombJames:
Currently, when a ChoiceSet has no valid choices, it says there are no suitable items, regardless of if there is an allowedDrops on the ChoiceSet. This change causes a homebrew drop box to appear in such a situation, allowing for allowedDrops to be used.

You can test this behaviour on Implement Adept. The choices are hardcoded to the existing implements, meaning if someone is using a homebrew implement, they have easy no way to set that implement as their adept, making automation predicated off that essentially impossible. With this change, the prompt window shows with a homebrew drop box.

### DIFF
--- a/src/module/rules/rule-element/choice-set/prompt.ts
+++ b/src/module/rules/rule-element/choice-set/prompt.ts
@@ -46,6 +46,7 @@ export class ChoiceSetPrompt extends PickAThingPrompt<string | number | object> 
             prompt: this.prompt,
             containsUUIDs: this.containsUUIDs,
             allowNoSelection: this.allowNoSelection,
+            allowedDrops: this.allowedDrops,
             selectMenu: this.choices.length > 9,
         };
     }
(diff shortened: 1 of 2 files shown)


stwlam:
This was already done

LebombJames:
I'm not exactly sure where or how, because adding Implement Adept with two homebrew implements spits out a big fat no valid options available.

stwlam:
https://github.com/foundryvtt/pf2e/pull/4200

LebombJames:
I think the issue is that change only fixes it for ChoiceSets which contain UUIDs. Implement Adept doesn't (it's values are just the names of the features), so accommodation for that.

This merge ensures the homebrew box is visible in such an occasion (the changes in the handlebars), and that the prompt actually appears in the first place if there's a valid allowedDrop

stwlam:
That's not an issue but a feature. There are no item drops when the choice set has no items.

LebombJames:
...Isn't that what the above PR fixed? This would simply being non-UUID choice sets in line with that. Even then, making the drop box appear if there is an allowedDrops is a useful feature 

With Implement Adept the choices are the existing implements, so if you don't have any (i.e homebrew) it'll close without you making a choice.

PR #4286, opened 2022-10-01
Reset actor and scene data when encounter turn changes

stwlam:


CarlosFdez:
private method not causing issues for newly created encounters?

stwlam:
it doesn't: the problem is caused by webpack and our constructor hack

CarlosFdez:
I'm impressed. I assumed something like _onCreate() would be vulnerable to it. Seems to be all the initialization stuff that breaks apart.

PR #4287, opened 2022-10-01
Fix Fodder Skeleton Speed RE

avagdumortesin:
Closes https://github.com/foundryvtt/pf2e/issues/4275

PR #4290, opened 2022-10-01
Add support for RE-based land speeds

### LINKED ISSUE #4275: Fodder Skeleton heritage does not alter base land speed correctly
### Description
The [fodder skeleton heritage](https://2e.aonprd.com/Heritages.aspx?ID=199) should set the base speed to 30ft, but it's not correctly altering it.

### Steps to Reproduce
1. Create a player character actor
2. Add skeleton ancestry
3. Add Fodder Skeleton heritage

### Screenshot
Skeleton Screenshot

stwlam:
Closes #4275, supersedes #4287

image

### DIFF
--- a/src/module/actor/creature/index.ts
+++ b/src/module/actor/creature/index.ts
@@ -709,24 +709,34 @@ export abstract class CreaturePF2e extends ActorPF2e {
         const rollOptions = this.getRollOptions(selectors);
 
         if (movementType === "land") {
-            const label = game.i18n.localize("PF2E.SpeedTypesLand");
-            const base = Number(systemData.attributes.speed.value ?? 0);
-            const statLabel = game.i18n.format("PF2E.SpeedLabel", { type: label });
+            const landSpeed = systemData.attributes.speed;
+            landSpeed.value = Number(landSpeed.value) || 0;
+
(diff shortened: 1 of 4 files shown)


src/module/actor/creature/index.ts
```diff
@@ -709,24 +709,34 @@ export abstract class CreaturePF2e extends ActorPF2e {
         const rollOptions = this.getRollOptions(selectors);
 
         if (movementType === "land") {
-            const label = game.i18n.localize("PF2E.SpeedTypesLand");
-            const base = Number(systemData.attributes.speed.value ?? 0);
-            const statLabel = game.i18n.format("PF2E.SpeedLabel", { type: label });
+            const landSpeed = systemData.attributes.speed;
+            landSpeed.value = Number(landSpeed.value) || 0;
+
+            const fromSynthetics = (this.synthetics.movementTypes[movementType] ?? []).map((d) => d() ?? []).flat();
```
CarlosFdez, line 715: Seems like this call is duplicated, you can probably lift it out of the if/else. And I guess the "get best logic" as well can probably be lifted out.

CarlosFdez: It can be handled in a separate refactor PR tbh. Having this fix in before a release is more important.

PR #4302, opened 2022-10-02
Add user setting to disable monochrome darkvision

stwlam:


CarlosFdez:
I suspect this would be better at the PC level somehow (in the token config). But I don't know the original request, maybe its for performance reasons.

stwlam:
I can't think of a reason why it should be at the PC level.

CarlosFdez:
Someone might wanna customize how it looks like or something, I dunno. Fetchling darkvision boon or something from zon kuthon. That's all that comes to mind.

PR #4329, opened 2022-10-03
Add Shoddy as an Other Tag for Armor, Equipment and Weapons.

### LINKED ISSUE #3793: There's no way to mark items as shoddy

xdy:


CarlosFdez:
Surprised other tags don't derive their type from the config.

CarlosFdez:
Would you be able to automate the effects of shoddy? If shoddy items can be suppressed, it might be worth doing it how crowbar does (where its a thing with a specific slug that can be suppressed by certain rule elements).

I think marking as shoddy alone isn't really useful, and I don't know if the effects of making shoddy work should be left to a module :^)

TikaelSol:
The system already handles improvised as well as the negation of the improvised penalty. So this is a very similar thing.

xdy:
My intention wasn't to leave automating it to a module, I'm just a believer in making incremental small changes. :)
I figured (based on Tikael's comment in the original issue) that this PR would be followed up (perhaps by me, perhaps by someone else) with a later PR that automated it.
Personally I think just the tag is useful enough for low-automation GMs.

CarlosFdez:
I should have responded. My bad.

I'm still not a fan of having some other tags automated and some not, especially when its easy to do and effectively the same thing. Leaving it like that is bound to create user reports eventually. I'll look into seeing if we can do this completely once I manage to find some time to do the whole thing in one fell swoop.

TikaelSol:
We also need an herbal tag for consumables, because Paizo likes their not-quite-traits. Maybe we rethink how we are constructing these tag lists entirely. As is this wouldn't be hard to build out, one slight issue with applying it similar to how improvised is applied now is that improvised is constructed in the roll itself which means that the improvised penalty doesn't reflect on the strike on the sheet, so if we are going to think about how to approach this we should also maybe find a way to do it better than that.

CarlosFdez:
I think its fine to add tags that can't be automated to just exist as is. Its just that in this case its pretty similar to another one that is already automated (thought perhaps not perfectly?)

stwlam:
Superseded by #4649

PR #3174, opened 2022-07-27
Add siege weapon actor type

surged20:
Adds a siege weapon actor and a sheet modeled off of the vehicle sheet style.
Icon is derived from https://game-icons.net/1x1/heavenly-dog/catapult.html

PR #4363, opened 2022-10-05
Add G&G siege weapon ammunition

surged20:
Note: all common and level 0 since not specified.

### DIFF
--- a/packs/data/equipment.db/alchemical-fire-barrel.json
+++ b/packs/data/equipment.db/alchemical-fire-barrel.json
@@ -1,0 +1,105 @@
+{
+    "_id": "l6fTNWsQc85AYpTm",
+    "img": "systems/pf2e/icons/equipment/consumables/other-consumables/black-powder-keg.webp",
+    "name": "Alchemical Fire Barrel",
+    "system": {
+        "activation": {
+            "condition": "",
+            "cost": 0,
+            "type": ""
+        },
+        "autoDestroy": {
+            "_deprecated": true,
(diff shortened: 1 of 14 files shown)


packs/data/equipment.db/alchemical-fire-barrel.json
```diff
@@ -0,0 +1,105 @@
...
+        "consumableType": {
+            "value": "ammo"
+        },
+        "consume": {
+            "_deprecated": true,
+            "value": ""
+        },
+        "containerId": null,
+        "description": {
+            "value": "<p>Ammunition for a firedrake.</p>"
```
stwlam, line 34: Are these descriptions in G&G?

surged20: No. The way they show ammunition is in the siege weapon statblock (presumably to conserve page count). e.g. **Ammunition** alchemical fire barrel (50 gp, 3 Bulk). I can remove it since it isn't RAW. Is that the best way to handle it?

stwlam: We don't usually improvise descriptions like that, but I'll leave it up to data entry to decide.

TikaelSol: Well, we sort of improvised most ammunition names themselves given the vague descriptions in the book. I assume these are intended to work as part of #3174 so I have not really dug into this yet, but I will this weekend.

PR #4369, opened 2022-10-05
Return false early when testing detection of GM-hidden placeable objects

### LINKED ISSUE #4359: Token movement stuttering and input drops
### Problem
Token movement stutters heavily on certain scenes. Inputs are often dropped during this stuttering. These performance issues are absent on simple scenes without any walls / lightning.

### How To Reproduce
1. Import Abomination Vaults premium module.
2. Load the `(A) - Gauntlight Ruins` scene, or any other complex scene in AV.
3. Add a player character actor with vision. Copy it a few times to make the issue more obvious.
4. Group select the tokens, and move them around using arrow keys.

### Video
https://user-images.githubusercontent.com/236326/193991251-86c15089-56c3-49ad-aaab-6813f9b7be0d.mp4

The video shows token movement on an Abomination Vaults scene which stutters, and then on an empty scene which is smooth.

### Profiler
[Profile Snapshot](https://profiler.firefox.com/public/d29d45f34gz7znacjj0zx7evkwtxq7xxdy7wvjg/calltree/?globalTrackOrder=0wn&hiddenGlobalTracks=1wl&hiddenLocalTracksByPid=23956-0w3~11992-0~21548-0~4716-0~21996-03&implementation=js&thread=s&v=7)
The large humps in the CPU graph correlate to token movement.

### Environment Info
Foundry Version: 10.286
PF2E Version: 4.2.2
Browsers Tested: Firefox and Chrome
Operating System: Windows
Modules: None active

These problems have happened to myself and all five of my players. We did not see these issues in Foundry v9 on the same world.

All players have desktop PCs with modern hardware/dedicated graphics cards.

Server issues can likely be ruled out, I saw this behavior originally on a dedicated server and was able to reproduce it on a locally hosted instance of Foundry.

stwlam:
Closes #4359.

This leaves such placeables still visible to GMs but cuts out the unnecessary performance cost of full detection testing against them.

### DIFF
--- a/src/module/canvas/perception/modes.ts
+++ b/src/module/canvas/perception/modes.ts
@@ -30,6 +30,18 @@ class VisionDetectionMode extends DetectionModeBasicSight {
         });
     }
 
+    /** Short-circuit test to false if token is GM-hidden */
+    override testVisibility(
+        visionSource: VisionSource<Token>,
+        mode: TokenDetectionMode,
+        config?: CanvasVisibilityTestConfig
+    ): boolean {
+        return (
+            !(config?.object instanceof PlaceableObject && config.object.document.hidden) &&
+            super.testVisibility(visionSource, mode, config)
+        );
+    }
+
     protected override _canDetect(visionSource: VisionSource<TokenPF2e>, target: PlaceableObject): boolean {
         if (!super._canDetect(visionSource, target)) return false;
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I am so surprised this isn't handled upstream.

PR #4374, opened 2022-10-06
Add critical field to FlatModifier form

CarlosFdez:
image


stwlam:
"Double on Critical Hit," "Critical Damage Only," "Don't Double on Crit?"

CarlosFdez:
People complained that without the damage label it was too confusing. Could change the last one to `Regular Damage on Crit` though.

stwlam:
The confusion may have been due to the naming of the other options. Poll with the above? :slightly_smiling_face:

PR #4375, opened 2022-10-06
Re-render actor directory when world actor's level is changed

stwlam:


CarlosFdez:
Flattening the context before supering it up seems odd but....whynot. If a module subclasses it it'll do it after the system does anyways.

stwlam:
It's because upstream doesn't do it since foundry had no reason to care about any properties that aren't top-level

PR #4379, opened 2022-10-07
Fix double-granting world items from ABC items

### LINKED ISSUE #4361: Ancestry features are added twice for homebrew ancestries
image

CarlosFdez:
Conversion to uuid has made the split between compendium/non-compendium no longer necessary. I guess we can see later if the extra migration becomes a problem for world items.

Fixes #4361

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -2132,7 +2132,8 @@
             "Rules": {
                 "Apply": "Apply",
                 "Close": "Close"
-            }
+            },
+            "WorldItemWarning": "This is a World Item and will not work if saved in a compendium"
         },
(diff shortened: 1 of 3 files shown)


static/lang/en.json
```diff
@@ -4911,6 +4911,7 @@
             },
             "Date": "{weekday}, {day} of {month}, {year} {era}"
         },
+        "WorldItem": "World Item",
```
stwlam, line 4914: 🥲

CarlosFdez: Other than WorldItemLabel I have no idea what to do with that.

PR #4398, opened 2022-10-08
Fix incorrect token names

TikaelSol:


packs/data/blood-lords-bestiary.db/nightmare-portal.json
```diff
@@ -46,7 +46,7 @@
     ],
     "name": "Nightmare Portal",
     "prototypeToken": {
-        "name": "RENAMEME"
+        "name": "Nightmare Portal"
```
stwlam, line 49: You can wipe the whole `prototypeToken` property in this case (which would happen on extract)

TikaelSol: Done

PR #4403, opened 2022-10-09
Add support for multiple IWR types per RE

stwlam:


CarlosFdez:
Good idea! I think having the sideeffect for this.type in getIWR() rather than somewhere else might be unsafe but I think it should be fine in the meantime

PR #4407, opened 2022-10-09
Fix flat modifier editor adding critical: false to newly added rules

CarlosFdez:


src/module/item/sheet/rule-elements/flat-modifier.ts
```diff
@@ -100,8 +100,8 @@ class FlatModifierForm extends RuleElementForm<FlatModifierSource> {
         // Flat Modifier types may have mutually exclusive properties
         delete formData[formData.type === "ability" ? "value" : "ability"];
 
-        // Critical is optional if empty string but empty isn't the same as false or true
-        if (formData.critical === "") {
+        // Critical is a tri-state of false, true, and undefined. It doesn't default to neither true or false
```
stwlam, line 103:
```suggestion
        // Critical is a tri-state of false, true, and null. It doesn't default to neither true or false
```

src/module/item/sheet/rule-elements/flat-modifier.ts
```diff
@@ -100,8 +100,8 @@ class FlatModifierForm extends RuleElementForm<FlatModifierSource> {
...
+        if (typeof formData.critical !== "boolean" && !formData.critical) {
```
stwlam, line 104: if it's not a boolean but also falsy?

CarlosFdez: In case foundry ever has "false" come in as the boolean false. Bit of paranoia and I'm not feeling lucky.

stwlam: This would yield the wrong result were the string "false" to come in

CarlosFdez: the string false is not a boolean and is not falsy, so its handled by the else branch.

CarlosFdez: That said, this will have the wrong result if `true` that is a boolean comes in. Now that I look at it again while I'm more away. I can do `!formData.critical && formData.critical !== false` if that's an easier to understand condition, and `formData.critical = formData.critical === "false" ? false : Boolean(formData.critical)` in the else.

CarlosFdez: I did the above, but then I did one more change the check for false even more explicit.

I've tested all of the ones I've put so far, though admittedly I tested the latest one the least. Seemed fine tho.

PR #4431, opened 2022-10-10
Create Statistic check/dc in constructor and full-load check/dc modifiers

### LINKED ISSUE #4414: AdjustModifier RE cannot always adjust modifiers on spell attack rolls and never on DCs
Easiest way to test it is to apply Frightened to a spellcaster, then try the following REs:

```json
{
    "key": "AdjustModifier",
    "suppress": true,
    "selector": "spell-dc",
    "slug": "frightened"
}
```

```json
{
    "key": "AdjustModifier",
    "suppress": true,
    "selector": "spell-attack",
    "slug": "frightened"
}
```

The penalties from frightened will not be suppressed.

If you change the selector to `"all"`, the spell DC will still have Frightened applied to it, but spell attack rolls will have them suppressed.

If you change the selector to `"spell-attack-dc"`, the spell attack rolls will have the Frightened penalties suppressed, but not if you change it to `"spell-attack-roll"` or `"attack-roll"`.

CarlosFdez:
Closes #4414

### DIFF
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -78,6 +78,9 @@ export class Statistic {
     /** If this is a skill, returns whether it is a lore skill or not */
     lore?: boolean;
 
+    check: StatisticCheck;
+    dc: StatisticDifficultyClass;
+
     constructor(public actor: ActorPF2e, data: StatisticData, public options?: RollOptionParameters) {
         this.#data = data;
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This can wait until after the next hotfix. The fix for the elite/weak issue is entirely separate from all of this. But it seems to work, so up to you if you wanna throw it in early.

PR #4433, opened 2022-10-10
Migrate item drops on actor sheet

stwlam:


CarlosFdez:
Migrations occuring in sheets is definitely a problem. We're going to need to revisit this at some point, hopefully if compendiums ever get an improvement that will let us better handle old data. Or via overrides in the actor class.

stwlam:
I don't particularly see a problem with the application ensuring it's passing off migrated item data, but it could be handled in `Item.createDocuments` instead.

CarlosFdez:
Its because of separation of concerns, and it creates a gotcha if anyone tries to create an actor sheet (unlikely obviously). This is a data level concern and should likely be handled by the document.

But for a hotfix, and especially because other migrations are sprinkled around, this is fine. Ideological structure can take a backseat. They can all be moved later, especially if we can ever solve the compendium migration problem once and for all (or if foundry does it).

CarlosFdez (dismissed):

Issue #4447, opened 2022-10-11
Hellknight Signifier Dedication bonus mod


Gucinnio:
The below bonus is only implemented for Deception not for Intimidation.
While wearing your signifer’s mask, you gain +1 circumstance bonus to Deception checks to Lie, Intimidation checks, and Deception DCs against Sense Motive.

PR #4456, opened 2022-10-12
Add intimidation bonus to hellknight signifier dedication.

TMunsie:
Addresses part of #4447. I'm not sure if we can do the rest, and not sure if we should add a toggle.

### DIFF
--- a/packs/data/feats.db/hellknight-signifer-dedication.json
+++ b/packs/data/feats.db/hellknight-signifer-dedication.json
@@ -57,6 +57,16 @@
                 "selector": "deception",
                 "type": "circumstance",
                 "value": 1
+            },
+            {
+                "key": "FlatModifier",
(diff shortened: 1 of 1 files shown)


packs/data/feats.db/hellknight-signifer-dedication.json
```diff
@@ -57,6 +57,16 @@
                 "selector": "deception",
                 "type": "circumstance",
                 "value": 1
+            },
+            {
+                "key": "FlatModifier",
+                "label": "wearing Signifier Mask",
+                "predicate": [
+                    "wearing-signifier-mask"
+                ],
```
TikaelSol, line 66: We should make an item for this actually.

TikaelSol:
Superseded by #4592, closing

Issue #4428, opened 2022-10-10
Links in Actor notes don't work in many scenarios


rlach:
Tested on: foundry v10.287
PF2e: 4.2.5
Modules: none

Public Notes section in NPC view can contain links to journals. When you give players permissions to view them with "Observer" level or higher those links work:

<img width="651" alt="Zrzut ekranu 2022-10-10 o 21 49 38" src="https://user-images.githubusercontent.com/5655000/194942665-b0a6e93f-f176-4f58-8e62-ae2d7691c0be.png">

But when the permissions level is "Limited" the links break, become non-clickable and all users can see is the the code behind the link:
<img width="651" alt="Zrzut ekranu 2022-10-10 o 21 49 25" src="https://user-images.githubusercontent.com/5655000/194942816-1b03aa4c-348b-4b41-abc2-4cf0e2bc5fec.png">

The same issue is happening with biography->Appearance section of PC sheets.
<img width="748" alt="Zrzut ekranu 2022-10-10 o 21 57 12" src="https://user-images.githubusercontent.com/5655000/194942975-16093bbd-a16b-46f7-8fca-e0658df7c176.png">

But unlike NPC sheets the links don't work regardless of permission level(here Observer):
<img width="748" alt="Zrzut ekranu 2022-10-10 o 21 59 56" src="https://user-images.githubusercontent.com/5655000/194943350-19258cfc-f41f-4db2-916b-1b87bc33ab28.png">

I tried to check if it is pf2e related or Foundy related, so I compared the behaviour to dnd5e system. While dnd5e has only single "Biography" section on PC and NPC sheets the links work regardless of permissions set for the character sheet, so I think this is pf2e issue.

All those links used to work in Foundry v9 and last pf2e version that supported v9.

Steps to reproduce(NPC):
1) As DM create a new NPC sheet
2) Create any journal entry
3) Open NPC sheet, click on "Prototype Token", check "Link Actor Data" and then click "Update Token"
4) Go to Notes, start editing, drag the journal entry on the edit field to create a link. Close the sheet to save the changes.
5) Right click on NPC in actors view, set user permissions for all players to "Limited"
6) Login as one of the players, open the NPC sheet.

The links are broken.

Steps to reproduce(PC)
1) As a DM create a new PC sheet
2) Create any journal entry
3) Go to biography, and start editing every sections there. In every section drag the journal entry on the edit field to make a link.
4) Close the sheet to save changes.
[...]

Issue #4465, opened 2022-10-13
HTML Enrichment for journal entries breaks in description of NPC with limited access


Rachell92:
I use the description of NPCs to add a journal entry for my players to keep notes in about them, previously giving my players Limited permissions over the NPCs was enough to get access to these journal entries, but the link appears to be broken now, just staying pure text.
I've tested the other permissions settings and it works fine both as observer and owner.
![image](https://user-images.githubusercontent.com/23698154/195535450-cb87d13c-ec03-4f30-9d47-416938ce387f.png)

PR #4467, opened 2022-10-13
Fix multiple instances of previously enriched content that was omitte

In3luki:
…d in the v10 conversion


In3luki:
I wasn't finished typing the description but the default action on enter press is submitting the PR. 🙃 
Closes #4428 and #4465

PR #4488, opened 2022-10-16
Disable tagify style sheets from modules

stwlam:


CarlosFdez:
awkward that we need to do this. I'd expect since our tagify elements are more specific, they'd override the module ones.

PR #4491, opened 2022-10-16
Remove double borders from frames and round corners

LebombJames:
Before: (Notice the black border touching the frame and the inner one containing the image)
imageimageimage

After:
imageimageimage

The rounding of the corners is just a minor aesthetic thing; I just thought it looked nicer but I'm not married to the idea

### DIFF
--- a/src/styles/actor/character/_crafting.scss
+++ b/src/styles/actor/character/_crafting.scss
@@ -187,13 +187,14 @@
                         height: 24px;
                         width: 24px;
                         margin: 2px 0;
-                        border-radius: 0;
+                        border-radius: 2px;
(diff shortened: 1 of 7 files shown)


src/styles/actor/character/_crafting.scss
```diff
@@ -187,13 +187,14 @@
...
+                        border-radius: 2px;
                         margin-left: 4px;
                         cursor: pointer;
                         .item-icon {
-                            max-width: 23px;
+                            max-width: 24px;
+                            border: none;
                         }
-
+
```
stwlam, line 197: Use editorconfig integration to prevent this

Issue #4502, opened 2022-10-18
Feature request: Add text search feature to inspect roll window


fraserstanton:
scrolling through the whole list can be a pain, some kind of text search would be nice?

CarlosFdez:
I'm not certain about a text sort, but I was consider sorting everything without : at the top.

stwlam:
Isn't it already sorted?

CarlosFdez:
Sorry. I'm not certain about a text search. I want to change the sorting to make certain types of roll options easier to find.

PR #4508, opened 2022-10-18
Put top level roll options at the top in the inspector

CarlosFdez:
image


CarlosFdez:
Bottom is fine too, if you want to disincentivize non-scoped roll options.

stwlam:
Is this to address (the motivation behind) #4502?

CarlosFdez:
It is, and its also something I've been meaning to do for a while because it has bit me before. That issue reminded me that I had to do it.

That said, its probably not what they want.

PR #4515, opened 2022-10-19
Introduce damage roll class, delegate roll drilldown rendering

CarlosFdez:
While this is better than currently (I hope), I'd like to at least make it possible to evaluate the damage roll class in some capacity before merging to see if this path can even be followed. Its up for now in case anyone has any input.


src/module/system/damage/roll-dialog.ts
```diff
@@ -159,44 +158,29 @@ export class DamageRollModifiersDialog extends Application {
...
-                            return { faces: d.faces ?? 0, result: r.result };
-                        })
-                    ),
-                    formula: partial,
-                    label: damageCategories[damageCategory as DamageCategory] ?? damageCategory,
-                    total: roll.total,
-                };
+                rollData.diceResults[damageType][damageCategory] = roll.dice.flatMap((d) =>
+                    d.results.map((r) => ({ faces: d.faces ?? 0, result: r.result }))
+                );
```
In3luki, line 175: This is currently a `number` array of roll results that is stored in the message flag. I wasn't sure if that is expected to be there by some modules so I left it in when I changed the render method a few weeks ago.
I guess changing it is the fastest way to find out. 😁

types/foundry/client/roll.d.ts
```diff
@@ -27,13 +27,13 @@ declare global {
      * console.log(r.total);    // 22
      */
     class Roll<TData extends RollData = RollData> {
```
stwlam, line 29: This should probably be `TOptions` now, right?

src/module/system/damage/roll-dialog.ts
CarlosFdez: It is currently that yeah. Unfortunately the render data and the result data are currently created simultaneously, and if we want a better separation of concerns, we need to be able to create the render data from the result data and with the current format that's not very possible. I'm just hoping it doesn't break.....too much....

CarlosFdez:
So two gotchas right now:
- DamageRoll has a "" formula and receives a damage object. Seems odd, but not sure of what to do here.
- I'm considering pre-rendering the roll instead of doing it the usual way (where it renders as part of getHTML()), since the format is liable to change. If we pre-render, format changes won't break compatibility.

stwlam:
Pre-rendering the roll is a circumvention of the core workflow; there'd need to be a very good justification for it

CarlosFdez:
I think the current justifications is that it pre-renders already on master, and it'll prevent old rolls breaking due to data model changes. We can leave a comment to swap to roll.toMessage() when we are further in the refactor.

Or we can say screw it, allow old rolls to break rendering and console vomit errors when we refactor later.

PR #4519, opened 2022-10-20
Fix close button for rules editing

KSops:
The close button is disabled for the rules editing element when looking at something from compendium, creating a scenario in which the user cannot navigate backwards.
This is done on foundry's side, so this PR attempts to remedy that by explicitly removing the disabled attribute from the close button.


kyles-ep:
Amended code to adhere to linter config

KSops:
Amended code to adhere to linter config

CarlosFdez:
Not the way I was imagining, but not a wrong way either. Sure! Thanks for the PR.

PR #4522, opened 2022-10-20
Add Book of the Dead mummy archetype

MSAbaddon:


packs/data/archetypes.db/mummy.json
```diff
@@ -0,0 +1,31 @@
+{
+    "_id": "1vxlKEYwMtPtl3hq",
+    "name": "Mummy",
```
TikaelSol, line 3: This journal isn't needed, we just have not removed this compendium yet. Got distracted by other things.

MSAbaddon: Removed

PR #4527, opened 2022-10-22
Add the option to open the effect's sheet directly from the Effects Panel

reonZ:
A new clicking option is added to the `Effects Panel` tooltips allowing to open the effect sheet using `Shift + Left Click`


CarlosFdez:
There's a part of me that's thinking that perhaps shift+click to open and then a [Send To Chat] on the title bar on the opened item sheet might be better. But I'd wait until other people agree with me before committing to that.

stwlam:
This is quite idiosyncratic and doesn't map to any other keyboard shortcut in the system. Maybe a separate button could be there, but the effects panel was never intended to be single buttons with four different operations depending on which modifier keys are held.

CarlosFdez:
Separate button could work. Some way of opening up the effect to make any tweaks might be useful, but I don't know about a unique modifier key (though I'm not opposed to it either). However if we have that, we could use a send to chat for all items anyways, so that would save us having to add a button/hotkey for it.

CarlosFdez:
I'm guessing most of us are in agreement for a "send to chat" button at the titlebar of the item sheet (I was unsure at first but I've been convinced). I think between that and extempore effects you should be able to have the featureset you want. Go ahead and do that if you're down for it.

reonZ:
This is originally a module i made for my players because we are very new to the game and always forget what conditions do, so the send to chat was helpful for us but i do agree that it may be out of place here.

I really like the idea of being able to open an effect/condition and think that it was a missing feature, having to open an actor sheet and go into the effects tab to then edit it was quite annoying and not being able to interact with the effect button for that felt counter-intuitive.

CarlosFdez:
Late response, but I think we're both ok with open item sheet, just disagree on the how or why.

I think shift + click to open sheet (and only that) would be fine, as it is the reverse of check dialog rolls (or exactly the same as check dialog rolls if the default for show dialog is flipped). But that's just me.

reonZ:
Alright, i made the changes:

- I removed the `Send to Chat` option.
- I changed the bind for the `Open Sheet` option to be `Shift + Click`

stwlam:
Superseded by #7037

PR #4530, opened 2022-10-22
Add player options from Crown of the Kobold King.

VestOfHolding:


VestOfHolding:
Hmm, I'll have to figure out why so many other label changes came into this.

PR #4535, opened 2022-10-23
Include domains in character/npc strike roll inspector

CarlosFdez:
image


stwlam:
Sort? :)

CarlosFdez:
I'd consider that a separate feature but sure I'll slip it in.

PR #4546, opened 2022-10-24
Fix 8th level permanent items rollable table

MSAbaddon:


packs/data/rollable-tables.db/8th-level-permanent-items.json
```diff
@@ -45,12 +45,12 @@
             "drawn": false,
             "img": "systems/pf2e/icons/equipment/held-items/rod-of-wonder.webp",
             "range": [
-                12,
+                13,
                 13
```
stwlam, line 49: This is going to break the table

PR #4551, opened 2022-10-25
Allow duration coercion to be optional

KSops:
### Problem
Right now auras force the applied effect duration to be unlimited and have no expiry style set. This means that even if you explicitly link to an effect that has a duration of N rounds, it will not be respected.
I suppose that the change was made because it was assumed auras should strictly be for aura-like effects. However, this strictness denies the possibility of using the aura system to apply effects to other tokens in a radius.
### Changes
Add a `noDurationCoercion` boolean field to the aura effect data type such that if it primitively evaluates to true, the coercion code is disabled and the original duration data from the effect (specified by uuid) is left alone.
e.g.
```
{
    "key": "Aura",
    "radius": 30,
    "effects": [{
        "affects": "allies",
        "events": ["enter"],
        "uuid": "Item.9Q68yHv607Y1EmSy",
        "removeOnExit": false,
        "noDurationCoercion": true
    }]
}
```
### Benefits
- Existing rules are not changed. Default when unspecified is still the original behaviour before this change
- Allows a Bard to cast `Inspire Courage` at different round counters easily by flickering the aura on and off (in combination with `"removeOnExit": false`)

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -313,8 +313,10 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
 
                 const source = mergeObject(effect.toObject(), { flags });
                 source.system.level.value = data.level ?? source.system.level.value;
-                source.system.duration.unit = "unlimited";
-                source.system.duration.expiry = null;
+                if (!data.noDurationCoercion) {
+                    source.system.duration.unit = "unlimited";
+                    source.system.duration.expiry = null;
+                }
                 toCreate.push(source);
             }
         }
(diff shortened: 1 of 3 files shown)


stwlam:
This was done as a temporary measure because expiring effects would degenerate into cycles of repeated application. Simply removing the forced indefinite duration with no other changes would simply reintroduce the issue.

KSops:
Sorry. For some reason I didn't pick up that a response/proposal was expected.

Instead of what's proposed in this PR, what if there was an "applyOnce" param on the aura (default false) where, if it is set to true, the effect is only granted once. This should alleviate the concern about re-application of effects. And when this flag is enabled, the duration is not tampered with.

PR #4553, opened 2022-10-25
Set the ID of a PC's worn armor as a roll option

stwlam:
I need it for a thing :)


apoapostolov:
Wouldn't the same be useful for the equipped shield?

stwlam:
I can't think of any uses off the top of my head, but maybe.

PR #4561, opened 2022-10-26
Ensure grant flags are set when duplicate grants are disallowed

stwlam:


src/module/rules/rule-element/grant-item/base.ts
```diff
@@ -223,6 +209,37 @@ class GrantItemRuleElement extends RuleElementPF2e {
...
+            onDelete:
+                this.onDeleteActions?.granter ?? setHasElement(PHYSICAL_ITEM_TYPES, grantee.type)
+                    ? "detach"
+                    : "cascade",
+        };
+
+        if (grantee instanceof ItemPF2e) {
+            // This is a previously granted item: update its grantedBy flag
+            // Don't await since it will trigger a data reset, possibly wiping temporary roll options
+            grantee.update({ "flags.pf2e.grantedBy": grantedBy }, { render: false });
```
CarlosFdez, line 237: I forgot, does foundry filter prevent updates if the diff is empty?

stwlam: It doesn't make it to the server, but this is a grant new item with a randomly-generated ID.

CarlosFdez: Then I guess even if somehow the logic ever tweaks so that this is used to "ensure" flags, it won't break from that.

PR #4578, opened 2022-10-26
Retrieve strike weapon item from damage as well

CarlosFdez:
image

image

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -1,7 +1,7 @@
 import { ActorPF2e } from "@actor";
 import { CriticalHitAndFumbleCards } from "./crit-fumble-cards";
 import { ItemPF2e } from "@item";
-import { ChatMessageDataPF2e, ChatMessageFlagsPF2e, ChatMessageSourcePF2e } from "./data";
+import { ChatMessageDataPF2e, ChatMessageFlagsPF2e, ChatMessageSourcePF2e, StrikeLookupData } from "./data";
 import { TokenDocumentPF2e } from "@scene";
 import { traditionSkills, TrickMagicItemEntry } from "@item/spellcasting-entry/trick";
 import { UserPF2e } from "@module/user";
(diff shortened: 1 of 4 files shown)


src/module/chat-message/document.ts
```diff
@@ -135,18 +134,18 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
         if (!actor?.isOfType("character", "npc")) return null;
 
         // Get the strike index from either the flags or the DOM. In the case of roll macros, it's in the DOM
-        const roll = this.rolls.at(0);
-        const strikeIndex = (() => {
-            if (roll instanceof StrikeAttackRoll) return roll.options.strike?.index;
+        const { index, altUsage } = (() => {
```
stwlam, line 137: Can this have a return type?

src/module/chat-message/document.ts
```diff
@@ -135,18 +134,18 @@ class ChatMessagePF2e extends ChatMessage<ActorPF2e> {
...
+            if (this.flags.pf2e.strike) return this.flags.pf2e.strike;
             const messageHTML = htmlQuery(ui.chat.element[0], `li[data-message-id="${this.id}"]`);
             const chatCard = htmlQuery(messageHTML, ".chat-card");
-            return chatCard?.dataset.strikeIndex === undefined ? undefined : Number(chatCard?.dataset.strikeIndex);
+            return {
+                index: chatCard?.dataset.strikeIndex === undefined ? undefined : Number(chatCard?.dataset.strikeIndex),
```
stwlam, line 142:
```suggestion
                index: chatCard?.dataset.strikeIndex === undefined ? null : Number(chatCard?.dataset.strikeIndex),
```

src/module/chat-message/document.ts
CarlosFdez: I can, but it does mean I'm returning -1 instead of undefined/null. Should be fine though.

PR #4579, opened 2022-10-26
Prepare character skill data sooner

CarlosFdez:
Prepare skills after perception but before AC, allowing the use of skills for AC computations (which some feats might need to use).


CarlosFdez:
Was on draft for a bit since I needed to fix something and I wanted to verify armor check penalties still worked. They do!

PR #4583, opened 2022-10-27
Increase reliability of rule source lookup in grant item

CarlosFdez:


src/module/rules/index.ts
```diff
@@ -105,7 +105,7 @@ class RuleElements {
             if (REConstructor) {
                 const rule = ((): RuleElementPF2e | null => {
                     try {
-                        return new REConstructor(data, item, options);
```
stwlam, line 108: Can this be part of options?

CarlosFdez: Options is passed to the item as a whole, and the source index needs to differ per rule in that item.

PR #4588, opened 2022-10-27
Allow GrantItem REs to assign a badge value to granted items

Cerapter:
https://user-images.githubusercontent.com/43446478/198371192-b2a6fc47-c853-4220-9acb-64334920bda2.mp4

This PR changes the behaviour of GrantItem REs so that they can now set the granted item's badge value if applicable (right now, only Effects and Conditions).

If "Reevaluate on Update" is ticked on, the granted item's badge value will update if something changes on the actor / item.

### DIFF
--- a/src/module/rules/rule-element/grant-item/base.ts
+++ b/src/module/rules/rule-element/grant-item/base.ts
@@ -4,7 +4,7 @@ import { ItemSourcePF2e } from "@item/data";
 import { ItemGrantDeleteAction } from "@item/data/base";
 import { MigrationList, MigrationRunner } from "@module/migration";
 import { isObject, pick, setHasElement, sluggify, tupleHasValue } from "@util";
-import { RuleElementPF2e, RuleElementSource } from "..";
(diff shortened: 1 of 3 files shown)


stwlam:
This is planned, but I'd like to approach it in a broader way to anticipate other alterations to an item.

CarlosFdez:
If I recall, there's a standing uncertainty over whether item alterations should be part of the GrantItem itself or defined in subclasses. Which at this point honestly, I'm fine with either if defining the item type in GrantItem is untenable or unwieldy. This is actually one of those "hard problems" you've stumbled into, one that is actually the reason we've been building out this functionality.

Cerapter:
So should this PR be just closed then?

CarlosFdez:
There was a bit of a discussion. Shark, what's your opinion on an "overrides" object with a "type" field that restricts the type of item being overriden, that explicitly handles specific overrides for the item?

stwlam:
Taking a less stapled-on approach to this

PR #4590, opened 2022-10-27
Adds pf2e flags with sourceId, uuid and type to used item chat message.

### LINKED ISSUE #4185: Add item data to "Use Item" chat messages
When using consumable items, the only indication of what item has been used is in the flavor text.

image
image

The message's `item` data returns `null`. 
Would be great if the item data was passed on to the message, much like when you post the item itself to chat below.

image

xdy:
Fixes #4185

### DIFF
--- a/src/module/item/consumable/document.ts
+++ b/src/module/item/consumable/document.ts
@@ -1,6 +1,6 @@
 import { TrickMagicItemPopup } from "@actor/sheet/trick-magic-item-popup";
 import { ItemPF2e, PhysicalItemPF2e, SpellcastingEntryPF2e, SpellPF2e, WeaponPF2e } from "@item";
-import { ItemSummaryData } from "@item/data";
+import { ItemSummaryData, ItemType } from "@item/data";
 import { TrickMagicItemEntry } from "@item/spellcasting-entry/trick";
 import { ValueAndMax } from "@module/data";
 import { LocalizePF2e } from "@module/system/localize";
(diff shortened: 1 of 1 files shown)


src/module/item/consumable/document.ts
```diff
@@ -119,16 +119,29 @@ class ConsumablePF2e extends PhysicalItemPF2e {
 
             // If using this consumable creates a roll, we need to show it
             const formula = this.system.consume.value;
+            const flags = {
+                pf2e: {
+                    origin: this
+                        ? {
```
stwlam, line 125: do you expect `this` to be undefined?

src/module/item/consumable/document.ts
```diff
@@ -119,16 +119,29 @@ class ConsumablePF2e extends PhysicalItemPF2e {
...
+                              type: this.type as ItemType,
+                          }
+                        : null,
+                },
+            };
             if (formula) {
                 new Roll(formula).toMessage({
                     speaker: ChatMessage.getSpeaker({ actor: this.actor }),
                     flavor: content,
+                    flags: flags,
```
stwlam, line 137:
```suggestion
                    flags,
```

src/module/item/consumable/document.ts
```diff
@@ -119,16 +119,29 @@ class ConsumablePF2e extends PhysicalItemPF2e {
...
                 });
             } else if (this.consumableType !== "ammo") {
                 ChatMessage.create({
                     user: game.user.id,
                     speaker: ChatMessage.getSpeaker({ actor: this.actor }),
                     content,
+                    flags: flags,
```
stwlam, line 144:
```suggestion
                    flags,
```

src/module/item/consumable/document.ts
xdy: Eh. Not really. Will fix.

PR #4608, opened 2022-10-28
Allow effects to be hidden from players

Cerapter:
https://user-images.githubusercontent.com/43446478/198745341-3c8c56de-92c4-4171-bce3-ff8ef09f15a7.mp4

This PR adds a feature that allows GM to apply effects to player characters, NPCs, and familiars, without revealing them to them.

- This is only for effects, but a way was paved for conditions if hiding conditions is ever desired.
- Effects are hidden from players on their characters' effects tabs, the effects panel on the top right, and the tokens on the scene.
- For the GM, effects that are hidden from players show up
  - with a dotted outline and `--visibility-gm-bg` colour in the PC effects tab,
  - with a `--visibility-gm-bg` drop shadow and a "SECRET" secondary tag on the effects panel,
  - plainly on tokens on the scene.
- Effects can quickly be toggled on and off on the effects tabs of characters.
- This does not hide any, well, effects of effects -- for example, a FlatModifier applied by a secret effect will still show up in modifier calculations, without being obfuscated in any way.

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -477,22 +477,15 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
         });
 
         // Change whether an effect is secret to players or not
-        for (const effects_sections of htmlQueryAll(html, ".effects-list") ?? []) {
-            for (const element of htmlQueryAll(effects_sections, "[data-action=effect-toggle-secret]") ?? []) {
-                element.addEventListener("click", async (event) => {
-                    const effectId = htmlClosest(event.currentTarget, "[data-item-id]")?.dataset.itemId;
-                    const effect = this.actor.items.get(effectId, { strict: true });
-                    /*
-            const target = $(event.currentTarget);
-            const parent = target.parents(".item");
-            const effect = this.actor?.items.get(parent.attr("data-item-id") ?? "");
-            */
-                    if (effect instanceof AbstractEffectPF2e) {
-                        const isSecret = effect.secret;
-                        await effect.update({ "system.secret": !isSecret });
-                    }
-                });
-            }
+        for (const element of htmlQueryAll(html, ".effects-list [data-action=effect-toggle-secret]") ?? []) {
+            element.addEventListener("click", async (event) => {
+                const effectId = htmlClosest(event.currentTarget, "[data-item-id]")?.dataset.itemId;
+                const effect = this.actor.items.get(effectId, { strict: true });
+                if (effect instanceof AbstractEffectPF2e) {
+                    const isSecret = effect.secret;
+                    await effect.update({ "system.secret": !isSecret });
+                }
+            });
         }
     }
 
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -477,15 +477,23 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
         });
 
         // Change whether an effect is secret to players or not
-        $html.find("a[data-action=effect-toggle-secret]").on("click", async (event) => {
+        for (const effects_sections of htmlQueryAll(html, ".effects-list") ?? []) {
+            for (const element of htmlQueryAll(effects_sections, "[data-action=effect-toggle-secret]") ?? []) {
+                element.addEventListener("click", async (event) => {
+                    const effectId = htmlClosest(event.currentTarget, "[data-item-id]")?.dataset.itemId;
+                    const effect = this.actor.items.get(effectId, { strict: true });
+                    /*
             const target = $(event.currentTarget);
             const parent = target.parents(".item");
             const effect = this.actor?.items.get(parent.attr("data-item-id") ?? "");
-            if (effect instanceof AbstractEffectPF2e) {
-                const isSecret = effect.secret;
-                await effect.update({ "system.secret": !isSecret });
+            */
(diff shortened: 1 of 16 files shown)


src/module/actor/creature/sheet.ts
```diff
@@ -475,6 +475,17 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
                 await effect.increase();
             }
         });
+
+        // Change whether an effect is secret to players or not
+        $html.find("a[data-action=effect-toggle-secret]").on("click", async (event) => {
```
CarlosFdez, line 480: We're kinda moving off of JQuery for these. Can you use vanilla JS and htmlClosest() for this? There should be examples if you follow the references?

src/module/apps/effects-panel.ts
```diff
@@ -62,6 +62,7 @@ export class EffectsPanel extends Application {
             actor,
             effects,
             conditions,
+            isGM: game.user.isGM,
```
CarlosFdez, line 65: I notice that actor sheets have user: { isGM } instead. (from using the debugger real quick). It might be best to imitate however the actor sheets inject that data in so that the templates more or less match.

src/module/item/abstract-effect/document.ts
```diff
@@ -12,6 +12,9 @@ export abstract class AbstractEffectPF2e extends ItemPF2e {
     abstract increase(): Promise<void>;
     abstract decrease(): Promise<void>;
 
+    /** If true, the AbstractEffect should be hidden from the user unless they are a GM */
+    abstract get secret(): boolean;
```
CarlosFdez, line 16: Instead of making this abstract and overriding to be return false in the condition, maybe make it return false here, and then leave it to subclasses to decide if they wanna allow secret versions or not.

static/templates/actors/familiar-sheet.html
```diff
@@ -213,7 +213,8 @@ <h4>{{localize "PF2E.EffectsLabel"}}</h4>
                     <ol class="effects-list item-list">
                         {{#each items as |item idx|}}
                         {{#if (eq item.type "effect")}}
-                            <li class="item effects" data-item-id="{{item._id}}">
+                        {{#if (or (not item.system.secret) @root.user.isGM)}}
+                            <li class="item effects {{#if item.system.secret}}secret{{/if}}" data-item-id="{{item._id}}">
```
CarlosFdez, line 217: Does {{#if item.secret}} work? I believe it should. If not that's definitely going to be annoying.

CarlosFdez (changes requested):

src/module/actor/creature/sheet.ts
Cerapter: Replaced with a combination of `htmlQueryAll` and `htmlClosest`.

src/module/apps/effects-panel.ts
Cerapter: Changed to that.

static/templates/actors/familiar-sheet.html
Cerapter: For whatever reason, it does not. The getter doesn't work, and if I do `{{log item}}` it clearly shows that it does not exist.

Familiar sheets, am I right.

src/module/item/abstract-effect/document.ts
Cerapter: Aye, changed to that.

src/module/actor/creature/sheet.ts
```diff
@@ -475,6 +475,25 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
                 await effect.increase();
             }
         });
+
+        // Change whether an effect is secret to players or not
+        for (const effects_sections of htmlQueryAll(html, ".effects-list") ?? []) {
+            for (const element of htmlQueryAll(effects_sections, "[data-action=effect-toggle-secret]") ?? []) {
```
CarlosFdez, line 481: Replace with a single loop for `htmlQueryAll(html, ".effects-list [data-action=effect-toggle-secret]")`. You should be able to do more or less the same queries you did with Jquery.

static/templates/actors/familiar-sheet.html
CarlosFdez: If its only like this in familiar sheets then feel free to ignore it. I'll need to go over this anyways when I work on familiars.

src/module/actor/creature/sheet.ts
```diff
@@ -475,6 +475,25 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
...
+                element.addEventListener("click", async (event) => {
+                    const effectId = htmlClosest(event.currentTarget, "[data-item-id]")?.dataset.itemId;
+                    const effect = this.actor.items.get(effectId, { strict: true });
+                    /*
+            const target = $(event.currentTarget);
+            const parent = target.parents(".item");
+            const effect = this.actor?.items.get(parent.attr("data-item-id") ?? "");
+            */
```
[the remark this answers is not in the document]
Cerapter: Oops.

src/module/actor/creature/sheet.ts
Cerapter: I appreciate the trust you put in my knowledge of JQuery. :sweat_smile:

CarlosFdez (dismissed):

CarlosFdez (approved):

PR #4613, opened 2022-10-29
Update herbal-item

### LINKED ISSUE #4412: Herbal Items unaccounted for for Herbalist Archetype
Travel Guide introduced a new trait-that-isn't-a-trait called "herbal items". These interact in a new way with Herbalist Archetype:

> Each of the plant items listed below has specific benefits, some of which apply when you use a specific action or activity.  Alchemical plant items are “herbal items,” meaning characters with the herbalist archetype (Advanced Player’s Guide) can use their archetype abilities with them.

These are Artevil Suspension, Curled Cure Gel, Drowsy Sun Eye Drops, Lady's Blessing Oil, Toothworth Extract and Yarrow-Root Bandage.

Currently, a Herbalist trying to put these in their crafting entry will get an alert that it does not match the `["alchemical", "healing"]` crafting item requirements.

KSops:
Fixes #4412

### DIFF
--- a/src/module/item/consumable/document.ts
+++ b/src/module/item/consumable/document.ts
@@ -6,8 +6,19 @@ import { ValueAndMax } from "@module/data";
 import { LocalizePF2e } from "@module/system/localize";
 import { ErrorPF2e } from "@util";
 import { ConsumableData, ConsumableType } from "./data";
+import { OtherConsumableTag } from "./types";
 
 class ConsumablePF2e extends PhysicalItemPF2e {
+    get otherTags(): Set<OtherConsumableTag> {
+        return new Set(this.system.traits.otherTags);
+    }
+
+    override prepareBaseData(): void {
+        super.prepareBaseData();
+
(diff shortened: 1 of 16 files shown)


stwlam (changes requested):
Dispense with the superfluous "-item" bit, and give it a full follow-through. The tag needs to be settable from consumable sheets.

stwlam (approved):

src/module/item/consumable/document.ts
```diff
@@ -6,8 +6,19 @@ import { ValueAndMax } from "@module/data";
...
+    get otherTags(): Set<OtherConsumableTag> {
+        return new Set(this.system.traits.otherTags);
+    }
+
+    override prepareBaseData(): void {
+        super.prepareBaseData();
+
+        this.system.traits.otherTags ??= [];
+    }
+
```
stwlam, line 21:
```suggestion
```
`otherTags` is in the model template, so it needn't be set during data preparation.

PR #4622, opened 2022-10-29
Draft: Build embedded items in packs data from a reference pack item where possible

In3luki:
This currently works based on the `core.sourceId` flag of embedded items so older actors would have to be touched up to support it.
Supports `Action`s and `Spell`s at the moment.

In extract:
- Check for the presence of a `sourceId` flag
- Load the corresponding pack item from disk
- Compare the source item with the pack item after stripping embedded properties based on item type
- If both are the same record the uuid and all embedded properties for storage

In build:
- Check if the `itemSource` is deflated
- Inflate the source based on the recorded uuid and overwrite/add the stored embedded properties

Problems:
- I'm using `JSON.stringify` to compare the embedded item with the pack item. A method to extract the difference between the two, like `diffObject`, would probably be better but would also be quite a bit more complex.
- Migrations. I've added a check in the migration runner to skip deflated item data assuming that the pack item would be properly migrated instead. Migrations of properties in deflated item data would need special casing.

I've attached an example `Actor` to show what the result could look like.

 I'm mainly looking for feedback if this even is the right approach.


stwlam:
We need to store spells and actions as references permanently for NPCs.

In3luki:
So  rather than rebuilding the item in the pack builder there should be in-game support for item references in `prepareEmbeddedItems`?

stwlam:
Maybe not right then, but there should be some asynchronous way to do it, to be called as needed.

In3luki:
That shouldn't be too difficult I think. The only problem I'm seeing is the performance for rendering an NPC sheet with many referenced items for the first time. Maybe the base compendium index could be utilized for that.

In3luki:
Closing this for now.
Using item references for NPCs is definitely possible but has two problems that need to be solved.
1. The references can't be part of `NPCSource#items` without overriding the `DataModel` validation for embedded items.
2. Rule Elements from item references need to be present in the NPC's data preparation. They also might reference values on the parent item so all item data needs to be available there.

The first can be solved by simply storing the reference data in `system.itemReferences` and merging that with the normal embedded items in data preparation.
The second would require asynchonously fetching all item references from the server when the NPC is prepared for the first time which could result in major performance problems.

PR #4629, opened 2022-10-30
Extract feats after opening/closing every sheet

stwlam:
Mostly expected results here, though some are worth double-checking


packs/data/feats.db/surprise-snare.json
```diff
@@ -36,8 +36,7 @@
             "custom": "",
             "rarity": "common",
             "value": [
-                "archetype",
-                "manipulation"
+                "archetype"
```
stwlam, line 39:
```suggestion
                "archetype",
                "manipulate"
```

PR #4631, opened 2022-10-30
Prevent `FormDataExtended` from generating arrays from disabled form elements

stwlam:


CarlosFdez:
This is a very....odd name for this PR.

stwlam:
I invite you to pick the name you fancy most.

CarlosFdez:
Depends. What's going on? The PR name looks like you're patching a foundry bug....but you're making a change to an element?

stwlam:
V10 will look for multiple elements with the same name and pull all the values into an array. But it grabs disabled form elements, too.

CarlosFdez:
And this is the only scenario where that happens?

CarlosFdez:
Prevent V10's FormDataExtended disabled form processing from clobbering feat data?

CarlosFdez:
Its only for feats in this case but sure fine

PR #4639, opened 2022-10-30
Show incapacitation note for all checks with the trait

fraserstanton:
Not just spells (as the old code did)


TikaelSol:
Renamed PR to match project guidelines.

PR #4655, opened 2022-10-31
Add the option to set effects as unidentified

reonZ:
A new option to set an effect `unidentified` is available, this will result in the hiding of the effect from the actor sheet, effects panel, token, combat tracker, etc for the players.

If the effect generates flat modifiers, those will be displayed as "Unidentified" in the tooltips, check modifiers dialog window and in chat card tags for the players.

When an effect is `unidentified`, no floaty text will be displayed over the token at creation or deletion of the effect for the players.

A checkbox input is available in the effect sheet sidebar for the GM and if the GM holds the `ctrl` key when an effect is 'drag & drop' onto a token or an actor sheet, the effect will automatically receive the `unidentified` state. 

Notes: 
- for a reason i don't understand, the `ctrl` key must be held before starting to drag the effect.
- there is more than likely a few scenarios i have not thought about and therefore not have taken care of yet.

### DIFF
--- a/.betterer.results
+++ b/.betterer.results
@@ -13,12 +13,12 @@ exports[`too-much-lint`] = {
     "types/foundry/client/collections/compendium-collection.d.ts:3957627828": [
       [210, 23, 3, "Unexpected any. Specify a different type.", "193409811"]
     ],
-    "types/foundry/client/documents/mixins/canvas-document-mixin.d.ts:4239377699": [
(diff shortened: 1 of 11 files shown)


CarlosFdez:
Effects being secret have already been merged into master. We can either close it, or you can rebase the modifier stuff onto the current implementation. Let me know what you wanna go with.

reonZ:
I will look into it then, we can close this one and i will open another one whenever if needed.

PR #4657, opened 2022-10-31
Expose self and target's ID as rollOptions

Cerapter:
Assuming a single target for a roll, this PR will make it so that `self:id:XYZ` and `target:id:ABC` are now available rollOptions to use.

`self:id` is available for targetless rolls as well.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -503,6 +503,9 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
             },
         });
 
+        // Always add the actor's ID to the rolloptions.
+        this.rollOptions.all[`self:id:${this.id}`] = true;
+
         this.setEncounterRollOptions();
     }
 
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -504,7 +504,9 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
         });
 
         // Always add the actor's ID to the rolloptions.
-        this.rollOptions.all[`self:id:${this.id}`] = true;
+        if (this.id) {
+            this.rollOptions.all[`self:id:${this.id}`] = true;
+        }
 
         this.setEncounterRollOptions();
     }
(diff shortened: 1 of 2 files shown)


stwlam:
could have sworn we already had this ...

src/module/actor/creature/index.ts
```diff
@@ -303,6 +303,9 @@ export abstract class CreaturePF2e extends ActorPF2e {
             };
         }
 
+        // Always add the actor's ID to the rolloptions.
+        this.rollOptions.all[`self:id:${this.id}`] = true;
```
Cerapter, line 307: This _could_ be moved to `ActorPF2e`, as it doesn't cause any problems, but that doesn't bring any new features with it.
Hazards and loot boxes won't have their ID grabbed for `target:id`, and the hazards' rolls are uninspectable due to an error anyways. Loot boxes don't usually roll.

stwlam (changes requested):
Set it in `ActorPF2e#prepareBaseData` right after the flag is initially added.

Cerapter:
> could have sworn we already had this ...

If this is not sarcasm, then not gonna lie, same. I thought it was already a feature. But the closest I can see is weapon IDs.

CarlosFdez:
Yeah checking the roll inspector, its not set. There's no self:id:blah field

src/module/actor/base.ts
```diff
@@ -503,6 +503,9 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
             },
         });
 
+        // Always add the actor's ID to the rolloptions.
+        this.rollOptions.all[`self:id:${this.id}`] = true;
+
```
stwlam, line 508: Check that they have an ID (temporary actors won't necessarily).

stwlam (approved):

stwlam:
Hm, only misgiving here is that synthetic actors will share IDs, which will lead to unwanted results if not anticipated.

Cerapter:
Can anything be done about that?

stwlam:
Not until V11 when synthetic actors (hopefully) go away for good.

stwlam:
I think we're going to have to wait until V11 for this sort of thing to be usable.

Cerapter:
No worries. Your PR #4659 already seemingly does what I wanted to do with this one, so I don't need this too bad. It might have some other uses, but marking is basically what I wanted to get outta it.

PR #4659, opened 2022-10-31
Add MarkTarget RE for tracking and referencing other tokens

stwlam:


src/module/rules/rule-element/mark-target/prompt.ts
```diff
@@ -0,0 +1,73 @@
...
+    prompt: string;
+
+    requirements: TargetRequirements | null;
+
+    #target?: TokenDocumentPF2e | null;
+
+    #resolve?: (value: TokenDocumentPF2e | null) => void;
+
+    constructor(params: PromptParameters) {
+        this.prompt = params.prompt ?? "Target a token by mouse-hovering one and pressing T. Press Escape to cancel.";
```
CarlosFdez, line 16: I think you forgot to localize this before making the PR?

stwlam: oh yeah, I had it sitting in a stash for a month and forgot where I left it 😅

PR #4669, opened 2022-11-02
Rename "rage" or "panache" predicate statements to "self:effect:*"

stwlam:


packs/data/extinction-curse-bestiary.db/kharostan.json
```diff
@@ -1534,25 +1534,19 @@
                     "value": ""
                 },
                 "rules": [
-                    {
-                        "domain": "all",
-                        "key": "RollOption",
-                        "option": "rage",
-                        "toggleable": true
-                    },
```
CarlosFdez, line 1542: Kharostan doesn't have a built in rage effect, looking at the actor. I think you might still need this.

TikaelSol: It just needs a link to the effect. Since it has a duration and they gain temp HP it shouldn't be a toggle.

Extinction Curse data is in awful shape but I'm a player in that AP and no one else has jumped at fixing it up.

SpartanCPA: I audited the first 2 books, but it's a pain to get you or Tim to review them while avoiding spoilers. I was hoping someone else could pick it up so I could be reviewer, but I'll get back to them post tokens if no one has by then.

PR #4674, opened 2022-11-03
Add android infiltrator automation

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/android-infiltrator.json
```diff
@@ -671,7 +671,42 @@
...
+                                "intensity": "4",
+                                "speed": "1",
+                                "type": "torch"
+                            },
+                            "bright": "0",
+                            "color": "#9b7337",
+                            "dim": "10",
+                            "shadows": "0.2"
+                        }
+                    }
```
TikaelSol, line 708:
```suggestion
                    {
                        "key": "TokenLight",
                        "value": {
                            "bright": "0",
                            "color": "#3b7355",
                            "dim": "10",
                            "shadows": "0.1",
                            "luminosity": 0.4
                        }
                    }
```
The torch animation looks weird in this case and I'd associate a more muted blue with a dim glow from an android personally. Here's what it looks like in darkness and dim light.

image

image

packs/data/pathfinder-bestiary-3.db/android-infiltrator.json
```diff
@@ -671,7 +671,42 @@
                 "requirements": {
                     "value": ""
                 },
-                "rules": [],
+                "rules": [
+                    {
+                        "domain": "all",
+                        "key": "RollOption",
+                        "option": "nanite-surge",
```
TikaelSol, line 678: The skill bonus and the glow have a different duration, so to do it properly it would need to be an effect for the glow and the toggle for the bonus.

MSAbaddon: Changed

PR #4677, opened 2022-11-03
Review B3 - huldra, hyakume

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/huldra.json
```diff
@@ -284,7 +285,7 @@
...
+                    "value": "<p><strong>Frequency</strong> once per day</p>\n<hr />\n<p><strong>Effect</strong> The huldra touches another creature to manipulate the creature's luck. The creature must attempt a @Check[type:will|dc:21] save.</p>\n<p>On a failure, the huldra chooses good luck or bad luck. If the huldra chooses good luck, the affected creature can roll twice on one d20 roll within the next minute and use the higher result; this is a fortune effect. If the huldra chooses bad luck, the creature must roll twice and use the lower result on its next d20 roll; this is a misfortune effect.</p>\n<hr />\n<p>@UUID[Compendium.pf2e.bestiary-effects.Effect: Manipulate Luck (Good)]{Effect: Manipulate Luck (Good)}</p>\n<p>@UUID[Compendium.pf2e.bestiary-effects.Effect: Manipulate Luck (Bad)]{Effect: Manipulate Luck (Bad)}</p>"
```
rsimoes, line 288: It hasn't been the practice to insert an HR before a spell effect link

stwlam, line 288: It hasn't been the practice to insert an HR before a spell effect link.

MSAbaddon: For bestiary effects it currently is the standard. I'm also fine without the hr th have all effect links look the same, but in this case we should update the other 240 entries as well.
image

stwlam: @TikaelSol has indicated we shouldn't do it. There should be a corpus-wide standard if there isn't one already.

MSAbaddon: Ok, I removed the hr. I'll create another MR that's remove the hr from other NPCs as well.

PR #4684, opened 2022-11-04
Fix shield HP updates not being reflected in token resource bars

### LINKED ISSUE #4568: Updating shield HP is not immediately reflected in the token resource bar
Shield HP can be used as a token resource and displayed as a bar. However, for linked tokens, when the shield HP is changed (e.g., via the character sheet), the resource bar is not updated immediately, but only once the token is hovered, or another similar interaction (e.g., reloading) is performed.

The reason is, that redrawing the bars is only triggered in the `_onUpdate` method of the `Actor` class in foundry core. Changing the shield HP however just changes the corresponding shield item, and thus does not cause the bars to be redrawn.

One way to solve it would be by also causing the bars to be redrawn when an owned shield item is changed. I have implemented this in one of my modules via the `updateItem` hook (https://github.com/ghost-fvtt/pf2-shields-token-bar/blob/main/modules/pf2-shields-token-bar.mjs#L90), but I think it should probably be done in pf2e directly instead.

I don't know that much about pf2e, but potentially, this could also affect other properties that are derived from items, which are usable in resources. In that case, it would probably make sense to extend the functionality to those kinds of items, too.

In3luki:
Fixed based on @ghost91 's suggestion.

Closes #4568

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -429,8 +429,9 @@ class TokenDocumentPF2e<TActor extends ActorPF2e = ActorPF2e> extends TokenDocum
             }
 
             // Update combat tracker with changed effects
-            if (this.combatant?.parent.active) ui.combat.render;
+            if (this.combatant?.parent.active) ui.combat.render();
         });
+        this.object.drawBars();
     }
 }
 
--- a/types/foundry/client/pixi/placeable-object/token.d.ts
+++ b/types/foundry/client/pixi/placeable-object/token.d.ts
@@ -251,7 +251,7 @@ declare global {
          * Refresh the display of Token attribute bars, rendering latest resource data
          * If the bar attribute is valid (has a value and max), draw the bar. Otherwise hide it.
          */
-        protected drawBars(): void;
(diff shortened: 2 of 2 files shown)


stwlam:
This should actually be done by core, but it's possible the bar redraw is done too early. I'm hesitant to add another blind redraw into the mix without making sure it doesn't break simultaneous re-rendering of other parts of a token, given other updates.

In3luki:
Updated.
This was actually caused by redirecting the `Actor#_onEmbeddedDocumentChange` call to `TokenDocumentPF2e#onActorEmbeddedItemChange` which did not redraw the token bars like the super method would do.
Also `Token#drawBars` was typed as `protected` which is incorrect.

stwlam:
Test it with a token resize and image swap all on one item.

In3luki:
Like this?

[token.webm](https://user-images.githubusercontent.com/41452412/201207496-7653faed-0075-41b3-8858-3450745bf3af.webm)

stwlam:
Nothing in the console?

In3luki:
Nope, no errors.

PR #4686, opened 2022-11-04
Review B3 - Kangaroo, Kappa, Kirin, Kishi

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/kirin.json
```diff
@@ -1284,7 +1401,37 @@
...
+                        "key": "Note",
+                        "outcome": [
+                            "success",
+                            "criticalSuccess"
+                        ],
+                        "predicate": [
+                            "rearing-thrust"
+                        ],
+                        "selector": "horn-attack",
+                        "text": "@Localize[PF2E.AttackEffects.KnockProne]",
```
TikaelSol, line 1430:
```suggestion
                        "text": "PF2E.AttackEffects.KnockProne",
```
We don't need to call to localize these, that is done during the processing of the note anyways.

PR #4688, opened 2022-11-04
Review B3 - Kitsune Trickster, Kokogiak, Kongamato, Kovintus Geomancer

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/kitsune-trickster.json
```diff
@@ -1650,15 +1652,15 @@
...
                 "rules": [],
-                "slug": null,
+                "slug": "change-shape",
                 "source": {
-                    "value": ""
+                    "value": "Pathfinder Bestiary"
                 },
                 "traits": {
-                    "custom": "",
+                    "custom": "[magical tradition]",
```
TikaelSol, line 1660:
```suggestion
                    "custom": "",
```
hmm, I suspect this should be removed entirely since we don't have a way to do custom traits on items anymore with the new editor.

PR #4707, opened 2022-11-05
Fix actor sidebar showing the actor level to users that lack permission

### LINKED ISSUE #4482: Limted NPC's levels are visible to players on the sidebar
Hey!

The issue with PF2 on V10 is: if a player can see an NPC in the sidebar (meaning "limited" permission level or higher), they can also see the NPC's level, which is not always desirable. It would be great if a DM could change it somehow so that the players don't see it. Because normally those who have have "limited" permission level for that NPC's character sheet, do not have access to that info and it'd be great if DMs could keep it that way.

Thanks in advance!

In3luki:
I'm not really happy with creating a new handlebars helper that is only used for this but it I believe it is the best solution.
As the actor rendering is done inside `templates/sidebar/document-directory.html` with an array of full `Document` instances the alternative would have been to subclass `SidebarDirectory` and override `SidebarDirectory#getData` to generate new render data for each actor, including the user permission, and passing that to a modfied template.

Closes #4482

### DIFF
--- a/src/module/apps/ui/actor-directory.ts
+++ b/src/module/apps/ui/actor-directory.ts
@@ -1,4 +1,5 @@
 import { ActorPF2e } from "@actor/base";
+import { htmlQueryAll } from "@util";
 
 /** Extend ActorDirectory to show more information */
 export class ActorDirectoryPF2e<TDocument extends ActorPF2e> extends ActorDirectory<TDocument> {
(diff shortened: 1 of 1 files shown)


stwlam:
We already subclass `ActorDirectory`

In3luki:
Classic case of brain fart. Updated.

stwlam:
Do we really need to copy over the folder template? Those are a bit of a pain to maintain.

In3luki:
I probably could just overwrite the `documents` property in the tree but that might lead to other problems. I'll take another look.

stwlam:
This could just be done from `activateListeners`, couldn't it? Hide the levels the user shouldn't see.

In3luki:
Oh, true. That would be a lot cleaner and less breaking. I'll update it after work.

In3luki:
Has some time during work. Last update?

PR #4708, opened 2022-11-05
Hide floaty text for unidentified effects

reonZ:
This simply make it so effects that are not shown to players also don't trigger floaty text onto the canvas.


src/module/canvas/token/object.ts
```diff
@@ -218,6 +219,17 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
...
 
+        /**
+         * If the floaty text is generated by an effect being created/deleted
+         * We do not display it if the effect is unidentified
+         */
+        if (!game.user.isGM && typeof params !== "number") {
+            const [change, document] = Object.entries(params)[0];
+            if (document.type === "effect" && (change === "create" || change === "delete")) {
+                if ((document as EffectPF2e).system.unidentified) return;
+            }
```
stwlam, line 230: Get params object and check for `"create" in effectParams`. If `document` is a document, type it as that rather than asserting it.

stwlam:
Why show it if a badge value is being updated?

reonZ:
That is a good question, there is probably no reason to check out the `updated`, though i don't recall seeing any floaty text for an updated modifier.

PR #4709, opened 2022-11-05
Allow the use of 'alt' key to add an unidentified effect

reonZ:
If the GM holds the `alt` key while dragging an effect onto a token or an actor sheet, the effect will be forcibly set to `unidentified`.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -692,6 +692,12 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
         const { actor } = this;
         const itemSource = item.toObject();
 
+        // Set effect to unidentified if ctrl key is held
+        if (game.user.isGM && itemSource.type === "effect") {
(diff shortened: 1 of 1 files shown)


stwlam:
Currently it's holding down ALT that does this with physical items.

reonZ:
Yeah, in my mind i was like "we use ctrl for blind rolls so i will go with ctrl", i can't believe how i didn't realize that alt was more appropriate even though i placed my code just next to the mystification part.

It has now been changed and uses `alt`.

src/module/actor/sheet/base.ts
```diff
@@ -692,6 +692,12 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
         const { actor } = this;
         const itemSource = item.toObject();
 
+        // Set effect to unidentified if alt key is held
+        if (game.user.isGM && itemSource.type === "effect") {
+            const altHeld = event.altKey || game.keyboard.isModifierActive(KeyboardManager.MODIFIER_KEYS.ALT);
+            if (altHeld) itemSource.system.unidentified = true;
```
stwlam, line 698:
```suggestion
            itemSource.system.unidentified ||= altHeld;
```

PR #4710, opened 2022-11-05
Make unidentifiable the modifiers of an unidentified effect

reonZ:
This is an attempt of making the modifiers generated by an `unidentified` effect as `unidentified` themselves for the players while displaying the regular information for the GM.

I am not familiar enough with the PF2e system source code to be able to find all the places i should work on for this matter, but i started nonetheless, for now, here is what i changed and could test:
- Modifiers tooltipster tooltips
- Breakdown html titles
- CheckModifiersDialog window
- CheckPF2e chat message tags
- DamageRollModifiersDialog chat message tags

I am willing to bet there are many other places i have missed and would be willing to go look there.

### DIFF
--- a/src/module/actor/modifiers.ts
+++ b/src/module/actor/modifiers.ts
@@ -140,6 +140,9 @@ class ModifierPF2e implements RawModifier {
     notes: string;
     hideIfDisabled: boolean;
 
+    unidentified: boolean;
+    originalLabel: string;
+
     /**
      * Create a new modifier.
      * Legacy parameters:
@@ -194,6 +197,10 @@ class ModifierPF2e implements RawModifier {
         // Force splash damage into being critical-only or not doubling on critical hits
         this.critical = this.damageCategory === "splash" ? !!params.critical : params.critical ?? null;
 
+        this.originalLabel = this.label;
+        this.unidentified = params.unidentified ?? false;
+        if (this.unidentified && !game.user.isGM) this.label = "Unidentified";
+
         if (this.force && this.type === "untyped") {
             throw ErrorPF2e("A forced modifier must have a type");
         }
@@ -203,8 +210,8 @@ class ModifierPF2e implements RawModifier {
     clone(options: { test?: Set<string> | string[] } = {}): ModifierPF2e {
         const clone =
             this.modifier === this.#originalValue
-                ? new ModifierPF2e(this)
-                : new ModifierPF2e({ ...this, modifier: this.#originalValue });
+                ? new ModifierPF2e({ ...this, label: this.originalLabel })
+                : new ModifierPF2e({ ...this, modifier: this.#originalValue, label: this.originalLabel });
         if (options.test) clone.test(options.test);
 
         return clone;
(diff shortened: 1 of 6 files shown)


src/module/rules/rule-element/flat-modifier.ts
```diff
@@ -127,6 +127,7 @@ class FlatModifierRuleElement extends RuleElementPF2e {
                     critical: this.critical,
                     hideIfDisabled: this.hideIfDisabled,
                     source: this.item.uuid,
+                    unidentified: this.item.isOfType("effect") && this.item.system.unidentified,
```
CarlosFdez, line 130: Rather than doing this, check if the item is an instanceof AbstractEffect and check for this.item.unidentified.

src/module/actor/modifiers.ts
```diff
@@ -140,6 +140,13 @@ class ModifierPF2e implements RawModifier {
     notes: string;
     hideIfDisabled: boolean;
 
+    unidentified: boolean;
+    originalLabel: string;
+
+    static get UNIDENTIFIED() {
+        return game.i18n.localize("PF2E.identification.Unidentified");
+    }
```
CarlosFdez, line 148: Is there a reason for this to be a getter?

reonZ: If i remember correctly, my thought process was that at class definition, `game.i18n` may not exist, but i didn't do any test on that, the fact that it is a getter itself makes it easy to change into a property at any time.

src/module/rules/rule-element/flat-modifier.ts
reonZ: very well

src/module/actor/modifiers.ts
```diff
@@ -140,6 +140,13 @@ class ModifierPF2e implements RawModifier {
     notes: string;
     hideIfDisabled: boolean;
 
+    unidentified: boolean;
```
stwlam, line 143: Make this `identified`

src/module/actor/modifiers.ts
stwlam: You can just localize this as needed. No need for a micro-optimization here.

src/module/actor/modifiers.ts
reonZ: Why ? we have worked with `unidentified` in the effect.

src/module/actor/modifiers.ts
reonZ: But everywhere it is used it is in need of `localization` event in the class constructor itself.

src/module/actor/modifiers.ts
CarlosFdez: I agree with reonZ here. identified is the neutral state. Unidentified is the special state.

CarlosFdez: Granted we can do `identified ??= true` during data preparation and that can work, though we'd need to also flip the default in effect. Not like null is a valid value.

CarlosFdez:
I'm gonna close this since its old (sorry!), but I think once we finally manage to have afflictions exist I'd love it if you could look into this again. The identified/unidentified thing might be easier to switch to "identified" once we have data models in place (and create a migration).

PR #4712, opened 2022-11-05
Delete Bone Croupier from EC compendium

SpartanCPA:
Reprinted in Book of the Dead with changes that act as errata.


TikaelSol:
Should note this for PDF to Foundry if the creature shows up in the adventure.

PR #4717, opened 2022-11-05
Compendium replace items and spells in EC Book 3

SpartanCPA:


SpartanCPA:
This was Book 2 - for changelog

PR #4728, opened 2022-11-06
Convert metagame visibility settings to booleans, change metagame prefix

CarlosFdez:
Under the hood it still uses the old user visibility stuff, we can knock those out as they come up / cause problems.

image

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -2852,19 +2852,11 @@
                 },
                 "ShowDC": {
                     "Name": "Show DCs on attacks and saves",
-                    "Hint": "Decide who can see the DC of a check.",
-                    "None": "None",
-                    "Gm": "GM",
-                    "Owner": "Owner",
-                    "All": "All"
+                    "Hint": "If enabled, players will see DCs of checks made against NPCs and other non-player-owned sources."
                 },
                 "ShowResults": {
                     "Name": "Show results on attacks and saves",
-                    "Hint": "Decide who can see the result of a check rolled from a chat card.",
-                    "None": "None",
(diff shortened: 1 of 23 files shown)


CarlosFdez:
If the prefix stuff bothers you, we can migrate all the . separators to _ so that we don't need it, then attempt to convert to boolean afterwards.

static/lang/en.json
```diff
@@ -2852,19 +2852,11 @@
                 },
                 "ShowDC": {
                     "Name": "Show DCs on attacks and saves",
-                    "Hint": "Decide who can see the DC of a check.",
-                    "None": "None",
-                    "Gm": "GM",
-                    "Owner": "Owner",
-                    "All": "All"
+                    "Hint": "Decide if players can see the DC of non-player checks."
```
stwlam, line 2855:
```suggestion
                    "Hint": "If enabled, players will see DCs of checks made against NPCs and other non-player-owned sources."
```

static/lang/en.json
```diff
@@ -2852,19 +2852,11 @@
...
                 },
                 "ShowResults": {
                     "Name": "Show results on attacks and saves",
-                    "Hint": "Decide who can see the result of a check rolled from a chat card.",
-                    "None": "None",
-                    "Gm": "GM",
-                    "Owner": "Owner",
-                    "All": "All"
+                    "Hint": "Decide if players can see the degree of success of a check rolled from a chat message."
```
stwlam, line 2859:
```suggestion
                    "Hint": "If enabled, players will see the degree of success of checks made against NPCs and other non-player-owned sources."
```

PR #4731, opened 2022-11-06
Allow cantrip deck creation from spell

### LINKED ISSUE #2256: [Feature Request] Cantrip Deck to allow linking of spells
In GitLab by @Unikatze on Apr 29, 2022, 15:02

Wondering if it's possible for the Cantrip Deck (SoM) to work similar to wands and scrolls, where you can assign a spell to it when you drag it into your inventory from the compendium.

KSops:
Resolves #2256


KSops:
I'm sure if there's demand for "other items" we can address it then.
As for demand for the deck, there's at least the original ticket author as well as myself. Maybe I'm playing my class wrong :P but I keep a few decks on my rogue for utility.

CarlosFdez:
I know right now you can simulate them by creating scrolls and adding uses to them. I guess this is a pretty unique case where its "pick any spell" I suppose.

Sure then. But for now might be best to lock it to first level spells.

src/module/actor/sheet/base.ts
```diff
@@ -719,7 +726,7 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
                     actor,
                     {},
                     async (heightenedLevel, itemType, spell) => {
-                        if (!(itemType === "scroll" || itemType === "wand")) return;
+                        if (!isSpellConsumableItemType(itemType)) return;
```
CarlosFdez, line 729: You should probably remove this check and make the function's type into SpellConsumableItemType.

src/module/item/consumable/spell-consumables.ts
```diff
@@ -31,26 +36,50 @@ const wandCompendiumIds: Record<number, string | undefined> = {
...
+function getNameForSpellConsumable(type: SpellConsumableItemType, spellName: string, heightenedLevel: number): string {
+    let name;
+    switch (type) {
+        case "cantrip-deck-5":
+            name = game.i18n.format("PF2E.CantripDeckFromSpell", { name: spellName });
+            break;
+        case "scroll":
+            name = game.i18n.format("PF2E.ScrollFromSpell", { name: spellName, level: heightenedLevel });
+            break;
+        default:
```
CarlosFdez, line 63: Don't use default for this. Use "wand". That way if anything is ever added, typescript will notify you that a change is required.

CarlosFdez:
You'll need to find a way to hide level selection when it changes to cantrip. You could look at the spellcasting create and edit dialog for implementation advice.

Instead of a new isSpellConsumable function, use tupleHasValue() or setHasElement() and have the value list somewhere. MAGIC_TRADITIONS can be searched as an example.

src/module/item/consumable/spell-consumables.ts
CarlosFdez: Even better, have a mapping from item type to localization entry. Then pull from that mapping.

stwlam:
> I'm sure if there's demand for "other items" we can address it then. As for demand for the deck, there's at least the original ticket author as well as myself. Maybe I'm playing my class wrong :P but I keep a few decks on my rogue for utility.

Demand for this is slight compared to "other items," and I'm not convinced it's worth hard-coding.

CarlosFdez:
The only reason I'm kinda entertaining it is because its basically just a different more restricted version of a scroll. This shouldn't happen too often, but ideally if we're gonna hardcode it, it'd be nice if it was cleaned up first.

That said, perhaps a better way of doing this is to add dropzones to scroll and wand details that show the embedded spell. That way the cantrip deck can just be an empty "scroll".

KSops:
Emulating this with empty scrolls affects a character's wealth (sure, that's editable but it's work), carried bulk, and missing item description about how cantrip decks work.

But if the decision is that request itself is not worth it then the original ticket should be marked as `won't fix` and we should close it accordingly with whatever justifications.

CarlosFdez:
I think what I plan to do is the dropzone once I find a bit more time (perhaps after I add custom damage type support), and we'll see how well that handles your use case. If it doesn't handle it well, we can consider adding additional options to this menu (after its cleaned up of course). I don't think its problematic to add dialog options for certain common cases so long as there is a non-hardcoded way to do it in the system.

KSops:
I was going over the rules and noticed cantrips can't be loaded into scrolls or wands. I'm thinking of leaving the wand scroll popup as it originally was then and adding a basic confirm dialog for creation of a cantrip. tldr; if it's a cantrip being dragged and dropped, switch to a confirm dialog that creates a cantrip deck, else go to original behaviour. Does that sound sensible?

CarlosFdez:
Thinking more on it, after a bunch of weeks, I don't disagree with the concept. I think the idea is fine. While we need to implement dropzones anyways for these sorts of items, this dialog is more of a user friendliness thing than a new feature persay.

That said, creating a whole new dialog type for just cantrips concerns me a bit. I would suggest renaming ScrollWandPopup to CastingItemCreateDialog and do it there if possible. Try to reduce the code to a minimum, so its not a pain to add more if we ever need to.

For the localizations, rather than creating new top level ones, it might be best to bunch them up. My recommendation is `PF2E.Item.Physical.FromSpell.X` where X is Cantrip5, Scroll, and Wand. Shark might ask you to change this later, he tends to have different suggestions on where to put these things. Feel free to remove the old localization entries for this.

There might eventually be more but that's what comes to mind. This code is part of the not-cleaned-up part of the codebase, so the end result should likely be less code than what you started with (or about par). Adding more is a sign that something probably went wrong.

KSops:
- Changed dialogs back into a single dialog
- Renamed dialog as suggested
- Changed localization strings
    - Wands are apparently not consumables so some strings were changed to call them "spellcasting items"
    - Reordered some strings in `en.json` to be sorted alphanumerically
    - Grouped item creation names as suggested

src/module/item/consumable/spell-consumables.ts
```diff
@@ -31,26 +41,36 @@ const wandCompendiumIds: Record<number, string | undefined> = {
...
+        case "cantripDeck5":
+            id = cantripDeckId;
+            break;
+        case "scroll":
+            id = scrollCompendiumIds[heightenedLevel] ?? null;
+            break;
+        default:
+            id = wandCompendiumIds[heightenedLevel] ?? null;
+    }
+    return id;
```
CarlosFdez, line 56: return directly from inside the case statements rather than assigning to a let. You could also alternatively chain a ternary.

In general try to avoid lets whenever possible.

src/module/actor/sheet/popups/casting-item-create-dialog.ts
```diff
@@ -0,0 +1,107 @@
...
+        ["wand", "PF2E.CastingItemCreateDialog.wand"],
+        ["cantripDeck5", "PF2E.CastingItemCreateDialog.cantripDeck5"],
+    ])
+);
+
+export class CastingItemCreateDialog extends FormApplication<ActorPF2e> {
+    onSubmitCallback: CastingItemCreateCallback;
+    spell: SpellPF2e;
+    formDataCache: FormOutputData;
+    isCantrip: boolean;
```
CarlosFdez, line 31: replacing all internal isCantrip references with this.spell.isCantrip will save you a few lines. There's no real need to cache that.

KSops: Just thought I'd save lookups but I guess it comes from a set, so it'll be negligible

static/templates/popups/casting-item-create-dialog.hbs
```diff
@@ -0,0 +1,26 @@
+<form autocomplete="off" onsubmit="event.preventDefault();">
+    <label>{{ localize "PF2E.CastingItemCreateDialog.label" }}</label>
+    {{#if (ne itemType "cantripDeck5")}}
+    <div class="form-group">
+        <p><label>{{ localize "PF2E.CastingItemCreateDialog.spellLevelLabel" }}</label>
+        <select name="level" data-dtype="Number">
+        {{#select level}}
+            {{#each validLevels}}
+            <option value="{{this}}">{{this}}</option>
```
CarlosFdez, line 9: you should probably match what heightenLevels does in spell-overlay.hbs

src/module/actor/sheet/popups/casting-item-create-dialog.ts
```diff
@@ -0,0 +1,105 @@
...
+        options.title = game.i18n.localize("PF2E.CastingItemCreateDialog.title");
+        options.template = "systems/pf2e/templates/popups/casting-item-create-dialog.hbs";
+        options.width = "auto";
+        options.submitOnChange = true;
+        options.closeOnSubmit = false;
+
+        return options;
+    }
+
+    override async getData(): Promise<FormApplicationData<ActorPF2e>> {
```
CarlosFdez, line 61: You should make the FormInputData the return type, and have it extend from FormApplicationData<ActorPF2e>. Rather than setting properties on the super, you should return a new object with the old properties splatted at the start. Most sheet classes more or less follow that pattern now.

static/templates/popups/casting-item-create-dialog.hbs
KSops: Which part are you referring to? Using a named local instead of `this`? Or a localization string?

src/module/actor/sheet/popups/casting-item-create-dialog.ts
```diff
@@ -0,0 +1,105 @@
...
+        };
+    }
+
+    override async _updateObject(event: Event, formData: FormOutputData) {
+        Object.assign(this.formDataCache, formData);
+        if (event.type === "submit") {
+            await this.formSubmit(event, formData);
+        } else {
+            await this.formUpdate(event, formData);
+        }
```
CarlosFdez, line 84: I don't think this method and the ones called after are big enough to need to split it up like this. I think what you can do is that event.type !== "submit", you do an await this.render() and then return. And then you do what's currently in formSubmit() afterwards.

KSops: Sure I can do that. Just thought I'd make it readable and clear so that someone can change behaviour in the future for a specific course of action.

static/templates/popups/casting-item-create-dialog.hbs
CarlosFdez: I mean ordinal values like 3rd Level, 4th Level. But I'm looking at the current dialog again, and this isn't a regression, so its ok to leave as is.

CarlosFdez:
I'm ok with this as written now. Thanks for working on it.

CarlosFdez (approved):

PR #4741, opened 2022-11-08
Correct placement of journal link for Curse Maelstrom

dogstarrb:


TikaelSol:
This is going to cause a conflict with your other PR.

PR #4750, opened 2022-11-09
Add merchant context for item transfer UI

### LINKED ISSUE #4729: Rephrase "move" to "buy" in merchant sheets when transacting between players and merchant sheet
When on a merchant sheet, if a player wants to buy something from a vendor and slides the object from the shop's inventory to his, FVTT asks the player in a **move** prompt "How many items do you want to **move**", when it should be _buy_ . 
This has caused many confusions in my group

Suggestion: Disimbiguate the wording, eventually add a prompt giving the amount of gold about to be payed for the sum of objects bought 
Eg: Buy Prompt: "You are about to buy X Lesser potions of healing for Y gold from N merchant. Are you sure?"

KSops:
Resolves #4729

As per the ticket, this will make UI strings more context-specific by using terms like "buy" and "purchase" instead of "move" for item transfers. Also adds a confirmation dialog only for purchasing from merchant.

This is probably a low-priority fix and probably needs eyes to make sure I made good choices in refactoring.


src/module/actor/sheet/base.ts
```diff
@@ -825,19 +826,32 @@ export abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorShee
...
+                { maxQuantity: sourceItemQuantity, lockStack: !stackable, isPurchase: isPaidTransaction },
                 (quantity, newStack) => {
-                    sourceActor.transferItemToActor(targetActor, item, quantity, containerId, newStack);
+                    if (isPaidTransaction) {
+                        purchaseConfirmationDialog(item, quantity, () => {
+                            sourceActor.transferItemToActor(targetActor, item, quantity, containerId, newStack);
+                        });
+                    } else {
+                        sourceActor.transferItemToActor(targetActor, item, quantity, containerId, newStack);
+                    }
```
KSops, line 842: Doesnt feel DRY but any ideas I have for addressing it compromises performance.

stwlam:
I have to think this could be done without such a large revision.

stwlam:
Superseded by #4775

PR #4756, opened 2022-11-09
Fix Inner Radiance Torrent Heightening

### LINKED ISSUE #4723: 2 round Heightened Inner Radiance Torrent damage calculation incorrect
Example Heightened to L3 should have 8d4 for the 2-3 action variant and 16d4 for the 2 round variant. Currently 2 round is only showing 12d4.

Rule text: "Heightened (+1) The initial damage, as well as the additional damage for the 2-round casting time, _each_ increase by 4d4..."

KSops:
Fixes #4723

### DIFF
--- a/packs/data/spells.db/inner-radiance-torrent.json
+++ b/packs/data/spells.db/inner-radiance-torrent.json
@@ -102,6 +102,13 @@
                             }
                         }
                     },
+                    "heightening": {
+                        "damage": {
+                            "0": "8d4"
+                        },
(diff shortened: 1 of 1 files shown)


stwlam:
I believe data entry intentionally made it this way due to SoM's author saying [it was a misprint](https://paizo.com/threads/rzs43gl6?Does-Inner-Radiance-Torrent-heighten-better#6) and will be corrected.

PR #4769, opened 2022-11-10
Review Ledalusca, Vine Leshy, Cactus Leshy, Seaweed Leshy from Bestiary 3

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/ledalusca.json
```diff
@@ -237,7 +237,7 @@
...
+                    "value": "<p>The ledalusca freezes their surface into a translucent, faintly colored ice sculpture matching the mirror image of a Medium creature or object they have seen before. While in this form, instead of a wave, they have the following Strike:</p>\n<p><strong>Melee</strong> <span class=\"pf2-icon\">a</span> ice shell [[/r 1d20+10 #Ice Shell]]{+10}/[[/r 1d20+5 #Ice Shell]]{+5}/[[/r 1d20+0 #Ice Shell]]{+0}, <strong>Damage</strong> [[/r {1d6+4}+{1d4}[cold]]]{1d6+4 physical damage plus 1d4 cold damage}</p>\n<p>The physical damage is a type that matches the new shape of the ledalusca's frozen shell. The ledalusca can revert to their liquid form by taking this action again.</p>"
```
TikaelSol, line 240: We can now use strike rule elements to enable this as a real strike on NPCs, see the (Kingmaker Spoiler) Avatar of the <spoiler> for how this is done, though that one needs its attack modifiers fixed to match the book.

MSAbaddon: Are you sure it is a  good idea to use strike rule elements for this even if it does not fully support npc strikes? For your example npc, even if the damage roll does the correct math, the npc sheet shows the wrong damage mod.
For the ledalusca I'd need to add a -1 flat modifier to the damage roll and also add an additional damage dice rule element to get the cold damage in.

TikaelSol: Then maybe we hold off until we can set up those strikes with base modifiers. It'll be a big project when we finally do get there. Either way at least make a regular strike for it since we want people to have access to automation for those rolls.

MSAbaddon: That'd be three strikes to do it correctly.
image
You sure it should get added? We currently have a lot of creatures with the above formating for additional strikes, e.g. all those change shape creatures with specific attack forms like the imp.

TikaelSol: Ugh, I guess we can add it to the pile of stuff to do later then

PR #4781, opened 2022-11-11
Allow effect badge counters to be set on drag

TikaelSol:
This will allow us to condense some effects as we can just use the badge number in the label similar to how conditions are handled.


CarlosFdez:
While we need to eventually consolidate condition/effect badges to be more or less the same thing, this code will likely be flagged by typescript during that conversion anyways, so this should be fine.

PR #4795, opened 2022-11-12
Review Mortics from Bestiary 3

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/gurgist-mauler.json
```diff
@@ -827,7 +830,40 @@
                 "requirements": {
                     "value": ""
                 },
-                "rules": [],
+                "rules": [
+                    {
+                        "domain": "all",
+                        "key": "RollOption",
+                        "option": "calculated-blow",
+                        "priority": 50,
```
TikaelSol, line 838:
```suggestion
```

MSAbaddon: Hm, no Idea why this gets added. I don't add it by hand at least.

PR #4796, opened 2022-11-13
Alter icon for secret inline checks

LebombJames:
image
First without the secret trait, second with

Once again, sorry github user `@Check`

### DIFF
--- a/src/styles/_colors.scss
+++ b/src/styles/_colors.scss
@@ -80,6 +80,7 @@ $adjusted-lower: #cc3311 !default;
   --inline-repost-bg: #{lighten(#ddd, 5%)};
   --visibility-gm-bg: #e8e8ef;
   --visibility-owner-bg: #ddebe1;
+  --blind-roll: #f5eaf5;
 
   /* Lighter / Darker */
   --light: #{$light-color};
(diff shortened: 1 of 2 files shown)


stwlam:
Great, now people are going to think the system is responsible for the monk module bug.

CarlosFdez:
has the monk module bug been fixed or going to go away anytime soon?

CarlosFdez:
The issue with the icon change (and I agree after discussing it) is that it doesn't look clickable. I'm not sure how well this will pan out, but what if you tint the button in the same color as the blind roll chat message?

LebombJames:
image

It's a tad subtle for my liking, but that's an issue with Foundry's choice of colour too.

CarlosFdez:
Perhaps if you also tinted the border a tad in that direction? Like a very desaturated dark purple (barely perceptable as purple).

I think subtle is fine though. Its not too important a player knows a roll will be secret. Especially now that secret rolls when clicked show in the dialog preset the roll mode (assuming they even have the dialog show up).

LebombJames:
image
Well, it's certainly barely perceptible even when enlarged in that image. But this'll do; I'll commit the changes

CarlosFdez:
That looks pretty decent. That'll do!

PR #4821, opened 2022-11-14
Review Phantoms and Planar Scions from Bestiary 3

MSAbaddon:


packs/data/bestiary-effects.db/effect-crush-chitin.json
```diff
@@ -43,4 +43,4 @@
         "unidentified": false
     },
     "type": "effect"
-}
```
TikaelSol, line 46: ?

MSAbaddon: That automatically happens when I export the bestiary effects. Must be a line break at last line that slipped in while I redid the merge request for the opossums.

PR #4843, opened 2022-11-16
Transmit context data to effects dropped from inline content links

stwlam:


CarlosFdez:
I'm actually somewhat bothered with how we're handling degree of success. Whenever we refer to "outcome" by that name in the codebase, we usually mean the resolved strings. Even in the context flag, outcome is the string result. But uniquely in the check, degreeOfSuccess is the numerical outcome.

In fact, we have gems like this in check.ts
```ts
        if (degree) {
            context.outcome = DEGREE_OF_SUCCESS_STRINGS[degree.value];
            context.unadjustedOutcome = DEGREE_OF_SUCCESS_STRINGS[degree.unadjusted];
            roll.options.degreeOfSuccess = degree.value;
        }
```

roll.outcome should probably convert back to maintain the naming convention maybe?

stwlam:
I wouldn't mind using magic numbers everywhere except for rule elements (for the sake of human readability).

CarlosFdez:
That's another alternative. In my head, outcome means the string, and degree of success means the magic number. If we change to magic numbers everywhere we can use the much nicer to read outcome variable, but we have to be quick to convert strings to numbers to avoid confusion.

That said, "outcome": 2 is a bit awkward to figure out for the console diver :P

PR #4844, opened 2022-11-16
Ensure all actors go through data re-preparation on totm setting change

stwlam:


CarlosFdez:
Might be worth making this a util at some point. I can imagine "reprepare all actors everywhere" might be a useful thing to have in general. I could use this for the campaign feats for example if I ever expose the UI for it.

PR #4859, opened 2022-11-17
Review Heracosphinx, Elder Sphinx, Cunning Fox, Feathered Bear from Bestiary 3

MSAbaddon:


packs/data/bestiary-effects.db/effect-guardians-aegis.json
```diff
@@ -0,0 +1,104 @@
...
+                                        "origin:trait:chaotic",
+                                        "origin:trait:evil"
+                                    ]
+                                }
+                            ]
+                        }
+                    }
+                ],
+                "selector": "saving-throw",
+                "slug": "guardians-aegis-1",
```
TikaelSol, line 49: Condense this to one simpler RE then use AdjustModifier to modify the bonus. It will end up being smaller overall because the predicates will be much simpler.

MSAbaddon: Thanks for the tip, didn't know that was possible. Please have a close look at the rule elements, neber used AdjustModifier before.

packs/data/bestiary-effects.db/effect-guardians-aegis.json
```diff
@@ -0,0 +1,79 @@
...
+                            {
+                                "and": [
+                                    "origin:trait:fiend",
+                                    "origin:trait:chaotic",
+                                    "origin:trait:evil"
+                                ]
+                            }
+                        ]
+                    }
+                ],
```
TikaelSol, line 54:
```suggestion
                "predicate": [
                    "origin:trait:fiend",
                    "origin:trait:chaotic",
                    "origin:trait:evil"
                ],
```
Demons fit this definition anyways so we can cut it way down.

PR #4861, opened 2022-11-17
Fix cha modifier of tallow guardian

kageru:
`null` seemed wrong to me, and I checked [pf2e tools](https://pf2easy.com/index.php?id=22710) to confirm (this compendium isn’t on aonprd yet).


TikaelSol:
Confirmed in the PDF.

PR #4876, opened 2022-11-19
Change color darkvision to an actor flag

TikaelSol:
Changes the hardcoded Fetchling check to an actor flag, and migrates that flag onto existing Fetchlings and Resonant Reflection of Life.


src/module/scene/token-document/document.ts
```diff
@@ -248,7 +248,7 @@ class TokenDocumentPF2e<TActor extends ActorPF2e = ActorPF2e> extends TokenDocum
             this.sight.range = basicDetection.range = defaults.range ?? null;
 
             // Temporary hard-coded fetchling check for initial release
```
CarlosFdez, line 250:
```suggestion
```

CarlosFdez:
Should probably check that it doesn't already have the rule. If the migration is run twice it might keep adding the flag right?

TikaelSol:
So something like
`if (getsColorDarkvision && !source.system.rules.some((r) => r.path === "flags.pf2e.colorDarkvision"))`
My IDE complains about path not existing on rule element source. I could do
`if (getsColorDarkvision && !source.system.rules.some((r) => r.key === "ActiveEffectLike"))`
But if someone has homebrewed anything with an AE like on either item it won't run. Though I guess that's not the worst thing since if they are homebrewing with AE likes they can definitely just copy paste the AE like to fix it.

CarlosFdez:
You could widen the type on the rules to Record<string, unknown>[]. So something like
`const rules: Record<string, unknown>[] = source.system.rules;`

CarlosFdez (approved):

PR #4877, opened 2022-11-19
Replace Broken with Hidden in token HUD

TikaelSol:
Broken only applies to objects, so we can ditch it and add Hidden to keep the HUD the same size as it is now.


CarlosFdez:
Vehicles :(

TikaelSol:
Yeah, but Hidden is going to come up far more often than vehicles. We could add both but the list is getting pretty large as is.

Far2Casual:
Definitely agree that Hidden is far more useful than Broken in the HUD.

Janonas:
Just to add, there is space for one more condition in the HUD without it adding a new row, so both could be possible, unless there is another candidate to add.

Janonas:
One thing that could also be done is a setting for the condition HUD to list the most commonly used conditions or the full list, like the setting for the icons.

Janonas:
im literally shaking and crying rn, is there a way to add it at only my end instead of it being pushed for everyone?

stwlam:
went with #8588 as a compromise

PR #4896, opened 2022-11-21
Restore origin context to saving throws

stwlam:


CarlosFdez:
Should it not add both origin and target options if both are provided?

stwlam:
If you can name a situation in which it'll make sense, sure.

CarlosFdez:
By that I mean, why the else if instead of two different ifs?

PR #4899, opened 2022-11-21
Prevent ChoiceSet from double renaming items

### LINKED ISSUE #4794: Ring of Energy Resistance bug
1. Create PC and Loot actors
2. Create Ring of Energy resistance to PC
3. Choose Acid
4. Give ring from PC to Loot and Back
5. Each time you repeat step 4 " (Acid)" will be added to name of ring

TikaelSol:
Closes #4794


stwlam:
This would prevent renaming in case of pre-selection by grants.

TikaelSol:
Ah, right forgot we did those pre-selections. Nevermind then.

PR #4902, opened 2022-11-21
Review Strix Kinmate, Sumbreiva, Swordkeeper from Bestiary 3

MSAbaddon:


src/scripts/config/traits.ts
```diff
@@ -401,6 +401,7 @@ const weaponTraits = {
     "deadly-2d12": "PF2E.TraitDeadly2D12",
     "deadly-3d12": "PF2E.TraitDeadly3D12",
     "deadly-4d12": "PF2E.TraitDeadly4D12",
+    death: "PF2E.TraitDeath",
```
TikaelSol, line 404: This doesn't look like it's used anywhere in this PR.

MSAbaddon: The sumbrevia's strikes has this trait

MSAbaddon: It is already within the strikes, it was only the entry in the coding that was missing.

PR #4915, opened 2022-11-22
Fix automatically-added item traits getting stored

stwlam:
image

need to loop back around and do the same for NPCs; will move things into helpers then


CarlosFdez:
Much better way than what I was thinking of doing.

PR #4930, opened 2022-11-23
Remove rule element labels from Staff of Providence and Expanded Healer's Tools

MSAbaddon:


stwlam:
Don't remove labels that provide the circumstance in which a bonus applies.

TikaelSol:
We have not been adding those labels and I'm not too keen on keeping them.

stwlam:
Why remove them? It is a straight downgrade in usability.

TikaelSol:
The staff can be localized, the healer's tools shouldn't really need a label.

MSAbaddon:
@stwlam Is it possible to make something like {item|name} - PF2E.SomeLocalizationString, {item|name} - @Localize[PF2E.SomeLocalizationString], or {item|name} - {PF2E|SomeLocalizationString} possible for rule element labels? That would allow us to create descriptive labels without blowing up re-en.json.

stwlam:
You can do that with a label, yes.

MSAbaddon:
Yeah, currently that doesn't work though, that's what I meant. A PF2E.XXX string only gets localized it if is the only string inside the label, @Localize is not active within the labels and it isn't possible to access localization strings in any other way.

stwlam:
What if `FlatModifier` supported a "sublabel" field to provide the circumstance?

MSAbaddon:
Sublabel sounds great :-)
Display in the roll dialogue would be <label> - <sublabel>?

stwlam:
Something like that, yeah. Go ahead and add a "sublabel" field; it'll just be ignored until support is added.

MSAbaddon:
Hm, just gave it a second thought - having a sublabel won't do e.g. for the Staff of Providence, since it covers two actions:
image

Getting this done with a sublabel would either require two separate rule elements or an entry "SenseDirectionSubsist" in the localization files. Best option would be to only have "Subsist" and "SenseDirection" as basic actions in the localization file and compose the label from multiple strings. What do you think?

simonward:
> @stwlam Is it possible to make something like {item|name} - PF2E.SomeLocalizationString, {item|name} - @Localize[PF2E.SomeLocalizationString], or {item|name} - {PF2E|SomeLocalizationString} possible for rule element labels? That would allow us to create descriptive labels without blowing up re-en.json.

For some languages, it’s difficult or impossible to assemble a localized string from substrings.

stwlam:
> For some languages, it’s difficult or impossible to assemble a localized string from substrings.

It'd be simple enough to arrange them separately, such as with the sublabel under the label.

PR #4963, opened 2022-11-25
Review Nosferatu from Bestiary 3

MSAbaddon:


SpartanCPA (commented):
Any reason you removed Dominate from the Innate Spells? The feature says they cast Dominate as a Divine Innate Spell

MSAbaddon:
> Any reason you removed Dominate from the Innate Spells? The feature says they cast Dominate as a Divine Innate Spell

Thought it's somewhat redundant since the ability already lists the spell including the inline button for the save. If you want to keep both I can add the spell again.

TikaelSol:
Yeah let's add it back.

MSAbaddon:
> Yeah let's add it back.

Done

PR #4971, opened 2022-11-26
Clean up DoS adjustment code, include saves vs incapacitation effects

stwlam:


CarlosFdez:
Conflicts!

While you're at it though, if you're down, treat wounds might look neater if you make the adjustments into private variables. RiskySurgeryAdjustment and MortalHealingAdjustment respectively.

Issue #2602, opened 2022-06-20
There seems to be no good way to handle non-general flat-footed conditions


iBoMbY:
Lets say for example I'm a Mastermind racket Rouge, and successfully identify a creature, then the target would be flat-footed for me until the start of my next turn (or 1 minute if critical success). 

Currently for the Rouge it is possible to use the toggle "Enable abilities that require a flat-footed target", which effects things like Sneak Attack damage dice, but this doesn't work to enable the AC -2 penalty to the DC on the attack roll.

I'm not exactly sure what the best way would be to handle this, though. Maybe some character specific effect item, very similar to the flat-footed condition, but with a predicate to make it only valid for attacks from a single character ID. But I don't think it's currently possible to do this with predicates, and also I see no existing way to generate character-specific effects. Of course it would be possible to hard-code this and link it to the 'target:condition:flat-footed', but I guess that would not be preferred.

TMunsie:
When we have conditional visibility then this would be a logical extension

PR #4973, opened 2022-11-26
Use proper flat-footed condition for TotM option

iBoMbY:
My proposal for a simple solution to use the correct Flat-Footed attack calculation (AC -2) when the TotM option "Enable abilities that require a flat-footed target" is set.

For this I also changed the isFlatFooted method to checkFlatFooted and return the dueTo with text, which should make things simpler in the end, when the other reasons get added.

### DIFF
--- a/src/module/actor/creature/index.ts
+++ b/src/module/actor/creature/index.ts
@@ -385,7 +385,9 @@ export abstract class CreaturePF2e extends ActorPF2e {
         rollOptions.all[`self:size:${sizeSlug}`] = true;
 
         // make sure "origin:target:condition:flat-footed" works ...
-        rollOptions.all["self:target:condition:flat-footed"] ??= rollOptions.all["target:condition:flat-footed"];
+        if (rollOptions.all["target:condition:flat-footed"]) {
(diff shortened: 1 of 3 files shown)


stwlam:
This doesn't make any sense given that there is no target when using the totm option.

CarlosFdez:
Sometimes there is a target when using the totm option. Outside of "theater of the mind", this option is also used in the case of situational flat footed situations set up by the player, when the gm is slow or doesn't have a fast toggle for flat footed. While the concession was labeled as theater of the mind, theater of the mind wasn't the only reason for wanting to keep the toggle (just the easiest one to explain).

That said, players often forget to disable the checkbox, and it will cause confusion as to why flat footed is being imposed. People may blame the system automation for flat footed rather than the toggle. Bare minimum the toggle would have to unset each round, and even then that might be awkward. That's why the toggle has never imposed the flat footed condition.

iBoMbY:
Well, currently it doesn't do what it says, and there still is no way to properly handle any non-general flat-footed condition. There are a million reasons to use this, any ranged Rouge needs this constantly, for example.

Also see: https://github.com/foundryvtt/pf2e/issues/2602

PR #4978, opened 2022-11-26
Accommodate injected (stringy) numbers in binary predication operations

stwlam:


CarlosFdez:
Can you give an example of what the injected stringy number is? Seems like its extracting numbers from a larger non-numerical stringy number using regex, so I'm wondering what its for.

stwlam:
"9"

CarlosFdez:
I'm thinking that I might be thrown off by the original logic, which is pretty confusing in what is even going on, that I didn't notice the new bit is the Number.isInteger() bit. Still don't, even after subsequent read-throughs. I think the only way to figure out what's doing is to employ the debugger....

Why does right accept any number but left accepts only integers? Shouldn't both be the same?

stwlam:
If you'd like to change that aspect, feel free to make a PR. Do you have questions about anything in this one?

CarlosFdez:
Should it not be `typeof Number(x) === "number"` for both halves? If the inconsistency is by design then it might be good to point that out and state why.

This bit is the part that hurty, but its old code, so if you wanna file it as legacy go for it:
`domainArray.flatMap((d) => (d.startsWith(left) ? Number(/:(-?\d+)$/.exec(d)?.[1]) : []));`

stwlam:
That is to allow comparisons like `{ "gt": ["item:level", 5] }`. "item:level:5" and "item:level:6" would both match.

CarlosFdez:
Sorry, my typeof check is incorrect. Would need to check `Number.isNaN(Number("aaa"))` or something akin to that

Right now as written, ` { "gt": ["6", 5.5] }` would work but not ` { "gt": ["6.5", 5.5] }`.

stwlam:
Only integer comparisons were needed at the time of implementation, and that is still the case today as far as I am aware. Thus it is simpler and a little more constrained.

CarlosFdez:
Wanna try removing `typeof right === "number"` and see if we're feeling lucky? Or is this intended to go on a hotfix and thus you'd rather play it safe and not assume we're always doing integer comparisons?

stwlam:
I'd like to see actual evidence of need before changing it.

CarlosFdez:
I guess then leave a comment that you're awaiting for evidence of need. I find it odd that a binary operator supports floating point right side values but not floating point left side values, and I know when revisited, it'll look like a bug. Might be best to leave a here be dragons no touchy for x reason type message if you're gonna leave it like that.

PR #5007, opened 2022-11-29
Change Kitsune Change Shape's name for added clarity

AlphaStarguide:
This change has two benefits.

1. make feature's name more clarity
2. Babele use "name" for searching translation. So it could fix [this problem](https://discord.com/channels/880968862240239708/880969304365994034/1045579414156611625)


AlphaStarguide:
Ops Sorry for that. I just really dont familiar with git but I'll google it.

PR #5011, opened 2022-11-30
Re-render effects panel when world time changes

stwlam:


CarlosFdez:
Luckily effect panel rendering is debounced, so it can be spammed as much we want.

PR #5016, opened 2022-11-30
Fix innate spells of Bugul Noz

TiloBuechsenschuss:
as per https://2e.aonprd.com/Monsters.aspx?ID=516


TikaelSol:
It looks like this is adding 3 uses to Wanderer's Guide, not Blindness

TiloBuechsenschuss:
> It looks like this is adding 3 uses to Wanderer's Guide, not Blindness

You're absolutely right, I managed to add the uses to the wrong spell. Doh, sorry.

PR #5024, opened 2022-12-01
Add data structure for regen abilities

MSAbaddon:
Thought it might be good to have regenerative abilities as actual data in the system.
Not sure though if that's the correct approach to it so before wasting too much time this request only contains the basic structure.


stwlam:
Does this have a feature that with the types?

stwlam:
This would be better done as a feature request, describing how the feature should work.

PR #5025, opened 2022-12-01
Have reevaluateOnUpdate also conditionally delete grants

stwlam:


CarlosFdez:
Given the move from the item.name fallback from the constructor to when its first created, is this liable to break existing grants? Grant item is kinda rough to figure out what's going on just by glancing at it.

PR #5026, opened 2022-12-01
Improve Dread Marshal Aura Effect

MSAbaddon:


stwlam:
This same PR has been made before and rejected because it isn't an appreciable improvement (and arguably it's even more tedious to use).

MSAbaddon:
Ok, thanks for the info - didn't know this came up before (although personally I'd prefer doing an additional click when adding the effect and get the damage bonus added correctly).

PR #5033, opened 2022-12-02
Remove ABC Manager

CarlosFdez:
I was requested not to include kits into this and "pretend they don't exist", but kits could reasonably be recoded in this way and it would simplify the code, though we do have to filter out the kit itself.

Names can swap. I couldn't think of what to call this other than a grant, and theoretically grant item could be refactored to use this mechanism if we really wanted to. But this should be a simple enough variation for use on specific familiars (theoretically staves too, if we want to allow spell editing).

That said, it doesn't fully work for afflictions, it kinda needs "re-evaluate" on update support for it to work properly...I think unless grant item expands to non-res, we gain the ability for items to inject REs, or we go a lot deeper with affliction support, it won't really be possible to do cleanly. Should work for specific familiars though.

The core is that one method drives "what do I add" and another method drives "what do I delete". What that means is up to the item itself.


src/module/item/base.ts
```diff
@@ -368,73 +368,106 @@ class ItemPF2e extends Item<ActorPF2e> {
...
-                // Pre-load this item's self: roll options for predication by preCreate rule elements
-                item.prepareActorData?.();
-
-                const rules = item.prepareRuleElements({ suppressWarnings: true });
-                for (const rule of rules) {
-                    const ruleSource = source.system.rules[rules.indexOf(rule)] as RuleElementSource;
-                    await rule.preCreate?.({ itemSource: source, ruleSource, pendingItems: nonKits, context });
+            // If any item we plan to add will add new items (such as ABC items), add those too
+            // When this occurs, keepId is switched on.
+            for (const item of items) {
```
stwlam, line 424:
```suggestion
            for (const item of [...items]) {
```

CarlosFdez: Good catch.

PR #5051, opened 2022-12-04
Update unmistakable-lore and experienced-professional

MrVauxs:


MrVauxs:
Note: The `lore-skill-check` selector is bugged on `@Check`s and thus currently the `AdjustDegreeOfSuccess` RE has little to no usage.

PR #5063, opened 2022-12-05
Fix description of Raging Athlete feat

stwlam:
Screenshot from 2022-12-05 14-51-24


TikaelSol:
Huh, this actually got a rewrite from the first printing apparently and we still had that one word in from the original description. Weird.

PR #5066, opened 2022-12-05
Unify "Evil or Good Only" spell restriction

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/adult-sovereign-dragon-spellcaster.json
```diff
@@ -2703,7 +2703,7 @@
                 }
             },
             "img": "systems/pf2e/icons/spells/detect-alignment.webp",
-            "name": "Detect Alignment (At Will) (Good or Evil Only)",
+            "name": "Detect Alignment (At Will) ((Evil or Good Only))",
```
TikaelSol, line 2706:
```suggestion
            "name": "Detect Alignment (At Will) ((Evil or Good Only))",
```
Ended up with extra parenthesis. Also looks like the books uses `Good or Evil` as the order.

image

I'd like to stick to Paizo's style, and if they have a conflicting style then I'd personally prefer to be "Good or Evil" instead of the other way around, since many English idioms use that ordering sop it sounds more natural than going in alphabetical order.

SpartanCPA: I agree

PR #5114, opened 2022-12-09
Remove unnecessary action cost type addition in npc sheet

CarlosFdez:


src/module/item/action/document.ts
```diff
@@ -36,7 +36,8 @@ class ActionItemPF2e extends ItemPF2e {
         const systemData = this.system;
 
         // Feat properties
-        const properties = [CONFIG.PF2E.actionTypes[systemData.actionType.value]].filter((property) => property);
+        const actionType = this.actionCost?.type ?? "passive";
+        const properties = [CONFIG.PF2E.actionTypes[actionType]].filter((property) => property);
```
stwlam, line 40: wot in ternation

CarlosFdez: Passive/Action/Reaction/Free show up on the properties of a feat/action chat message. This comes from either the system data, or the actionCost helper (which should probably be renamed to activationTime?). Pre-emptively, I changed it to use the helper so that later steps are a bit easier.

That said, the filter is probably silly and we should probably cull it....

PR #5132, opened 2022-12-10
Update style of inlines with repost buttons

Dorako:
Current (top button is hovered):
CleanShot 2022-12-10 at 14 45 17@2x
In AV premium module:
CleanShot 2022-12-10 at 14 47 40@2x

New:
CleanShot 2022-12-10 at 14 34 55@2x
on hovering the top icon:
CleanShot 2022-12-10 at 14 35 19@2x
In AV premium module:
CleanShot 2022-12-10 at 14 49 08@2x

### DIFF
--- a/src/styles/_globals.scss
+++ b/src/styles/_globals.scss
@@ -127,11 +127,6 @@ a.foundry-href {
     padding: 1px 0.2em 0;
     white-space: nowrap;
     word-break: break-all;
-
-    &.with-repost {
-        border-right: none;
-        padding-right: 0;
-    }
 }
 
 /* ----------------------------------------- */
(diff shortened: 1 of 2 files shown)


stwlam:
What about ones without the repost button?

Dorako:
> What about ones without the repost button?

Like these?
Current:
CleanShot 2022-12-10 at 16 21 16@2x
New:
CleanShot 2022-12-10 at 16 23 11@2x


I note that inline buttons are rendered slightly different in the chat and journals, causing alignment and overlapping issues in both the old and new styles.
Is there a technical reason inlines aren't using display: inline-flex? Would make it a lot easier to center the elements regardless of the fontsize/font family used.

TikaelSol:
Does this hit any snags with premium content?

Dorako:
> Does this hit any snags with premium content?

AV journals are pictured above. I'd say the introduction of the button styling krangled the AV journals in the first place, and this new styling undoes some of that.

I don't own any of the other modules to test with, and the Player's guide preview modules don't include any inlines with repost.

PR #5159, opened 2022-12-12
Add roll inspector support for casted spells

CarlosFdez:
The sooner we can get off of SpellPF2e#toMessage() the better.

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -35,7 +35,7 @@ export const ChatCards = {
             const strikeAction = message._strike;
             if (strikeAction && action?.startsWith("strike-")) {
                 const context = message.flags.pf2e.context;
-                const altUsage = context && context.type !== "damage-roll" ? context?.altUsage : null;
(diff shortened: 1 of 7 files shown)


src/module/chat-message/listeners/cards.ts
```diff
@@ -35,7 +35,7 @@ export const ChatCards = {
             const strikeAction = message._strike;
             if (strikeAction && action?.startsWith("strike-")) {
                 const context = message.flags.pf2e.context;
-                const altUsage = context && context.type !== "damage-roll" ? context?.altUsage : null;
+                const altUsage = context && "altUsage" in context ? context?.altUsage : null;
```
stwlam, line 38:
```suggestion
                const altUsage = context && "altUsage" in context ? context.altUsage : null;
```

stwlam:
What is meant by "add roll inspector support?" I can already inspect the attack and damage rolls of casted spells.

CarlosFdez:
It enables the roll inspector for dc computation. I adjusted it so only DC spells receive it. There's no modifiers at all, but the rest are visible

image

PR #5166, opened 2022-12-13
Add special prone-condition option to Cover effect

stwlam:
Prone with prone-cover vs melee attack:
image

Prone with prone-cover vs ranged attack:
image


stwlam:
Oh, yeah, got ahead of myself! I guess we could move the bonus over to the cover effect as a fourth option that only shows if the recipient is already prone.

PR #5178, opened 2022-12-13
Work around lack of top-level trait roll options for spell attack rolls

stwlam:


CarlosFdez:
We can add those top level traits if preferred?

stwlam:
Maybe, but can it wait until after release? I'd like to do a scan of any unintended side effects.

PR #5186, opened 2022-12-14
Add the Hidden condition to the token hud

Janonas:
The hidden condition comes up often in combat if you have a rogue or any stealthy PC/Enemy. Despite it lacking integration, we have concealed and dazzled in the token hud which also lack integration. It has value in remiding the GM and players who is hidden and who is not, without having to go through the conditions compedium. Plus there is a module which automates the flat checks to creatures with this condition to which it would be a boon.


TikaelSol:
Overlaps with #4877, so we only need to merge one of the two.

stwlam:
Duplicates efforts of #4877

PR #5194, opened 2022-12-15
Fix sources and remove junk data from Blood Lords book 6

TikaelSol:
The macro I used to set the sources apparently had the item and actor style source path swapped, so the junk source details needs removed from both actors and items, and the correct source added for items.


SpartanCPA (approved):

SpartanCPA:
Something is breaking the pipeline

PR #5223, opened 2022-12-17
Implement custom damage types

CarlosFdez:
image

image

No color support and no uncategorized support. But I think this might a good enough start?


stwlam:
What happens if you input an invalid icon class?

stwlam:
Also, title text on hovering the custom-typed instances?

CarlosFdez:
If the invalid icon doesn't start with fa or has a space, it gets wiped to empty space. Anything else it should leave a blank space.

As for title text, almost....
image

But I caught this separately
image

stwlam:
Could there be an tooltip'd info box with information on what inputs are accepted? Just looking at it, a typical user won't know what to input, knowing nothing about font awesome.

CarlosFdez:
Not against it, though given this is a feature used by a subset of users, I'm not sure to what degree we should have these bells and whistles. But if you think we should then sure I'll give it a shot.

The roll flavor stuff is going to be tricky though. It seems there's a set of configured "lower case" variants of all the type. Not sure if we should have a flavor textbox as well, or have it automatically added via toLowerCase() for now.

stwlam:
I'm ok with just lower-casing it

stwlam:
Also, can we do the standard `hb_` prefix for this?

CarlosFdez:
Not against the hb_ prefix, though I should probably show the slug in the form somewhere even if its not editable.

Waiting until you finalize what you're doing with persistent/splash before continuing. There's some time remaining at least.

CarlosFdez:
And I believe this is done.
image

PR #5235, opened 2022-12-18
Add support for migrating journal entries

stwlam:


packs/scripts/run-migration.ts
```diff
@@ -185,7 +197,11 @@ async function migrate() {
 
         // skip journal entries, rollable tables, and macros
```
CarlosFdez, line 198: How accurate is this comment now?

PR #5240, opened 2022-12-19
Fix circular imports in @actor/character

stwlam:


CarlosFdez:
seems something in the weapon's import path does some instanceof checks instead of isOfType() but ohwell.

PR #5251, opened 2022-12-19
Improve Alchemist Research Field Homebrew Support

velara:
Revise Alchemist Fields (Bomber, Chirurgeon, Mutagenist, Toxicologist) to add flags with the UUIDs for the various choice class freatures (field discovery, etc.)

Update field discovery, greater field discovery, perpetual infusions, perpetual perfection and perpetual potency to grant item based on the flags set by the field.

Update Research Field item allowedDrops predicate so that homebrew items were recognized.


velara:
If this ends up being an acceptable approach I'll go through and do the same for the other classes so we've got a nice standard system.

packs/data/classfeatures.db/chirurgeon.json
```diff
@@ -21,7 +21,38 @@
...
+                "path": "flags.pf2e.alchemist.greaterFieldDiscovery",
+                "value": "Compendium.pf2e.classfeatures.JJcaVijwRt9dsnac"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.alchemist.perpetualPerfection",
+                "value": "Compendium.pf2e.classfeatures.YByJ9O7oe8wxfbqs"
+            }
+        ],
```
stwlam, line 55: Instead of multiple AE-likes, have a single one that sets an object

velara: You blew my mind a little. I didn't realize I could pass objects with AE-likes.

stwlam:
This looks good, but it's going to need a migration so as to not break all existing alchemist PCs.

TikaelSol:
`npm run lint:fix` should fix the pipeline, and you'll need to switch this to be migration 812.

velara:
I've updated the migration to be 812 and got it passing on the checks. Thanks for the help turns out I've forgotten most of what I learned from my last larger non-data entry commit a year+ back. 

As per our conversation on the discord about being unable to migrate grant items I've only migrated the four research fields which set the flags. 

I then ran the migration on a test world with a level 1, 10, and 20 alchemist of each of the four fields. They all migrated successfully. The higher level alchemists all kept the old version of their Field Discovery and other similar items, however if I increased the level they could the new versions of any of the features they did not have. Ex, Perpetual Perfection being added to a low level alchemist, and were able to find the flag and grant the appropriate item for the field successfully.

Down levelling them and re-levelling them up replaced all the old versions of the class features with the new ones. Which is all the things I can think of going wrong, so hopefully this migration should keep anyone's characters from being broken by my alchemist changes.

velara:
I've updated to be Migration 814 and merged all the existing changes into my branch.

velara:
Updated to be Migration 815, and fixed a mistake I'd made when I corrected merge conflicts, where I accidentally selected the older version of the research field text.

stwlam:
Could you run the migration and push up the results?

velara:
I’m away from home this weekend, but I will migrate the compendiums when I’m back and push the results.

velara:
I've run the migration and committed the pack updates. Some seem to be unrelated to this migration, so I wonder if they just got missed during a previous migration.

packs/data/iconics.db/fumbus-level-5.json
```diff
@@ -2772,137 +2772,6 @@
             },
             "type": "feat"
         },
-        {
```
stwlam, line 2775: How did this get deleted?

velara: I totally forgot to include this change in the commit write up. Given this was the one instance I knew in the compendium of an outdated version of the Bomber field, I manually updated Fumbus to use the updated version of the bomber field, by dropping Fumbus to 4th level and then back to 5th.

stwlam: Ah, gotcha. Sounds good, then!

PR #5271, opened 2022-12-21
Add support for AE-like subtractions/array-element removals

stwlam:


src/module/rules/rule-element/adjust-strike.ts
```diff
@@ -132,36 +136,39 @@ class AdjustStrikeRuleElement extends AELikeRuleElement {
...
                                     const changeTraitScore = getTraitScore(changeValue);
 
-                                    // If the new trait's score is higher than the existing trait's score, then remove the existing one
-                                    // otherwise just return out as we don't want to add the new (lesser) trait
+                                    // If the new trait's score is higher than the existing trait's score, then remove
+                                    // the existing one otherwise just return out as we don't want to add the new
+                                    // (lesser) trait
                                     if (changeTraitScore > existingTraitScore) {
-                                        traits.value.findSplice((trait) => trait === existingTraitMatch[0]);
+                                        traits.findSplice((trait) => trait === existingTraitMatch[0]);
```
CarlosFdez, line 160: Is it better to return here as well?

stwlam: Sorry, what? I don't think I changed anything here

CarlosFdez: You added an if else to the bottom. If this branch is hit, it continues on to the if/else. Which I think get ignored, because it'll fail the if, but I think you can change this entire if/else to an if with a return afterwards.

stwlam: image

CarlosFdez: yes, I mean that you can probably remove the else and just return. Unless its important for if (this.mode === "add" && !traits.includes(change)) if/else is intentional to happen even if you get in here.

JDCalvert: @stwlam @CarlosFdez the reason the `return` was in an `else` was so that we didn't add the new trait if the _new_ trait had a lower "score" than the _existing_ trait (e.g. trying to add `deadly-d6` when you already had `deadly-d10`).

Moving the `return` out of the `if` means that if we're trying to add `deadly-d8` and we already have `deadly-d6`, then we'll remove the `deadly-d6` trait and then we `return` out and never add the `deadly-d8` trait. I've reverted this change in #5866.

PR #5275, opened 2022-12-21
Added support for spell damage traits, modifier breakdown, and damage instances

CarlosFdez:
image


CarlosFdez:
Healing hands still work
image

Still no support for adding damage dice, but at this point its not a lot of work to add I think, and its a new feature anyways.

PR #5279, opened 2022-12-21
Replace empty action icon in Compendium Browser spell tab with a new icon for longer actions

In3luki:
Looks like `getActionIcon` no longer returns the mystery man as a default image so this had to be adjusted too. Showing the one action icon has always been wrong so this should be better.

Before:
before

After:
after


stwlam:
I wish we had an icon for that, like a clock.

In3luki:
I found https://game-icons.net/1x1/delapouite/solar-time.html . Other possible candidates are a clock tower or an alarm clock but nothing really fits.

image

In3luki:
The "sun rays" are a bit jaggy. Not sure how to fix that.

image

PR #5284, opened 2022-12-21
Fix tokens with animated art disappearing on reload/scene switch

### LINKED ISSUE #5283: Animated tokens disappear on reload/scene switch
This is a weird one! I put together a fancy new animated token for a PFS character. However, when I actually use this as token art, the token seems to flicker once and then vanish entirely - selection border included - when reloading the page or switching scenes. Upon selecting another, non-vanished token, it reappears. When there are multiple animated tokens, only the one that was placed first seems to disappear - though all of them flicker for a moment when loading the scene.

Some other observations and details:
- This happens in Firefox (108.0.1), Chromium (108.0.5359.124), and Electron (whichever version foundry bundles) though there are subtle differences - on Firefox the tokens seem to vanish entirely as described above, while on Chromium and Electron the selection borders seem to remain visible for me.
- This only happens in worlds with the pf2e system - I could not reproduce the flickering nor the disappearance in a world using the dnd5e system
- This is only reproducible in worlds using the pf2e system, and all other modules are disabled
- I was able to reproduce this both with my local development setup (local copy of foundry v10 build 291, pf2e system built from HEAD at time of writing) as well as my main instance that I use for games (foundry v10 build 291 hosted on oracle cloud behind a reverse proxy and cloudflare, pf2e system version 4.5.1)
- This happens both when using token art from the foundry data directory (e.g. `tokens/animated_token.webm`) as well as when hotlinking the file over the web (e.g. `https://static.dmarcuse.me/kaylee_token2_animated.webm`)

Here's a video of the issue in Firefox:
https://user-images.githubusercontent.com/5429337/209010586-3052c362-fcb0-4b5d-a654-7b3513727551.mp4

And here's a copy of the animated token itself:
https://user-images.githubusercontent.com/5429337/209010930-f37e0f1f-65ec-4522-8154-0f6e69dab591.webm

dmarcuse:
When redrawing tokens during a scene, animated tokens (and perhaps other tokens with art that loads slowly?) would sometimes have `visible` set to `false` incorrectly. The `isVisible` getter seems to always behave correctly though, so this simple one-line change fixes this issue.

Tokens that have been hidden by the GM or are out of sight due to walls/darkness still seem to work properly with this PR, but this may have other implications I'm not aware of, and I'm happy to change this PR if necessary.

Closes #5283

### DIFF
--- a/src/scripts/hooks/canvas-ready.ts
+++ b/src/scripts/hooks/canvas-ready.ts
@@ -25,16 +25,14 @@ export const CanvasReady = {
 
             // Redraw tokens
             for (const token of canvas.tokens.placeables) {
-                let redrawTrigger = Promise.resolve();
+                const redrawTrigger = token.isVideo
+                    ? new Promise((resolve) => {
+                          // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
+                          // invisible if drawn immediately after they begin playing
+                          token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50), { once: true });
+                      })
+                    : Promise.resolve();
 
-                // Animated tokens sometimes disappear if the underlying video isn't given a moment to begin playing
-                if (token.isVideo) {
-                    redrawTrigger = new Promise((resolve) => {
-                        // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
-                        // invisible if drawn immediately after they begin playing
-                        token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50), { once: true });
-                    });
-                }
                 redrawTrigger.then(async () => {
                     const { visible } = token;
                     await token.draw();
--- a/src/scripts/hooks/canvas-ready.ts
+++ b/src/scripts/hooks/canvas-ready.ts
@@ -32,7 +32,7 @@ export const CanvasReady = {
                     redrawTrigger = new Promise((resolve) => {
                         // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
                         // invisible if drawn immediately after they begin playing
-                        token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50));
+                        token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50), { once: true });
                     });
                 }
                 redrawTrigger.then(async () => {
--- a/src/scripts/hooks/canvas-ready.ts
+++ b/src/scripts/hooks/canvas-ready.ts
@@ -23,14 +23,24 @@ export const CanvasReady = {
                 CONFIG.MeasuredTemplate.defaults.angle = canvas.scene.hasHexGrid ? 60 : 90;
             }
 
-            Promise.resolve().then(async () => {
-                // Redraw tokens
-                for (const token of canvas.tokens.placeables) {
+            // Redraw tokens
+            for (const token of canvas.tokens.placeables) {
+                let redrawTrigger = Promise.resolve();
+
+                // Animated tokens sometimes disappear if the underlying video isn't given a moment to begin playing
+                if (token.isVideo) {
+                    redrawTrigger = new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


stwlam:
I guess we'll see

stwlam:
One thing you can try testing with is tokens with different ephemeral transformations like size, light, and texture.

dmarcuse:
Hmm... I did some more testing, and while Enlarge, Light, and custom effects to change token images all worked as expected, it seems that player users (but not GMs) still occasionally have animated tokens disappear. I'm not really sure what's going on under the hood here - some sort of race condition between loading the token art and the properties?

dmarcuse:
Alright, I've put together a more robust fix. It seems the root cause is related to tokens being drawn before the video has had a chance to start playing, so I've updated the PR to instead wait for the `play` event (plus an extra 50ms for safety) before redrawing animated tokens. With this, I can no longer reproduce the issue as either a player or a GM, and other features like walls, lighting, token hiding, and temporary token size/image changes seem to work as expected.

src/scripts/hooks/canvas-ready.ts
```diff
@@ -23,14 +23,24 @@ export const CanvasReady = {
...
+            // Redraw tokens
+            for (const token of canvas.tokens.placeables) {
+                let redrawTrigger = Promise.resolve();
+
+                // Animated tokens sometimes disappear if the underlying video isn't given a moment to begin playing
+                if (token.isVideo) {
+                    redrawTrigger = new Promise((resolve) => {
+                        // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
+                        // invisible if drawn immediately after they begin playing
+                        token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50));
```
stwlam, line 35: This is going to continuously add event listeners to the token.

stwlam: This is been in place in one form or another since V9, so maybe it's not needed anymore. Test the cases I mentioned before without the redraw present. If they still work, we can remove it entirely.

dmarcuse: > This is going to continuously add event listeners to the token.

I'm not quite sure what you mean - `new Promise` only calls the callback once, so only one event listener is added. The event listener should only ever fire once since the user cannot control the video element and it loops continuously, and `setTimeout` also fires only once.

I'll configure the listener with `{ once: true }` just to be safe, in case there's an edge case I'm missing with the video playback.

> This is been in place in one form or another since V9, so maybe it's not needed anymore. Test the cases I mentioned before without the redraw present. If they still work, we can remove it entirely.

Things seem to mostly work. The only issue I noticed is that if the enlarge effect is applied by a different user while you're viewing a different scene, switching back to the original scene sometimes shows tokens with the wrong size until you hover over/select a token. This seems like a minor enough bug that it might be fine to just remove this code anyways in the interest of simplifying things.

src/scripts/hooks/canvas-ready.ts
```diff
@@ -23,14 +23,24 @@ export const CanvasReady = {
...
+
+                // Animated tokens sometimes disappear if the underlying video isn't given a moment to begin playing
+                if (token.isVideo) {
+                    redrawTrigger = new Promise((resolve) => {
+                        // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
+                        // invisible if drawn immediately after they begin playing
+                        token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50), { once: true });
+                    });
+                }
+                redrawTrigger.then(async () => {
```
stwlam, line 38:
```suggestion
                const redrawTrigger = token.isVideo
                    ? new Promise((resolve) => {
                          // Wait until 50ms after the video starts playing to be safe - sometimes they'll still be
                          // invisible if drawn immediately after they begin playing
                          token.sourceElement.addEventListener("play", () => setTimeout(resolve, 50), { once: true });
                      })
                    : Promise.resolve();

                redrawTrigger.then(async () => {
```

dmarcuse:
Thank you!

PR #5310, opened 2022-12-24
Draft: Add variants to Dragon Breath and remove Dragon Breath duplicate spells.

xdy:


TikaelSol:
Do we have any NPCs with these spells?

xdy:
Hmm. Good question. I'll draft this and replace on any npcs in a separate commit on this PR.

xdy:
Just one npc (with two variants), Yabin the Just from FotRP.
But, I've realized that this PR has the wrong approach to do this. The selection should be made automatically when one *gets* the spell, not every time it's cast. Which I have no clue how to do.
So, I'll close the PR.

PR #5311, opened 2022-12-24
Add safeguards to `GrantItemRuleElement#reevaluateOnUpdate`

stwlam:
- Force `replaceSelf` and `allowDuplicate` to be false
- Fail validation if no predicate is included


CarlosFdez:
Funny enough, reevaluateOnUpdate is a silly hack I do when I manually edit an RE to have a grant item, so that it can actually grant the item when added to an item already on an actor. Shame to lose that, but I guess we should hope for a real solution eventually.

PR #5316, opened 2022-12-25
Add support for setting border and fill alphas to the Auras RE

### LINKED ISSUE #3359: Suggestion: Have the ability to hide/adjust the transparency of the auras
It would be a great thing to have for those situations where there becomes a bit too much noise on the screen, but still have the great aura utility.

xdy:
E.g.
```{"key":"Aura","radius":10,"slug":"test-aura","alphas":{"border":0.23,"fill":0.42}}```
or
```{"key":"Aura","radius":10,"slug":"test-aura","colors": {"border":"#ff00ff", "fill": "#00ff00"}, "alphas":{"border":0.23,"fill":0.42}}```

The previous defaults of border 0.75 and fill 0 are used if the alphas property is not set.

I chose to keep alphas as a separate property from colors so as to avoid the need for any migration, and to keep them simpler.

### DIFF
--- a/src/module/canvas/token/aura/renderer.ts
+++ b/src/module/canvas/token/aura/renderer.ts
@@ -1,4 +1,4 @@
-import { AuraColors, AuraData } from "@actor/types";
+import { AuraData, AuraPart } from "@actor/types";
 import { TokenPF2e } from "..";
 import { EffectAreaSquare } from "../../effect-area-square";
 import { ItemTrait } from "@item/data/base";
@@ -19,8 +19,11 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
     /** Traits associated with this aura: used to configure collision detection */
     traits: Set<ItemTrait>;
 
-    /** Border and fill colors in hexadecimal */
-    private colors: TokenAuraColors;
+    /** Border color and alpha */
+    private border: TokenAuraPart;
+
+    /** Fill color and alpha. Named auraFill so as to not override Graphics.fill */
+    private auraFill: TokenAuraPart;
 
     /** Standard line thickness for circle shape and label markers */
     static readonly LINE_THICKNESS = 3;
@@ -29,7 +32,8 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
         super();
 
         this.token = params.token;
-        this.colors = this.#convertColors(params.colors);
+        this.border = this.#convertPart(params.border, 0.75);
+        this.auraFill = this.#convertPart(params.fill, 0);
         this.radius = params.radius;
         this.radiusPixels = 0.5 * this.token.w + (this.radius / (canvas.dimensions?.distance ?? 0)) * canvas.grid.size;
         this.traits = new Set(params.traits);
@@ -90,8 +94,8 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
         this.visible = false;
         if (!this.shouldRender) return;
 
-        this.beginFill(this.colors.fill, 0)
-            .lineStyle(AuraRenderer.LINE_THICKNESS, this.colors.border, 0.75)
+        this.beginFill(this.auraFill.color, this.auraFill.alpha ?? 0)
+            .lineStyle(AuraRenderer.LINE_THICKNESS, this.border.color, this.border.alpha ?? 0.75)
             .drawCircle(this.token.w / 2, this.token.h / 2, this.radiusPixels)
             .endFill();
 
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by stwlam** [comment]
That data structure is super gross. :(

**Comment by stwlam** [comment]
Also, I don't think this would address the spirit of the feature request. Of course editing a rule element would achieve toning down the opacity of an aura, but that shouldn't be the ad hoc workflow.

**Comment by xdy** [comment]
> That data structure is super gross. :(
True enough. I can't think of a better one though.

**Comment by xdy** [comment]
> Also, I don't think this would address the spirit of the feature request. Of course editing a rule element would achieve toning down the opacity of an aura, but that shouldn't be the ad hoc workflow.

Also true, edited the first comment to remove the "Fixes..." part of it. Will push a new commit without it.

**Comment by xdy** [comment]
Would be easy enough to change to something like this if you think it'd be better:
```{
  "key": "Aura",
  "radius": 10,
  "slug": "test-aura",
  "aura": {
    "border": {
      "color": "#ff00ff",
      "alpha": 0.23
    },
    "fill": {
      "color": "#00ff00",
      "alpha": 0.42
    }
  }
}
```
Would need a migration though.

xdy:
Should probably move "radius" inside "aura" also in that case.

stwlam:
The `aura` bit probably isn't necessary. `border` and `fill` can be at the top level.

xdy:
Sounds good. I'll change it to this when I have the time:
```
{
  "key": "Aura",
  "slug": "test-aura",
  "radius": 10,
  "border": {
    "color": "#ff00ff",
    "alpha": 0.23
  },
  "fill": {
    "color": "#00ff00",
    "alpha": 0.42
  }
}
```

xdy:
New data structure pushed, it seems no migration is needed as the system has no auras with colors.

CarlosFdez:
my first thought is that setting aura appearances outta be like setting templates. It should have a similar data structure to templates, so much so that modules that affect templates that target pf2e could reasonably artsy it up.

Its not actually templates, so it'd never be automatic, but the closer in appearance the API is the neater such a thing can be.

xdy:
Similar enough, imho (template example from measured-template.ts):
```{
  t: "cone",
  user: game.user._id,
  x: 1000,
  y: 1000,
  direction: 0.45,
  angle: 63.13,
  distance: 30,
  borderColor: "#FF0000",
  fillColor: "#FF3366",
  texture: "tiles/fire.jpg"
}```
t, x, y, direction,angle,distance are covered by key and radius for auras.
User is implicit as it's on an RE on an item on an actor owned by a user.
borderColor and fillColor are covered by border.color and fill.color for auras.
Alphas are new of course.
And, I'm seriously considering looking at adding textures to auras. Would be my first foray into pixi land, so would like take forever, and might fail entirely. Either way, not something for this PR.

**Comment by stwlam** [comment]
> New data structure pushed, it seems no migration is needed as the system has no auras with colors.

Users do

File: src/module/canvas/token/aura/renderer.ts [hunk 1]
```diff
@@ -111,29 +115,32 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
...
+        const defaults = {
+            color: Number(foundry.utils.Color.fromString(user.color ?? "#0000000")) || 0,
+            alpha: defaultAlpha,
+        };
+        if (part) {
+            const color = foundry.utils.Color.fromString(part.color);
             return {
-                border: Number(foundry.utils.Color.fromString(colors.border)) || 0,
-                fill: Number(foundry.utils.Color.fromString(colors.fill)) || userColor,
+                color: Number(color ?? defaults.color),
```
stwlam, line 139: `color` is always going to be non-nullish

xdy:
I can't see why the colors field gets set to null instead of deleted. It's probably something obvious, but, I'm stuck.

xdy:
> I can't see why the colors field gets set to null instead of deleted. It's probably something obvious, but, I'm stuck. I _think_ I've done the same as in other migrations, but, obviously not.

I got it working. Seems that the -= dance isn't needed anymore. At least for this migration. :)

TikaelSol:
migration 811 got taken already, and Avery has 812 looking in good shape so this will need to be 813 at least.

TikaelSol:
Well shark just took 812 too. Sorry, big migration time.

xdy:
I've changed to use migration 814, I guess we'll see if that to gets taken. :)

CarlosFdez:
I get the feeling this is gonna get more of a looksee after IWR is much more stable 🙃

xdy:
Yeah, I'll just look in on occasion when I happen to notice a conflict. :)

xdy:
Pushed as 818, safe to ignore for a bit. :)

stwlam:
I'm not sure why this should be a per-RE thing. People who are annoyed by black circles aren't annoyed by only some of them.

CarlosFdez:
Similar feeling. Any such properties should be added to look pretty, which is template would resolve. If this is because some auras are annoying, perhaps this should be resolved more globally?

xdy:
I figured the usecases for this are:
Accessibility - more or less visible auras, making the auras visible without hovering.
Improved appearance/homebrew (especially if/when textured auras become a thing) - 'that aura is a cloud of smoke, so I'll make it's fill less transparent'

CarlosFdez:
For accessibility, doing this globally might be best. Either as something you feel for sure is a better set of defaults, or start the beginnings of the "look and feel" section that allows you to add additional specific changes to existing battleforms and auras as such (such as textures, image changes, etc).

For the latter, I'd feel more comfy if you put textures along with everything.

xdy:
Hmm. Might as well close this PR then as I don't think I'll get around to doing anything about textures anytime soon. For one thing, I haven't a clue where to start. :)

PR #5317, opened 2022-12-25
Fix token effects not redrawing when conditions icons are changed

CarlosFdez:


src/module/canvas/status-effects.ts
```diff
@@ -45,36 +45,15 @@ export class StatusEffects {
...
-
         CONFIG.PF2E.statusEffects.iconDir = iconDir;
         CONFIG.PF2E.statusEffects.lastIconTheme = chosenSetting;
         this.#updateStatusIcons();
+        await resetAndRerenderActors();
+        if (canvas.scene) {
+            for (const token of canvas.scene.tokens) {
+                token.object.drawEffects();
+            }
+        }
```
stwlam, line 56:
```suggestion
        if (canvas.ready) {
            for (const token of canvas.tokens.placeables) {
                token.drawEffects();
            }
        }
```

PR #5323, opened 2022-12-26
Implement RE form for fast healing and regeneration

CarlosFdez:
image


CarlosFdez:
Need to make fast healing chat message pretty somehow. Will give that a shot later.

stwlam:
The deactivation probably shouldn't be from the list of weaknesses. Some of them wouldn't make enough sense to be implementable.

CarlosFdez:
Got a thought? The deactivations list are mostly for localization, not for implementing. They can accept arbitrary strings (the whitelist isn't enforced), and what we're likely to do is that some are automated and some are not.

PR #5345, opened 2022-12-28
Remove `value` property from loot actor description

In3luki:
This fixes the HTML sanitation error and allows loot actors to be created again.

### DIFF
--- a/src/module/migration/migrations/810-loot-description-value.ts
+++ b/src/module/migration/migrations/810-loot-description-value.ts
@@ -1,0 +1,20 @@
+import { ActorSourcePF2e } from "@actor/data";
+import { MigrationBase } from "../base";
+
+/** Remove value property from loot actor description */
+export class Migration810LootDescriptionValue extends MigrationBase {
+    static override version = 0.81;
+
+    override async updateActor(source: ActorSourcePF2e): Promise<void> {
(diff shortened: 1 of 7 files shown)


src/module/migration/migrations/810-loot-description-value.ts
```diff
@@ -0,0 +1,20 @@
...
+import { MigrationBase } from "../base";
+
+/** Remove value property from loot actor description */
+export class Migration810LootDescriptionValue extends MigrationBase {
+    static override version = 0.81;
+
+    override async updateActor(source: ActorSourcePF2e): Promise<void> {
+        if (source.type === "loot") {
+            const details: MaybeWithValue = source.system.details;
+            if (typeof details.description === "object") {
```
stwlam, line 11: use `instanceof Object` just in case of the extremely unlikely scenario that `description` is `null`

src/module/migration/migrations/810-loot-description-value.ts
```diff
@@ -0,0 +1,20 @@
...
+                details.description = details.description.value;
```
stwlam, line 12:
```suggestion
                details.description = String(details.description.value ?? "");
```
Additional paranoia

PR #5351, opened 2022-12-29
Create persistent damage conditions from drag drop inline rolls

CarlosFdez:
I am blanking on visual changes to make it obvious its drag/droppable though. 

We will also need a migration for persistent damage inline rolls (the previous migration kinda skipped it, some still have the {1}[persistent,acid] syntax). The migration will also need to remove all embedded persistent damage conditions following an inline roll (which was never necessary in the module do not put that evil on me!)

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -67,6 +67,12 @@ class TextEditorPF2e extends TextEditor {
             anchor.innerHTML = `${icon.outerHTML}${label}`;
             anchor.dataset.tooltip = roll.formula;
             anchor.dataset.damageRoll = "";
+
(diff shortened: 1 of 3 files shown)


src/module/system/text-editor.ts
```diff
@@ -67,6 +67,12 @@ class TextEditorPF2e extends TextEditor {
             anchor.innerHTML = `${icon.outerHTML}${label}`;
             anchor.dataset.tooltip = roll.formula;
             anchor.dataset.damageRoll = "";
+
+            const isPersistent = !roll.instances.some((i) => !i.persistent);
```
stwlam, line 71:
```suggestion
            const isPersistent = roll.instances.length > 0 && roll.instances.every((i) => i.persistent);
```

PR #5352, opened 2022-12-29
Include damage modifiers and dice in roll inspector

CarlosFdez:
Note that currently we don't have a way to tell which modifiers were used in which outcomes, since damage templates do every category at once. That said, with the critical property being shown, I think people can still deduce it anyways.

image

### DIFF
--- a/src/module/chat-message/chat-roll-details.ts
+++ b/src/module/chat-message/chat-roll-details.ts
@@ -1,3 +1,5 @@
+import { BaseRawModifier, DamageDicePF2e } from "@actor/modifiers";
+import { addSign } from "@util";
 import { ChatMessagePF2e } from ".";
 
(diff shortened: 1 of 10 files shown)


stwlam:
You can't check that they're enabled?

CarlosFdez:
For a specific outcome, no. Only for predicate based ones. That said, its still a bit lacking, and I don't think it'll be perfect unless we refactor the damage pipeline further. For example, sneak attack is omitted entirely rather than disabled.

image

Its because weapon/melee attempts to calculate each outcome simultaneously via the same set of modifiers rather than independently.

CarlosFdez:
I identified why sneak attack is omitted entirely instead of disabled. Its because deferred damage dice return null for ignored ones.

PR #5357, opened 2022-12-30
Allow all super duper damage categories for flat modifier re form and spells

CarlosFdez:


src/module/item/spell/data.ts
```diff
@@ -22,7 +22,7 @@ type SpellDamageCategory = keyof ConfigPF2e["PF2E"]["damageCategories"];
 
 export interface SpellDamageType {
     value: DamageType;
-    subtype?: "persistent" | "splash";
+    subtype?: DamageCategoryMain;
```
stwlam, line 25: There isn't anything "main" about these. Energy is just as much of a category as precision.

CarlosFdez: I suspected this would be the hangup. Yeah they're definitely not, but its closer than DamageCategoryMutuallyExclusive. They're also not really primary either.

stwlam: Did you suspect it because I already said as much? :stuck_out_tongue: 

Special, complementary, parallel--but not "main"

CarlosFdez: They're not special either, adamantine is just as special as precision. Everything is complementary to everything, everything is parallel to everything.

stwlam: In the context of the form, it's an additional category added to what would otherwise just be the category of the damage type (or no category if the damage type lacks one).

CarlosFdez: could do DamageCategory and DamageCategoryAdditional :P

CarlosFdez: The reason why is because this damage category, the one I called "main", is the single choice top level one that damage dice / flat modifier / spells use. In that context, it is the main one we use.

The alternative way of thinking of it is that all other damage categories are "bonus". So either we call the new one DamageCategoryMain, or we promote the new one to DamageCategory and call the old one something new (either DamageCategoryBonus or DamageCategoryAdditional).

PR #5395, opened 2023-01-01
Remove superfluous main category and fix negative spell modifiers

CarlosFdez:
Fixes spell breakdowns and negative ability modifiers. Removes the "energy" from the below:
image


CarlosFdez:
so the superfluous "energy" category was causing negative modifiers to be listed as 1d4 + ( - 1[energy]) which caused it do some wonkiness. That said, combine terms probably should be smarter and be able to handle the parse the second expression. Maybe for a future update.

PR #5396, opened 2023-01-01
Add custom damage types to IWR

CarlosFdez:
Not sure if bugfix or enhancement.

### DIFF
--- a/src/module/system/settings/homebrew/menu.ts
+++ b/src/module/system/settings/homebrew/menu.ts
@@ -5,7 +5,7 @@ import { LocalizePF2e } from "@module/system/localize";
 import { htmlClosest, htmlQuery, htmlQueryAll, isObject, objectHasKey, pick, sluggify, tupleHasValue } from "@util";
 import Tagify from "@yaireo/tagify";
 import { PartialSettingsData, SettingsMenuPF2e, settingsToSheetData } from "../menu";
-import { isHomebrewFlagCategory, prepareCleanup } from "./helpers";
+import { isHomebrewCustomDamage, isHomebrewFlagCategory, prepareCleanup } from "./helpers";
 import "@yaireo/tagify/src/tagify.scss";
 import {
     BASE_DAMAGE_TYPES_TO_CATEGORIES,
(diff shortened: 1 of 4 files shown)


src/module/system/settings/homebrew/menu.ts
```diff
@@ -280,6 +281,11 @@ class HomebrewElements extends SettingsMenuPF2e {
             DAMAGE_TYPE_ICONS[slug] = data.icon?.substring(3) ?? null; // icons registered do not include the fa-
             CONFIG.PF2E.damageTypes[slug] = data.label;
             CONFIG.PF2E.damageRollFlavors[slug] = data.label.toLocaleLowerCase();
+
+            for (const collection of [immunityTypes, weaknessTypes, resistanceTypes]) {
+                const collectionWidened: Record<string, string> = collection;
+                collectionWidened[slug] = data.label.toLocaleLowerCase();
```
stwlam, line 287: Should sluggify and also clean up when it's removed

CarlosFdez: The slug is already sluggified. The flavor shouldn't be though. As for cleaning up, I think deleting entire entries from existing monsters is kinda too much. I think the `undefined 10` is actually better.

stwlam: The migration process should take care of it. If a GM removes a homebrew thingy, what chance is there that they still want it?

CarlosFdez: They could have renamed it for example rather than removing it. That still counts as a removal.

We can hold till tomorrow to decide what to do with this after you're spent with the new year. Personally I'm tempted on making it such that IWR data can be detected as invalid and deactivated.

stwlam: New year till June

CarlosFdez: Done, (with removing the config). I had to add reading from flags since it was going to require some fundamental changes, and I wanted to be sure I had it all right.

Turns out the current traits have a potential bug, if I'm reading it right, where if a user adds a trait matching a module and then removes it, it migrates incorrectly. Avoided that bug for damage types, in exchange for it being a little weird looking.

CarlosFdez:
Slipped in a small persistent damage regression since I caught that with the custom damage types.

PR #5410, opened 2023-01-01
Have `DamageRoll#total` return a minimum of 1

In3luki:
Should the minimum damage override be noted somewhere?

image


CarlosFdez:
Note that this is not true for IWR. How does this interact with resistances?

PR #5423, opened 2023-01-02
Add support for condition immunity

stwlam:
image

Screenshot from 2023-01-01 22-29-08


src/module/actor/loot/document.ts
```diff
@@ -38,6 +38,11 @@ class LootPF2e extends ActorPF2e {
         return false;
     }
 
+    /** It's a sturdy box. */
```
CarlosFdez, line 41: You actually slayed me with this one!

PR #5441, opened 2023-01-02
Fixes for ancestors oracle curse effects not applying correctly

AFigureOfBlue:
I'm running an Ancestors oracle and ran into some challenges with the curse's effects not applying properly. Here's as many of the fixes as I could figure out how to implement.

* "flag" needed to be changed to "rollOption" in order for any of the other rule elements to actually apply.
* Added failure notes for Perception checks when not on the Skillful ancestor.
* Don't apply Battle ancestor's attack bonus to spell attacks (it should only be for Strikes)
* No longer limit the bonus from the Spellcasting ancestor to damaging effects (because it's also supposed to apply to healing).

The only elements of the curse that still aren't handled properly after these changes is that the damage bonus from the Spellcasting ancestor should apply to any non-cantrip spells, not just those cast from spell slots (that is, it should apply to focus spells, innate spells, etc.) I wasn't readily able to figure out on my own how to structure a predicate to that effect. Additionally (though lower priority), there isn't a note provided about the flat check being needed when Casting a Spell when not on the Spellcasting ancestor.

### DIFF
--- a/packs/data/feat-effects.db/effect-ancestor-curse.json
+++ b/packs/data/feat-effects.db/effect-ancestor-curse.json
@@ -118,7 +118,7 @@
                         ]
                     }
                 ],
-                "selector": "mundane-attack",
+                "selector": "strike-attack-roll",
                 "text": "PF2E.OracleCurses.Ancestor.MinorFailure",
                 "title": "{item|name}"
             },
@@ -160,7 +160,7 @@
                         ]
                     }
                 ],
-                "selector": "mundane-attack",
+                "selector": "strike-attack-roll",
                 "text": "PF2E.OracleCurses.Ancestor.ModerateFailure",
                 "title": "{item|name}"
             },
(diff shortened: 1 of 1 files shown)


packs/data/feat-effects.db/effect-ancestor-curse.json
```diff
@@ -238,7 +280,7 @@
                         ]
                     }
                 ],
-                "selector": "attack",
+                "selector": "mundane-attack",
```
TikaelSol, line 283:
```suggestion
                "selector": "strike-attack-roll",
```

packs/data/feat-effects.db/effect-ancestor-curse.json
```diff
@@ -146,7 +160,7 @@
...
```
TikaelSol, line 163:
```suggestion
                "selector": "strike-attack-roll",
```

packs/data/feat-effects.db/effect-ancestor-curse.json
```diff
@@ -118,7 +118,7 @@
...
```
TikaelSol, line 121:
```suggestion
                "selector": "strike-attack-roll",
```

AFigureOfBlue:
Good to know about `strike-attack-roll`! It wasn't in the wiki's list of available selectors, but I'll keep it in mind going forward.

PR #5446, opened 2023-01-02
Define Organsight damage as Precision Damage

### LINKED ISSUE #5422: Organsight grants extra damage via damage dice, but it is not labelled precision damage

SpartanCPA:
Closes #5422

### DIFF
--- a/packs/data/spell-effects.db/spell-effect-organsight.json
+++ b/packs/data/spell-effects.db/spell-effect-organsight.json
@@ -39,6 +39,7 @@
                 "value": -2
             },
             {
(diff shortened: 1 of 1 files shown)


packs/data/spell-effects.db/spell-effect-organsight.json
```diff
@@ -39,6 +39,7 @@
                 "value": -2
             },
             {
+                "damageType": "precision",
```
stwlam, line 42:
```suggestion
                "category": "precision",
```

PR #5469, opened 2023-01-03
Fix blocking of selected roll mode from chat log

stwlam:


TMunsie (approved):

types/foundry/client/documents/chat-message.d.ts
```diff
@@ -203,7 +203,7 @@ declare global {
     }
 
     interface ChatMessageModificationContext extends DocumentModificationContext {
-        rollMode?: RollMode;
+        rollMode?: RollMode | "roll";
```
CarlosFdez, line 206: What is this roll mode but not a roll mode thing?

stwlam: "roll" isn't a roll mode but a cue to Foundry to get the roll mode from elsewhere (chat log select menu or core setting).

PR #5483, opened 2023-01-04
Fix bomb notes

MSAbaddon:


stwlam:
Those can be removed

MSAbaddon:
What do you mean? Are those no longer needed?

TikaelSol:
Persistent and splash damage will be handled by the system as part of the weapon damage. Not all bomb notes will go away as there may be other effects but most of them will vanish when we tackle them.

MSAbaddon:
Ok, understood. On the other hand, I guess it won't hurt to update the strings as long as they are still in use? Once they can get removed/reduced to what is actually neccessary you already have them up to date.

static/lang/re-en.json
```diff
@@ -67,117 +67,117 @@
...
-                    "success": "<strong>Effect</strong> The bomb also deals [[/r 2d6[persistent,acid]]] and [[/r (2[splash])[acid]]]{2 acid splash damage}",
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*2d6)[persistent,acid]]] and [[/r (2[splash])[acid]]]{2 acid splash damage}"
+                    "success": "<strong>Effect</strong> The bomb also deals [[/r 2d6[persistent,acid]]] and [[/r (2[splash])[acid]]]{2 acid splash damage}.",
+                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*2d6)[persistent,acid]]] and [[/r (2[splash])[acid]]]{2 acid splash damage}."
                 }
             },
             "AlchemistsFire": {
                 "Greater": {
-                    "success": "<strong>Effect</strong> The bomb also deals [[/r 3[persistent,fire]]] and [[/r (3[splash])[fire]]]{3 fire splash damage}",
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r 2*3[persistent,fire]]] and [[/r (3[splash])[fire]]]{3 fire splash damage}"
```
stwlam, line 89: This communicates that it's standard crit damage.

MSAbaddon: You mean the 2*3?
tbh, I'm not sure how to handle this correctly. 2*3[persistent,fire] creates an incorrect damage formula, (2*3)[persistent, fire] creates an incorrect button label:

image
image
image

[[/r (2*3)[persistent,fire]]]{2*3 persistent fire} would do the job, but that's a lot of custom labels.

stwlam: Don't worry about the labels; we'll take care of that: they just need parentheses

MSAbaddon: Ok, changed

PR #5509, opened 2023-01-05
Fix deadly variant of Clawsong spell effect

Far2Casual:
At the moment the rule element for the Deadly part adds a D8 dice all the time to the damage roll, instead of just adjusting the strike to give it Deadly (D8). This will fix it.
Before: 
image
After:
image

### DIFF
--- a/packs/data/spell-effects.db/spell-effect-clawsong-deadly-d8.json
+++ b/packs/data/spell-effects.db/spell-effect-clawsong-deadly-d8.json
@@ -18,15 +18,11 @@
         },
         "rules": [
             {
-                "critical": "true",
-                "damageType": "slashing",
-                "diceNumber": 1,
-                "dieSize": "d8",
-                "key": "DamageDice",
-                "selector": "strike-damage",
-                "traits": [
-                    "deadly-d8"
-                ]
+                "definition": ["item:slug:claw"],
+                "key": "AdjustStrike",
+                "mode": "add",
+                "property": "weapon-traits",
+                "value": "deadly-d8"
             }
         ],
         "source": {


TikaelSol:
We can (and should) consolidate these effects into one fully automated effect, and add the old effect to the list of effects needed to be migrated away when we finish consolidation.

Far2Casual:
@TikaelSol I agree. I could try and do the rule element part, but I'm unsure about the migration part. If you have a few pointers, I could try to do a new PR for it ? Or I could open a new PR with the consolidated effect, and leave the migration part ?

TikaelSol:
Leave the migration for now, either of these PRs is fine to just update to the consolidation. Then we can check off Clawsong from the tracking issue list https://github.com/foundryvtt/pf2e/issues/4144

Far2Casual:
So I've digged deeper into this one and #5512 and in both cases, I'm bumping into system limitations that make merging those effets harder than it looks. I'll comment on both PR to illustrate.

Making a RE like this does not work at the moment:
```{
    "definition":["item:slug:claw"],
    "key":"AdjustStrike",
    "mode":"add",
    "predicate": ["item:trait:versatile-p"],
    "property":"weapon-traits",
    "value":"deadly-d8"
}```

Despite the trait "versatile-p" being present in the roll options, the predicate just does not seem to work. 
![image](https://user-images.githubusercontent.com/63196064/211142467-671859b0-6fbc-49b5-b521-2b354c82b7b4.png)

It's specific to item options, as if the predicate is on anything else (like, class:sorcerer), then the RE is applied.
![image](https://user-images.githubusercontent.com/63196064/211142508-95c990a9-ec07-4ab7-b229-a524d7dd79cf.png)

Another problem is for the dice upgrade. It's supposed to go "up to D6 or up to D8 if it's already D6". If someone's a Lizard (with Claws) and the feat Razor Claws bumping the die to D6, casting the spell on him should allow him to raise to D8. This is done via a RE at the moment and works, but in the roll options, the dice given stays "D4" (which leads a weird display, see under). So again, I think I can't make a predicate looking if it's D6 or not, because it's not targetable.
![image](https://user-images.githubusercontent.com/63196064/211142706-0f4b8ab2-87b6-4052-8329-c83ccb7b36e0.png)


So unfortunately, that consolidation of effects goes beyond my competence 😬. I'll leave it up to you to decide if it's worth correcting the current effects to make them work while the system is updated to allow consolidation.

PR #5523, opened 2023-01-06
Create persistent damage conditions when applying damage

stwlam:


CarlosFdez:
Conflicts!

Ideally I'd like some way to alter / cancel the stuff, since if it turns out to be awkward in practice I'd like the option of preventing persistent damage application in a module to create a separate button (I'd give it a bit to decide if its worth it), but I'd need to dig into the pipeline anyways to come up with a good idea, and it'd likely be its own PR anyways once I both have a good idea and have sufficient motivation (best to give it a shot first imo).

PR #5529, opened 2023-01-06
Only increase weapon dice number from striking/ABP if originally 1

stwlam:
Striking Rune: "A striking rune [...], increasing the weapon damage dice it deals to two instead of one"
Devastating Attacks: "At 4th level, your weapon and unarmed Strikes deal two damage dice instead of one."


TMunsie:
This seems consistent with what is written. PM approved

PR #5543, opened 2023-01-07
Update language of legacy flaws option

stwlam:


CarlosFdez:
This works, but what's your opinion on keeping voluntary as is and adding a type field to ancestry.system.voluntary? For non-legacy, ancestry.system.voluntary.flaws can be the only thing in there, but legacy adds the boost field and caps the flaws to 2.

stwlam:
We can do that when we add non-legacy flaws.

CarlosFdez:
You could do that, though you'd have to change the buttons back to voluntary flaws since they'd be writing to the same location (only the boost and the max allowed would differ).

CarlosFdez (approved):

PR #5549, opened 2023-01-07
Delete only top-level class features when class is changed

CarlosFdez:


stwlam:
This will leave behind class features granted by class features

CarlosFdez:
Aren't those generally cleaned up by the grant item system? Class features often grant skill items and those are cleaned up.

From a rudimentary test on an alchemist and thaumaturge, it also cleaned up the nested class features. Is it worth adding a getter to feat to get the root of a grant chain?

PR #5550, opened 2023-01-07
Move damage colors to global SCSS variables

In3luki:


src/styles/_colors.scss
```diff
@@ -29,6 +29,26 @@ $tertiary-light-color: lighten($tertiary-color, 10) !default;
...
+$damage-force: #6300aa;
+$damage-good: #9d730a;
+$damage-lawful: #402600;
+$damage-mental: midnightblue;
+$damage-negative: #00001f;
+$damage-piercing: $damage-physical;
+$damage-poison: #5b7332;
+$damage-positive: #ffffe0;
+$damage-slashing: $damage-physical;
+$damage-sonic: darkcyan;
```
stwlam, line 51: Make them CSS variables if you're going to do that

In3luki: I've added CSS variables based on the SCSS variables. Apparently functions like `color.adjust` don't work with CSS variables.

stwlam: Ohhh, right, right.

stwlam: Ehh, let's nix that, then, sorry. All it'll do if anyone tries to override them is not work for some reason.

src/styles/_colors.scss
```diff
@@ -105,4 +125,25 @@ $adjusted-lower: #cc3311 !default;
...
+  --damage-good: $damage-good;
+  --damage-lawful: $damage-lawful;
+  --damage-mental: $damage-mental;
+  --damage-negative: $damage-negative;
+  --damage-physical: $damage-physical;
+  --damage-piercing: $damage-piercing;
+  --damage-poison: $damage-posion;
+  --damage-positive: $damage-positive;
+  --damage-slashing: $damage-slashing;
+  --damage-sonic: $damage-sonic;
```
stwlam, line 148:
```suggestion
```

PR #5559, opened 2023-01-07
Fixed the Daemonic Infector "Cleric Domain Spells" spellcasting entry from innate to focus

chepseskhaf:


stwlam:
These need to be recreated: the data is stored differently between focus spells and innate spells.

chepseskhaf:
I juste recreated the spellcasting entry in Foundry and got the same thing. I only reordered the properties to match the old data.

PR #5572, opened 2023-01-08
Restore the variant Devourer and Bog Mummy from AV

silvative:


silvative:
Fingers crossed on this one...

PR #5608, opened 2023-01-09
Add predicate to disrupting rune effect

stwlam:


src/module/item/physical/runes.ts
```diff
@@ -528,6 +529,7 @@ export const WEAPON_PROPERTY_RUNES: Record<WeaponPropertyRuneType, WeaponPropert
                     damageType: "positive",
                     diceNumber: 2,
                     dieSize: "d6",
+                    predicate: [{ or: ["target:mode:undead", { not: "target" }] }],
```
Far2Casual, line 532: could we consider the same for the brilliant rune (which also does specifically 1d4 damage vs undead) ?

Issue #5439, opened 2023-01-02
Double Number of Dice on Critical Hit Rule Doesn't Function 


HarryCuddletrousers:
![Screenshot 2023-01-02 154144](https://user-images.githubusercontent.com/40306844/210279987-1be2b621-9757-48a9-8d6d-c770fb330ddb.png)
![Screenshot 2023-01-02 154255](https://user-images.githubusercontent.com/40306844/210279992-abbf29d4-ebcb-46c1-bedc-3559eb1601be.png)

jdwpom:
It probably doesn't need to be said, but for the sake of completeness, this issue arose with v4.6.0. "Double the Dice" works fine on all versions of 4.5.

PR #5610, opened 2023-01-09
Fix 'double number of dice' option when rolling critical damage rolls

cdharma3:
Proposed fix for issue #5439. 

Modifies `formula.ts` to take into account the critical damage option. Notably modifies the `createDamageFormula` function to double the base damage dice and modifier, and the damage dice and modifiers for any extra rolls that aren't persistent or splash damage.

Also modifies the `sumExpression` function to not prepend the multiplication if the number of damage dice are already being doubled.


stwlam:
The challenge of reimplementing dice doubling isn't the the doubling itself but maintaining a separate group of dice that can be accurately removed upon damaging a target immune to critical hits.

cdharma3:
Oh gotcha! I can take a look at that as well, unless there's a fix already in the works?

stwlam:
It's definitely going to get fixed. I or someone else just needs to dive into it.

PR #5630, opened 2023-01-10
Show icon for precision damage

stwlam:
Screenshot from 2023-01-10 03-10-50


CarlosFdez:
Seems this also has some bugfixes and UI niceties as well, like skipping unnecessary () and something regarding better crit display of crits.

PR #5693, opened 2023-01-14
Add more specificity to the CSS selectors for Critical Hit/Fumble Journals

JamzTheMan:
More specificity to the CSS selectors added so the more generic CRB Journal Styles in Dorako UI do not override the look and feel of Critical Hit/Fumble cards.

Also allows the System to specify CSS for Journals without overriding the these specific types of journal entries.

Before:
image

After:
image

### DIFF
--- a/src/styles/system/journal/_index.scss
+++ b/src/styles/system/journal/_index.scss
@@ -1,5 +1,6 @@
-@import "critical-fumble-deck";
-
-.journal-page-content {
-    @include tinymce-system-specific;
+.journal-entry-page {
+    .journal-page-content {
+        @include tinymce-system-specific;
+        @import "critical-fumble-deck";
+    }
 }
--- a/src/styles/system/journal/critical-fumble-deck.scss
+++ b/src/styles/system/journal/critical-fumble-deck.scss
@@ -1,115 +1,152 @@
-.fumble-deck {
-    h1 {
-        border: 2px solid;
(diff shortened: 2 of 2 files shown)


stwlam:
Why would we alter CSS to escape a module's restyling instead of requesting that the module simply not break it?

CarlosFdez:
To be fair, the current styling is too unspecific, kinda to the point of absurdity. Would it work to instead change journal/_index.scss to import the critical fumble deck stuff from inside of .journal-page-content?

stwlam:
Sure, that's a more concise fix.

JamzTheMan:
To answer the why first, @CarlosFdez already basically covered it. It was less to "fix" a module issue and more to "allow" modules to override journal styling, but also in the future, let the system also add journal styling without clobbering the cards. The style of the cards will probably never match the CRB type styling and need to stand on their own so adding specificity helps here.

For the solution, I've never messed with scss before, but this seems to work? I needed both levels .journal-entry-page && .journal-page-content specified, so I just moved the content block under the new page block. It doesn't seem to affect the tinymce css (which also seems deprecated?)

Is this ok or is there a better way?

```scss
.journal-entry-page {
    .journal-page-content {
        @include tinymce-system-specific;
        @import "critical-fumble-deck";
    }
}
```

stwlam:
Yeah, that works.

PR #5699, opened 2023-01-14
Restore crit dice doubling

stwlam:
Maybe delay merge?

### DIFF
--- a/src/module/system/damage/roll.ts
+++ b/src/module/system/damage/roll.ts
@@ -3,7 +3,7 @@ import { DamageRollFlag } from "@module/chat-message";
 import { UserPF2e } from "@module/user";
 import { DegreeOfSuccessIndex } from "@system/degree-of-success";
 import { RollDataPF2e } from "@system/rolls";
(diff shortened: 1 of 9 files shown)


src/module/system/damage/roll.ts
```diff
@@ -492,11 +492,24 @@ class DamageInstance extends AbstractDamageRoll {
...
 
         const { head } = this;
-        return head instanceof ArithmeticExpression || head instanceof Grouping ? head.critImmuneTotal : this.total;
+
+        // Get the total with all damage-doubling removed
+        const undoubledTotal =
+            head instanceof ArithmeticExpression || head instanceof Grouping ? head.critImmuneTotal : this.total;
+
+        if (game.settings.get("pf2e", "critRule") === "doubledamage") {
+            return undoubledTotal;
```
CarlosFdez, line 505: it seems you check for the crit rule twice, once when building the total, and again here in the damage roll. Since it being double dice seems to be baked into the formula itself, perhaps this is something that should be set in roll options.

PR #5700, opened 2023-01-14
Fix `maxPersistentAfterIWR` for bleed damage

stwlam:


CarlosFdez:
You did warn me of my methodology before, saying I was lucky that there was no monster that has specific IWR vs persistent damage yet. Only fair we incorporate such things across the board :P

PR #5714, opened 2023-01-15
Add damage dice support to spells

CarlosFdez:
image
image
image

### DIFF
--- a/src/module/actor/modifiers.ts
+++ b/src/module/actor/modifiers.ts
@@ -642,8 +642,8 @@ class DiceModifierPF2e implements BaseRawModifier {
             : null;
 
         this.predicate = new PredicatePF2e(params.predicate ?? []);
-        this.enabled = this.predicate.test([]);
-        this.ignored = !this.enabled;
+        this.enabled = params.enabled ?? this.predicate.test([]);
+        this.ignored = params.ignored ?? !this.enabled;
     }
 }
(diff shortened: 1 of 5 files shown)


CarlosFdez:
I removed sneak attack from the PR. If this requires a clean magical trickster implementation, then we will need to refactor sneak attack, and that'll require a migration. Aka performed separately.

src/module/actor/modifiers.ts
```diff
@@ -642,8 +642,8 @@ class DiceModifierPF2e implements BaseRawModifier {
             : null;
 
         this.predicate = new PredicatePF2e(params.predicate ?? []);
-        this.enabled = this.predicate.test([]);
-        this.ignored = !this.enabled;
+        this.enabled = params.enabled ?? this.predicate.test([]);
+        this.ignored = params.ignored ?? !this.enabled;
```
stwlam, line 646: How about we just remove the empty predicate test?

CarlosFdez: we could replace it for testing if the predicate is empty instead. I think the intent is to say "if its not tested and it needs to be, default to disabled". Might break something if we don't have anything at all here.

stwlam: They get tested in @system/damage/weapon, so it's covered there.

Issue #40, opened 2019-10-14
Output skill proficiency of player when rolling a skill


cswendrowski:
In GitLab by @Peenicks on Oct 14, 2019, 07:01

This would mostly be useful for traps when detecting them and disabling them, having a reference for their proficiency.

Issue #22, opened 2019-08-29
Configure static weapon damage


cswendrowski:
In GitLab by @overdox on Aug 29, 2019, 18:25

Blowgun does a static 1P damage, currently there is no way configure that for said weapon.

Issue #21, opened 2019-08-29
proficiency with single item types


cswendrowski:
In GitLab by @overdox on Aug 29, 2019, 18:20

Currently there is no way to set proficiency to specific items, gained from feats or in the case of clerics, from their deities favored weapon.

Issue #20, opened 2019-08-28
Feat actions - missing option


cswendrowski:
In GitLab by @overdox on Aug 28, 2019, 06:35

Currently you can only select "action" on feats, but as some feats can take 2 or 3 actions, we need a way to enable a feat to take any number of actions.

PR #5734, opened 2023-01-15
Add effect with an RE for "You are Flat-Footed until end of your …

xdy:
…next turn.", used in the Critical Fumble Deck cards #19, #20, #21, #22 and #40.

### DIFF
--- a/packs/data/criticaldeck.db/critical-fumble-deck-19.json
+++ b/packs/data/criticaldeck.db/critical-fumble-deck-19.json
@@ -13,7 +13,7 @@
             "src": null,
             "system": {},
             "text": {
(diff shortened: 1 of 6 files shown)


TikaelSol:
Don't know what I think about granting conditions.

xdy:
Me neither. I saw it was done in a few other places, and it works, so I figured I'll send in a PR.
Fwiw, removing flatfooted due to this sort of thing is something I always forget, so at least for me it'd be useful.

xdy:
Though, if this is acceptable the effect should probably have this too:
```
                "onDeleteActions": {
                    "grantee": "restrict"
                },
```

xdy:
Added onDeleteActions as per above.

PR #5736, opened 2023-01-15
Have Aquatic Combat effect grant Flat-Footed condition 

stwlam:


TikaelSol:
You are only flat footed if you don't have a swim speed.

PR #5762, opened 2023-01-16
Update scene navigation, scene controls, and app UI

Dorako:
Notably no changes* to Hotbar, Sidebar, Chat Messages, Placeables HUD, Effect Panel in this PR.
(*Slight changes due to global style overwrites)

There's been quite a few eyes on these changes, since they're used as-is in recent version of Dorako UI, and I've gotten no reports of these changes breaking anything. 

Pictures of premium modules not breaking:
CleanShot 2023-01-16 at 20 02 21@2x
CleanShot 2023-01-16 at 20 12 51@2x
CleanShot 2023-01-16 at 20 17 27@2x

### DIFF
--- a/src/styles/_colors.scss
+++ b/src/styles/_colors.scss
@@ -1,4 +1,4 @@
-@use 'sass:color';
+@use "sass:color";
 
 /* ----------------------------------------- */
 /* Color definitions                         */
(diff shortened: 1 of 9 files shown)


src/styles/_colors.scss
```diff
@@ -1,3 +1,5 @@
+@use 'sass:color';
```
stwlam, line 1:
```suggestion
@use "sass:color";
```

PR #5776, opened 2023-01-17
Always add class DCs for PCs, refine appearance

stwlam:
This removes the exclusion of untrained class DCs when populating `system.proficiencies.classDCs`. A character with only a single untrained class DC won't have it appear on their sheet, but it will be present and available for the occasional mechanical necessity. Class DC options when creating spellcasting entries will only include trained (or better) class DCs, and will appear as "Class DC ([class name])"


stwlam:
Ezren:
image

Ezren comes down with kleptomania:
image

PR #5781, opened 2023-01-17
Revert "Update scene navigation, scene controls, and app UI"

stwlam:
Reverts foundryvtt/pf2e#5762


CarlosFdez:
Since this will come up, unfortunately it did reduce contrast by more than we would have liked, and changes to core application windows like the scene switcher and settings configuration would not be taken well by our foundry overlords.

Changes to the side buttons/macros/playerlist, and sidebar are alright, but they need to be strict improvements. A hard to read actor list on some maps, especially for people with poor vision, is a regression.

That said, I personally did enjoy replacing the thick natural foundry border for the drop shadow, and the higher border radius. Though internally someone did raise a good point that changes to the UI should make it more pathfinder, and not just be for the sake of it.

Thank you still for the original PR though.

PR #5782, opened 2023-01-17
Reinstate red selection background for canvas controls from Dorako PR

CarlosFdez:
image


stwlam:
BUT WHAT ABOUT THE :blue_square:

CarlosFdez:
Good eye, 🟦 reinstated

CarlosFdez:
Dorako's original version had this issue, but I think untoggled control (dark blue) and selected control (red) look way too similar to each other. I think it might be handy to have the untoggled button's background color be marginally transparent (only being fully opaque when selected).

image

Let me know if this is preferred and I'll change it right up. Otherwise feel free to merge as is.

PR #5788, opened 2023-01-17
Simplify inline bleed damage rolls

TikaelSol:
Since the other style is still functional I don't think we need to migrate. This just shaves a bit off of the compendium size.


stwlam:
Whatever helps the busybodies stop fixating on this. :stuck_out_tongue:

PR #5797, opened 2023-01-18
Update heal companion spell

MSAbaddon:


TikaelSol:
So this is just removing the description from the variants?

MSAbaddon:
Yes. According to Shark it shouldn't be possible to add different descriptions to spell variants. The description within the variants are out of date.

PR #5807, opened 2023-01-19
Allow for decimal scales in TokenImage

LebombJames:
Currently `scale` only takes integers, which can be hard to set up properly for tokens with pop-out elements.

### DIFF
--- a/src/module/rules/rule-element/token-image.ts
+++ b/src/module/rules/rule-element/token-image.ts
@@ -21,7 +21,7 @@ export class TokenImageRuleElement extends RuleElementPF2e {
             this.value = null;
         }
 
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/token-image.ts
```diff
@@ -21,7 +21,7 @@ export class TokenImageRuleElement extends RuleElementPF2e {
             this.value = null;
         }
 
-        if (typeof data.scale === "number" && Number.isInteger(data.scale)) {
+        if (typeof data.scale === "number" && Number(data.scale)) {
```
stwlam, line 24:
```suggestion
        if (typeof data.scale === "number" && data.scale > 0) {
```

PR #5816, opened 2023-01-19
Fix the success effect of lift nature's caul

rectulo:


rectulo:
thanks to Dalvyn for this catch

TikaelSol:
image

This was errata's in the hardcover version. Sickened is too powerful for a success effect so it makes sense.

PR #5821, opened 2023-01-20
Keep tabs from shifting in item sheet when changing to rules tab

CarlosFdez:
image

image


CarlosFdez:
Another part of me is considering removing the X Summary header entirely regardless of whether there's a sidebar or not.

PR #5826, opened 2023-01-21
Fix `item:damage-dice:N` roll options being set according to the number of faces rather than dice

dmarcuse:
There seems to have been a minor copy/paste error, causing this roll option to be set for the number of faces of the damage die rather than the number of dice. This causes an issue with the rule elements in the Student of the Staff feat, causing the arcane cascade critical damage to be rolled incorrectly.


stwlam:
Working as intended

dmarcuse:
Are you sure? This caused a regression for the pf2e system rule elements for Student of the Staff, and is different from the previous behavior which counted the number of damage dice - so this could be breaking things subtly for other modules/user-created rule elements as well

TikaelSol:
This is definitely wrong

image

src/module/item/weapon/document.ts
```diff
@@ -185,9 +185,9 @@ class WeaponPF2e extends PhysicalItemPF2e {
                 [`reload:${this.reload}`]: !!this.reload,
                 [`damage:type:${damage.type}`]: true,
                 [`damage:category:${damage.category}`]: !!damage.category,
-                [`damage:die:number:${damage.dice.number}`]: !!damage.dice.faces,
+                [`damage:die:number:${damage.dice.number}`]: !!damage.dice.number,
```
stwlam, line 188:
```suggestion
                [`damage:die:number:${damage.dice.number}`]: !!damage.dice.faces,
```

src/module/item/weapon/document.ts
```diff
@@ -185,9 +185,9 @@ class WeaponPF2e extends PhysicalItemPF2e {
...
                 [`damage:die:faces:${damage.dice.faces}`]: !!damage.dice.faces,
-                [`damage-dice:${damage.dice.faces}`]: !!damage.dice.faces,
+                [`damage-dice:${damage.dice.number}`]: !!damage.dice.number,
```
stwlam, line 190:
```suggestion
                [`damage-dice:${damage.dice.number}`]: !!damage.dice.faces,
```

dmarcuse:
Thank you!

PR #5846, opened 2023-01-21
Include value of "0" in `getActionIcon` helper

stwlam:


CarlosFdez:
I think I vaguely remember a situation where this actually doesn't do what we want, but I can't recall it exactly, so I'm down to be reminded by new bug reports if that is indeed the case.

PR #5848, opened 2023-01-21
Debounce document resets and rerenders on settings changes

stwlam:


CarlosFdez:
It might be good for to make these functions add to a set or map (keyed by id) in memory and then call a flush() function that is debounced. Debounced by its nature, if interrupted, will actually cancel the previous function call. Therefore if you try to render A/B/C and then call it again with B,D before the first timeout ends, it will not refresh A nor C.

PR #5852, opened 2023-01-22
Add a dialog to modifiy damage roll modifiers

In3luki:
It's only enabled in development  for now and can be opened by holding shift while pressing a damage button.
The UI could still use some refinement but it's looking ok in my opinion.

Weapon Damage:

Weapon_Normal

Weapon_Critical

Weapon_Sneak

Weapon_Deadly

Splash_Persistent

Spell Damage:

Spell_Damage

Spell_Damage2

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -104,6 +104,7 @@ import { CharacterFeats } from "./feats";
 import { createForceOpenPenalty, createShoddyPenalty, StrikeWeaponTraits } from "./helpers";
 import { CharacterHitPointsSummary, CharacterSkills, CreateAuxiliaryParams, DexterityModifierCapData } from "./types";
 import { CHARACTER_SHEET_TABS } from "./values";
+import { eventToRollParams } from "@scripts/sheet-util";
 
 class CharacterPF2e extends CreaturePF2e {
     /** Core singular embeds for PCs */
@@ -1815,32 +1816,37 @@ class CharacterPF2e extends CreaturePF2e {
                     return "";
                 }
 
-                const damage = WeaponDamagePF2e.calculate({
+                const outcome = method === "damage" ? "success" : "criticalSuccess";
+                const { self, target, options } = context;
+                const damageContext: DamageRollContext = {
+                    type: "damage-roll",
+                    sourceType: "attack",
+                    self,
+                    target,
+                    outcome,
+                    options,
+                    domains,
+                    ...eventToRollParams(params.event),
+                };
+
+                if (params.getFormula) {
+                    damageContext.skipDialog = true;
+                }
+
+                const damage = await WeaponDamagePF2e.calculate({
                     weapon: context.self.item,
                     actor: context.self.actor,
                     actionTraits: context.traits,
                     proficiencyRank,
-                    options: context.options,
                     weaponPotency,
+                    context: damageContext,
                 });
                 if (!damage) return null;
 
-                const outcome = method === "damage" ? "success" : "criticalSuccess";
                 if (params.getFormula) {
                     const formula = damage.damage.formula[outcome];
                     return formula ? new DamageRoll(formula).formula : "";
                 } else {
-                    const { self, target, options } = context;
-                    const damageContext: DamageRollContext = {
-                        type: "damage-roll",
-                        sourceType: "attack",
-                        self,
-                        target,
-                        outcome,
-                        options,
-                        domains: damage.domains,
-                    };
-
                     return DamagePF2e.roll(damage, damageContext, params.callback);
                 }
             };
--- a/src/module/actor/helpers.ts
+++ b/src/module/actor/helpers.ts
@@ -12,7 +12,7 @@ import {
 } from "@module/rules/helpers";
 import { TokenDocumentPF2e } from "@scene";
 import { CheckPF2e, CheckRoll } from "@system/check";
-import { DamagePF2e, DamageType, WeaponDamagePF2e } from "@system/damage";
+import { DamagePF2e, DamageRollContext, DamageType, WeaponDamagePF2e } from "@system/damage";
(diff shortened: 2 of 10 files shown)


In3luki:
Updated to support added spell damage dice.

CarlosFdez:
Bothered me at first that spell damage became async but I suppose character strikes work like that too huh?

That said, check modifiers dialog is displayed in CheckPF2e, but this is displayed when computing formulas. That bothers me a little bit but I can imagine why it'd be a major pain in the rear to convert all that, especially considering how inconsistent instance assignment is.

CarlosFdez:
The label seems to have superflous text. Not sure if you need +2d6 precision if another column already lists it.

In3luki:
> Bothered me at first that spell damage became async but I suppose character strikes work like that too huh?
> 
> That said, check modifiers dialog is displayed in CheckPF2e, but this is displayed when computing formulas. That bothers me a little bit but I can imagine why it'd be a major pain in the rear to convert all that, especially considering how inconsistent instance assignment is.

I wanted to keep it a getter but there is no other way to intercept the modifiers before the forumla is created.

> The label seems to have superflous text. Not sure if you need +2d6 precision if another column already lists it.

The labels are created somewhere else. I can look into stripping the damage dice from it if you want.

CarlosFdez:
Ah, the labels are a symptom of our awkward breakdown creation pipeline. Don't worry about it then.

As for making spell damage a method, we need to do that anyways for outcomes.

In3luki:
Now uses `eventToRollParams` which makes it adhere to the check roll show dialog setting.

src/module/actor/character/document.ts
```diff
@@ -1815,32 +1816,37 @@ class CharacterPF2e extends CreaturePF2e {
...
+                    type: "damage-roll",
+                    sourceType: "attack",
+                    self,
+                    target,
+                    outcome,
+                    options,
+                    domains,
+                };
+
+                const rollParams = params.event ? eventToRollParams(params.event) : {};
```
CarlosFdez, line 1831: Would there be any consequences here if eventToRollParams() allowed event to be optional, and purely used the settings in that situation?

In3luki: Probably not. I'll check it out.

CarlosFdez: I could imagine you could ... it into the damage context, then you'd only need to tweak skipDialog specifically.

CarlosFdez:
Would it be cleaner to create the DamageRollContext in spell's getFormula() then return it instead of returning modifiers/damage dice/etc? Those returns exist to create the damage context in rollDamage().

In3luki:
Yeah, spell damage preparation is a bit all over the place. I can take another look at it.

src/module/system/damage/weapon.ts
```diff
@@ -27,16 +27,24 @@ import { DegreeOfSuccessIndex, DEGREE_OF_SUCCESS } from "@system/degree-of-succe
...
+    static async calculateStrikeNPC(
         attack: MeleePF2e,
         actor: NPCPF2e | HazardPF2e,
         actionTraits: TraitViewData[] = [],
         proficiencyRank = 0,
-        options: Set<string> = new Set()
-    ): WeaponDamageTemplate | null {
+        options: Set<string> = new Set(),
+        context: DamageRollContext
+    ): Promise<WeaponDamageTemplate | null> {
```
CarlosFdez, line 47: If you're adding new params, might as well convert it to an object like WeaponDamagePF2e#calculate(). That said, isn't options part of context?

In3luki: I'm not sure what the intention was there anymore. The options are indeed part of the context.

In3luki:
- Made `eventToRollParams` event parameter optional.
- Moved `DamageRollContext` and `SpellDamageTemplate` creation to `SpellPF2e#getDamage`.
- Changed `WeaponDamagePF2e.calculateStrikeNPC` params to an object.

src/module/item/spell/document.ts
```diff
@@ -826,34 +867,13 @@ class SpellPF2e extends ItemPF2e {
             if (variant) return variant.rollDamage(event);
         }
 
-        const spellDamage = this.damage;
+        const rollParams = eventToRollParams(event);
+        const spellDamage = await this.getDamage({ ...rollParams });
```
CarlosFdez, line 871:
```suggestion
        const spellDamage = await this.getDamage(eventToRollParams(event));
```

CarlosFdez:
this: Embedded<SpellPF2e> is working fine? I remember removing those because it caused nonesense with subclassing.

src/module/actor/helpers.ts
```diff
@@ -337,22 +338,31 @@ function strikeFromMeleeItem(item: Embedded<MeleePF2e>): NPCStrike {
...
+                type: "damage-roll",
+                sourceType: "attack",
+                self,
+                target,
+                outcome,
+                options,
+                domains,
+            };
+
+            const rollParams = params.event ? eventToRollParams(params.event) : {};
```
CarlosFdez, line 352: Missed a spot :^)

src/module/item/spell/document.ts
```diff
@@ -324,6 +349,8 @@ class SpellPF2e extends ItemPF2e {
             // Add damage dice to instances
             for (const dice of damageDice) {
                 if (dice.override) continue;
+                // Skip dice that were ignored in the dialog
+                if (dice.ignored) continue;
```
CarlosFdez, line 353:
```suggestion
                if (dice.override || dice.ignored) continue;
```

CarlosFdez: Well, this is likely to change anyways when I move all of this to the damage pipeline. Don't worry about it.

In3luki:
Updated

PR #5863, opened 2023-01-23
Debounce aura-detection methods

stwlam:


src/module/actor/base.ts
```diff
@@ -107,6 +107,14 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
             }
 
             super(data, context);
+
+            // Add debounced checkAreaEffects method
+            Object.defineProperty(this, "checkAreaEffects", {
```
CarlosFdez, line 112: It might be a bit odd, but maybe you can do `checkAreaEffects = foundry.utils.debounce(checkAreaEffects.bind(this), 50)` at the instance level?

stwlam: That's effectively what this does, except it doesn't appear in `Object.keys`

PR #5866, opened 2023-01-23
Fix AdjustStrike trait upgrade

### LINKED ISSUE #5775: Mark for Death feat's Toggle is not upgrading deadly dice for qualifying weapons
My player is using a war razor (deadly d6) so when the Mark for Death Toggle is set, it should upgrade to deadly d8, and its not.  It still rolls the d6.  I'm also not seeing the backstabber trait show up anywhere, but that may or may not be a separate issue.

JDCalvert:
In the recent update to allow `remove` and `subtract` weapon traits using the `AdjustStrike` rule element, an issue was introduced where, if a scorable trait was added (e.g. fatal-d12) and there was already a trait with the same name (e.g. fatal-d10) on the weapon, then the existing trait would be removed but the new trait would not be added.

Fixes #5775.

### DIFF
--- a/src/module/rules/rule-element/adjust-strike.ts
+++ b/src/module/rules/rule-element/adjust-strike.ts
@@ -159,6 +159,7 @@ class AdjustStrikeRuleElement extends AELikeRuleElement {
                                     if (changeTraitScore > existingTraitScore) {
                                         traits.findSplice((trait) => trait === existingTraitMatch[0], change);
                                     }
+
                                     return;
                                 }
                             }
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/adjust-strike.ts
```diff
@@ -158,9 +158,9 @@ class AdjustStrikeRuleElement extends AELikeRuleElement {
                                     // (lesser) trait
                                     if (changeTraitScore > existingTraitScore) {
                                         traits.findSplice((trait) => trait === existingTraitMatch[0]);
+                                    } else {
+                                        return;
                                     }
```
stwlam, line 163:
```suggestion
                                        traits.splice(traits.indexOf(existingTraitMatch[0]), 1, change);
                                    }

                                return;
```

stwlam: That works, thanks!

PR #5870, opened 2023-01-23
Defer to core.rollMode setting when assigning most roll modes

stwlam:


CarlosFdez:
Since when did "roll" grab the core rollmode? That's neat, wonder how I missed that one.

PR #5881, opened 2023-01-24
Autoroll affliction damage on stage change

CarlosFdez:
Also refactors spell damage construction (though I'm still iffy on whether build should throw or warn and return null) and fixes some affliction issues.

Once we decide what the selectors should be, we can add retrieving modifiers and what-not.

### DIFF
--- a/src/module/item/affliction/data.ts
+++ b/src/module/item/affliction/data.ts
@@ -41,9 +41,9 @@ interface AfflictionOnset {
 }
 
 interface AfflictionDamage {
-    value: string;
+    formula: string;
     type: DamageType;
-    category?: DamageCategoryUnique;
+    category?: DamageCategoryUnique | null;
 }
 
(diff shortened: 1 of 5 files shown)


CarlosFdez:
This should be fine to go now.

src/module/item/affliction/data.ts
```diff
@@ -42,7 +42,7 @@ interface AfflictionOnset {
 
 interface AfflictionDamage {
     value: string;
-    type: DamageType;
+    damageType: DamageType;
```
stwlam, line 45: I liked `type` :(

CarlosFdez: I agree, but everything from rule elements to melee damage uses damageType :(

CarlosFdez: Converted back to type, we'll migrate everything else later.

PR #5889, opened 2023-01-24
Add slugs to attack effects for Extinction Curse 

Far2Casual:
Follow up to #5861


TikaelSol:
These slugs should probably be added automatically on extract. Wonder why so many are missing, they actually have an impact on automation outside of localization.

PR #5890, opened 2023-01-24
Add slugs to attack effects for Agents of Edgewatch

Far2Casual:
Follow-up to #5861


TikaelSol:
Renamed to fit with project standards. We can probably just lump any others into one PR.

PR #5894, opened 2023-01-25
Complete Migration 816

velara:
run-migration.ts was missed when adding the migration information for Migration 816. Sorry. 

Updated run-migration.ts and re-ran npm run migrate. Pushed changes to one NPC from the Kingmaker Bestiary


velara:
Files I explicitly tried to avoid committing somehow ended up in the push. Closed.

PR #5904, opened 2023-01-25
Have `RuleElementPF2e` extend `DataModel`

stwlam:
This includes both `RuleElementPF2e` getting a `DataModel` parent as well as initial work on writing schema definitions for subclasses. It makes use of a `LaxSchemaField` class that doesn't prune unrecognized fields so that the remaining conversion work can be done over time.


CarlosFdez:
Conflicts :(

PR #5919, opened 2023-01-26
Review EC book 6 NPCs (part 2 of 6)

roxim5:


packs/data/extinction-curse-bestiary.db/convergent-kendley-nathrael.json
```diff
@@ -1504,16 +1593,24 @@
                         "key": "RollOption",
                         "label": "An ally with Convergent Tactics is within reach of the target",
```
SpartanCPA, line 1594: We should @Localize this

SpartanCPA (approved):

PR #5929, opened 2023-01-26
Cache the first `TokenAura.squares` getter result instead of recreating it each time it is called

### LINKED ISSUE #5905: NPCs with large Aura areas cause lag during PC movement
NPCs with larger aura rule elements enabled (dragons Frightful Presence) cause lag when moving PC tokens on a scene with multiple walls and multiple actors. 

Doing any of the following reduces/eliminates the lag:
- Removing Walls
- Removing Actors (except for 1 PC and the Dragon)
- Reducing the area of the Aura RE 

Exported Scene:
[fvtt-Scene-fort-solstice-gates (1).zip](https://github.com/foundryvtt/pf2e/files/10502976/fvtt-Scene-fort-solstice-gates.1.zip)

Video:

https://user-images.githubusercontent.com/89946334/214670613-a7de5252-226f-4346-afe4-ee7e1cf7d481.mp4

In3luki:
I'm not 100% sure this has no negative side effects but it seems to be working fine.

Closes #5905

### DIFF
--- a/src/module/scene/token-document/aura/index.ts
+++ b/src/module/scene/token-document/aura/index.ts
@@ -67,8 +67,7 @@ class TokenAura implements TokenAuraData {
 
     /** The squares covered by this aura */
     get squares(): EffectAreaSquare[] {
-        this.#squares ??= getAreaSquares(this);
-        return this.#squares;
+        return (this.#squares ??= getAreaSquares(this));
     }
 
     /** Does this aura overlap with (at least part of) a token? */
--- a/src/module/scene/token-document/aura/index.ts
+++ b/src/module/scene/token-document/aura/index.ts
@@ -23,6 +23,8 @@ class TokenAura implements TokenAuraData {
     /** Does this aura affect its emanating token? */
     private includesSelf: boolean;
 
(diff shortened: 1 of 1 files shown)


In3luki:
This is major performance improvement because the getter calls `getAreaSquares` which in turn calls `canvas.walls.checkCollision` for each aura grid square. Multiplied by the number of tokens in the scene that can lead to thousands or even tens of thousands of collision checks.

src/module/scene/token-document/aura/index.ts
```diff
@@ -65,7 +67,8 @@ class TokenAura implements TokenAuraData {
 
     /** The squares covered by this aura */
     get squares(): EffectAreaSquare[] {
-        return getAreaSquares(this);
+        this.#squares ??= getAreaSquares(this);
+        return this.#squares;
```
stwlam, line 71:
```suggestion
        return (this.#squares ??= getAreaSquares(this));
```

stwlam:
I _think_ this should work: adding a wall to a scene would trigger a data reset for the enter scene, including embedded tokens.

PR #5934, opened 2023-01-26
Add item level display to item directory

LebombJames:
To bring the item sidebar in line with the actor one, and to improve UX to see item levels at a glance, particularly for creating homebrew.
image

### DIFF
--- a/src/module/apps/sidebar/index.ts
+++ b/src/module/apps/sidebar/index.ts
@@ -1,4 +1,5 @@
 export { ActorDirectoryPF2e } from "./actor-directory";
+export { ItemDirectoryPF2e } from "./item-directory";
 export { ChatLogPF2e } from "./chat-log";
 export { CompendiumDirectoryPF2e } from "./compendium-directory";
 export { EncounterTrackerPF2e } from "./encounter-tracker";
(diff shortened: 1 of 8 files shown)


stwlam:
I think an easy gain would be to use the same term as on the sheet for each item type: a consumable isn't "Level 0" but "Item 0," for example.

CarlosFdez:
I feel like trying to centralize how the item sheet and this does it somehow might be more prudent. The sheet sets itemType in getData() but perhaps there's nothing wrong with setting CONFIG.PF2E.Item.typeHeader for every single item type to define what it maps to (with many of them being Item). Then the sheets can pull from that.

CarlosFdez:
Actually, right after sending it, I just remembered that feat items show Level for features, and Feat for feats, and spells show Cantrip/Focus/Spell depending on what type of spell it is. What I suggested won't work, but also highlights why its going to be important to centralize that to avoid such mistakes.

stwlam:
We can just merge this and then look at doing it for both actors and items

stwlam:
oh wait it's already done ... ehh we fix as we find stuff

LebombJames:
Yeah I thought making a centralised function would be better to unify both the header and the sidebar, but didn't wanna rock the boat too much (plus I wasn't exactly sure where to do so)

PR #5949, opened 2023-01-27
Add Silence spell effect

TikaelSol:


packs/data/spells.db/silence.json
```diff
@@ -23,7 +23,7 @@
...
+            "value": "<p>The target makes no sound, preventing creatures from noticing it using hearing alone. The target can't use sonic attacks, nor can it use actions with the auditory trait. This prevents it from casting spells that include verbal components.</p>\n<p>@UUID[Compendium.pf2e.spell-effects.Spell Effect: Silence]{Spell Effect: Silence}</p>\n<hr />\n<p><strong>Heightened (4th)</strong> The spell effect emanates from the touched creature, silencing all sound in or passing through a 10- foot radius and preventing any auditory and sonic effects in the affected area. While within the radius, creatures are subject to the same effects as the target. Depending upon the position of the effect, a creature might notice the lack of sound reaching it (blocking off the noise coming from a party, for example).</p>"
```
SpartanCPA, line 26: There should be a fourth level aura effect that grants the original effect as well.

TikaelSol: I thought that, but this is one of those 'no aura trait' things. Paizo may need to clarify this.

SpartanCPA: It probably doesn't have the aura trait because the base casting wouldn't have one. But I can see what you mean.

SpartanCPA (approved):

PR #5961, opened 2023-01-27
Add support for condition/effect alterations from GrantItem REs

stwlam:


src/module/rules/rule-element/grant-item/rule-element.ts
```diff
@@ -193,6 +206,52 @@ class GrantItemRuleElement extends RuleElementPF2e {
...
+        if (grantedSource.type !== "condition" && grantedSource.type !== "effect") {
+            this.failValidation(`unable to alter "${grantedSource.name}"${sourceId}: must be condition or effect`);
+            return false;
+        }
+
+        const hasBadge =
+            grantedSource.type === "condition"
+                ? typeof grantedSource.system.value.value === "number"
+                : grantedSource.type === "effect"
+                ? grantedSource.system.badge?.type === "counter"
```
CarlosFdez, line 238: Even if we get more badge types, changing the value should still be ok (badge type only declares that it can be incremented/decrement) but we can cross that bridge when we get there.

PR #5984, opened 2023-01-28
Implement rolling persistent damage recovery from effect panel

CarlosFdez:
image

### DIFF
--- a/src/module/apps/effects-panel.ts
+++ b/src/module/apps/effects-panel.ts
@@ -2,7 +2,7 @@ import { ActorPF2e } from "@actor";
 import { AbstractEffectPF2e, EffectPF2e } from "@item";
 import { AfflictionPF2e } from "@item/affliction";
 import { EffectExpiryType } from "@item/effect/data";
-import { htmlQueryAll } from "@util";
+import { htmlQuery, htmlQueryAll } from "@util";
 import { FlattenedCondition } from "../system/conditions";
 
 export class EffectsPanel extends Application {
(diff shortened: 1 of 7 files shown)


CarlosFdez:
An alternative is that instead of all condition.system.persistent checks, we can make a new PersistentSomethingOrOther class that gets created during data prep that is top-level, so we can roll recovery via condition.persistent.rollRecovery()

stwlam:
Does it roll damage or damage and recovery?

CarlosFdez:
Recovery. For damage my thoughts are either left click on image to roll damage for just that persistent damage, or left click on image to pop open the editor and then have a "roll persistent damage" button. I haven't decided yet :(

stwlam:
The current chafing from userland is over not being able to process persistent damage outside combat. If this is intended to address that, shouldn't it simply reproduce what normally happens at the end of an affected actor's turn?

CarlosFdez:
Its not, its only intended to address:
> You can take steps to help yourself recover from persistent damage, or an ally can help you, allowing you to attempt an additional flat check before the end of your turn. This is usually an activity requiring 2 actions, and it must be something that would reasonably improve your chances (as determined by the GM). For example, you might try to smother a flame or wash off acid. This allows you to attempt an extra flat check immediately, but only once per round.

We need a different solution for that chaff, and I'm not quite sure yet.

stwlam:
That's well handled during combat by the existing recovery buttons in chat, though.

CarlosFdez:
Its not. If you haven't taken the persistent damage yet, you can't even attempt to recover currently. You kinda have to either roll manually, or end your turn and don't apply the damage, then go back to your turn.

And if you have taken the persistent damage already, you have to scroll up the entire chat log to the previous round to attempt to recover again. You can make the attempt before you even take the damage.

CarlosFdez:
This was motivated because I remembered an episode I was crit by a ghoul alchemist in a certain AP, and we were frantically trying to wash it off so I wouldn't die to it. At the time I had to manually `/r d20`.

I kinda have an idea for the other chaff, I might settle with left click = edit persistent, and then add a roll all persistent damage in that dialog button somewhere.

PR #5990, opened 2023-01-28
Use modifiers and damage dice for applicable crit-spec effects

stwlam:


src/module/system/damage/weapon.ts
```diff
@@ -391,14 +391,32 @@ class WeaponDamagePF2e {
             }
         }
 
-        // Roll notes
+        // Critical specialization effects
+        const critSpecEffect = ((): CritSpecEffect => {
+            if (!weapon.isOfType("weapon")) return [];
+
+            // If an alternate critical specialization effect is available, apply it only if there is also a
+            // qualifying non-alternate
```
xyzzy42, line 399: Why require this?  Is there a reason to prevent a custom crit spec being applied to a weapon that wasn't assigned to an existing weapon group?  I.e., someone wants to make a new weapon that isn't in a group and give it a new kind of crit spec effect too.

The only use I can see of alternate crit spec text is Spike Launcher, which should work fine without the limitation.

src/module/system/damage/weapon.ts
```diff
@@ -391,14 +391,32 @@ class WeaponDamagePF2e {
...
+            const critSpecs = actor.synthetics.criticalSpecalizations;
+            const standard = critSpecs.standard.map((cs): CritSpecEffect => cs(weapon, options)).pop();
+            const alternate = critSpecs.alternate.map((cs): CritSpecEffect => cs(weapon, options)).pop();
```
xyzzy42, line 402: Doesn't this process all the critspecs and then just remove the last one from the array and throw away are the ones before it?  It should only be necessary to process the last one.

```
const standard = critSpecs.standard.slice(-1)[0]?.(weapon, options);
```

stwlam: Good catch, but the fix needs to be a bit different.

src/module/system/damage/weapon.ts
stwlam: An `alternate` is a crit spec that requires one already have the critical specialization feature.

src/module/system/damage/weapon.ts
```diff
@@ -391,14 +391,38 @@ class WeaponDamagePF2e {
...
+            // qualifying non-alternate
+            const critSpecs = actor.synthetics.criticalSpecalizations;
+            const standard = critSpecs.standard
+                .map((cs) => cs(weapon, options))
+                .filter((ce): ce is CritSpecEffect => !!ce)
+                .pop();
+            const alternate = critSpecs.alternate
+                .map((cs) => cs(weapon, options))
+                .filter((ce): ce is CritSpecEffect => !!ce)
+                .pop();
```
xyzzy42, line 408: Oh I see, you wanted the last one _that didn't evaluate to null_.  It's still possible to not process all the ones before the final one.

```
const standard = critSpecs.standard.reduceRight((res, cs) => res ?? cs?.(weapon, options), null);
```
This starts from the end and as soon as a cs that non-false is found it stops evaluating them.

src/module/system/damage/weapon.ts
xyzzy42: I see, the RE really has two different functions.  One is grant crit spec for the strike.  The other is to define an alternate crit spec for the strike, but not grant it.  It does only one at a time.

To create a new weapon, one would put a crit spec RE on the weapon that defines its crit spec, with a predicate to match item.id.  Then some other crit spec RE on a feat or feature would grant crit spec to some kind of group (simple, two-handed, dwarf, etc.) that the weapons happens to be in.

src/module/system/damage/weapon.ts
stwlam: If you see opportunities to micro-optimize, so make a PR

PR #5993, opened 2023-01-28
Added pf2e.chatMessageSave hook.

Lej:
So, I have been trying to create a module that makes it easier to get an overview of save results as well as to select involved tokens:
image

I have not been successful in doing what I wanted without adding some hooks to PF2E.
I have no idea if this is reasonable or not, or if there's some way to implement this without changing PF2E.
Do this make any sense, or am I going about this in the wrong way?
Just thought I should check before I spend more time on this.

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -9,6 +9,21 @@ import { onRepairChatCardEvent } from "@system/action-macros/crafting/repair";
 import { LocalizePF2e } from "@system/localize";
 import { ErrorPF2e, sluggify, tupleHasValue } from "@util";
 import { ChatMessagePF2e } from "..";
+import { DegreeOfSuccessString } from "@system/degree-of-success";
+import { CheckRoll } from "@system/check/roll";
+import { TokenPF2e } from "@module/canvas";
+
+export interface ChatMessageSaveEventArgs {
+    event: JQuery.ClickEvent;
+    button: HTMLElement;
+    message: ChatMessagePF2e;
+    actor: ActorPF2e | null | undefined;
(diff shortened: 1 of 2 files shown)


stwlam:
You can reliably find such messages by checking the check context flag:
image

Lej:
> You can reliably find such messages by checking the check context flag: image

@stwlam While this is helpful, I don't think it solves the main problem I am having. I want to, given a message corresponding to a save rolled from another message, find:
1. The message that contains the button that was clicked to roll the save
2. The button that was clicked to roll the save

And vice versa, given a message with save buttons, find any messages that resulted from clicking those buttons.
Is this possible?

stwlam:
What is the information you're wanting to acquire from the other message and its button?

Lej:
> What is the information you're wanting to acquire from the other message and its button?

1. Something to group saves message on. E.g. I want to be able to differentiate save message that come from two different fireball chat message.
2. Something humanly readable to differentiate which save button in the other message that was clicked (since there can be multiple such buttons; default spell save button and/or inline saves). If could e.g. be the text of the clicked save button, or the button itself.
3. Any damage rolls resulting from the other message, and something humanly readable to distinguish which button was pressed to roll the damage (e.g. damage button text).

I want this to be able to, given a message containing save and/or damage buttons, be able to present all rolls resulting from pressing those buttons.

For each such save rolled, I also want to be able to find the actors/tokens that rolled those saves so that I can easily select them, apply damage to them, or similar (this should already be possible I think).

stwlam:
We can give you the DOM event associated with the roll, if present.

Lej:
> We can give you the DOM event associated with the roll, if present.

I think that should work, the event should contain a reference to the triggering element.

PR #6006, opened 2023-01-29
Fix several base speed REs

### LINKED ISSUE #5999: Wings of Air Effect RE looks to be incorrect.
Due to wording in the feat, I believe that the Base Speed RE should look like this.
image

TikaelSol:
Closes #5999


stwlam:
This break four more instead of fixing the reported one.

TikaelSol:
Ah, went the wrong way on them

PR #6007, opened 2023-01-29
Add autocomplete key options based on schema in RE codemirror editor

CarlosFdez:


src/module/item/sheet/base.ts
```diff
@@ -282,9 +282,14 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
         const editingRuleElement = this.editingRuleElement;
         if (editingRuleElement) {
             const ruleText = JSON.stringify(editingRuleElement, null, 2);
+            const schema = RuleElements.all[String(editingRuleElement.key)]?.defineSchema();
```
stwlam, line 285:
```suggestion
            const schema = RuleElements.all[String(editingRuleElement.key)]?.schema;
```

PR #6011, opened 2023-01-29
Add perform action macro

nikolaj-a:
For the Perform skill action, add action implementation to the system and expose it on the pf2e namespace of the game object, add several action macros to call that system implementation with the most commonly used variants from the Core Rulebook, and also add links to execute the action implementation from the action item description.

Perform variants and traits


stwlam:
I wonder if this could be reduced to the combination of traits involved.

TikaelSol:
There's things that give explicit bonuses to Perform (Oratory) or (Winds), so that info is also necessary.

PR #6014, opened 2023-01-29
Add support for effect counter labels and formula badges

CarlosFdez:
Labels exist in data yet but they're not shown. We can probably add it to the tooltip under the name?

image

image

image

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -2215,9 +2215,16 @@
                 },
                 "ShowTokenIcon": "Show token icon?",
                 "Unidentified": "Unidentified?",
-                "AddBadge": "Add Counter",
+                "AddBadge": "Add",
+                "AddLabel": "Add Label",
+                "LabelPlaceholder": "Stage Label",
                 "BadgeType": {
-                    "Counter": "Counter"
+                    "counter": "Counter",
+                    "formula": "Formula",
+                    "value": "Fixed Value"
+                },
+                "ExtraSelections": {
+                    "Labels": "Counter (Labeled)"
                 }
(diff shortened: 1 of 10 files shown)


CarlosFdez:
I can add a devise a strategem implementation as part of this or separately, but functioning it looks like this

https://user-images.githubusercontent.com/1286721/215319326-0dd56869-d32e-4cf5-86c1-24cdb4e5b074.mp4

Note that the "value" badge type cannot be manually created, it only exists as a result of the formula evaluation.

static/lang/en.json
```diff
@@ -2201,9 +2201,16 @@
...
+                "AddBadge": "Add",
+                "AddLabel": "Add Label",
+                "LabelPlaceholder": "Stage Label",
                 "BadgeType": {
-                    "Counter": "Counter"
+                    "counter": "Counter",
+                    "formula": "Formula",
+                    "value": "Value"
+                },
+                "ExtraSelections": {
```
CarlosFdez, line 2212: This fundamentally bothers me and I'm willing to try a replacement.

stwlam: "Fixed value" communicates it a little better

CarlosFdez: For Mild/Moderate/Severe or is this about Value? I could buy it for Value.

My umbrage was at having an ExtraSelections section 🙃

stwlam: yeah what the fuck is that, and why is it bactrian'd

stwlam: Yes for value

PR #6018, opened 2023-01-29
Remove extra data attribute in effects panel template, tidy class

stwlam:


CarlosFdez:
I think I handled this in #6020 (ran into it also while working on something), and I think I also handled not increasing the effect stage when clicking in the tooltip as well.

PR #6020, opened 2023-01-29
Open persistent damage dialog when left clicking effect panel icon

CarlosFdez:


src/module/apps/effects-panel.ts
```diff
@@ -72,11 +73,15 @@ export class EffectsPanel extends Application {
         super.activateListeners($html);
         const html = $html[0]!;
 
-        for (const iconElem of htmlQueryAll(html, "div[data-item-id]")) {
+        for (const effectEl of htmlQueryAll(html, ".effect-item[data-item-id]")) {
+            const id = effectEl.dataset.itemId;
```
stwlam, line 77:
```suggestion
            const id = effectEl.dataset.itemId;
```
`itemId`?

CarlosFdez: sure!

PR #6023, opened 2023-01-29
Add a secret GM notes field on items

nikolaj-a:
Add a GM notes field to the common item template, which is shown only to GM users in the description tab of all item sheets.

The specific use-case right now is for forged documents that need a secret DC to discover the forgery for passive observers, but the field itself should be usable in many situations.

image


stwlam:
How big is it when it's empty?

nikolaj-a:
Same size. I found it to be a little hard to adjust the size properly. Maybe it can be collapsed a bit using `:has()` or by converting the description tab to a grid?

I picked the current size by looking at this:
image

With the default item sheet window size, the editor buttons take up two rows making it a _super_ tight fit 😐

nikolaj-a:
Another option is to add a GM tab. Not my preferred option due to the reduced visibility. But it does give room for more GM-y things.

stwlam:
Maybe it can default to `null` and have minimal profile until the GM clicks +?

nikolaj-a:
Let me try that.

nikolaj-a:
For some reason the `:has(...)` selector doesn't work properly with descendant classes 🤷 This is the most compact I could make it, since the editor has to be present in the DOM for `Sheet#activeEditor(...)` to work.

image

CarlosFdez:
You could check hazards to see how to make them compact. I had to do some shenaniganry there.

nikolaj-a:
After taking a look at the hazard sheet, I decided to go in a slightly different direction and put the GM notes at the top and let the entire description panel flow normally, instead of having an always-visible, height-adapting block at the bottom.
image

src/module/item/sheet/base.ts
```diff
@@ -379,6 +383,16 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
                 this.item.update({ [`system.variants.-=${index}`]: null });
             });
         }
+
+        for (const section of htmlQueryAll(html, "[data-action=edit-section]")) {
+            section.addEventListener("click", (event) => {
+                if (!(event.target instanceof HTMLElement)) return;
```
stwlam, line 389: Is this not the same element as `section`?

static/templates/items/item-sheet.hbs
```diff
@@ -82,11 +82,37 @@
...
             <!-- Item Description -->
             <div class="tab item-description" data-tab="description">
-                {{#if (not isVariant)}}
-                    {{editor enrichedContent.description target="system.description.value" button=true owner=owner editable=editable}}
-                {{else}}
-                    {{{enrichedContent.description}}}
-                {{/if}}
+                <div class="item-description-layout">
+                    {{#if user.isGM}}
+                        <div class="section">
```
stwlam, line 87: A `div` with a class of "section" is rather perverse. :)

src/module/item/sheet/base.ts
nikolaj-a: It's _supposed_ to be the little 🗒️ icon in the header. I might be able to attach the listener to the section instead if that's better.

static/templates/items/item-sheet.hbs
nikolaj-a: Haha, good point! 😂 I'll change it to use `<section>` instead.

src/module/item/sheet/base.ts
nikolaj-a: Moved the listener to the container instead.

shemetz:
Can you please change the "**GM Notes**" header to look different while it's empty?  I imagine that 99% of items will not have any GM notes, and for all of those cases the extra text and separating line take up a lot of space, along with drawing the eye.

I would prefer it if the GM Notes element didn't take any lines at all, as long as it's empty;  and the GM would need to click some button in order to add GM notes, with that button being unobtrusive and somewhere on the side or the top of the dialog window.

nikolaj-a:
Compacted the description tab further when there are no GM notes:
Item sheet without GM notes

Issue #6057, opened 2023-01-31
Expose `GhostTemplate`


theripper93:
I need to override the drawPreview method for 3D Canvas, but that is currently impossible since the `GhostTemplate` class is not exposed anywhere

PR #6058, opened 2023-01-31
Expose GhostTemplate class in game.pf2e namespace

nikolaj-a:
Closes https://github.com/foundryvtt/pf2e/issues/6057

### DIFF
--- a/src/global.d.ts
+++ b/src/global.d.ts
@@ -43,8 +43,9 @@ import { ModuleArt } from "@system/module-art";
 import { CustomDamageData, HomebrewTag, HomebrewTraitSettingsKey } from "@system/settings/homebrew";
 import { TextEditorPF2e } from "@system/text-editor";
 import { sluggify } from "@util";
-import { CombatantPF2e, EncounterPF2e } from "./module/encounter";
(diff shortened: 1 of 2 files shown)


stwlam:
Hmm do we gotta? :\

nikolaj-a:
I mean, we don't _have_ to. But unless we have plans to change the implementation in the near future I don't see why not. Apparently it's needed for some module integration 🤷

CarlosFdez:
I should dig a bit, but I suspect there's a better way to do this, perhaps as a static variable in MeasuredTemplatePF2e or elsewhere. We expose MeasuredTemplatePF2e but apparently that's not enough to alter the draw preview.

CarlosFdez:
I'm double checking and having a separate GhostTemplate I think was wrong from the beginning. The way 5e has it where they're both the same class, only differentiated by drawPreview(), is probably the more correct approach to begin with. We should probably address that and that'll fix the issue in a much cleaner way I think.

nikolaj-a:
Closed in favor of https://github.com/foundryvtt/pf2e/pull/6078

PR #6069, opened 2023-01-31
Have Diverse Weapon Expert upgrade advanced weapon proficiency to trained

AFigureOfBlue:
The Diverse Weapon Expert feat already used rule elements to set up the new proficiency for simple and martial weapons, but was missing the proficiency for advanced weapons.


stwlam:
Put PR titles in imperative mood rather than framing them like issue reports

AFigureOfBlue:
> Put PR titles in imperative mood rather than framing them like issue reports

Thanks for the note! I will aim to do so going forward.

PR #6074, opened 2023-02-01
Add convenience getter for retrieving effects' originating actors and items

stwlam:


src/module/item/abstract-effect/document.ts
```diff
@@ -12,6 +14,28 @@ export abstract class AbstractEffectPF2e extends ItemPF2e {
     abstract increase(): Promise<void>;
     abstract decrease(): Promise<void>;
 
+    /** Get the actor that created this effect */
+    get origin(): { actor: ActorPF2e | null; item: Embedded<ItemPF2e> | null } {
+        const [actorOrToken, item] = this.isOfType("affliction", "effect")
```
CarlosFdez, line 19: given that condition overrides this getter, I don't think you have to check the type right?

stwlam: Only afflictions and effects have a `context` in their system data

CarlosFdez: Typescript thing it is!

Issue #5078, opened 2022-12-06
Expanded Compendium Browser CSS Bug


MrVauxs:
Happens without modules.
![image](https://user-images.githubusercontent.com/32039708/205944334-6793218a-7e25-4059-983b-da858e21f921.png)

PR #6083, opened 2023-02-01
Fix placement of save button in compendium browser on resizing

Tythyrn:
Fixes foundryvtt/pf2e#5078

This pushes the save button to the bottom of the screen so that the user can expand the compendium to as large as they'd like without getting a lot of white space.

Note: This is my first contributing PR to a GitHub project that isn't mine so let me know if I need to include anything else!

image


TikaelSol:
Changed the title to be in line with the system standard. PR titles should finish the sentence "If merged this PR will _____"

PR #6091, opened 2023-02-02
Implement broken reduction for vehicle AC and fortitude saves

CarlosFdez:


CarlosFdez:
I suspect the styling for adjusted-lower might need some help on contrast though:
image

PR #6099, opened 2023-02-02
Add support for categorizing NPC damage partials

stwlam:
Has a side effect of fixing NPC damage breakdowns

### DIFF
--- a/static/templates/items/melee-details.hbs
+++ b/static/templates/items/melee-details.hbs
@@ -1,37 +1,46 @@
-<div class="form-group">
+<div class="form-group attack-modifier">
     <label>{{localize "PF2E.NPCWeaponAttackLabel"}}</label>
-    <input type="number" name="system.bonus.value" value="{{data.bonus.value}}" data-dtype="Number"/>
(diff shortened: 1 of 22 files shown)


static/templates/items/melee-details.hbs
```diff
@@ -1,37 +1,46 @@
...
+                </select>
+            </div>
+            <div class="form-group category">
+                <label>{{localize "PF2E.Category"}}</label>
+                <select name="system.damageRolls.{{key}}.category">
+                    {{#select (lookup damageRoll "category")}}
+                        <option value=""></option>
+                        <option value="persistent">{{localize "PF2E.ConditionTypePersistentShort"}}</option>
+                        <option value="precision">{{localize "PF2E.Damage.Precision"}}</option>
+                        <option value="splash">{{localize "PF2E.TraitSplash"}}</option>
```
CarlosFdez, line 30: You should be able to map this from sheet data if you pass in pick(CONFIG.PF2E.damageCategories, DAMAGE_CATEGORIES_UNIQUE). That's what I did for afflictions for example.

Definitely would like a rename for DAMAGE_CATEGORIES_UNIQUE though. That only came about because main was unavailable.

PR #6121, opened 2023-02-03
Add more tweaks to PC header alignment

CarlosFdez:
image


CarlosFdez:
I removed the bottom before but I never removed the top padding. The decreased gap and the (slightly) better left alignment made it kinda hard to notice there was some oddness though.

I can fully remove the level bottom padding if you think it looks better. THis is how it would look like if I were to do that. 
image

PR #6132, opened 2023-02-03
Fix token distance measurement in 3D space

### LINKED ISSUE #4824: token.distanceTo returning inaccurate measurements for elevated tokens.
I've been testing token.distanceTo (I've implemented it here in TTA for ease of viewing, but I get the same results in the console with all modules disabled). 

Things mostly seem to work as you'd expect when everyone is at the same elevation, but if tokens are at different elevations the distance seems to increase a lot faster than one would expect. I've tested with ranged attacks and confirmed that those are using these same measurements for range penalties.

image

In the above example the one token is only 5ft up, but the distance between it and the token on the ground increases by 40 feet. While there should only be a single diagonal required in this measurement. 

Replication: measure distanceTo between two tokens at different elevations.

Drental:
closes #4824

removed Pythagoras of 3 diagonals in favor of 3d counting of minecraft blocks.

### DIFF
--- a/src/module/canvas/helpers.ts
+++ b/src/module/canvas/helpers.ts
@@ -18,33 +18,7 @@ function measureDistanceRect(
         return canvas.grid.measureDistance(r0, r1);
     }
 
-    const gridWidth = canvas.grid.grid.w;
-
-    // Return early if the rectangles overlap
-    const rectanglesOverlap = [
-        [r0, r1],
-        [r1, r0],
-    ].some(([rA, rB]) => rB.right > rA.left && rB.left < rA.right && rB.bottom > rA.top && rB.top < rA.bottom);
-    if (rectanglesOverlap) return 0;
-
-    // Snap the dimensions and position of the rectangle to grid square units
-    const snapBounds = (rectangle: PIXI.Rectangle, { toward }: { toward: PIXI.Rectangle }): PIXI.Rectangle => {
-        const roundLeft = rectangle.left < toward.left ? Math.ceil : Math.floor;
-        const roundTop = rectangle.top < toward.top ? Math.ceil : Math.floor;
-
-        const left = roundLeft(rectangle.left / gridWidth) * gridWidth;
-        const top = roundTop(rectangle.top / gridWidth) * gridWidth;
-        const width = Math.ceil(rectangle.width / gridWidth) * gridWidth;
-        const height = Math.ceil(rectangle.height / gridWidth) * gridWidth;
-
-        return new PIXI.Rectangle(left, top, width, height);
-    };
-
-    // Find the minimum distance between the rectangles for each dimension
-    const r0Snapped = snapBounds(r0, { toward: r1 });
-    const r1Snapped = snapBounds(r1, { toward: r0 });
-    const dx = Math.max(r0Snapped.left - r1Snapped.right, r1Snapped.left - r0Snapped.right, 0) + gridWidth;
-    const dy = Math.max(r0Snapped.top - r1Snapped.bottom, r1Snapped.top - r0Snapped.bottom, 0) + gridWidth;
+    const { dx, dy } = measureDistanceRectDimensions(r0, r1);
 
     return measureDistanceOnGrid({ dx, dy }, { reach });
 }
--- a/src/module/canvas/helpers.ts
+++ b/src/module/canvas/helpers.ts
@@ -50,6 +50,52 @@ function measureDistanceRect(
 }
 
 /**
+ * Measure the minimum distance between two rectangles on x and y Dimensions
+ * @param r0      The origin rectangle
+ * @param r1      The destination rectangle
+ */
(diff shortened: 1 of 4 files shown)


src/module/canvas/helpers.ts
```diff
@@ -46,7 +66,7 @@ function measureDistanceRect(
     const dx = Math.max(r0Snapped.left - r1Snapped.right, r1Snapped.left - r0Snapped.right, 0) + gridWidth;
     const dy = Math.max(r0Snapped.top - r1Snapped.bottom, r1Snapped.top - r0Snapped.bottom, 0) + gridWidth;
 
-    return measureDistanceOnGrid({ dx, dy }, { reach });
+    return { dx: dx, dy: dy };
```
stwlam, line 69:
```suggestion
    return { dx, dy };
```

src/module/canvas/token/object.ts
```diff
@@ -327,27 +328,29 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
...
-                target.bounds.height,
-                Math.floor((targetDimensions.height / gridDistance) * gridSize)
-            ),
+        distance.dz = measureDistanceRectDimensions(xzPlane.self, xzPlane.target).dy;
+
+        const nx = Math.ceil(Math.abs(distance.dx / gridSize));
+        const ny = Math.ceil(Math.abs(distance.dy / gridSize));
+        const nz = Math.ceil(Math.abs(distance.dz / gridSize));
+
+        // ingore the lowest difference
```
stwlam, line 337:
```suggestion
        // Ignore the lowest difference
```

src/module/canvas/helpers.ts
```diff
@@ -18,14 +18,34 @@ function measureDistanceRect(
...
+
+    return measureDistanceOnGrid({ dx, dy }, { reach });
+}
+
+/**
+ * Measure the minimum distance between two rectangles on x and y Dimensions
+ * @param r0      The origin rectangle
+ * @param r1      The destination rectangle
+ */
+function measureDistanceRectDimensions(r0: PIXI.Rectangle, r1: PIXI.Rectangle): { dx: number; dy: number } {
```
stwlam, line 31: Given how little `measureDistanceRect` does now, does it even need to exist anymore?

Drental: I will look into where it is used and if that can be migrated. Chances are Probably yes.

Drental: refactored so that `measureDistanceRect` doesn't exist anymore. Tellme how wrong my method to achieve that was :)

also played around with gridless so it isn't pure dumb anymore.

src/module/canvas/helpers.ts
```diff
@@ -20,33 +26,75 @@ function measureDistanceRect(
 
     const gridWidth = canvas.grid.grid.w;
 
+    let dx = 0;
+    let dy = 0;
+    let dz = 0;
```
stwlam, line 31: No `let`s, son!

src/module/canvas/token/object.ts
```diff
@@ -297,57 +297,25 @@ class TokenPF2e extends Token<TokenDocumentPF2e> {
...
-        const hypotenuse = Math.sqrt(Math.pow(distance.xy, 2) + Math.pow(Math.max(distance.xz, distance.yz), 2));
-
-        return Math.floor(hypotenuse / gridDistance) * gridDistance;
+        return measureDistanceRect(this.bounds, target.bounds, {
+            reach,
+            elevation0: Math.floor((selfElevation / gridDistance) * gridSize),
+            height0: Math.floor((selfDimensions.height / gridDistance) * gridSize),
+            elevation1: Math.floor((targetElevation / gridDistance) * gridSize),
+            height1: Math.floor((targetDimensions.height / gridDistance) * gridSize),
+        });
```
stwlam, line 318: This signature is kind of silly: 0 and 1 correspond with `this` and `target`

Drental: should I rename the signature or pass this and target instead?

stwlam: Renaming 0 and 1 to self and target doesn't make it less silly :stuck_out_tongue:

stwlam: Pass the pairs of tokens instead

Drental: got it. moved :)

src/module/canvas/helpers.ts
```diff
@@ -10,7 +10,15 @@ import { TokenPF2e } from "./token";
 function measureDistanceRect(
```
stwlam, line 10: Now rename to `measureDistanceCuboid` :)

stwlam (changes requested):

src/module/canvas/helpers.ts
Drental: renamed and references fixed.
Running another test

Drental: https://user-images.githubusercontent.com/74384067/218332565-e4bfea3a-c2af-4234-96bb-8728bd45875f.MP4

stwlam (approved):

PR #6141, opened 2023-02-04
Automate Dread Striker

TikaelSol:
I don't like the roll option name, but that's easily changed. If we had roll options for initiative value and skill rolled this same method could be used for Surprise Attack.

I'll need to write a migration, which won't be hard just need a finalized version of this.

Also need to check if the target is blanket immune to being flat footed, though not sure how to go about that one immediately.

### DIFF
--- a/packs/data/feats.db/dread-striker.json
+++ b/packs/data/feats.db/dread-striker.json
@@ -21,7 +21,16 @@
         "prerequisites": {
             "value": []
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "option": "self:override-target-flat-footed",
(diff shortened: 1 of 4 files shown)


stwlam:
Would rather do this with a more general Relative effect RE

TikaelSol:
What other conditions are granted relative like this? I can only think of flat footed.

stwlam:
https://2e.aonprd.com/Feats.aspx?ID=3883

TikaelSol:
That's overriding mental immunity for a specific action, not granting a separate condition based on the circumstance. But if you want to tackle it in some more generic way that's fine.

stwlam:
It would be an effect that removes an immunity

PR #6154, opened 2023-02-05
Add resistance rule element to Fire Resistance feat

KSops:
Add missing RE for Fire Resistance feat
https://2e.aonprd.com/Feats.aspx?ID=2973

### DIFF
--- a/packs/data/feats.db/fire-resistance.json
+++ b/packs/data/feats.db/fire-resistance.json
@@ -34,6 +34,11 @@
                 "selector": "saving-throw",
                 "type": "circumstance",
                 "value": 1
+            },
+            {
+                "key": "Resistance",
+                "type": "fire",
+                "value": "max(1,floor(@actor.level/2))"
             }
         ],
         "source": {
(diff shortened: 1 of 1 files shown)


packs/data/feats.db/fire-resistance.json
```diff
@@ -34,6 +34,11 @@
                 "selector": "saving-throw",
                 "type": "circumstance",
                 "value": 1
+            },
+            {
+                "key": "Resistance",
+                "type": "fire",
+                "value": "max(1,floor(@actor.level/2))"
```
stwlam, line 41: The `max` shouldn't be there

stwlam (changes requested):

packs/data/feats.db/fire-resistance.json
TikaelSol, line 41:
```suggestion
                "value": "floor(@actor.level/2)"
```

KSops:
Wasn't sure if the max was some convention or not based on other existing REs like packs/data/feats.db/nothing-but-fluff.json
Thanks for tweaking

stwlam:
> Wasn't sure if the max was some convention or not based on other existing REs like packs/data/feats.db/nothing-but-fluff.json Thanks for tweaking

Some abilities and effects set a minimum of 1, but this one doesn't.

PR #6156, opened 2023-02-05
Clean up remaining data.data and actor.data.type references

xyzzy42:


xyzzy42:
There is also something in hotbar.ts,
```ts
                const itemId = data.id ?? (isObject<{ _id?: unknown }>(data.data) ? data.data._id : null);
                const uuid = data.uuid;
                const prefix =
                    typeof data.pack === "string"
                        ? `Compendium.${data.pack}`
                        : typeof data.actorId === "string"
                        ? `Actor.${data.actorId}.Item`
                        : "Item";
                const item = await fromUuid(uuid ?? `${prefix}.${itemId}`);
```
As far as I could tell, by randomly dragging different things from different places, `data.uuid` is always set and neither `data.id` nor `data.data` is ever set.  But maybe there is a possibility I didn't try that still uses the non-uuid code path.

PR #6163, opened 2023-02-05
Release Notes 4.7.0

stwlam:


In3luki:
The damage dialog should be left out or marked as dev only.

PR #6165, opened 2023-02-05
Extend refreshFromCompendium to keep additional NPC settings

MSAbaddon:


stwlam:
I don't know what you mean by  "NPC settings"

MSAbaddon:
Basically, I used the function to update spells on npcs for a whole compendium and the additions helped reducing manual fixes afterwards. Yet, I didn't want to make the additional things default to the function so I thought I add the option to do it one way or the other.

src/module/item/base.ts
```diff
@@ -213,7 +213,7 @@ class ItemPF2e extends Item<ActorPF2e> {
     }
 
     /** Pull the latest system data from the source compendium and replace this item's with it */
-    async refreshFromCompendium(): Promise<void> {
+    async refreshFromCompendium(keepNPCSettings = false): Promise<void> {
```
stwlam, line 216: This isn't a setting

src/module/item/base.ts
```diff
@@ -243,8 +243,45 @@ class ItemPF2e extends Item<ActorPF2e> {
             const { containerId, quantity } = currentSource.system;
             mergeObject(updates, expandObject({ "system.containerId": containerId, "system.quantity": quantity }));
         } else if (currentSource.type === "spell") {
-            // Preserve spellcasting entry location
-            mergeObject(updates, expandObject({ "system.location.value": currentSource.system.location.value }));
+            if (!keepNPCSettings) {
+                // Preserve spellcasting entry location
+                mergeObject(updates, expandObject({ "system.location.value": currentSource.system.location.value }));
```
stwlam, line 248: Don't ever use `expandObject`, and just make it do the right thing instead of requiring an argument to be passed.

MSAbaddon: Not sure if I correctly understood what you meant, sorry :-(
So instead of adding an option to do it one way or the other, I should just change it the way I believe it works better?
Regarding expandObject - well, I just copied what was alread there - so instead of doing this I should replace the expandObject entirely and just merge an object I created before?

PR #6167, opened 2023-02-05
Localize and fix traits for the save of Stunning Fist

TikaelSol:
Plus bonus extract noise.

override traits needed to be added because monk punches become magical but stunning fist isn't and monsters should not get their +1 vs magic bonus on it.


stwlam:
Try pulling: all that stuff is already fixed

PR #6176, opened 2023-02-05
Replace explicit NPC crit spec text with PF2E.Item.Weapon.CriticalSpecialization.*

xyzzy42:
Doesn't change note REs into damage REs, e.g. for pick.
Doesn't change the brawling notes, as the class DC based inline save doesn't work for NPCs.

### DIFF
--- a/packs/data/feats.db/garudas-squall.json
+++ b/packs/data/feats.db/garudas-squall.json
@@ -28,13 +28,10 @@
         },
         "rules": [
             {
-                "key": "Note",
-                "outcome": [
-                    "criticalSuccess"
-                ],
-                "selector": "bow-weapon-group-damage",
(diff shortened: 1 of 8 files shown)


xyzzy42:
There is a mistake in graveknight-champion.  The greatpick does slashing damage, but the crit spec does piercing.

And the crit spec does 2 fire too, which is a bit odd, since normally property runes don't count as a weapon damage dice.

packs/data/feats.db/garudas-squall.json
```diff
@@ -33,7 +33,7 @@
                     "criticalSuccess"
                 ],
                 "selector": "bow-weapon-group-damage",
-                "text": "If the target of the critical hit is adjacent to a surface, it gets stuck to that surface by the missile. The target is @UUID[Compendium.pf2e.conditionitems.Immobilized]{Immobilized} and must spend an Interact action to attempt a @Check[type:athletics|dc:10] check to pull the missile free; it can't move from its space until it succeeds. The creature doesn't become stuck if it is incorporeal, is liquid (like a water elemental or some oozes), or could otherwise escape without effort.",
+                "text": "PF2E.Item.Weapon.CriticalSpecialization.bow",
```
stwlam, line 36: This should just be a CritSpec RE for bows

xyzzy42: Didn't notice it was a feat.  Fixed.

PR #6180, opened 2023-02-06
Allow NPC melee categorized partial without matching base

CarlosFdez:
image


CarlosFdez:
Turns out I already did the refactor that allowed this, I was just missing a step....

PR #6189, opened 2023-02-06
Normalize "skill" actions by adding missing trait

### LINKED ISSUE #5851: Skill Actions Missing Skill Trait
The "Balance", "Borrow a Spell" and "Demoralize" actions don't have the Skill trait in the compendium:
image

I am not 100% sure this is not as intended, but as far as I can tell all other Skill actions have the trait even if it is not the case in the book, so I think this is an error

KSops:
Closes #5851

Even though this may not be strictly correct, imho it's better to at least have normalized behaviour.

### DIFF
--- a/packs/data/actions.db/balance.json
+++ b/packs/data/actions.db/balance.json
@@ -26,7 +26,8 @@
             "custom": "",
             "rarity": "common",
             "value": [
-                "move"
+                "move",
+                "skill"
             ]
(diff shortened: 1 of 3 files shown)


TikaelSol:
I don't know how I feel about this one. I guess we can do it until we do something better for skill actions and then we can fix them all at once.

PR #6192, opened 2023-02-06
Hide "setInitiative" button for blind rolls

### LINKED ISSUE #6030: Issue: Players are able to see Blind Rolls using the "Use as Initiative" button.
My players are able to see the results of blind rolls by clicking on the "Set as initiative" button. An example is on the screenshot attached; The player rolled a blind check, but by clicking on the button, it set their initiative as that value, making the blind roll useless.
Is there a way to not show this button when a roll is blind? Or maybe show "???" as the initiative value?

Player's view:
image

GM's view:
image

FoundryVTT v10.291
PF2 v4.6.8

KSops:
Closes #6030

### DIFF
--- a/src/module/chat-message/listeners/set-as-initiative.ts
+++ b/src/module/chat-message/listeners/set-as-initiative.ts
@@ -5,7 +5,7 @@ export const SetAsInitiative = {
         const { context } = message.flags.pf2e;
         if (
             message.token &&
-            (message.isAuthor || game.user.isGM) &&
+            ((message.isAuthor && !message.blind) || game.user.isGM) &&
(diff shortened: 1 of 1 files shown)


stwlam:
If a PC or other creature rolls stealth for initiative, their initiative roll is used as both a sneak check and their actual initiative roll. We should hide their placement in the tracker for cases like this, leaving it to the GM to reveal it when the creature's turn starts.

KSops:
Is this something that can be done from the system-level? If so could you outline how this might be achieved?

CarlosFdez:
I don't think rolling stealth for initiative is a secret check though.

PR #6216, opened 2023-02-07
Fix application of "deadAtZero" setting

stwlam:
For some reason, `Actor#update` no longer returns itself following the update if the actor is synthetic.


CarlosFdez:
Why are you using the return result without checking it then?

stwlam:
It returns "itself," just not the same reference

CarlosFdez:
Ah, so you're not working around it, its just preventing you from doing an even nicer solution.

PR #6227, opened 2023-02-07
Fix name of Thorns battle form translation and update Hidden/Murderous Thorn(s) feats

xyzzy42:
The BattleForm RE uses "Thorns".  It's written "thorns" in the book.  But the translation entry was "Thorn".

### DIFF
--- a/packs/data/feats.db/hidden-thorn.json
+++ b/packs/data/feats.db/hidden-thorn.json
@@ -34,9 +34,9 @@
                 "group": "knife",
                 "img": "systems/pf2e/icons/unarmed-attacks/spine.webp",
                 "key": "Strike",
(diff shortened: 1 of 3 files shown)


stwlam:
Are there any other REs using that translation key?

TikaelSol:
Hidden Thorn does.

xyzzy42:
Hidden **Thorn**: "You gain a **thorns** unarmed attack that deals 1d6 piercing damage."
Murderous **Thorns**: "If you have the Hidden Thorn feat, your **thorn** attack gains the deadly d6 trait."

I love how the feat name is the opposite of the attack name, for maximum inconsistency.

I could update both feats to be thorns, the slug in the former and the item predicate in the latter.

stwlam:
Go ahead and make it all "thorns"

PR #6228, opened 2023-02-07
Allow ranged BattleForm RE strikes and make Green Man's Thorns ranged

xyzzy42:


packs/data/spell-effects.db/spell-effect-nature-incarnate-green-man.json
```diff
@@ -45,7 +45,9 @@
                                 "modifier": 6
                             },
                             "group": "brawling",
+                            "maxRange": 100,
                             "modifier": 34,
+                            "range": 100,
                             "traits": [
                                 "unarmed",
                                 "range-100"
```
stwlam, line 53:
```suggestion
                            "range": 100,
                            "traits": [
                                "unarmed"
```

xyzzy42: These range traits are only for NPC attacks?

Once this is removed, then the range no longer shows up on the strike in the actions tab.

However, I do have another PR to display range increment on strikes.  Maybe something like that is better than the trait?  Or perhaps the trait should be synthetically generated from the range data?

Issue #5586, opened 2023-01-08
Druid Feat Order Explorer


dootxdoot:
When you select class Feat Order Explorer it's not prompting to ask which order to join.

PR #6248, opened 2023-02-08
Automate Order Explorer Feat

avagdumortesin:
Fixes https://github.com/foundryvtt/pf2e/issues/5586.

### DIFF
--- a/packs/data/feats.db/order-explorer.json
+++ b/packs/data/feats.db/order-explorer.json
@@ -86,7 +86,6 @@
                 "flag": "second-order",
                 "key": "ChoiceSet",
                 "prompt": "PF2E.SpecificRule.Druid.DruidicOrder.Prompt",
-                "recordSlug": true,
                 "rollOption": "feat:order-explorer"
             },
             {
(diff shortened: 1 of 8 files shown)


TikaelSol:
We'll need to rewrite the orders as well to not grant their other benefits like granting a skill training / etc. Also we should do the first level feat choice, which will just have to be a manually built list since we won't be able to query that one (and don't want to add more queries right now anyways).

packs/data/feats.db/order-explorer.json
```diff
@@ -22,7 +22,78 @@
...
+                                "not": "feature:wild-order"
+                            }
+                        ],
+                        "value": "Compendium.pf2e.classfeatures.Wild Order"
+                    }
+                ],
+                "flag": "second-order",
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Druid.DruidicOrder.Prompt",
+                "recordSlug": true,
```
stwlam, line 89:
```suggestion
```

PR #6264, opened 2023-02-08
Fix REs on Zombie Dedication and Numb feats, add Deteriorated effect

stwlam:


packs/data/feat-effects.db/effect-deteriorated.json
```diff
@@ -0,0 +1,45 @@
...
+            "value": -1
+        },
+        "level": {
+            "value": 2
+        },
+        "rules": [
+            {
+                "key": "GrantItem",
+                "onDeleteActions": {
+                    "granter": "detach"
```
stwlam, line 23:
```suggestion
                    "granter": "detach",
                    "grantee": "cascade"
```

PR #6272, opened 2023-02-09
Have status effects panel operate on all controlled tokens

stwlam:
This actually works out quite cleanly:
- If no token is controlled, right-clicking on a token also controls it.
- Right-clicking any token among a set of controlled tokens maintains the controlled status of all tokens.
- If a token isn't among those controlled when right-clicking it, controlled tokens are cleared, and the right-clicked token is then controlled.

image

### DIFF
--- a/src/module/canvas/status-effects.ts
+++ b/src/module/canvas/status-effects.ts
@@ -1,6 +1,6 @@
 import { LocalizePF2e } from "@system/localize";
 import { StatusEffectIconTheme } from "@scripts/config";
-import { ErrorPF2e, fontAwesomeIcon, objectHasKey } from "@util";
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Not a bad idea! Though I can imagine it getting a little hairy if one token has frightened one and then you click on another token (which by default in foundry doesn't deselect the first one) also apply frightened to that one, thus making the original token frightened 2.

stwlam:
By default in Foundry it does deselect the first one.

CarlosFdez:
I remember it used to work that way without left click deselect, but I guess that's been changed. Neat, then this shouldn't cause any issues!

PR #6291, opened 2023-02-10
Fix `fromUUIDs` never returning cached compendium documents

In3luki:
`fromUuidSync` alawys returns the compendium index entry instead of the document (even if it is cached locally) so the cache check didn't work at all.

```js
...
  if ( collection instanceof CompendiumCollection ) {
    doc = doc ?? collection.index.get(documentId);
    if ( doc ) doc.pack = collection.collection;
  } else {
...
````

### DIFF
--- a/src/util/from-uuids.ts
+++ b/src/util/from-uuids.ts
@@ -10,13 +10,15 @@ async function fromUUIDs(uuids: string[]): Promise<ClientDocument[]> {
     const actors: ActorPF2e[] = [];
     const items: ItemPF2e[] = [];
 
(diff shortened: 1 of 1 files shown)


stwlam:
I'd rather have a better `fromUuidSync` than just not utilize the cache

In3luki:
`fromUuid` adds the requested documents to the local `CompendiumCollection` so `fromCompendiumUuidSync` returns the cached document the next time the uuid is requested.
We could also 🐵 🩹 `fromUuidSync` to return the cached document instead.

stwlam:
Oh, I see, you did exactly that! Sorry, morning brain! :blush:

PR #6297, opened 2023-02-11
Fix scroll restoration including sheet item summaries

CarlosFdez:


CarlosFdez:
I'll try to see how much of it I can convert to not-jquery when I look into conditions later (and see if I can't clean up the mess some), but since this is HTML from _renderInner(), I don't know if we can really do that.

stwlam:
Yeah don't struggle too much with it. Sometimes upstream jquery makes it a losing fight.

PR #6299, opened 2023-02-11
Add RE to Magical Trickster to automate sneak attack damage on spell attacks

Far2Casual:


stwlam:
I think this wasn't implemented because @CarlosFdez still had some work to do to make it function correctly.

CarlosFdez:
For it to work, existing sneak attack needs to be refactored to store the damage dice somewhere, or the sneak attack feature needs to be the one to implement magical trickster, and it being an archetype is what decides the damage dice. There are many sources of sneak attack scattered across many archetypes that need to be resolved for this to work correctly.

stwlam (changes requested):
Needs to pull from single source of sneak attack damage rather than reimplementing sneak attack rules

stwlam:
Superseded by #6406

PR #6304, opened 2023-02-11
Expand crit spec support to NPC attacks with linked weapons

stwlam:


src/module/rules/rule-element/crit-spec.ts
```diff
@@ -68,6 +68,8 @@ class CritSpecRuleElement extends RuleElementPF2e<CritSpecRuleSchema> {
                 });
                 const bonusValue = ABP.isEnabled(this.actor)
                     ? ABP.getAttackPotency(this.actor.level)
+                    : weapon.isOfType("melee")
+                    ? weapon.linkedWeapon?.system.runes.potency ?? 0
```
CarlosFdez, line 72: This looks suspicious but also if it causes a bug it'll be easy to notice so might as well!

PR #6316, opened 2023-02-12
Don't automatically add Magic trait to weapons with tradition traits

stwlam:


CarlosFdez:
I was confused for a second but I'm assuming its to avoid double magical.

Issue #1776, opened 2021-10-11
Crafting tab formulas can't be re-ordered by dragging them


cswendrowski:
In GitLab by @danielagardeazabal on Oct 11, 2021, 24:16

Crafting tab formulas can't be re-ordered by dragging them. They are stuck in the order they were put onto the sheet.

PR #6324, opened 2023-02-12
Improve character sheet crafting tab

### LINKED ISSUE #6086: Add confirmation popup when deleting a formula
On a PC's crafting tab, the Delete Item button doesn't have a popup to confirm if you actually want to delete it. This is especially dangerous because the Craft button is right next to it.

In3luki:
- Some CSS improvements for better header alignment.
- Removed a big chunk of `jQuery` from the tabs listeners.
- All Formulas are now sortable.
- All Formula items can now be sent to chat by clicking on the item image.
- All Formulas now support expanding their item summary.
- Deleting or unpreparing a formula now opens a confirmation dialog before deletion.
- The Compendium Browser buttons now open with the maximum level of the header they were clicked in.

Before:
Before

After:
Recording 2023-03-05 at 17 12 44

Closes #6086, #1776

### DIFF
--- a/src/module/actor/character/crafting/formula.ts
+++ b/src/module/actor/character/crafting/formula.ts
@@ -82,6 +82,7 @@ export class CraftingFormula implements CraftingFormulaData {
 
 export interface CraftingFormulaData {
     uuid: ItemUUID;
+    sort?: number;
     dc?: number;
     batchSize?: number;
     deletable?: boolean;
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -15,6 +15,7 @@ import {
     ErrorPF2e,
     getActionIcon,
     groupBy,
+    htmlClosest,
     htmlQuery,
     htmlQueryAll,
     isObject,
@@ -22,13 +23,15 @@ import {
     setHasElement,
     tupleHasValue,
 } from "@util";
+import { UUIDUtils } from "@util/uuid-utils";
 import { CharacterPF2e } from ".";
 import { CreatureSheetPF2e } from "../creature/sheet";
 import { AbilityBuilderPopup } from "../sheet/popups/ability-builder";
 import { ManageAttackProficiencies } from "../sheet/popups/manage-attack-proficiencies";
 import { AutomaticBonusProgression } from "./automatic-bonus-progression";
 import { CharacterConfig } from "./config";
-import { CraftingFormula, craftItem, craftSpellConsumable } from "./crafting";
+import { CraftingFormula, CraftingFormulaData, craftItem, craftSpellConsumable } from "./crafting";
+import { PreparedFormulaData } from "./crafting/entry";
 import { CharacterProficiency, CharacterSkillData, CharacterStrike, MartialProficiencies } from "./data";
 import { CharacterSheetData, ClassDCSheetData, CraftingEntriesSheetData, FeatCategorySheetData } from "./data/sheet";
 import { PCSheetTabManager } from "./tab-manager";
@@ -164,10 +167,9 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
         const hasQuickAlchemy = !!(
             this.actor.rollOptions.all["feature:quick-alchemy"] || this.actor.rollOptions.all["feat:quick-alchemy"]
         );
-        const useQuickAlchemy = hasQuickAlchemy && flags.quickAlchemy;
 
         sheetData.crafting = {
-            noCost: flags.freeCrafting || useQuickAlchemy,
+            noCost: flags.freeCrafting,
             hasQuickAlchemy,
             knownFormulas: formulasByLevel,
             entries: await this.prepareCraftingEntries(),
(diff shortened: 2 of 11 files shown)


stwlam:
Which is the quick alchemy toggle?

In3luki:
> Which is the quick alchemy toggle?

The old one is the checkbox on top in the before image. The new buttons are the little flasks in the `Known Formulas` list that only appear if the character has quick alchemy and the item has the `alchemical` and `consumable` traits.
Clicking a button will create one item (regardless of the quantity next to it) and deduct one infused reagent.

In3luki:
Fixed `isItemUUID` reference and made the name grid a little bit wider.

stwlam:
Is there enough room to stack it over a lightning bolt? :)

https://fontawesome.com/docs/web/style/stack

src/module/actor/character/crafting/formula.ts
```diff
@@ -78,10 +78,15 @@ export class CraftingFormula implements CraftingFormulaData {
     get description(): string {
         return this.item.description;
     }
+
+    get quickAlchemy(): boolean {
```
CarlosFdez, line 82: I don't think this is the right place for it. There are ways of being able to quick alchemy certain formulas that are actually not alchemical. This is probably something that should be driven by the crafting entries instead.

https://2e.aonprd.com/Feats.aspx?ID=123

This feat makes the potions into elixers, but I suspect the transformation of potion -> elixer would be done as part of the quick crafting pipeline itself. Unless you got a different idea, such as storing data at the formula level on whether or not its an improbably elixer or not?

In3luki: Of course there has to be an exception... and they don't even get the alchemical trait. 
Problem is that there is currently no way to alter a formula permanently as they are strict references to a compendium item.
Maybe with a choice set on the feat to select potions that sets a flag on the actor which could be utilized to collect all quick alchemy items in `getData`.

I think I'll revert to the toggle while I think about it.

CarlosFdez: Being an elixir forces the alchemical trait, as per the elixir trait I think.

I don't think you have to change what you filter at least for this draft, probably only where you filter it. Perhaps a crafting entry can have a canQuickAlchemy(item) and then you allow it if it at least one matches. Or maybe the getter is fine.

Checking via the crafting entry would also handle the cases where you have advanced alchemy but not quick alchemy (herbalist).

In3luki: Crafting entries currently only include "prepared" formulas for daily crafting, but quick alchemy can be used with any known formula, so putting the check on the entry doesn't really make sense.
The`hasQuickAlchemy` check happens in the template so the button only shows up when the character has the action.

But leaving this change in will make it impossible to craft items from the feat you postetd because the formula traits can't be altered. One solution would be to have a `Known Formulas` list per crafting entry where formula traits could be added based on character feats.

CarlosFdez: You mighta misunderstood. I don't mean via the formula in crafting entry, I meant maybe a crafting entry could have a method that receives a formula that then lets you know if it can do quick alchemy with that formula (regardless of whether its been added there or not).

Crafting Entries (terrible name) are basically advanced alchemy features. You use quick alchemy with an advanced alchemy feature. I think it makes sense for them to decide if a formula can be quick alchemy'd. In fact, you can't quick alchemy at all if there are zero crafting entries.

In3luki: I see. I didn't realize crafting entries are just used for advanced alchemy. That still leaves the question how a potion formula could be flagged as eligible for quick alchemy if it's base on a compendium reference.
For now I've reverted back to the global toggle to not make those edge cases impossible.

In3luki:
Updated. Also updated title and description and added issues that will be closed by this PR.
Should be ready to merge now.

static/lang/en.json
```diff
@@ -1435,9 +1435,13 @@
             "ExpendFormula": "Expend Formula",
             "NoEligibleEntry": "No eligible crafting entry was found.",
             "QuickAddTitle": "Add this formula to one or more crafting entries",
+            "RemoveFormulaDialogQuestion": "Are you sure you want to remove the formula for: '{name}'?",
```
stwlam, line 1438:
```suggestion
            "RemoveFormulaDialogQuestion": "Are you sure you want to remove the formula for {name}?",
```

static/lang/en.json
```diff
@@ -1435,9 +1435,13 @@
...
+            "RemoveFormulaDialogTitle": "Remove Formula",
             "ToggleFreeCrafting": "Toggle Free Crafting",
             "ToggleQuickAlchemy": "Quick Alchemy",
-            "UndeletableTooltip": "This crafting formula is part of another item and cannot be individually deleted."
+            "UndeletableTooltip": "This crafting formula is part of another item and cannot be individually deleted.",
+            "UnprepareFormulaDialogQuestion": "Are you sure you want to unprepare the formula for: '{name}'?",
```
stwlam, line 1443:
```suggestion
            "UnprepareFormulaDialogQuestion": "Are you sure you want to unprepare the formula for {name}?",
```

src/module/actor/sheet/base.ts
```diff
@@ -511,23 +513,27 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
...
+                                return [new ItemProxyPF2e(formula.item.toObject(), { parent: this.actor }), true];
+                            }
+                        }
+                        throw ErrorPF2e(`Invalid UUID [${itemId}]!`);
+                    }
+                    return [this.actor.items.get(itemId, { strict: true }), false];
+                })();
+
+                if (item.isOfType("physical") && !item.isIdentified) return;
+                await item.toMessage(undefined, { create: true, data: { fromFormula } });
```
stwlam, line 534:
```suggestion
                if (!item.isOfType("physical") || item.isIdentified) {
                    await item.toMessage(undefined, { create: true, data: { fromFormula } });
                }
```

src/module/actor/character/sheet.ts
```diff
@@ -572,6 +574,156 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
         }
 
         // CRAFTING
+        const craftingTab = htmlQuery(html, ".tab.crafting");
+
+        for (const element of htmlQueryAll<HTMLLIElement>(craftingTab, "li.formula-item")) {
+            const quant = htmlQuery<HTMLInputElement>(element, "input[data-action=enter-quantity]");
```
stwlam, line 580: We can spell out `quantity` to the bitter end and be ok

In3luki:
Updated

src/module/actor/character/sheet.ts
```diff
@@ -572,6 +574,155 @@ class CharacterSheetPF2e extends CreatureSheetPF2e<CharacterPF2e> {
...
+                        const index = element.dataset.itemIndex;
+                        await craftingEntry.setFormulaQuantity(Number(index), itemUUID ?? "", newValue);
+                        return;
+                    }
+                    this.formulaQuantities[formula.uuid] = Math.max(newValue, minBatchSize);
+                    this.render();
+                });
+            }
+
+            const craftButton = htmlQuery(element, "a[data-action=craft-item]");
```
stwlam, line 627: Was this a single button or one for each formula?

In3luki: That's the 🔨 button on each formula.

PR #6327, opened 2023-02-12
Fix target of levitate spell in carbuncle to Self only

manfred7168:
Fix carbuncle (of bestiary 2): Levitate target has to be "Self only"

### DIFF
--- a/packs/data/pathfinder-bestiary-2.db/carbuncle.json
+++ b/packs/data/pathfinder-bestiary-2.db/carbuncle.json
@@ -188,7 +188,7 @@
                     "value": false
                 },
                 "target": {
-                    "value": "1 unatteded object or willing creature"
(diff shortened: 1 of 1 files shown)


stwlam:
Please state what exactly you're fixing in the PR title

manfred7168:
Changed.

Gesendet von Mail für Windows

Von: Shark that walks like a man
Gesendet: Sonntag, 12. Februar 2023 21:52
An: foundryvtt/pf2e
Cc: Manni; Author
Betreff: Re: [foundryvtt/pf2e] Data entry: Fix carbuncle (PR #6327)

Please state what exactly you're fixing in the PR title

TikaelSol:
We don't normally change the details of spells like this beyond changing the name, though maybe we should.

PR #6331, opened 2023-02-12
Fix level of advanced thoughtform

rectulo:


stwlam:
Could you combine PRs like these instead of making multiple for back-to-back one-character changes?

rectulo:
When I found one, I fix it : a "Fire and forget" thing during my work on the french translation.

PR #6339, opened 2023-02-13
Change left click effect panel functionality for persistent to roll damage

CarlosFdez:
image

### DIFF
--- a/src/module/apps/effects-panel.ts
+++ b/src/module/apps/effects-panel.ts
@@ -1,14 +1,18 @@
 import { ActorPF2e } from "@actor";
 import { AbstractEffectPF2e, EffectPF2e } from "@item";
 import { AfflictionPF2e } from "@item/affliction";
-import { PersistentDialog } from "@item/condition/persistent-damage-dialog";
 import { EffectExpiryType } from "@item/effect/data";
+import { TokenDocumentPF2e } from "@scene";
 import { htmlQuery, htmlQueryAll } from "@util";
 import { FlattenedCondition } from "../system/conditions";
 
 export class EffectsPanel extends Application {
+    private get token(): TokenDocumentPF2e | null {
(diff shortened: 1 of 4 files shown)


src/module/apps/effects-panel.ts
```diff
@@ -1,12 +1,16 @@
...
 import { htmlQuery, htmlQueryAll } from "@util";
 import { FlattenedCondition } from "../system/conditions";
 
 export class EffectsPanel extends Application {
+    private get token(): TokenDocumentPF2e | null {
+        return canvas.tokens.controlled[0]?.document ?? game.user?.character?.token ?? null;
+    }
+
     private get actor(): ActorPF2e | null {
         return canvas.tokens.controlled[0]?.actor ?? game.user?.character ?? null;
```
stwlam, line 15: I suppose this could pull from the token getter?

CarlosFdez: it could, though character.token.actor is a pretty weird roundabout retrieval.

stwlam: It's going to fail to get a token basically every time for players if they don't have anything controlled

CarlosFdez: Tested, doing this will introduce a bug for players on scenes without tokens for regular effect panel usages.

game.user.character has a value but not game.user.character.token.actor

stwlam: `Actor#token` is always either a synthetic actor or null

CarlosFdez: wait, I thought you were saying to get the actor from the token? Are you saying vice versa (token from actor)?

Vice versa should work, there would be a minor bug if an actor is a linked actor and has multiple tokens in the scene, but that's such a minor rare thing that I think I don't mind.

PR #6362, opened 2023-02-14
Tagify remaining Compendium Browser trait filters

In3luki:
Also changes the trait filter from `some` to `every` to work as expected when multiple traits are selected.


stwlam:
I think the expectation, at least by now, is that it be `every` :)

In3luki:
> I think the expectation, at least by now, is that it be `every` :)

Did you mean `some`? I can only say that I was always irritated when I selected two or more traits and it wasn't filtered down to only those.

PR #6368, opened 2023-02-14
Allow item slugs to be set from RE panel

stwlam:
Also replaces usage of `navigator.clipboard` with usage of `game.clipboard`

Screenshot from 2023-02-14 12-35-32
Screenshot from 2023-02-14 12-35-46


CarlosFdez:
I've been needing to revisit this since using navigator.clipboard caused problems. Neat.

src/module/item/sheet/base.ts
```diff
@@ -457,9 +493,11 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
...
 
     protected override async _updateObject(event: Event, formData: Record<string, unknown>): Promise<void> {
-        // Avoid setting a baseItem of an empty string
-        if (formData["system.baseItem"] === "") {
-            formData["system.baseItem"] = null;
+        // Avoid setting a baseItem or slug to an empty string
+        for (const key of ["baseItem", "slug"]) {
+            if (formData[`system.${key}`] === "") {
+                formData["system.${key}"] = null;
+            }
```
CarlosFdez, line 500: Isn't this already handled by preUpdate?

PR #6370, opened 2023-02-14
Ensure minimum value and max for PC XP

### LINKED ISSUE #6360: When current XP is a negative number it gets doubled each time the Character Sheet is open
When current XP is a negative number it gets doubled each time the Character Sheet is open.

Steps to reproduce:
1. Create a new actor so it has 0 experience
2. Type -10 (or any other negative number) into the current experience
3. The actor now has -10 experience
4. Close the character sheet
5. Open it again, the character has -20 experience. Each time you reopen it the negative experience gets doubled (so the next are -40, -80, -160 etc)

A use case for negative experience is when leveling a character before they get full 1000 exp. I do it sometimes for NPCs travelling with the party that are usually down on exp and will level up within the first hour of next session not to waste time adding them feats but also remembering how much exp they actually have

stwlam:
Closes #6360

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -2035,6 +2035,11 @@ class CharacterPF2e extends CreaturePF2e {
             }
         }
 
+        // Ensure minimum XP value and max
+        const xp = changed.system?.details?.xp ?? {};
+        if (typeof xp.value === "number") xp.value = Math.max(xp.value, 0);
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I think maximum might be necessary for overflowing. I think a lot of GMs that track XP don't allow you to level until you rest for the night, so you need to keep tracking additional XP you earn until you finally manage to level.

That said I generally do milestone leveling so I can't be certain.

PR #6379, opened 2023-02-15
Remove JQuery for focus point toggle, reuse listener for carry type menu

stwlam:


CarlosFdez:
Was wondering if this was a micro-optimization, but I guess its not that much of one, at least if nothing as changed regarding javascript's optimization

https://stackoverflow.com/a/80927

PR #6389, opened 2023-02-16
Fix condition/effect item summaries, and remove more JQuery

CarlosFdez:


CarlosFdez:
GSAP is an agnostic animation lib that foundry comes with a copy of, that will allow us a bit of shiny without the JQuery.

PR #6395, opened 2023-02-16
Flatten unnessary nesting of if block

stwlam:


CarlosFdez:
Hah, it used to be nested for options.instant, but I forgot to give it a second go when I managed to get the instant branch down to a single line.

PR #6404, opened 2023-02-16
Support building action subtitle from actor and target

nikolaj-a:
Allow an action to dynamically build the subtitle of the action chat card - which will usually be the check type, like "Mercantile Lore Check" - from the actor performing the action and the target of the action. This will be helpful for the Discover and Influence actions that might often rely on lore skills where we need the convert the slug to the translated name from either of those actors.

### DIFF
--- a/src/module/system/action-macros/helpers.ts
+++ b/src/module/system/action-macros/helpers.ts
@@ -98,14 +98,6 @@ export class ActionMacroHelpers {
         }
 
         for (const actor of rollers) {
-            let title = "";
-            if (options.actionGlyph) {
-                title += `<span class="pf2-icon">${options.actionGlyph}</span> `;
-            }
-            title += `<b>${game.i18n.localize(options.title)}</b>`;
-            title += ` <p class="compact-text">(${game.i18n.localize(options.subtitle)})</p>`;
-            const content = (await options.content?.(title)) ?? title;
-
             const targetOptions = targetActor?.getSelfRollOptions("target") ?? [];
             const selfToken = actor.getActiveTokens(false, true).shift();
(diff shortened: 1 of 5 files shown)


src/module/system/action-macros/helpers.ts
```diff
@@ -120,6 +112,18 @@ export class ActionMacroHelpers {
...
+                title += `<span class="pf2-icon">${options.actionGlyph}</span> `;
+            }
+            title += `<b>${game.i18n.localize(options.title)}</b>`;
+            const subtitle =
+                typeof options.subtitle === "function"
+                    ? options.subtitle?.({ actor: selfActor, target: targetActor })
+                    : options.subtitle;
+            title += ` <p class="compact-text">(${game.i18n.localize(subtitle)})</p>`;
+            const content = (await options.content?.(title)) ?? title;
+
```
stwlam, line 126: - Prefer IIFE over `let`
- Prefer `<strong>` over `<b>`
- You touched it, so now it has to go in a template or at least be cobbled together via DOM API :slightly_smiling_face:

stwlam (changes requested):

stwlam (approved):

static/templates/chat/action/header.hbs
```diff
@@ -0,0 +1,9 @@
+<h4 class="action">
```
stwlam, line 1: Is the `h4` styled such that a `strong` element will still be bolder?

nikolaj-a: Yes - bolder is better! Actually I just kept to what we have now, since I didn't want to go hunt down CSS stuff as well.

PR #6410, opened 2023-02-17
Brush up several high level druid feats

TikaelSol:
Clearly we need to do a pass of the core classes...


packs/data/feats.db/primal-aegis.json
```diff
@@ -10,7 +10,7 @@
             "value": null
         },
         "description": {
-            "value": "<p>You surround yourself with a thick field of protective primal energy. You and allies within 30 feet of you gain resistance equal to your Wisdom modifier to acid, cold, electricity, fire, negative, and positive damage.</p>"
```
stwlam, line 13: Aura? :)

TikaelSol: Done, now just make `@item.origin` work on auras

PR #6414, opened 2023-02-17
Fix regression with character sheet action headers

CarlosFdez:


CarlosFdez:
I accidentally made them all red and bolded all the time which was apparently wrong. Also made the animation a bit nicer for strikes/actions by moving the padding/'margin to the main style there.

PR #6415, opened 2023-02-17
Add action roll notes informing swashbucklers when they gain panache

stwlam:
image


denDAY04:
Is there a way to get this change to automatically pick up on actors that were made prior to the release of this feature? I noticed my swashbuckler wasn't getting this message, and after some extensive digging I found out that the actor had no rollNotes in their synthetic class property. 

The tedious fix is to re-added the class to the character sheet the proper rollNotes were added. But I wonder if there's a better/smarter/automated way?

PR #6423, opened 2023-02-17
standardize hardness data entry

### LINKED ISSUE #6421: data entry bug: Spiral Centurion has Hardness in the wrong field
image

TMunsie:
Standardizes hardness into hardness field instead of as a HP note.

Closes #6421

### DIFF
--- a/packs/data/menace-under-otari-bestiary.db/animated-armor-bb.json
+++ b/packs/data/menace-under-otari-bestiary.db/animated-armor-bb.json
@@ -259,7 +259,7 @@
                 "value": ""
             },
             "hardness": {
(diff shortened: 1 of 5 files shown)


packs/data/menace-under-otari-bestiary.db/animated-armor-bb.json
```diff
@@ -259,7 +259,7 @@
                 "value": ""
             },
             "hardness": {
-                "value": null
+                "value": 9
```
stwlam, line 262: We should make this line up with hazards and vehicles (`attributes.hardness`, no `value` object)

PR #6436, opened 2023-02-18
Add `and`,`or` and `not` conditions to the Compendium Browser trait filter

In3luki:
Recording 2023-02-18 at 18 06 57

### DIFF
--- a/src/module/apps/compendium-browser/tabs/action.ts
+++ b/src/module/apps/compendium-browser/tabs/action.ts
@@ -75,10 +75,8 @@ export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
         const { checkboxes, multiselects } = this.filterData;
 
         // Traits
-        const selectedTraits = multiselects.traits.selected.map((s) => s.value);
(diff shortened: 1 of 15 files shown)


stwlam:
Maybe radios with "all of above" and "any of above"? For the "not" bit, how about a :no_entry_sign: next to the :x:?

In3luki:
I tried a 🚫 element at first but I can't get it into the tag without breaking the styling of the whole thing. 😫

In3luki:
image

stwlam:
> image

I think we can eschew the "Condition" part.

In3luki:
image

src/module/apps/compendium-browser/tabs/action.ts
```diff
@@ -98,6 +95,7 @@ export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
             },
             multiselects: {
                 traits: {
+                    condition: "and",
```
stwlam, line 98:
```suggestion
                    conjunction: "and",
```

In3luki: Done

PR #6443, opened 2023-02-18
Fix bulk of Mithral Waffle Iron

Drental:
https://2e.aonprd.com/Equipment.aspx?ID=275 Mithral Objects have their bulk reduced by 1 and go to Light bulk from 1 Bulk


stwlam:
Thank you for this crucial update.

stwlam (approved):

PR #6445, opened 2023-02-19
Rename Phuthi to a generic statblock name

SpartanCPA:


TikaelSol:
Is there a reason to make this a separate actor instead of generalize Phuthi?

SpartanCPA:
@TikaelSol merged and renamed

PR #6463, opened 2023-02-20
Remove two unicodes in GM screen

AlphaStarguide:
As title, these unnecessary unicodes would make trouble when localizing system with weblate.


stwlam:
The entire file is unicode. Can you say which characters were removed?

AlphaStarguide:
OK, I removed these control codes.

the monster can reactivate it by spending the normal spellcasting actions the spell `\u001` crequires.</p>


is the level of the monster in `\u001f` icting the disease. The disease follows the rules for afflictions.</p>

stwlam:
I believe the second one should have "inflicting" as the word.

AlphaStarguide:
!!!You are right. Fixed.

Issue #3957, opened 2022-09-13
Request: Save filter selections in the 'Source' section of the compendium browser


mistamichal:
It would be good if the GM and/or players could save the selections so that we only have to see items, spells, feats, etc. from the sourcebooks we are using for the game and not have to re-select them each time we play. It would also avoid any cases of: for example -  a player creating a character and choosing a feat that's not in an allowed rulebook for the campaign.

PR #6476, opened 2023-02-20
Allow GM to select allowed sources for the compendium browser

glont:
This adds a new world setting, to save, which entries should to be displayed in the compendium browser by specifying wanted sources. For the GM a new section in the "Load Packs"/Settings page is added to set this up. By default this behaves as before, letting all entries show up.

In the bellow screenshots it shows the equipment tab before setting up the new settings, setting up the settings and then the filtered equipment list.

Equipment Before:

1_equipment_before

New Settings:

2_settings_example

Equipment After, note the changed sources filter. Also the Alchemical Crossbow is not showing up anymore:

3_equipment_after

This mostly works, but there are at least the following issues:

- [x] ~Settings UI is pretty bad, though it only concerns GMs~. It is still not great, but it should be ok. I would expect a GM would not change this setting all that often, after they have it set up the first time. Of course they may want to change it in case new content came out.
    - [x] ~Checking 200 Checkboxes for if you want the source or not is annoying.~ Sources can be filtered via a text box and toggled on and off in bulk, this should mostly solve this.
    - [x] ~After changing the sources, the GM will see the sources as slugs of already loaded documents until they reload the application.~ This is fixed
- [ ] Now the reason some information is not shown can be that either the pack is not loaded OR the source is not allowed. Not sure how to avoid this easily. It still makes sense to disallow whole packs, to improve loading.
- [x] New Sources are automatically marked allowed on settings screen by default, there is a setting to change this
    - [x] ~this could be an additional setting~. Added additional setting
    - [x] ~GM still needs to got to the settings page and click save, so that new sources become active.~ This is fixed
- [x] ~Items with no sources are currently displayed, this could be an additional setting~. Added additional setting
- [x] ~Using translated content adds additional checkboxes to the settings UI, because now there are new sources. This is mostly fine, but GMs may need to switch languages to set this up. I'm not sure if there are many groups using different languages, but who knows.~ The sources settings can be deleted, which should remove non existent sources from the settings UI.
- [x] Nonexistent sources are not deleted. This makes it possible to use multiple languages, ~but also means the list will get longer, the more is translated.~ The GM can now delete / reset the sources settings, which will go back to the defaults of allowing all sources, including known sources, new / unknown sources and empty sources.
- [x] ~Loading the settings page is slower, because all sources need to be found. I did only test this on my local machine. It mostly did not feel worse, but that could be different, depending on the situation (server performance, internet connection, client performance, ...)~ Tested on a VM for both client and server, where foundry itself is barely usable, and the impact is negligible.
- [x] ~Normal loading did seem to be not really impacted, but again I did not test this excessively.~ Tested on a VM for both client and server, where foundry itself is barely usable, and the impact is negligible.

It may be possible to add the new world setting to the existing CompendiumBrowserSettings. I'm not sure if that is the best idea, it would at least require a new migration and It may make sense to split the configuration up for ease of use / better ui/ux, later on anyways. I'm also not super great with creating Types in TypeScript (yet) and did not want to do to many changes on the existing code if I can avoid that.

See also: #3957

### DIFF
--- a/src/module/apps/compendium-browser/index.ts
+++ b/src/module/apps/compendium-browser/index.ts
@@ -2,12 +2,12 @@ import { KitPF2e, PhysicalItemPF2e } from "@item";
 import { ActionTrait } from "@item/action/index.ts";
 import { ActionType } from "@item/data/base.ts";
 import { BaseSpellcastingEntry } from "@item/spellcasting-entry/index.ts";
-import { ErrorPF2e, htmlQueryAll, isObject, localizer, objectHasKey } from "@util";
+import { ErrorPF2e, htmlQuery, htmlQueryAll, isBlank, isObject, localizer, objectHasKey, sluggify } from "@util";
 import { getSelectedOrOwnActors } from "@util/token-actor-utils.ts";
 import { UserPF2e } from "@module/user/document.ts";
 import Tagify from "@yaireo/tagify";
 import noUiSlider from "nouislider";
-import { BrowserTabs, PackInfo, SortDirection, TabData, TabName } from "./data.ts";
+import { BrowserTabs, PackInfo, SortDirection, SourceInfo, TabData, TabName } from "./data.ts";
 import { Progress } from "./progress.ts";
 import * as browserTabs from "./tabs/index.ts";
 import {
@@ -30,6 +30,13 @@ class PackLoader {
         Item: Record<string, { pack: CompendiumCollection; index: CompendiumIndex } | undefined>;
     } = { Actor: {}, Item: {} };
 
+    loadedSources: string[] = [];
+    sourcesSettings: CompendiumBrowserSources;
+
+    constructor() {
+        this.sourcesSettings = game.settings.get("pf2e", "compendiumBrowserSources");
+    }
+
     async *loadPacks(
         documentType: "Actor" | "Item",
         packs: string[],
@@ -37,6 +44,7 @@ class PackLoader {
     ): AsyncGenerator<{ pack: CompendiumCollection<CompendiumDocument>; index: CompendiumIndex }, void, unknown> {
         this.loadedPacks[documentType] ??= {};
         const localize = localizer("PF2E.CompendiumBrowser.ProgressBar");
+        const sources = this.#getSources();
 
         const progress = new Progress({ steps: packs.length });
         for (const packId of packs) {
@@ -56,8 +64,9 @@ class PackLoader {
                     const firstResult: Partial<CompendiumIndexData> = index.contents.at(0) ?? {};
                     // Every result should have the "system" property otherwise the indexFields were wrong for that pack
                     if (firstResult.system) {
-                        this.setModuleArt(packId, index);
-                        data = { pack, index };
+                        const filteredIndex = this.#createFilteredIndex(index, sources);
+                        this.#setModuleArt(packId, filteredIndex);
+                        data = { pack, index: filteredIndex };
                         this.loadedPacks[documentType][packId] = data;
                     } else {
                         ui.notifications.warn(
@@ -76,13 +85,129 @@ class PackLoader {
     }
 
     /** Set art provided by a module if any is available */
-    private setModuleArt(packName: string, index: CompendiumIndex): void {
+    #setModuleArt(packName: string, index: CompendiumIndex): void {
         if (!packName.startsWith("pf2e.")) return;
         for (const record of index) {
             const uuid: CompendiumUUID = `Compendium.${packName}.${record._id}`;
             const actorArt = game.pf2e.system.moduleArt.map.get(uuid)?.img;
             record.img = actorArt ?? record.img;
         }
+    }
+
+    #getSources(): Set<string> {
+        const sources = new Set<string>();
+        for (const source of Object.values(this.sourcesSettings.sources)) {
+            if (source?.load) {
+                sources.add(source.name);
+            }
+        }
+        return sources;
+    }
+
+    #createFilteredIndex(index: CompendiumIndex, sources: Set<string>): CompendiumIndex {
+        if (sources.size === 0) {
+            // Make sure everything works as before as long as the settings are not yet defined
+            return index;
+        }
+
+        if (game.user.isGM && this.sourcesSettings.ignoreAsGM) {
+            return index;
+        }
+
+        const filteredIndex: CompendiumIndex = new Collection<CompendiumIndexData>();
+        const knownSources = Object.values(this.sourcesSettings.sources).map((value) => value?.name);
+
+        for (const document of index) {
+            const source = this.#getSourceFromDocument(document);
+            const blank = isBlank(source);
+
+            if (
+                (blank && this.sourcesSettings.showEmptySources) ||
+                sources.has(source) ||
+                (this.sourcesSettings.showUnknownSources && !blank && !knownSources.includes(source))
+            ) {
+                filteredIndex.set(document._id, document);
+            }
+        }
+        return filteredIndex;
+    }
+
+    async updateSources(packs: string[]): Promise<void> {
+        await this.#loadSources(packs);
+
+        for (const source of this.loadedSources) {
+            const slug = sluggify(source);
+            if (this.sourcesSettings.sources[slug] === undefined) {
+                this.sourcesSettings.sources[slug] = {
+                    load: this.sourcesSettings.showUnknownSources,
+                    name: source,
+                };
+            }
+        }
+
+        // Make sure it can be easily displayed sorted
+        this.sourcesSettings.sources = Object.fromEntries(
+            Object.entries(this.sourcesSettings.sources).sort((a, b) => a[0].localeCompare(b[0], game.i18n.lang))
+        );
+    }
+
+    async #loadSources(packs: string[]): Promise<void> {
+        const localize = localizer("PF2E.CompendiumBrowser.ProgressBar");
+        const progress = new Progress({ steps: packs.length });
+
+        const loadedSources = new Set<string>();
+        const indexFields = ["system.details.source.value", "system.source.value"];
+        const knownDocumentTypes = ["Actor", "Item"];
+
+        for (const packId of packs) {
+            const pack = game.packs.get(packId);
+            if (!pack || !knownDocumentTypes.includes(pack.documentName)) {
+                progress.advance("");
+                continue;
+            }
+            progress.advance(localize("LoadingPack", { pack: pack?.metadata.label ?? "" }));
+            const index = await pack.getIndex({ fields: indexFields });
+
+            for (const element of index) {
+                const source = this.#getSourceFromDocument(element);
+                if (source && source !== "") {
+                    loadedSources.add(source);
+                }
+            }
+        }
+
+        progress.close(localize("LoadingComplete"));
+        const loadedSourcesArray = Array.from(loadedSources).sort();
+        this.loadedSources = loadedSourcesArray;
+    }
+
+    #getSourceFromDocument(document: CompendiumIndexData): string {
+        // There are two possible fields where the source can be, check them in order
+        if (document.system?.details?.source?.value) {
+            return document.system.details.source.value;
+        }
+
+        if (document.system?.source?.value) {
+            return document.system.source.value;
+        }
+
+        return "";
+    }
+
+    reset(): void {
+        this.loadedPacks = { Actor: {}, Item: {} };
+        this.loadedSources = [];
+    }
+
+    async hardReset(packs: string[]): Promise<void> {
+        this.reset();
+        this.sourcesSettings = {
+            ignoreAsGM: true,
+            showEmptySources: true,
+            showUnknownSources: true,
+            sources: {},
+        };
+        await this.updateSources(packs);
     }
 }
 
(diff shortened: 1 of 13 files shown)


stwlam:
I think a better approach may be to separate source filtering from source inclusion. It may still be desirable to filter sources on the fly, but a GM may want to blacklist some sources. This could be done in what is currently the "packs" tab.

glont:
@stwlam

I hope I did understand what you mean.

I will look into that, it seems to me that I would need to filter the data of what is loaded in [loadPacks](https://github.com/foundryvtt/pf2e/blob/d30ec6c1e41a98d1813df7d96851d0f32424c19d/src/module/apps/compendium-browser/index.ts#L57), because one compendium can have entries from multiple sources. 

I would need to get a list of all the existing sources from all compendiums, and let the DM check or uncheck those on the "settings"/packs tab.

tehguitarist:
One way to improve UI/UX might be to add subheadings. PFS adventures, AP Adventures, Core Rulebooks, Standalone Adventures etc, which could be entirely toggled on and off. That way a user could enable JUST core rulebooks and the adventure they're running (if any) quickly.

glont:
That could work, I will look into it, especially how to make it so that translated sources will be added to the right groups. Maybe that data could be provided by the system and overridden by the translation module. And in the worst case it would just not be grouped.

glont:
I did not find an easy way to group the sources. Instead I added a text input, which can be used to filter the list of sources and with a checkbox beside the input all visible sources can be toggled on or off.

glont:
@stwlam I implemented it, so that a list of sources is saved, where for each source it knows if it should be included or not, similar to how the Packs already work. We may need to change the "Load Packs" tab to settings, and I did only limited test about how this impacts loading times for GMs, but from those test it did not seem to matter.

I'm not sure about the code quality, as I said I'm pretty new to TypeScript.

TikaelSol:
You may want to submit a small data entry PR so you have one merged PR to the system and pipelines will auto run for your code. Just adding an icon to some piece of equipment from Treasure Vault would do.

src/module/apps/compendium-browser/index.ts
```diff
@@ -81,6 +91,131 @@ class PackLoader {
             record.img = actorArt ?? record.img;
         }
     }
+
+    #getAllowedSources() {
```
stwlam, line 95: Add return types to all these signatures

glont: I did that, rebased and also fixed a smaller issue I found while testing.

glont:
I think this is at a pretty good point now, I tested it today a bit more and did not see anything weird anymore and it seemed to work well, but that may very well be tunnel vision. If there are still things I should clean-up, like adding return types, please let me know.

The last two open Tasks in the description of this pull request, are things I wanted to mention, because they could create some confusion. There is maybe a need for some documentation / help text in the settings page. Besides that, what should be the next steps?

glont:
Fixed merge conflicts and force-pushed again. I still think this is ready to be merged, but if there are still things to fix, please let me know.

FYI: @stwlam

src/module/apps/compendium-browser/index.ts
```diff
@@ -331,14 +478,82 @@ class CompendiumBrowser extends Application {
...
+                const toggleElement = form.querySelector("input[data-action='setting-sources-toggle-visible']");
+                toggleElement?.addEventListener("click", (event) => {
+                    const value = (event.currentTarget as HTMLInputElement)?.checked;
+                    const allElements = form.querySelectorAll(
+                        "label[data-element='setting-source']:not(.hidden) input[type='checkbox']"
+                    );
+                    for (const element of allElements as NodeListOf<HTMLInputElement>) {
+                        element.checked = value;
+                    }
+                });
```
In3luki, line 540: This needs to be improved but I'm running out of time right now. Some pointers:
- `toggleElement` should be at the top so it can be reused in the `sourceSearch` listener instead of redefining it.
- Remove the single quotes from the selectors, they are not needed.
- Use `querySelector<HTMLInputElement>` (or better `htmlQuery<HTMLInputElement>`) instead of asserting the type.
- Try to avoid using `not` in the css selectors. The whole thing can probably be done in a single loop by checking `element.classList.has("hidden")` and checking `element.dataset.name` for the search value.

I will look at the rest of the PR in a few hours.

src/module/apps/compendium-browser/index.ts
```diff
@@ -70,14 +80,142 @@ class PackLoader {
...
+
+        if (document.system?.source?.value) {
+            return document.system.source.value;
+        }
+
+        return "";
+    }
+
+    #setDefaults(): void {
+        if (!objectHasKey(this.sources, "allowed")) {
```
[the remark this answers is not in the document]
glont: You are right, I will add the default where it is needed and remove this method fully. I think I was just stupid and did not realize that I could give a real Object in the `game.settings.register` method.

src/module/apps/compendium-browser/index.ts
```diff
@@ -70,14 +80,142 @@ class PackLoader {
...
         for (const record of index) {
             const uuid: CompendiumUUID = `Compendium.${packName}.${record._id}`;
             const actorArt = game.pf2e.system.moduleArt.map.get(uuid)?.img;
             record.img = actorArt ?? record.img;
         }
     }
+
+    #getAllowedSources(): Set<string> {
+        const sources = new Set<string>();
+        for (const [_key, source] of Object.entries(this.sources.allowed)) {
```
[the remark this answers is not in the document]
glont: Fixed that

src/module/apps/compendium-browser/index.ts
```diff
@@ -70,14 +80,142 @@ class PackLoader {
...
+                this.sources.allowed[slug] = {
+                    load: this.sources.allowUnknownSources,
+                    name: source,
+                };
+            }
+        }
+
+        // Make sure it can be easily displayed sorted
+        this.sources.allowed = Object.fromEntries(
+            Object.entries(this.sources.allowed).sort((a, b) => a[0].localeCompare(b[0]))
```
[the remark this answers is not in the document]
glont: Fixed

src/module/apps/compendium-browser/index.ts
```diff
@@ -70,14 +80,142 @@ class PackLoader {
...
+
+        return "";
+    }
+
+    #setDefaults(): void {
+        if (!objectHasKey(this.sources, "allowed")) {
+            this.sources.allowed = {};
+        }
+
+        if (!objectHasKey(this.sources, "allowEmptySources")) {
```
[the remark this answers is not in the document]
glont: As said in the previous suggestion, this method was removed completely.

src/module/apps/compendium-browser/index.ts
glont: All of this makes sense, will work on this. I think I was kind of tunnel visioned after looking into the css selector for the checking if the data element has the value.

src/module/apps/compendium-browser/index.ts
```diff
@@ -331,14 +466,70 @@ class CompendiumBrowser extends Application {
...
+                const sourceSettings = htmlQueryAll<HTMLElement>(form, "label[data-element=setting-source]");
+
+                sourceSearch?.addEventListener("input", () => {
+                    const value = sourceSearch.value?.trim().toLocaleLowerCase(game.i18n.lang);
+
+                    for (const element of sourceSettings) {
+                        const shouldBeHidden =
+                            !isBlank(value) &&
+                            !isBlank(element.dataset.name) &&
+                            element.dataset.name.toLocaleLowerCase(game.i18n.lang).indexOf(value) === -1;
```
In3luki, line 500:
```suggestion
                    for (const element of sourceSettings) {
                        const name = element.dataset.name?.toLocaleLowerCase(game.i18n.lang);
                        const shouldBeHidden = !isBlank(value) && !isBlank(name) && !name.includes(value);
```
Other than that it looks good to me now.

glont: This is updated.

glont:
Rebased & Force Pushed to fix merge conflicts.

glont:
Rebased & Force Pushed to fix merge conflicts again.

@stwlam: Could I please get some feedback? At least a "we do not want this", or similar would be good to know.

static/lang/en.json
```diff
@@ -2653,6 +2653,15 @@
...
             },
+            "CompendiumBrowserSources": {
+                "AllowEmptySources": "Allow entries with empty sources to be displayed.",
+                "AllowUnknownSources": "Allow entries with unknown sources to be displayed.",
+                "DeleteAllInfo": "This can not be undone, but you may reconfigure this. The settings will be set to allow all sources, which include known sources, empty sources and unknown sources.",
+                "DeleteAllQuestion": "Are you sure you want to delete the allowed sources settings?",
+                "DeleteAllTitle": "Delete allowed sources settings",
+                "Hint": "Settings to exclude sources from loading",
+                "Name": "Compendium Browser Allowed Sources"
+            },
```
stwlam, line 2664: To make it clear that this controls what players can see, can the language here be changed to make that explicit? E.g., "Player Access" instead of "Compendium Browser Allowed Sources"

glont: Currently it also filters for the GM, so "Player Access" may not be perfect. I will think about a better name.

glont: @stwlam I did two things:

1. I added a setting to allow GMs to ignore filtering, which is also now the default.
2. I changed and added new strings on the UI, see the updated screenshot in description.

In the end I used "Player Access", because I did not find a better name for this, but I just used it for the list of sources.

stwlam: It's fine to not treat this as a permission-related feature for players, but the language in the labels shouldn't include terms  like "allowed" or "permitted" in that case.

glont: @stwlam I changed the labels, I hope this is better. I also did some refactoring of the code to remove "allowed" in variable names, to avoid confusion.

Also I updated / added screenshots to show an example for the equipment screen, before and after applying the settings for the filtering + the new settings section.

glont:
Rebased and fixed conflicts, also squashed commits too make rebasing less tedious.

static/lang/en.json
```diff
@@ -2679,6 +2679,17 @@
...
             },
+            "CompendiumBrowserSources": {
+                "DeleteAllInfo": "This can not be undone, but you may reconfigure this. The settings will be set to allow all sources, which include known sources, empty sources and unknown sources.",
+                "DeleteAllQuestion": "Are you sure you want to delete the filter sources settings?",
+                "DeleteAllTitle": "Delete filter sources settings",
+                "Hint": "Settings to display only entries with specified sources in the compendium browser.",
+                "IgnoreAsGM": "Let GMs ignore filtering and see all entries.",
+                "Name": "Compendium Browser Specify Sources",
+                "ShowEmptySources": "Do not filter entries with empty sources.",
+                "ShowUnknownSources": "Do not filter entries with unknown sources.",
```
stwlam, line 2690: What is an unknown source?

glont: It is a source, which was not yet in the system when the settings where created, so it is missing in the list of sources. For Example:

- There are two sources s1 and s2 in the system at the moment. 
- We set it up that s1 is loaded and s2 is not loaded. 
- In an update s3 is introduced to the system.
 
In this example s3 is an unknown source, we do not have yet made a decision if s3 specifically should be loaded or not.

Maybe new sources would be better?

static/lang/en.json
```diff
@@ -2679,6 +2679,17 @@
                 "Hint": "Settings to exclude packs from loading",
                 "Name": "Compendium Browser Packs"
             },
+            "CompendiumBrowserSources": {
+                "DeleteAllInfo": "This can not be undone, but you may reconfigure this. The settings will be set to allow all sources, which include known sources, empty sources and unknown sources.",
+                "DeleteAllQuestion": "Are you sure you want to delete the filter sources settings?",
+                "DeleteAllTitle": "Delete filter sources settings",
+                "Hint": "Settings to display only entries with specified sources in the compendium browser.",
+                "IgnoreAsGM": "Let GMs ignore filtering and see all entries.",
+                "Name": "Compendium Browser Specify Sources",
```
stwlam, line 2688:
```suggestion
                "Name": "Included Sources",
```

static/lang/en.json
```diff
@@ -2679,6 +2679,17 @@
                 "Hint": "Settings to exclude packs from loading",
                 "Name": "Compendium Browser Packs"
             },
+            "CompendiumBrowserSources": {
+                "DeleteAllInfo": "This can not be undone, but you may reconfigure this. The settings will be set to allow all sources, which include known sources, empty sources and unknown sources.",
+                "DeleteAllQuestion": "Are you sure you want to delete the filter sources settings?",
+                "DeleteAllTitle": "Delete filter sources settings",
```
stwlam, line 2685:
```suggestion
                "DeleteAllQuestion": "Are you sure you want to reset all settings for included sources?",
                "DeleteAllTitle": "Reset Settings",
```

static/lang/en.json
```diff
@@ -2679,6 +2679,17 @@
                 "Hint": "Settings to exclude packs from loading",
                 "Name": "Compendium Browser Packs"
             },
+            "CompendiumBrowserSources": {
+                "DeleteAllInfo": "This can not be undone, but you may reconfigure this. The settings will be set to allow all sources, which include known sources, empty sources and unknown sources.",
```
stwlam, line 2683:
```suggestion
                "DeleteAllInfo": "All compendium documents will become browseable, including those with empty sources.",
```

PR #6534, opened 2023-02-22
Fix ganzi martial artist abilities

MSAbaddon:


packs/data/pathfinder-bestiary-3.db/ganzi-martial-artist.json
```diff
@@ -1899,50 +1899,11 @@
             },
             "type": "action"
         },
-        {
-            "_id": "o5w3DGu6fR9R74V6",
-            "img": "systems/pf2e/icons/actions/Passive.webp",
-            "name": "Variable Resistance",
```
TikaelSol, line 1905: Shouldn't they have this? Or do they have multiple?

MSAbaddon: There is a second ability "Ganzi resistances" that also includes an effect.

PR #6570, opened 2023-02-24
Add NPCs for PFS #4-10

SpartanCPA:


SpartanCPA:
`Compulsion` on Fascinating Lecture is not a real trait. The level 1-2 hazards have the `Aberration` trait, even though they are Hazards

PR #6571, opened 2023-02-24
Fix inclusion of full PC weapon-damage selectors in roll inspector

stwlam:


CarlosFdez:
Funny enough the name made me think it was the opposite, that you were trying to fix it from including the full list.

PR #6574, opened 2023-02-24
Add rule elements to NPCs with Fast Healing

SpartanCPA:


TikaelSol:
I think Supe wanted to add a UI field for this instead but just living on the Fast Healing item is easier and makes sense right now. It would also make migration relatively easy since it would be structured data.

packs/data/abomination-vaults-bestiary.db/volluk-azrinae.json
```diff
@@ -3323,7 +3323,13 @@
                 "requirements": {
                     "value": ""
                 },
-                "rules": [],
+                "rules": [
+                    {
+                        "key": "FastHealing",
+                        "type": "fast-healing",
```
TikaelSol, line 3329:
```suggestion
```
We should be able to leave the type out, since it defaults to fast healing instead of regeneration.

CarlosFdez:
> I think Supe wanted to add a UI field for this instead but just living on the Fast Healing item is easier and makes sense right now. It would also make migration relatively easy since it would be structured data.

I did want this, and to auto-generate a rule element, but since then Stwlam has become adamant on rule elements not being auto-generated. I think the most we can do is expose existing fast healing / regeneration in the UI but not be created via there.

TikaelSol:
So have the presence of the fast healing or regeneration features create a field that could be edited? Then they would just have some actor data reference for the value to pick up that field? That would be pretty neat actually.

CarlosFdez:
I wasn't thinking of it being editable (just visual), but theoretically it could be, at least if we had a way of handling resolvables (and brackets, if it supports brackets). For non-brackets I have a few ideas but I suspect brackets to be impossible.

stwlam:
> I did want this, and to auto-generate a rule element, but currently stwlam is adamant on rule elements not being auto-generated. I think the most we can do is expose existing fast healing / regeneration in the UI but not be created via there.

I don't really have a strong opinion on rule elements being autogenerated: they just need to be on items.

packs/data/abomination-vaults-bestiary.db/volluk-azrinae.json
SpartanCPA: It was added by the system itself, but sure, I can take it out.

PR #6582, opened 2023-02-24
Add automatic calculation of price and level in armor based on runes and materials and the ability to make specific magic armor

Zullock:
Changed the armor to be more in line with the weapons.

Now the armor automatically calculates the price and level based on runes and materials.

Added checkbox to make the armor a Specific Magic Armor.

Modified the existing specific magic armor to be counted as such.

Added missing armor runes and prepared the data structure to eventually make some automations (like getting resistance from the energy-resistance runes or only show the runes appropriate for the armor in question).

image


stwlam:
Thanks for the submission, but we're taking runes in a different direction rather than copy-pasting code from weapons.

Zullock:
I understand that it's not the long term plan to have it this way, but right now we are missing the runes from TV in armors, the data entry for specific magic armor and I think there is some value to having this meanwhile. Unless we are talking that this is going to be changed in a very short period of time, of course.

stwlam:
It's planned for next major release. Are you on Discord at all? It'd be good to coordinate on major feature expansions like this.

PR #6584, opened 2023-02-24
Prevent effect sheet from submitting on change of badge-type select

stwlam:


CarlosFdez:
It doesn't even have a name property, why in the hell is it submitting? Funny enough I saw no evidence of it until I looked more closely and saw the <form> element flash.

Perhaps this is a thing we should do in the base actor sheet where all selects that don't have a name property should prevent the submission?

stwlam:
> It doesn't even have a name property, why in the hell is it submitting? Funny enough I saw no evidence of it until I looked more closely and saw the `<form>` element flash.
> 
> Perhaps this is a thing we should do in the base actor sheet where all selects that don't have a name property should prevent the submission?

I don't remember these submitting, but maybe we normally have a handler for unnamed select elements.

stwlam:
> Perhaps this is a thing we should do in the base actor sheet where all selects that don't have a name property should prevent the submission?

I'd be worried about that spawning other unintended consequences. Maybe for 4.9?

PR #6608, opened 2023-02-25
Highlight grid area of `MeasuredTemplatePF2e` previews

In3luki:
I'm pretty sure this used to be a thing so it's either a bugfix or an enhancement.

Before:
before

After:
after


stwlam:
What's with the darker lines in the bottom half?

In3luki:
No idea. This is the wall for reference:

image

In3luki:
Also happens on master with the normal measured template:

image

PR #6613, opened 2023-02-25
Limit disabling of REs that include ChoiceSets with no selection

stwlam:


CarlosFdez:
This is tough to understand from the PR name. Glancing at the code, is it that you're limiting to those that pass the predicate? Aka skipping disabling from choice sets that fail predication?

PR #6614, opened 2023-02-25
Show repeating ammo icon again

CarlosFdez:
image


stwlam:
Wow I never knew that was a thing.

CarlosFdez:
For capacity I wanna do multiple bullet icons with a crosshair selection and mouseover tooltips. But I never found the time. Perhaps with Rotary Bow I'll be pressured enough to do it.

PR #6624, opened 2023-02-26
Add return types to `ActorSpellcasting` getters & methods

stwlam:


CarlosFdez:
The return type of recharge() isn't quite correct (if we manage to get spellcasting entries to REs and actor data that will be very wrong) but we can cross that bridge when we get there.

PR #6631, opened 2023-02-26
Add confirmation popup when deleting a formula

glont:
When deleting other options like items or spells a confirmation popup is already used, this updates the crafting tab, so formulas are not accidentally deleted.

See also: #6086

### DIFF
--- a/static/templates/actors/character/tabs/crafting.hbs
+++ b/static/templates/actors/character/tabs/crafting.hbs
@@ -44,7 +44,7 @@
                     </li>
                     <!-- Add formula items for each formula level -->
                     {{#each section as |craftedItem i|}}
-                        <li class="item formula-item" data-formula-lvl="{{lvl}}" data-item-id="{{craftedItem.uuid}}" data-is-formula="true">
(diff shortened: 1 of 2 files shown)


static/templates/actors/character/tabs/crafting.hbs
```diff
@@ -44,7 +44,7 @@
                     </li>
                     <!-- Add formula items for each formula level -->
                     {{#each section as |craftedItem i|}}
-                        <li class="item formula-item" data-formula-lvl="{{lvl}}" data-item-id="{{craftedItem.uuid}}" data-is-formula="true">
+                        <li class="item formula-item" data-formula-lvl="{{lvl}}" data-item-id="{{craftedItem.uuid}}" data-item-name="{{craftedItem.name}}" data-is-formula="true">
```
stwlam, line 47: Can this depend entirely on the UUID?

glont: @stwlam I think there is a bigger pull request open for work on the crafting tab, which is #6324 and there as far as I can tell it also adds the confirmation popup, so I would just close this smaller pull request and let the larger work fix this up. Also adding a comment in the issue mentioned above, to inform that this is worked on.

PR #6636, opened 2023-02-27
Replace ritual spellcasting entries with in-memory ones

stwlam:


static/templates/actors/spellcasting-spell-list.hbs
```diff
@@ -83,7 +83,7 @@
                                 {{!-- Innate spells replace range for uses --}}
                                 {{#unless entry.isInnate}}
                                     {{spell.system.range.value}}
-                                {{else if uses}}
+            {{else if uses}}
```
CarlosFdez, line 86: ??

stwlam: Sorry, I think my editor isn't expecting these cursed `unless ... else if` blocks

PR #6658, opened 2023-02-27
Change default label of perception @Checks to "Perception"

stwlam:
(from "Perception Check")


TikaelSol:
Do we want to grab flat checks while we're at it? They have the same issue.

PR #6672, opened 2023-02-28
Enforce tagify whitelist by default

CarlosFdez:


src/util/tags.ts
```diff
@@ -42,6 +42,7 @@ function tagify(
         return null;
     }
 
+    enforceWhitelist ??= true;
```
stwlam, line 45: Why not up top?

CarlosFdez: No particular reason, just seemed pretty. I can move it.

src/util/tags.ts
```diff
@@ -36,6 +36,8 @@ function tagify(
     input: HTMLInputElement | null,
     { whitelist, maxTags, enforceWhitelist }: TagifyOptions = {}
 ): Tagify<TagRecord> | null {
+    enforceWhitelist ??= true;
```
stwlam, line 39: hiiigher

CarlosFdez: Funny enough I independently noticed what you meant. I think I wasn't aware you could set defaults for objects destructured in the params.

PR #6692, opened 2023-03-01
Don't rerender actor sheets every turn change

stwlam:
This turned out to be extremely annoying for anyone doing something in their actor sheet when a turn changes


CarlosFdez:
Something something damn handlebars

PR #6703, opened 2023-03-01
Add rudimentary party sheet start

CarlosFdez:
We can leave this as a draft until there's more to it. It really doesn't have a lot right now, some bits are unlocalized until we finally decide on what they will be, and some ideas may change.

image


CarlosFdez:
Superceded by #7289

PR #6704, opened 2023-03-01
Add rule element to Surprise Attack class feature

stwlam:


packs/data/classfeatures.db/surprise-attack.json
```diff
@@ -21,7 +21,26 @@
...
+                "key": "EphemeralEffect",
+                "predicate": [
+                    "encounter:round:1",
+                    {
+                        "lt": [
+                            "self:participant:initiative:rank",
+                            "target:participant:initiative:rank"
+                        ]
+                    }
+                ],
```
TikaelSol, line 35: `...if you roll Deception or Stealth for initiative...` Need to account for this too.

stwlam: whoops, right!

PR #6707, opened 2023-03-02
Set icon for energy beam, add persistent damage

stwlam:
plus extract normalization along for the ride

image


stwlam:
Went ahead and added its persistent damage too

PR #6721, opened 2023-03-02
Generate `target` roll options before extracting `EphemeralEffect`s

In3luki:
This came up while I was testing the combination of a `MarkToken` with an `EphermeralEffect` predicated on the mark.
Fully replacing the original roll options with the new options from the clone probably is the safest way to do it.

Test Feat (Mastemind Racket):
```json
{
  "key": "EphemeralEffect",
  "predicate": [
    "target:mark:effect-mastermind"
  ],
  "selectors": [
    "attack",
    "strike-damage"
  ],
  "uuid": "Compendium.pf2e.conditionitems.AJh5ex99aV6VTggg"
}
```
```json
{
  "key": "MarkToken",
  "slug": "effect-mastermind"
}
```

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -798,6 +798,18 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
                 ? [`origin:distance:${distance}`, `target:distance:${distance}`]
                 : [null, null];
 
+        // Target roll options
+        const getTargetRollOptions = (actor?: ActorPF2e | null): string[] => {
+            const targetOptions = actor?.getSelfRollOptions("target") ?? [];
+            if (targetToken && targetOptions.length > 0) {
+                targetOptions.push("target"); // An indicator that there is a target of any kind
+                const mark = this.synthetics.targetMarks.get(targetToken.document.uuid);
+                if (mark) targetOptions.push(`target:mark:${mark}`);
+            }
+            return targetOptions;
+        };
+        const targetRollOptions = getTargetRollOptions(targetToken?.actor);
+
         // Get ephemeral effects from this actor that affect the target while being attacked
         const targetEphemeralEffects = await extractEphemeralEffects({
             affects: "target",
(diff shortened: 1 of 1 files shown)


stwlam:
What's the "this" that came up?

src/module/actor/base.ts
```diff
@@ -798,14 +798,25 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
...
 
+        // Target roll options
+        const setBaseTargetOptions = (targetOptions: string[]): void => {
+            if (targetToken && targetOptions.length > 0) {
+                targetOptions.push("target"); // An indicator that there is a target of any kind
+                const mark = this.synthetics.targetMarks.get(targetToken.document.uuid);
+                if (mark) targetOptions.push(`target:mark:${mark}`);
+            }
+        };
+        let targetOptions = targetToken?.actor?.getSelfRollOptions("target") ?? [];
```
stwlam, line 809: No `let`

In3luki:
> What's the "this" that came up?

Oh, right. The `target` predicate doesn't work because the roll option is set after the ephemeral effect is already resolved.

src/module/actor/base.ts
In3luki: Done

src/module/actor/base.ts
```diff
@@ -798,14 +798,26 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
                 ? [`origin:distance:${distance}`, `target:distance:${distance}`]
                 : [null, null];
 
+        // Target roll options
+        const getTargetRollOptions = (actor?: ActorPF2e | null): string[] => {
```
stwlam, line 802:
```suggestion
        const getTargetRollOptions = (actor: Maybe<ActorPF2e>): string[] => {
```

src/module/actor/base.ts
```diff
@@ -798,14 +798,26 @@ class ActorPF2e extends Actor<TokenDocumentPF2e, ItemTypeMap> {
                 ? [`origin:distance:${distance}`, `target:distance:${distance}`]
                 : [null, null];
 
+        // Target roll options
+        const getTargetRollOptions = (actor: Maybe<ActorPF2e>): string[] => {
+            const targetOptions = actor?.getSelfRollOptions("target") ?? [];
+            if (targetToken && targetOptions.length > 0) {
```
stwlam, line 804:
```suggestion
            if (targetToken) {
```

PR #6729, opened 2023-03-04
Unify isLocked getter

CarlosFdez:


stwlam:
Wouldn't this be incorrect for any restricting grants?

CarlosFdez:
There's no behavior change. The reason for it in condition is that if the condition happens to be locked for another reason, it will shortcircuit and say its locked even if there's no restricting grant.

In the affliction work, its possible for it to be locked for another reason, but right now it can't be.

PR #6746, opened 2023-03-05
Add ammunition rule elements to linked weapons

stwlam:


nikolaj-a (dismissed):
👍

CarlosFdez:
how well does this work with toggle option? Or the rule editor in general?

CarlosFdez:
I think its fine if they don't work, so long as the result isn't terrible awful disaster.

stwlam:
Tested with a toggle: it just fails to find the rule element on the weapon and bails. Verboten rule elements like that could be handled from consumable items, and the user scolded.

CarlosFdez:
Perfect! I think it'll be good if it gets supported, but no rush on that.

PR #6753, opened 2023-03-05
Build out `EphemeralEffect` item alteration and add support for multiple alteration properties to the mixin

In3luki:
- Added new `ItemAlterationValueField` that allows `string`, `number` or `null` values for now.
- Built out the `alteration` schema in `EphemeralEffect` using the new field.
- Added rudimentary support for other allowed alteration properties which can be defined per class that uses the mixin by overriding the `this.allowedAlterations` array.
- Split off `applyAlteration` method from `applyAlterations` so it's possible to iterate through the alterations in a subclass without that happening in `applyAlterations` again.
- Added parameters to pass `resolvables` or an already `resolvedValue` to `applyAlteration`.

### DIFF
--- a/src/module/rules/rule-element/grant-item/rule-element.ts
+++ b/src/module/rules/rule-element/grant-item/rule-element.ts
@@ -58,6 +58,7 @@ class GrantItemRuleElement extends RuleElementPF2e {
         this.grantedId = this.item.flags.pf2e.itemGrants[this.flag ?? ""]?.id ?? null;
 
         this.alterations = data.alterations && this.isValidItemAlteration(data.alterations) ? data.alterations : [];
+        this.allowedAlterations = ["badge-value"];
     }
 
     static ON_DELETE_ACTIONS = ["cascade", "detach", "restrict"] as const;
--- a/src/module/rules/rule-element/mixins.ts
+++ b/src/module/rules/rule-element/mixins.ts
@@ -1,12 +1,22 @@
 import { ItemSourcePF2e } from "@item/data";
+import { tupleHasValue } from "@util";
 import { AELikeChangeMode } from "./ae-like";
 import { RuleElementPF2e } from "./base";
 import { RuleElementSchema } from "./data";
 
 /** A mixin for rule elements that allow item alterations */
 abstract class WithItemAlterations<TSchema extends RuleElementSchema> {
+    /** Every class using the mixin must define its allowed alterations here */
+    allowedAlterations: AllowedItemAlteration[] = [];
+
     static apply<TSchema extends RuleElementSchema>(Class: typeof RuleElementPF2e<TSchema>): void {
-        for (const methodName of ["itemCanBeAltered", "isValidItemAlteration", "applyAlterations"] as const) {
+        for (const methodName of [
+            "itemCanBeAltered",
+            "isValidItemAlteration",
+            "isValidItemAlterationValue",
+            "applyAlteration",
+            "applyAlterations",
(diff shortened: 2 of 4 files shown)


src/module/rules/rule-element/grant-item/rule-element.ts
```diff
@@ -58,6 +58,7 @@ class GrantItemRuleElement extends RuleElementPF2e {
         this.grantedId = this.item.flags.pf2e.itemGrants[this.flag ?? ""]?.id ?? null;
 
         this.alterations = data.alterations && this.isValidItemAlteration(data.alterations) ? data.alterations : [];
+        this.allowedAlterations = ["badge-value"];
```
stwlam, line 61: Can this be static?

src/module/system/schema-data-fields.ts
```diff
@@ -100,4 +100,28 @@ class PredicateField<
...
+    static override get _defaults() {
+        return mergeObject(super._defaults, {
+            required: true,
+            nullable: true,
+            initial: null,
+        });
+    }
+
+    protected _cast(value: unknown): string | number | null {
+        return typeof value === "string" || typeof value === "number" || value === null ? value : null;
```
stwlam, line 119:
```suggestion
        return value;
```

src/module/system/schema-data-fields.ts
```diff
@@ -100,4 +100,28 @@ class PredicateField<
...
+    }
+
+    protected override _validateType(value: unknown): boolean {
+        return typeof value === "string" || typeof value === "number" || value === null;
```
stwlam, line 123:
```suggestion
        return ["string", "number"].includes(typeof value) || value === null;
```

src/module/rules/rule-element/mixins.ts
```diff
@@ -1,12 +1,22 @@
...
 abstract class WithItemAlterations<TSchema extends RuleElementSchema> {
+    /** Every class using the mixin must define its allowed alterations here */
+    allowedAlterations: AllowedItemAlteration[] = [];
+
     static apply<TSchema extends RuleElementSchema>(Class: typeof RuleElementPF2e<TSchema>): void {
-        for (const methodName of ["itemCanBeAltered", "isValidItemAlteration", "applyAlterations"] as const) {
+        for (const methodName of [
+            "itemCanBeAltered",
+            "isValidItemAlteration",
+            "isValidItemAlterationValue",
```
stwlam, line 16: I imagine this could be rolled into a single method

src/module/system/schema-data-fields.ts
In3luki: Shouldn't it cast an invalid value to valid one?

stwlam: No, it should if the value isn't valid.

In3luki: DataField:
```js
  /**
   * Cast a non-default value to ensure it is the correct type for the field
   * @param {*} value       The provided non-default value
   * @returns {*}           The standardized value
   * @protected
   */
  _cast(value) {
    throw new Error(`Subclasses of DataField must implement the _cast method`);
  }
```
NumberField:
```js
  /** @override */
  _cast(value) {
    return Number(value);
  }
  ```

stwlam: Yeah, that's not great since a `string` or `ActorPF2e` isn't a number. A `NumberField` shouldn't take anything in the entire world and coerce it into one.

PR #6784, opened 2023-03-07
Add scope to roll option rule element

nikolaj-a:
Add a scope field to the RollOption rule element, which will allow restricting a roll option toggle to specific UI regions. If scope is omitted or set to "actions-tab", the toggle will show up in the default region on the Actions tab.

### DIFF
--- a/src/module/rules/rule-element/roll-option.ts
+++ b/src/module/rules/rule-element/roll-option.ts
@@ -64,6 +64,12 @@ class RollOptionRuleElement extends RuleElementPF2e<RollOptionSchema> {
     static override defineSchema(): RollOptionSchema {
         return {
             ...super.defineSchema(),
+            scope: new fields.StringField({
+                required: false,
+                nullable: false,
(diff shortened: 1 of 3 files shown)


src/module/rules/rule-element/roll-option.ts
```diff
@@ -64,6 +64,7 @@ class RollOptionRuleElement extends RuleElementPF2e<RollOptionSchema> {
     static override defineSchema(): RollOptionSchema {
         return {
             ...super.defineSchema(),
+            scope: new fields.StringField({ required: false, nullable: false }),
```
stwlam, line 67: Can this be given an `initial` of the actions tab? And maybe also a `choices` of what is currently supported (two so far?)

nikolaj-a: Until the Treat Wounds PR is done and merged only `actions-tab` is valid. I'll add the initial and choices.

PR #6791, opened 2023-03-07
Change Rule Element UI setting to one that controls player access

stwlam:


CarlosFdez:
Conflicts!

That said it'd be neat if somehow we could add these sorts of things to the permissions dialog, so that trusted players can get access but not regular players for example.

stwlam:
We can't add anything to that

CarlosFdez:
What a shame, that'd be the perfect place.

PR #6799, opened 2023-03-07
Add a critical damage overlay for spell damage

In3luki:
The critical damage overlay is a full replacement for the normal damage data so it should be the absolute critical damage.
A critical damage override can be loaded with `spell.loadVariant({ criticalDamage: true })` if one is available.

Acid Splash:
image

image

Acid Arrow:
image

image

### DIFF
--- a/src/module/item/spell/data.ts
+++ b/src/module/item/spell/data.ts
@@ -50,6 +50,9 @@ interface SpellSystemSource extends ItemSystemSource {
     damage: {
         value: Record<string, SpellDamage>;
     };
+    critical?: {
+        value: Record<string, SpellDamage>;
+    };
     heightening?: SpellHeighteningFixed | SpellHeighteningInterval;
     overlays?: Record<string, SpellOverlay>;
     save: {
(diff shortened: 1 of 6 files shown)


stwlam:
How would this work with saving-throw spells?

In3luki:
Wouldn't it be the same as attacks? I assumed spells would get a critical damgage button like strikes and the critical damage method would check if an overlay is available and use that.

CarlosFdez:
it wouldn't. People don't want to reroll for each degree of success. Fireball does half on success, not half formula.

In3luki:
I didn't intend for this to be a full solution for spell critical damage, just for the spells that don't scale directly from the normal formula. Another one of those would be [Shocking Grasp](https://2e.aonprd.com/Spells.aspx?ID=283).
The UI is probably a bit to prominent for such a small use case. Maybe it could be hidden behind another checkbox in the details.

CarlosFdez:
I think we either need to be able to define scaling by outcome on a per damage basis (critical damage only, basic, maybe a multiplier by outcome row), or be able to refer to base damage rows in other outcomes. That should create a structure thatll scale into supporting most use cases

PR #6833, opened 2023-03-09
Refactor initial filtering of Compendium Browser tabs

In3luki:
This replaces the current clunky (and not type aware) initial filtering implementation. The initial filter is now a copy of the initialized `filterData` of a `CompendiumBrowserTab` that can be requested by calling `CompendiumBrowserTab#getFilterData`.
The returned object is fully typed (this could still be improved by narrowing the type of `CheckboxData.options`) and can be savely mutated.
`CompendiumBrowser#openTab` now expects a full `filterData` object as an optional second parameter.
A tab can now also be opened by calling `CompendiumBrowserTab#open` which also accepts initial `filterData` for that tab as an optional parameter.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -394,7 +394,9 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
         $html.find(".sell-all-treasure button").on("click", (event) => this.onSellAllTreasure(event));
 
         // Inventory Browser
-        $html.find(".inventory-browse").on("click", (event) => this.onClickBrowseEquipmentCompendia(event));
+        for (const link of htmlQueryAll(html, ".inventory-browse")) {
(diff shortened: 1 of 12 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -633,24 +633,28 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
...
         for (const filterCode of checkboxesFilterCodes) {
             const splitValues = filterCode.split("-");
             if (splitValues.length !== 2) {
-                console.error(
-                    `Invalid filter value for opening the compendium browser:\n'${JSON.stringify(
-                        checkboxesFilterCodes
-                    )}'`
-                );
+                console.error(`Invalid filter value for opening the compendium browser: "${filterCode}"`);
                 return;
```
stwlam, line 644: Can this be changed to an `ErrorPF2e` throw?

In3luki: Done. I've also improved the sheet listeners a bit more.

PR #6855, opened 2023-03-10
Add list of core icons to check against while building packs

stwlam:


CarlosFdez:
Probably should have some documentation on how to populate the icons, the very few times they update that list.

Issue #6858, opened 2023-03-10
Overriding a Deadly weapon's damage type does not update the additional critical damage


cuyima:
Foundry Version: 10.291
PF2E Version: 4.8.3

When creating a rule to override a weapon's damage type, the additional critical damage by the Deadly trait is not updated to match the  base damage.
![image](https://user-images.githubusercontent.com/52758909/224427022-b69e64ae-2015-4e2e-b57f-2a50cab457d4.png)

I tried the following rules and neither affected the Deadly damage:
```json
{
  "definition": [
    "item:slug:soulforged-partizan"
  ],
  "key": "DamageDice",
  "selector": "strike-damage",
  "override": {
    "damageType": "{item|flags.pf2e.rulesSelections.planarPainDamage}"
  }
}
{
  "definition": [
    "item:slug:soulforged-partizan"
  ],
  "key": "DamageDice",
  "selector": "damage",
  "override": {
    "damageType": "{item|flags.pf2e.rulesSelections.planarPainDamage}"
  }
}
```

stwlam:
Deadly dice don't necessarily increase with increasing weapon die size: that must be done separately

cuyima:
I am not trying to increase the damage dice i am trying to make them match the overridden damage type of the weapon.

stwlam:
Sorry, I completely misread this!

PR #6869, opened 2023-03-11
Override damage type of deadly dice when main damage type is changed

cuyima:
When the the damage type of a weapon is changed, the deadly dice should use the same damage type (not size!) as the base. Otherwise, if using something like Soulforger's Planar Pain on a weapon with the Deadly trait, the deadly die will never do elemental damage when it should. See #6858
Before:
Screenshot 2023-03-11 142035
After:
Screenshot 2023-03-11 142124

### DIFF
--- a/src/module/system/damage/weapon.ts
+++ b/src/module/system/damage/weapon.ts
@@ -604,10 +604,9 @@ class WeaponDamagePF2e {
                 base.dieSize = override.override?.dieSize ?? base.dieSize;
                 base.damageType = override.override?.damageType ?? base.damageType;
                 base.diceNumber = override.override?.diceNumber ?? base.diceNumber;
-                damage.dice.forEach((die) => {
-                    if (die.slug.startsWith("deadly-"))
-                        die.damageType = override.override?.damageType ?? die.damageType;
-                });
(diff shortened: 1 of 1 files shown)


cuyima:
broke the PR by renaming the branch. sorry.

src/module/system/damage/weapon.ts
```diff
@@ -604,6 +604,10 @@ class WeaponDamagePF2e {
                 base.dieSize = override.override?.dieSize ?? base.dieSize;
                 base.damageType = override.override?.damageType ?? base.damageType;
                 base.diceNumber = override.override?.diceNumber ?? base.diceNumber;
+                damage.dice.forEach((die) => {
+                    if (die.slug.startsWith("deadly-"))
+                        die.damageType = override.override?.damageType ?? die.damageType;
+                });
```
stwlam, line 610:
```suggestion
                for (const die of damage.dice.filter((d) => d.slug.startsWith("deadly-"))) {
                    die.damageType = override.override?.damageType ?? die.damageType;
                }
```

stwlam (changes requested):
You have your fork protected, so I'm not able to commit the requested change.

stwlam (approved):

PR #6870, opened 2023-03-11
Disable PF2e's GM Vision if the GM Vision module is enabled

dev7355608:
The GM Vision module (https://foundryvtt.com/packages/gm-vision) increases brightness similar to PV's GM Vision.

I also added missing checks to PF2e' GM Vision setting `onChange` handler: without these checks it's possible to toggle PF2e's GM Vision via marco as non-GM and if the GM Vision or Perfect Vision module is enabled.

### DIFF
--- a/src/module/system/settings/index.ts
+++ b/src/module/system/settings/index.ts
@@ -265,9 +265,16 @@ export function registerSettings(): void {
         default: false,
         type: Boolean,
         onChange: (value) => {
-            const color = value ? CONFIG.PF2E.Canvas.darkness.gmVision : CONFIG.PF2E.Canvas.darkness.default;
-            CONFIG.Canvas.darknessColor = color;
-            canvas.colorManager.initialize();
+            if (
+                canvas.ready &&
+                game.user.isGM &&
+                !game.modules.get("gm-vision")?.active &&
+                !game.modules.get("perfect-vision")?.active
+            ) {
+                const color = value ? CONFIG.PF2E.Canvas.darkness.gmVision : CONFIG.PF2E.Canvas.darkness.default;
+                CONFIG.Canvas.darknessColor = color;
+                canvas.colorManager.initialize();
+            }
         },
     });
 
--- a/src/scripts/hooks/ready.ts
+++ b/src/scripts/hooks/ready.ts
@@ -83,6 +83,7 @@ export const Ready = {
             if (
                 canvas.ready &&
                 game.user.isGM &&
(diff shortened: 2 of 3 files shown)


TMunsie:
This is highly unlikely to get merged. Coding in "if my module is present" exceptions is a very bad precedent, but I'll let the code leads weigh in on it first. If there is something to be exposed, that's one thing that can be handled. But "do something specific if module X or Y is present" is going the wrong direction. Coding should be foundry, then system, then module. Not Foundry, then module, then system.

dev7355608:
I opened this PR only because the exception for PV was already in there. The real issue here is that the concept of GM Vision is system-agnostic: therefore it should be a module and not a feature of any system.

stwlam:
This doesn't really make any sense: players can use a macro to change the CONFIG values directly. And thank you for sharing your feelings on what should and shouldn't be a module, but you're just going to have to live with it.

dev7355608:
@stwlam: You're the one that added the exception for PV when you added GM Vision to the system. I was just following your example. I wouldn't have bothered otherwise to resolve this conflict. I wasn't even aware of PF2e GM Vision and the exception it made for PV until recently when randomly I stumbled over it.

There are things that make more sense as a module. That's just a fact. GM Vision is something all systems can benefit from, and it doesn't need a system-specific implementation. So it should be either a module or a core feature, but not part of any system.

And I can live perfectly fine with it, because I don't play PF2e.

stwlam:
I don't really care what you think is just a fact, but I'm happy to merge this if you remove unrelated cruft.

dev7355608:
Why immediately close the PR then instead of asking for this part to be removed? That gives the impression that all of it is rejected. I included this "unrelated cruft" because of two reasons:
- I assumed that GM Vision is only supposed to be used by GMs because of the name, and isn't supposed to be a hidden setting for non-GMs.
- I suspected it wouldn't work for non-GMs properly anyway, because, when a non-GM logins in, the brightness adjustment of GM Vision is not applied if the GM Vision is enabled. So if the player would try to turn GM Vision via `game.settings.get("pf2e", "gmVision", true)`, the `onChange` handler shouldn't be executed if the setting was already true. So the player would have to set it to false first and then again to true to turn GM Vision on. As it turns out the `onChange` handler of client (but not world) settings are called every time even if the value of the setting didn't change. I assume this is a core bug (I reported it already). So once/if this core bug is fixed, PF2e probably needs to make some changes to make GM Vision properly work for non-GMs again.

PR #6880, opened 2023-03-11
Divine Disharmy Action macro added

El-Castillo:
Created a macro for Divine Disharmony thaumaturge class feat, in line with the tamper class feat for Inventor.

Used a modified method from the Escape action macro to automatically detect and roll the highest modifier the actor has for the action.


stwlam:
Sorry, but we aren't expanding this collection.

PR #6887, opened 2023-03-12
automate Thaumaturge regalia through GrantItem

El-Castillo:
Selecting "Regalia" as a thaumaturge implement grants a "Regalia" item. This item grants the regalia benefits that can be automated when wielded, checking the flags set by the choices built into the class features to upgrade the abilities appropriately.


stwlam:
image

PR #6889, opened 2023-03-12
Change order of roll notes for action macros

nikolaj-a:
Move the extra roll notes for action macros after the roll notes for the used statistic. These extra notes are usually related to the degree of success and makes most sense at the bottom of the list.


nikolaj-a:
Found an alternative way to deal with this.

PR #6975, opened 2023-03-15
Remove type definitions of `ActorData` and `ItemData` subclasses

stwlam:
No runtime changes

### DIFF
--- a/src/module/actor/vehicle/sheet.ts
+++ b/src/module/actor/vehicle/sheet.ts
@@ -2,7 +2,7 @@ import { ActorSheetPF2e } from "../sheet/base";
 import { VehiclePF2e } from "@actor/vehicle";
 import { ErrorPF2e, getActionIcon, htmlClosest, htmlQuery, htmlQueryAll } from "@util";
 import { ActorSheetDataPF2e } from "@actor/sheet/data-types";
-import { ActionItemData } from "@item/action";
+import { ActionItemPF2e } from "@item";
 
 export class VehicleSheetPF2e extends ActorSheetPF2e<VehiclePF2e> {
     static override get defaultOptions(): ActorSheetOptions {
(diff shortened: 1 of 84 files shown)


src/module/actor/vehicle/sheet.ts
```diff
@@ -40,19 +40,18 @@ export class VehicleSheetPF2e extends ActorSheetPF2e<VehiclePF2e> {
...
+        };
 
         // Actions
         for (const item of this.actor.itemTypes.action.sort((a, b) => a.sort - b.sort)) {
-            const itemData = item.toObject(false);
+            const itemData = item.toObject(false) as unknown as ActionItemPF2e;
             const img = getActionIcon(item.actionCost);
             const actionType = item.actionCost?.type ?? "free";
-            actions[actionType].actions.push({ ...itemData, img });
+            actions[actionType].actions.push({ ...itemData, img } as ActionItemPF2e);
```
CarlosFdez, line 54: I don't think item.toObject(false) is an ActionItemPF2e :(

stwlam: So you think it's a thing that went away after V9?

CarlosFdez: its more that the type of the map could be anything. Why do an as unknown as ActionItemPF2e rather than using the return type of toObject()?.

stwlam: Because it's also not the source data. The ideal type is closer to the Document than the source

CarlosFdez: So long as these get fixed soonish sure. It just bugs me that I just got this out of any-land and now we're returning right back into it.

CarlosFdez (approved):

PR #6998, opened 2023-03-16
Add light- and heavy-barding armor categories

stwlam:


CarlosFdez:
there's a part of me that would like them to not be immutable incase its an AC and someone wants to tweak it, but ACs are kinda just using regular armor and calling it barding anyways.

PR #7000, opened 2023-03-16
Propagate aura level to effect level by default

JDCalvert:
When creating an effect from an aura, use the aura's level as the effect's level, unless overridden when defining the effect in the aura's rule element.

Also removed the explicit level set in the Protective Ward aura as it's no longer needed.

### DIFF
--- a/src/module/actor/types.ts
+++ b/src/module/actor/types.ts
@@ -36,6 +36,7 @@ type SaveType = (typeof SAVE_TYPES)[number];
 
 interface AuraData {
     slug: string;
+    level?: number;
     radius: number;
     effects: AuraEffectData[];
     colors: AuraColors | null;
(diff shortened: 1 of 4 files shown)


src/module/actor/types.ts
```diff
@@ -36,6 +36,7 @@ type SaveType = (typeof SAVE_TYPES)[number];
 
 interface AuraData {
     slug: string;
+    level?: number;
```
stwlam, line 39:
```suggestion
    level: number | null;
```

src/module/rules/rule-element/aura.ts
```diff
@@ -57,6 +57,7 @@ export class AuraRuleElement extends RuleElementPF2e {
         if (typeof Number.isInteger(radius) && radius > 0 && radius % 5 === 0) {
             const data = {
                 slug: this.slug,
+                level: this.item.system.level?.value,
```
stwlam, line 60:
```suggestion
                level: this.item.system.level?.value ?? null,
```

PR #7006, opened 2023-03-16
Move floaty text handling to abstract effect

CarlosFdez:
Reducing the code for affliction condition automation

### DIFF
--- a/src/module/item/effect/document.ts
+++ b/src/module/item/effect/document.ts
@@ -200,7 +200,6 @@ class EffectPF2e extends AbstractEffectPF2e {
         return super._preUpdate(changed, options, user);
     }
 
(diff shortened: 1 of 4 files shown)


src/module/item/effect/document.ts
```diff
@@ -200,29 +200,12 @@ class EffectPF2e extends AbstractEffectPF2e {
...
-    ): void {
-        super._onCreate(data, options, userId);
-
-        if (!this.flags.pf2e.aura || game.combat?.started) {
-            this.actor?.getActiveTokens().shift()?.showFloatyText({ create: this });
-        }
-    }
-
-    /** Show floaty text when this effect is deleted from an actor */
+    /** Show floaty text when this effect is deleted from an actor (if unlinked) */
```
stwlam, line 203:
```suggestion
```

PR #7014, opened 2023-03-17
Automate the skill selection of Marshal Dedication

TikaelSol:
Supersedes #6992

### DIFF
--- a/packs/data/feats.db/marshal-dedication.json
+++ b/packs/data/feats.db/marshal-dedication.json
@@ -44,6 +44,96 @@
                     "mental",
                     "visual"
                 ]
+            },
+            {
+                "adjustName": false,
+                "choices": [
+                    {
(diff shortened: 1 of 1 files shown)


simonward (approved):

packs/data/feats.db/marshal-dedication.json
```diff
@@ -44,6 +44,96 @@
...
+                        "value": "itm-trained"
+                    },
+                    {
+                        "label": "PF2E.SkillItm",
+                        "predicate": [
+                            "skill:itm:rank:1"
+                        ],
+                        "value": "itm-expert"
+                    }
+                ],
```
stwlam, line 79: This is going to fail out the entire choice set and then the rest of the rule elements on the item if they're expert or better on both diplomacy and intimidation.

PR #7019, opened 2023-03-18
Add ability to define if an affliction's condition is linked

CarlosFdez:
image

### DIFF
--- a/static/templates/items/affliction-details.hbs
+++ b/static/templates/items/affliction-details.hbs
@@ -15,7 +15,7 @@
 </ol>
 
 {{#each stages as |stage stageId|}}
-    <ol class="form-list subsection" data-stage-id="{{stageId}}">
+    <ol class="form-list subsection affliction-stage" data-stage-id="{{stageId}}">
         <header>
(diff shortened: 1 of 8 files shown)


CarlosFdez:
I can changed it from linked to locked if that makes more sense. Kinda at a standstill regarding what to call it.

I also moved afflictions to the top because otherwise the location of the effect icon keeps hopping around when you increase the stage as conditions are added or removed. At the top is better UX.

static/templates/items/affliction-details.hbs
```diff
@@ -80,6 +80,9 @@
                         {{/if}}
                         {{#if @root.options.editable}}
                             <div class="item-controls">
+                                <a class="link {{#if condition.linked}}linked{{/if}}" data-action="condition-link">
+                                    <i class="fas fa-fw {{#if condition.linked}}fa-link{{else}}fa-link-slash{{/if}}"></i>
+                                </a>
                                 <a class="remove" data-action="condition-delete"><i class="fa fa-fw fa-times"></i></a>
```
stwlam, line 86: These could use some title text, especially the Linked toggle

PR #7034, opened 2023-03-19
Filter compendium results based on action "Browse" button's context

Novynn:
The intent of this PR is to make the "Browse" buttons on the Actions tab of the player sheet more intuitive.

The PR fulfills this by filtering the resulting compendium list based on which button was pressed. In order to do so I needed to add the "Action Type" checkbox filter to the action compendium so that actions / reactions / free actions were filtered correctly.

I expect there to be feedback on this as this is my first submission and my primary motivation was getting it working for my upcoming sessions.


TikaelSol:
Be sure to run `npm run lint:fix` and `npm run test`.

It may also help to submit a small data entry PR to allow the pipelines to run for you automatically.

PR #7043, opened 2023-03-19
Set `initiativeStatistic` flag on combatant when using `actor#attributes.initiative.roll`

In3luki:


src/module/actor/creature/document.ts
```diff
@@ -512,6 +512,7 @@ abstract class CreaturePF2e extends ActorPF2e {
                 const updateTracker = args.updateTracker ?? true;
                 if (updateTracker) {
                     game.combat?.setInitiative(combatant.id, roll.total);
+                    await combatant.setFlag("pf2e", "initiativeStatistic", proficiency);
```
stwlam, line 515: never use `setFlag`

In3luki: done. should be safe to skip rendering for that update so I added that too.

stwlam: Unfortunately this will only prevent the item sheet from rendering

PR #7063, opened 2023-03-21
Generate static returns of `ClientDocumentMixin`

stwlam:
plus modest Foundry types refactor and dependency upgrade to typescript 5.0.2


CarlosFdez:
I'm not a fan of including changes to # in this PR, but might as well PR and see what breaks later.

PR #7066, opened 2023-03-21
Add variants to Lay on Hands spell

TikaelSol:


TikaelSol:
Actually, something is up with making spell variants. Will dig into that and revisit.

PR #7068, opened 2023-03-22
Remove unused handlebars helpers, improve type annotations of remaining

stwlam:


CarlosFdez:
times is still in use, check for {{#times for all hits

PR #7078, opened 2023-03-22
Fix spell variant interval heightening creation, chat card display for saves and action icons for variants with no action override

In3luki:
- Variant buttons for spells that deal no damage are now displayed.
- Creation of interval heightening entries was using `data` as fallback property which lead to the update being silently dropped for variants. Not sure why.
- Variants that had the same action time as their parent were not displaying any action icon. Now they show the parents action icon instead.
- Updating of non-embedded variants was failing because of the `options.parent` check in `SpellPF2e#update`. Removing that should have no side effects.

image

image

### DIFF
--- a/src/module/item/spell/document.ts
+++ b/src/module/item/spell/document.ts
@@ -721,20 +721,14 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
         }
 
         const variants = this.overlays.overrideVariants
-            .flatMap((variant): SpellVariantChatData => {
-                const overlayIds = [...variant.appliedOverlays!.values()];
-                const actions = (() => {
-                    const actionIcon = getActionIcon(variant.system.time.value, null);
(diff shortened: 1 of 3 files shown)


src/module/item/spell/document.ts
```diff
@@ -725,7 +725,7 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
                 const overlayIds = [...variant.appliedOverlays!.values()];
                 const actions = (() => {
                     const actionIcon = getActionIcon(variant.system.time.value, null);
-                    return variant.system.time.value !== this.system.time.value && actionIcon ? actionIcon : null;
+                    return actionIcon;
```
stwlam, line 728: You can remove the IIFE now. :)

In3luki: Updated. I don't know how I didn't notice that. 😄

PR #7090, opened 2023-03-23
Enforce slugginess of `ActorPF2e#toggleCondition` argument at runtime

stwlam:
This method is frequently used in user macros and leads to aberrant results when the argument isn't an actual slug


CarlosFdez:
The only thing that comes to mind is maybe toggling persistent damage off, but I think that's handled elsewhere, so prob fine.

PR #7091, opened 2023-03-23
Bug 7087: IWR - Handle 0 value typed damage

Rapida11428:
Updated the IWR module to check if the roll was increasedFrom zero rather than defaulting to the instance type damage.

Tests ran:
1d0 fire vs. resistance to fire (0 dmg) / immunity to fire (0 damage)/ resistance to all (0 damage)/ no resist (1 damage)
1d0 untyped vs. resistance to fire (1 dmg) / immunity to fire (1 damage)/ resistance to all (0 damage)/ no resist (1 damage)


stwlam:
superseded by #7092

PR #7101, opened 2023-03-24
Skip building packs if all local pack data files are unchanged

In3luki:
This records the last modified time of all pack files and compares to an exisiting record of the last build run if available.
Only tested on windows. Collection of file stats takes <1 second on my test machine.
It could also be expanded to only build changed packs.

This should have no effect on CI/CD scripts because the `.pack-changes.json` file doesn't exist there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ packs/temp-data
 static/packs/
 types/foundry/node_modules/
 types/foundry/package-lock.json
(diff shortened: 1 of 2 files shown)


stwlam:
You can just use `node_modules/.bin/webpack` if you don't want to rebuild packs

In3luki:
That's way too easy... 😛

Issue #1152, opened 2021-03-06
Inventory should be more easily sortable


cswendrowski:
In GitLab by @yoshisman8 on Mar 6, 2021, 15:23

If at all possible, having a simple button next to the Add and Search/Open compendium buttons on the Inventory tab that sorts the items in that group alphabetically would be a really welcomed addition for people like me who like seeing things sorted alphabetically.

Possible place for said button:

![image](http://gitlab.com/hooking/foundry-vtt---pathfinder-2e/uploads/036567309d3e43d0b25306e2e6b84622/image.png)

zarmstrong:
leaving this comment here since it was on the dupe issue:

Additionally, all columns sortable would be great. Sort by bulk, by equipped status, etc.

Grammernatzi:
Very much still wish I had this after four years

PR #7111, opened 2023-03-25
Inventory Sorting (name, quantity, weight, sell value)

Rapida11428:
Added the ability to sort inventory on 'item-name', 'item-quantity', 'item-weight', and 'item-sell-value'.

Tested on NPC inventory, Loot/Merchant sheet, and PC inventory.

Only thing I'm unhappy with right now is changing quantity can shift the sort order. I think this is "ok", but would like feedback.

For the actual sorting, there is probably a better way to do it - currently iterating over the lists at least twice (sort, then forEach).

(Sort of based on request in: https://github.com/foundryvtt/pf2e/issues/1152)


Rapida11428:
I think this should probably be extended to show sort buttons on the sheets, but I think this is a decent stand-alone change so didn't attempt it yet.

CarlosFdez:
Lots of things wrong with this one unfortunately.
1) No JQuery, we're swapping to raw DOM events
2) You don't need a bunch of events, datasets are a thing
3) Changing data and then refreshing the sheet and handling it during sheet rendering is a better approach if you're trying to do a permanent sort (which it looks like you're trying to do). If this is just a resort  that has to be done every so often though, I believe there's a built in foundry method for it.

Rapida11428:
1. Ah, good to know. I was basing it on other examples in the file.
2. Makes sense, I'm already using a data-set so can just leverage that.
3. I'll need to figure this out - new to how rendering the sheet works.

Rapida11428:
Examples of the UI here: https://imgur.com/a/TXOTwi7

Rapida11428:
@CarlosFdez - I wasn't able to figure out the re-sort method you mentioned. Any advice there would be appreciated.

stwlam:
Closing due to foundry sorting being replaced with system solution

PR #7121, opened 2023-03-25
Use CRB language for creature identification info on NPC sheet

stwlam:
Before:
image

After:
image


CarlosFdez:
Should be ok to start looking at this soon, once the conflicts from the vite stuff is resolved!

CarlosFdez:
So I think its fine and I'm down to try it, though I am slightly concerned since implicitly the book is talking about "more" applicable skills, not just any applicable lore skill, but it is what it is.

PR #7149, opened 2023-03-26
Expand system action list to support metadata

nikolaj-a:
Expand the existing `game.pf2e.actions` object to be an actual `Collection<Action>` instance, able to contain action implementations to be bundled with the system. Some will be just plain data-driven actions, while others with more complexity will be custom implementations. The existing action functions are still available, but should be considered deprecated going forward to lessen the maintenance burden.

Included are three examples for action implementations, two SingleCheckActions (`hide` and `sneak`) as well as a custom action implementation (`trip`).

image


stwlam:
I didn't catch any use of `Collection` facilities: any reason to not just make it a `Map`?

src/scripts/set-game-pf2e.ts
```diff
@@ -24,26 +24,34 @@ import {
...
+        Object.entries({
             encouragingWords,
             raiseAShield,
             restForTheNight,
             earnIncome: showEarnIncomePopup,
             steelYourResolve,
             treatWounds,
             ...ActionMacros,
-        };
+        }).forEach(([name, action]) => {
```
stwlam, line 52: No `forEach` 😊

nikolaj-a: Fixed!

PR #7166, opened 2023-03-27
Fix regressions with rolling initiative

### LINKED ISSUE #7164: Initiative regressions introduced after 4.9.3
- Initiative can't be rolled from sheets without first adding a combatant separately
- Initiative rolled from sheets always shows the roll dialog
- The `initiativeStatistic` combatant flag is no longer recorded under any circumstance

In3luki:
- A combatant is created on rolling initiative if one doesn't exist already
- Roll params are passed from the character sheet again
- The `initiativeStatistic` is always set for characters and npcs

Closes #7164

### DIFF
--- a/src/module/actor/initiative.ts
+++ b/src/module/actor/initiative.ts
@@ -31,25 +31,9 @@ class ActorInitiative {
         this.statistic = statistic;
     }
 
-    async roll(args: InitiativeRollParams): Promise<InitiativeRollResult | null> {
+    async roll(args: InitiativeRollParams = {}): Promise<InitiativeRollResult | null> {
         // Get or create the combatant
-        const combatant = await (async (): Promise<CombatantPF2e<EncounterPF2e> | null> => {
-            if (!game.combat) {
-                ui.notifications.error(game.i18n.localize("PF2E.Encounter.NoActiveEncounter"));
-                return null;
-            }
-            const token = this.actor.getActiveTokens().pop();
-            const existing = game.combat.combatants.find((combatant) => combatant.actor === this.actor);
-            if (existing) {
-                return existing;
-            } else if (token) {
-                await token.toggleCombat(game.combat);
-                return token.combatant ?? null;
-            } else {
-                ui.notifications.error(game.i18n.format("PF2E.Encounter.NoTokenInScene", { actor: this.actor.name }));
-                return null;
-            }
-        })();
+        const combatant = await CombatantPF2e.fromActor(this.actor, false);
         if (!combatant) return null;
 
         if (combatant.hidden) {
@@ -57,12 +41,16 @@ class ActorInitiative {
         }
 
         const roll = await this.statistic.roll(args);
-        if (!roll) return null;
+        if (!roll) {
+            // Render combat sidebar in case a combatant was created but the roll was not completed
+            game.combats.render(false);
+            return null;
+        }
 
         // Update the tracker unless requested not to
         const updateTracker = args.updateTracker ?? true;
         if (updateTracker) {
-            game.combat?.setInitiative(combatant.id, roll.total);
+            combatant.encounter.setInitiative(combatant.id, roll.total);
         }
 
         return { combatant, roll };
--- a/src/module/apps/sidebar/chat-log.ts
+++ b/src/module/apps/sidebar/chat-log.ts
@@ -3,9 +3,10 @@ import { ArmorPF2e } from "@item";
 import { TokenPF2e } from "@module/canvas";
 import { ChatMessagePF2e } from "@module/chat-message";
 import { applyDamageFromMessage } from "@module/chat-message/helpers";
+import { CombatantPF2e } from "@module/encounter";
 import { CheckPF2e } from "@system/check";
 import { DamageRoll } from "@system/damage/roll";
-import { ErrorPF2e, fontAwesomeIcon, htmlClosest, htmlQuery, objectHasKey } from "@util";
+import { fontAwesomeIcon, htmlClosest, htmlQuery, objectHasKey } from "@util";
 
 export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
     /** Replace parent method in order to use DamageRoll class as needed */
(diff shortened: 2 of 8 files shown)


src/module/encounter/document.ts
```diff
@@ -145,6 +145,20 @@ class EncounterPF2e extends Combat {
         await this.update({ turn: this.turns.findIndex((c) => c.id === currentId) });
     }
 
+    override async setInitiative(id: string, value: number): Promise<void> {
+        const combatant = this.combatants.get(id, { strict: true });
+        if (combatant.actor?.isOfType("character", "npc")) {
+            return game.combat?.setMultipleInitiatives([
```
stwlam, line 151:
```suggestion
            return this.setMultipleInitiatives([
```

src/module/encounter/combatant.ts
```diff
@@ -36,6 +37,24 @@ class CombatantPF2e<
...
+        if (!game.combat) {
+            ui.notifications.error(game.i18n.localize("PF2E.Encounter.NoActiveEncounter"));
+            return null;
+        }
+        const token = actor.getActiveTokens().pop();
+        const existing = game.combat.combatants.find((combatant) => combatant.actor === actor);
+        if (existing) {
+            return existing;
+        } else if (token) {
+            await token.toggleCombat(game.combat);
```
stwlam, line 51: Can we use `game.combat.createEmbeddedDocuments` here so that we can control rendering? Not rendering would be desirable in some callers' cases.

In3luki: Updated. I've set it to render by default.

In3luki: Also added the `sceneId` to the `Combatant` source. It seems to work without it but core `TokenLayer#toggleCombat` uses it so it's probably safer to include it.

src/module/apps/sidebar/chat-log.ts
```diff
@@ -308,8 +306,16 @@ export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
                         );
                         return;
                     }
+                    if (!actor) return;
+                    const combatant = await CombatantPF2e.fromActor(actor);
+                    if (!combatant) return;
+                    const value = message.rolls.at(0)?.total ?? 0;
+
+                    await game.combat!.setInitiative(combatant.id, value);
```
stwlam, line 314:
```suggestion
                    await combatant.encounter.setInitiative(combatant.id, value);
```

src/module/apps/sidebar/chat-log.ts
```diff
@@ -308,8 +306,16 @@ export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
                         );
                         return;
                     }
+                    if (!actor) return;
+                    const combatant = await CombatantPF2e.fromActor(actor);
+                    if (!combatant) return;
```
stwlam, line 311:
```suggestion
```

src/module/actor/initiative.ts
```diff
@@ -60,9 +45,8 @@ class ActorInitiative {
         if (!roll) return null;
 
         // Update the tracker unless requested not to
-        const updateTracker = args.updateTracker ?? true;
         if (updateTracker) {
-            game.combat?.setInitiative(combatant.id, roll.total);
+            game.combat!.setInitiative(combatant.id, roll.total);
```
stwlam, line 49:
```suggestion
            combatant.encounter.setInitiative(combatant.id, roll.total);
```

src/module/chat-message/listeners/set-as-initiative.ts
```diff
@@ -37,35 +35,22 @@ export const SetAsInitiative = {
...
                 if (!actor) return;
-
-                const combatant = ((): CombatantPF2e<EncounterPF2e> | null => {
-                    const existing = game.combat.combatants.find((c) => c.actor === actor);
-                    if (existing) return existing;
-                    ui.notifications.error(game.i18n.format("PF2E.Encounter.NotParticipating", { actor: actor.name }));
-                    return null;
-                })();
+                const combatant = await CombatantPF2e.fromActor(actor);
                 if (!combatant) return;
```
stwlam, line 49:
```suggestion
```

src/module/chat-message/listeners/set-as-initiative.ts
```diff
@@ -37,35 +35,22 @@ export const SetAsInitiative = {
...
+                const value = message.rolls.at(0)?.total ?? 0;
 
-                const statistic =
-                    message.flags.pf2e.context?.domains.find(
-                        (s): s is SkillLongForm | "perception" =>
-                            setHasElement(SKILL_LONG_FORMS, s) || s === "perception"
-                    ) ?? null;
-                const value = message.rolls[0].total;
+                await game.combat!.setInitiative(combatant.id, value);
```
stwlam, line 52:
```suggestion
                await combatant?.encounter.setInitiative(combatant.id, value);
```

src/module/actor/initiative.ts
```diff
@@ -31,25 +31,10 @@ class ActorInitiative {
...
-                return existing;
-            } else if (token) {
-                await token.toggleCombat(game.combat);
-                return token.combatant ?? null;
-            } else {
-                ui.notifications.error(game.i18n.format("PF2E.Encounter.NoTokenInScene", { actor: this.actor.name }));
-                return null;
-            }
-        })();
+        const combatant = await CombatantPF2e.fromActor(this.actor, updateTracker);
```
stwlam, line 37: The point of having the render option is so that we don't render twice, which will happen here.

src/module/apps/sidebar/chat-log.ts
```diff
@@ -308,8 +306,16 @@ export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
                         );
                         return;
                     }
+                    if (!actor) return;
+                    const combatant = await CombatantPF2e.fromActor(actor);
```
stwlam, line 310:
```suggestion
                    const combatant = await CombatantPF2e.fromActor(actor, false);
```

src/module/apps/sidebar/chat-log.ts
In3luki: This would always show the notification that the initiative was set even if there is no active encounter.

PR #7172, opened 2023-03-28
Move ammunition select to below strike name on PC sheet

stwlam:
Before:
Screenshot from 2023-03-27 22-28-31

After:
Screenshot from 2023-03-27 22-36-57

### DIFF
--- a/src/styles/actor/character/_actions.scss
+++ b/src/styles/actor/character/_actions.scss
@@ -8,9 +8,7 @@
         }
 
         li.strike {
-            .item-name {
-                align-items: flex-start;
-            }
+            gap: 3px;
 
             .ammo-remaining {
                 white-space: nowrap;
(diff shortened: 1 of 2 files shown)


CarlosFdez:
That's going to lead to big stacks of buttons for combination weapons. And will look kinda bad for repeating ammo (and capacity once we get there, since capacity will need a way to load into the current chamber).

This will need fixing though, though I don't know of the fix. That said Rounds (X) has been a real PITA for merchant sheets, I've been having to make a new Rounds item that has no label just to avoid it.

My immediate thoughts: 
- do what you did but move magazine ammo to the right of the rounds display, but that'll create inconsistency with bombs.
- move the strike icon into the name row (and make it smaller) to give more everything more horizontal room.

I'm not really sure on anything though and would love more professional UI/UX opinions.

PR #7198, opened 2023-03-30
Continue broadening scope of `ActorPF2e#getCheckContext`

stwlam:


src/module/actor/base.ts
```diff
@@ -940,28 +965,49 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
+                    const statistic = targetActor?.saves?.[targetedDC]?.dc;
+                    return statistic ? { scope, statistic, slug: targetedDC, value: statistic.value } : null;
+                }
+                // A skill DC
+                default: {
+                    const statistic = targetActor?.skills?.[targetedDC]?.dc;
+                    return statistic ? { scope, statistic, slug: targetedDC, value: statistic.value } : null;
+                }
+            }
+        })();
```
CarlosFdez, line 1008: How well would it work to use getStatistic() for this? I can imagine ac/armor won't work well, but the rest seem doable by that method at a glance.

src/module/system/statistic/index.ts
```diff
@@ -329,7 +329,13 @@ class StatisticCheck {
             const isValidAttacker = actor.isOfType("creature", "hazard");
             const isAttackItem = item?.isOfType("weapon", "melee", "spell");
             if (isValidAttacker && isAttackItem && ["attack-roll", "spell-attack-roll"].includes(this.type)) {
-                return actor.getCheckRollContext({ item, domains, statistic: this, options: new Set() });
+                return actor.getCheckContext({
+                    item,
+                    domains,
+                    statistic: this,
+                    targetedDC: "armor",
```
CarlosFdez, line 336: why armor over ac?

stwlam: Following convention with the rest

PR #7199, opened 2023-03-30
Add support for ephemeral effects on damage application

stwlam:


CarlosFdez (approved):
This has been sitting here a while, so only going to approve it just in case you want to wait on this for whatever reason.

PR #7201, opened 2023-03-30
Make the sort by helper generate a comparator instead

CarlosFdez:


CarlosFdez:
I personally have no preference but this does chain better so you might prefer it.

This is something that came to mind while working on something else that needs to use this function. If there's a better name in mind or you prefer it how it is let me know.

PR #7217, opened 2023-03-31
Remove problematic bestiary ability link from Reflection of Life

roxim5:


packs/data/campaign-effects.db/effect-reflection-of-life-fast-healing.json
```diff
@@ -18,6 +18,7 @@
         },
         "rules": [
             {
+                "details": "",
```
stwlam, line 21: ?

PR #7225, opened 2023-03-31
Adjust "Formulated Sunlight" quantity and charges to match rations

zachjb:
Since the quantity and charges were swapped, it counted as 7 separate objects for purposes of bulk and wealth calculations.

### DIFF
--- a/packs/data/equipment.db/formulated-sunlight.json
+++ b/packs/data/equipment.db/formulated-sunlight.json
@@ -8,8 +8,8 @@
         },
         "baseItem": null,
         "charges": {
-            "max": "1",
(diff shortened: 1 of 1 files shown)


packs/data/equipment.db/formulated-sunlight.json
```diff
@@ -8,8 +8,8 @@
         },
         "baseItem": null,
         "charges": {
-            "max": "1",
-            "value": "1"
+            "max": "7",
+            "value": "7"
```
stwlam, line 12:
```suggestion
            "max": 7,
            "value": 7
```

PR #7240, opened 2023-04-01
Enhance inventory sorting with `SortableJS`

In3luki:
[sortable-inventory.webm](https://user-images.githubusercontent.com/41452412/229311409-0ae583dc-e78e-4f05-8527-3c0d6271a8ad.webm)

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -44,6 +44,7 @@ import { IWREditor } from "./popups/iwr-editor";
 import { RemoveCoinsPopup } from "./popups/remove-coins-popup";
 import { CraftingFormula } from "@actor/character/crafting";
 import { UUIDUtils } from "@util/uuid-utils";
+import Sortable, { type SortableEvent } from "sortablejs";
 
 /**
  * Extend the basic ActorSheet class to do all the PF2e things!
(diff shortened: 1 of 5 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -589,6 +598,80 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
             ]);
         });
 
+        // Inventory sorting
+        const inventoryList = htmlQueryAll(html, "section.inventory-list, ol[data-container-type=actorInventory]").find(
+            (e) => e !== null
+        );
```
stwlam, line 604: Shouldn't this just be an `htmlQuery`? `htmlQueryAll` can't return an array with `null` elements.

In3luki: You're right. I was going from the assumption that two selectors needed to be resolved with ``htmlQueryAll` but `htmlQuery` works just fine.

stwlam:
Could dropping an item directly onto a container row (rather than its contents) easily be handled to place it in the target container?

src/module/item/physical/document.ts
```diff
@@ -318,6 +319,106 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
+            // If theses two items are stackable, combine them
+            if (await this.stackWith(relativeTo)) return;
+        }
+        if (toContainer) {
+            if (!toContainer.isOfType("backpack")) {
+                throw ErrorPF2e(`The toContainer item is not a container item!`);
+            } else if (!toContainer.actor) {
+                throw ErrorPF2e(`The toContainer item is not an owned item!`);
+            }
+        }
```
stwlam, line 377: I suspect most of this could be handled with type checking

In3luki:
> Could dropping an item directly onto a container row (rather than its contents) easily be handled to place it in the target container?

Fixed. It was working before but must have broken when I added item stacking.
- Dropping on a container puts the item into the container.
- Dropping in container contents puts the item relative to the contained item.

src/module/item/physical/document.ts
In3luki: This is intended for modules or macros that might pass in the wrong item types.

In3luki:
I *hope* everything is working now:

[sortable-inventory.webm](https://user-images.githubusercontent.com/41452412/230380733-cb895a8b-2a70-4769-8507-4c7cf16ee1a4.webm)

PR #7255, opened 2023-04-02
Update Singular Expertise to use crossbow-damage selector

CrimsonPolar:
Removes the predicate and uses the "crossbow-damage" selector instead.


stwlam:
That would make it work for just the crossbow-proper. Hand crossbows, etc. would no longer function.

CrimsonPolar:
Understood, thank you for clarifying!

Issue #6382, opened 2023-02-15
With the Action Macro Repair - Crafting it's not possible to repair other things than items


sorrythatuserbla:
With the current state of the action macro it is only possible to repair items with that (and it works perfect afaik) but not to repair construct companions or make other repair checks. 
When clicking "Cancel" the blue roll window will pop up but even confirming that will not result in any roll which seems like a bug.

sorrythatuserbla:
Here is a small example of the bug.
https://user-images.githubusercontent.com/78597675/224001817-16ae43e5-0448-4817-bb3d-fed7cb6e8cc0.mp4

PR #7259, opened 2023-04-02
Check if the item is null when repairing an item

keyneston:
## What

Currently if you hit cancel or the [x] button before providing an item it will continue and attempt to roll a skill check. But then will not post the results of the skill roll.

This change makes it so that this won't happen. e.g. if the cancel or [x] button are pressed on the previous display.

Ideally this should do more. Such as rolling skill, and then having various repair buttons available for the player to select depending upon the DM's guidance. That said that is outside of my skill level in terms of JS and my knowledge of the project, so you are getting this tiny bug fix.

## Other Issues

Relates to: #6382

### DIFF
--- a/src/module/system/action-macros/crafting/repair.ts
+++ b/src/module/system/action-macros/crafting/repair.ts
@@ -11,6 +11,12 @@ async function repair(options: RepairActionOptions) {
     const item =
         options.item ?? (options.uuid ? await fromUuid(options.uuid) : await SelectItemDialog.getItem("repair"));
 
+    if (item === null) {
+        // Don't do anything if passed a null item.
+        // TODO: do something smart here. Like possibly rolling the result and the recovered amount both.
+        return;
+    }
+
     // ensure specified item is a valid crafting target
     if (item && !(item instanceof PhysicalItemPF2e)) {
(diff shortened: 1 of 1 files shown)


src/module/system/action-macros/crafting/repair.ts
```diff
@@ -11,6 +11,12 @@ async function repair(options: RepairActionOptions) {
     const item =
         options.item ?? (options.uuid ? await fromUuid(options.uuid) : await SelectItemDialog.getItem("repair"));
 
+    if (item === null) {
+        // Don't do anything if passed a null item.
+        // TODO: do something smart here. Like possibly rolling the result and the recovered amount both.
+        return;
+    }
+
```
stwlam, line 19: This would be handled better in the check right below it.

nikolaj-a:
We might want to hold on for this one until it has been converted to an actual action class. That will change the signature a bit to allow returning a promise.

src/module/system/action-macros/crafting/repair.ts
keyneston: I tried that but it it had two issues.

1) You then get an error from where `item.name` is called because item == null. This can be worked around but makes the following block a bit more messy.
2) This is then returning an error everytime "cancel" is called rather than just exiting out. Which seems like a mistake to me.

Issue #1956, opened 2022-01-02
Tracking Issue: Data entry for ChoiceGrant improvements


cswendrowski:
In GitLab by @TikaelSol on Jan 1, 2022, 22:26

This is a tracking issue for what items have had ChoiceSet/GrantItem added to them

- [x] Classes/Class Features
- - [x] Alchemist
- - [x] Barbarian
- - [x] Bard
- - [x] Champion
- - [x] Cleric
- - [X] Druid
- - [X] Fighter
- - [x] Gunslinger
- - [x] Inventor
- - [x] Investigator
- - [x] Magus
- - [x] Monk
- - [x] Oracle
- - [X] Ranger
- - [X] Rogue
- - [x] Sorcerer
- - [x] Summoner
- - [X] Swashbuckler
- - [X] Witch
- - [X] Wizard

- [x] Class Feats
- - [x] Alchemist
- - [x] Barbarian
- - [x] Bard
- - [x] Champion
- - [x] Cleric
- - [x] Druid
- - [x] Fighter
- - [x] Gunslinger
- - [x] Inventor
- - [x] Investigator
- - [x] Magus
- - [x] Monk
- - [x] Oracle
- - [x] Ranger
- - [x] Rogue
- - [x] Sorcerer
- - [x] Summoner
- - [x] Swashbuckler
- - [x] Witch
- - [x] Wizard

- [ ] Other Feats
- - [x] Skill Feats
- - [x] General Feats
- - [x] Ancestry Feats
- - [x] Heritages
- - [ ] Advanced multiclass feats

- [ ] Spell Effects
- - [x] A - J
- - [ ] K - Z
- [ ] Equipment

TikaelSol:
Not sure why I closed this, it is still open

Lioheart:
I looked for a way to manually create this feat, but it is impossible without the appropriate changes to the system.

You need to create a self:halflevel variable (along the lines of self:level). This will make it possible to use it in the feat.
![obraz](https://github.com/foundryvtt/pf2e/assets/39154364/e7bfe211-367b-45f9-98a7-2c299e87fbf7)

For example, in the Advanced Trickery feat when I inserted the following condition, I could select level feat up to the character's level (example for a 14th level character):
```
"lte": [
                "item:level",
                "self.level"
              ]
```
![obraz](https://github.com/foundryvtt/pf2e/assets/39154364/0b567ce3-7a6f-406b-9f62-f54b59e7d7c8)

[...]

PR #7260, opened 2023-04-02
Merge Lay on Hands into one spell

LebombJames:
May need a migration or 68


LebombJames:
Due to the deletion of the undead version of the spell and effect, some migration is needed to not break the spell on NPCs. Should be done as part of #1956

packs/data/spell-effects.db/spell-effect-lay-on-hands.json
```diff
@@ -17,10 +18,26 @@
         },
         "rules": [
             {
+                "hideIfDisabled": true,
                 "key": "FlatModifier",
+                "predicate": [
+                    "self:mode:living"
```
stwlam, line 24: This should predicate on what the target is rather than the caster

stwlam: Both also need to pivot on negative healing.

LebombJames: How exactly does negative healing interact with it? The healing would be affected, but that's taken care of by the spell variants. How does that relate to the AC bonus/penalty?

I imagine the predicates would be `living` OR NOT `negative healing`, and `undead` OR `negative healing`?

Issue #5850, opened 2023-01-22
QOL UX feature: right-click to 'unset' actor sheet fields


anathemamask:
reporting on a UX painpoint for my players fresh to the system

it would be convenient if a contextMenu() click on an actor sheet's fields that are set by drag-and-drop prompted to remove that field.

Drag and drop setting of class, for example, applies that class but removes the compendium browser button for applying another, so if you accidentally grab the wrong class you cannot remove it, having to open the compendium from another sheet or the compendium menu to resolve this isn't an easy UX to intuit.

An alternative would be to make right-clicks open the compendium for the appropriate section, as an alternative.

CarlosFdez:
At the very least, dragging a new class does remove the previously added class and all class features added by the previous one, so until this exists, people can fix any such errors by dragging the class they actually want onto their sheet.

PR #7269, opened 2023-04-02
Add ability to delete Ancestry/Heritage/Class/Background/Deity from PC sheet

JulieWinchester:
This PR resolves https://github.com/foundryvtt/pf2e/issues/5850, though it implements a different solution than the one suggested in the issue (right-clicking to delete field value).

- For Ancestry, Heritage, Class, Background, and Deity fields on the PC sheet, the "edit item" icon has been dropped in favor of a "delete item" icon. This delete item icon uses the standard delete item code used elsewhere in the system, so after clicking, it pops up a confirm box to verify the deletion. 
- To retain the edit/view functionality of the previous edit item icon, field values have been made clickable to bring up the edit item interface. Mouse switches to cursor when hovering over the value, and the values have the same hover behavior as feat names (change to red color, slight text-shadow). For what it's worth, I did play around with keeping both edit-item and delete-item icons, but it seemed to take up a bit too much space. 
- Repeated template code for the four main fields has been replaced by a partial. This could easily be dropped, or it could be taken further by modifying how the `sheetData` properties for these fields get put together, which would reduce the need for so many manually specified partial sub-properties in the template code. But for now, this PR modifies the template structure while leaving `sheetData` alone.

For more info, I'm on discord as JellyfishJail#3956. I'll put a couple of pictures to demonstrate the changes. 

Screen Shot 2023-04-02 at 4 52 52 PM

With mouse hovering over Ancestry field value
Screen Shot 2023-04-02 at 4 53 03 PM


TikaelSol:
We need to be able to open the item for editing still.

JulieWinchester:
> We need to be able to open the item for editing still.

Clicking the item's text label opens the item for editing. In my last screenshot my mouse (not visible in the screenshot sadly) is hovering over the "Aasimar" heritage text label, and clicking it would open the heritage item.

TikaelSol:
May be best to make it a hamburger menu that opens to edit / remove

JulieWinchester:
Sure, that sounds good. Do you have thoughts about the icon appearing on hover (which is what the delete icon does currently) vs being present always?

JulieWinchester:
@TikaelSol this latest update uses a three-dot tooltip menu to bring up explicit Edit Item and Delete Item options. I've added a screenshot below. 

It doesn't use any of the disappear-until-hover behavior that the previous version had, though I can add that back in. 

I used the three dots icon for the menu instead of something like three horizontal bars because three dots are used elsewhere to signal a modal menu for PC sheet tab customization, and also because the three-bar icon is used elsewhere on the PC sheet for moving/ordering spellcasting sheet sections. 

Screen Shot 2023-04-09 at 6 18 53 PM

stwlam:
We might want to switch to using Foundry's context menu API for the UI component, but the feature itself looks good. Thanks!

static/lang/en.json
```diff
@@ -351,6 +351,7 @@
         "AlignmentNG": "Neutral Good",
         "AlignmentTitle": "Your alignment",
         "AllSavesLabel": "All Saves",
+        "Ancestry": "Ancestry",
```
stwlam, line 354: Use "ITEM.TypeAncestry" for this instead of adding a new localization key.

JulieWinchester:
@stwlam Just pushed an update to use the Foundry ContextMenu API. It still behaves like before - so the context menu comes up on left-click (not right click as is default for ContextMenu), like with other uses of the three-dot menu in the character sheet, etc. ContextMenu is a bit finicky in some ways, like I wasn't able to figure out how to slap `item-edit` or `item-delete` classes on context menu elements in order to take advantage of existing event listeners. So there's a bit more finagling to accomplish the actual edit and delete behavior, but hopefully not too much.

Screen Shot 2023-04-13 at 9 17 51 PM

PR #7289, opened 2023-04-05
Add initial work on Party Sheet

CarlosFdez:
image
image
image
image


stwlam:
Thinking on what would be useful for a GM to see about the party at a glance, I'm thinking we could jettison saving throw modifiers and ability scores, move hitpoints under the profile pictures, and then utilize the space to the right for top skills and languages.

CarlosFdez:
Eventually the languages at the top (not shown) will, when moused over, show who speaks the language. I don't think languages per player is useful given that will exist.

Might as well show all skills, or strip out almost everything so you can see all 4-5 players simultaneously.

CarlosFdez:
This seems to occur for ancestries that get "upgraded" by a heritage
image

This doesn't seem to happen if it occurs for things that get darkvision innately like fetchling.

Not sure if we should consider cleanup nor where, but I think this might end up being a bit of a problem. Would like to handle it separately from this PR though depending on where we do it.

CarlosFdez:
> This seems to occur for ancestries that get "upgraded" by a heritage image
> 
> This doesn't seem to happen for things that get darkvision innately like fetchling.
> 
> Not sure if we should consider cleanup nor where, but I think this might end up being a bit of a problem if we leave it as is. Would like to handle it separately from this PR though depending on where we do it.

This is now handled by the sheet. Some actors having only darkvision but others having both that and low light may be inconsistent, but its normal for PF2e. However its redundant for the sheet, so its now removed at the sheet level.

static/assets/sheet/accent-decoration.png
stwlam, line 1: webp!

stwlam:
Conflicts!

PR #7299, opened 2023-04-06
render the dice to use in the roll window

asuffield:
Previously, the button would show "Roll (+9)"; after this change, it will show "Roll (1d20+9)", correctly resolving substitutions and fate as options are changed.

Do a little refactoring and cleanup while we're here.


stwlam:
Mind posting a screenshot for this?

PR #7307, opened 2023-04-07
Use `declare` instead of non-nullish assertions on class fields 

stwlam:
Eventually we'll need to flip on `useDefineForClassFields` in tsconfig.json, which will actually emit class fields in the transpile result. Using `declare` will prevent that, preserving the current behavior.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -87,33 +87,33 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
     private constructed = true;
 
     /** Is this actor preparing its embedded documents? Used to prevent premature data preparation of embedded items */
-    preparingEmbeds?: boolean;
+    preparingEmbeds = false;
 
     /** Handles rolling initiative for the current actor */
-    initiative?: ActorInitiative;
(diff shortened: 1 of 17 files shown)


src/module/actor/base.ts
```diff
@@ -543,6 +543,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
 
     protected override _initialize(): void {
         this.constructed ??= false;
+        this._itemTypes = null;
```
CarlosFdez, line 546: I don't know if handling it here is safe. _itemTypes is a cache, so it needs to be killed every data prep, not just the first.

stwlam: `_initialize` is what calls `prepareData`

CarlosFdez: Seems to indeed be the case after testing, but that does mean the comment is possibly wrong?

> Perform one-time initialization tasks which only occur when the Document is first constructed.

CarlosFdez: Since the only thing that gave me pause is foundry docs being a liar liar pants on fire I can push button

PR #7322, opened 2023-04-08
Fix dragging of content links in inventory item description

In3luki:
~I'm not really happy with creating another `Sortable` instance but this is the only way I could get it to work.~
Found a better way.


stwlam:
Could there be a handle selector of `"*:not(.clicky-place)"`?

PR #7324, opened 2023-04-08
Fix action promise resolving

nikolaj-a:
Fix action promises resolving before all results have been collected, and update the Drop Prone and Stand actions to utilise that correct resolving to apply or remove the Prone condition accordingly.

### DIFF
--- a/src/module/actor/actions/simple.ts
+++ b/src/module/actor/actions/simple.ts
@@ -102,4 +102,4 @@ class SimpleAction extends BaseAction<SimpleActionVariantData, SimpleActionVaria
     }
 }
 
-export { SimpleAction, SimpleActionVariantData };
+export { SimpleAction, SimpleActionResult, SimpleActionVariant, SimpleActionVariantData, SimpleActionUseOptions };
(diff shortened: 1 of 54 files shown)


src/module/actor/actions/simple.ts
```diff
@@ -47,7 +53,8 @@ class SimpleActionVariant extends BaseActionVariant {
             actors.push(...getSelectedOrOwnActors());
         }
         if (actors.length === 0) {
-            return ui.notifications.warn(game.i18n.localize("PF2E.ActionsWarning.NoActor"));
+            const reason = game.i18n.localize("PF2E.ActionsWarning.NoActor");
+            return Promise.reject(reason);
```
stwlam, line 57: Prefer an exception throw over a promise rejection if nothing is going to catch it with `Promise.catch()`

nikolaj-a: Fixed, and it made me realise that we should never show those UI notifications from within those action implementations. I've moved the UI notifications into the functions used by the current action macros instead, so we don't have a functionality gap until those functions are deprecated and removed.

The action implementations will return a rejected promise instead, and it's up to the caller, like the sheets, to trigger the UI notifications in case of an error.

PR #7326, opened 2023-04-08
Convert Thievery skill actions

nikolaj-a:
Disable Device, Palm an Object, Pick a Lock, and Steal.

Also support adding modifiers, predicated or otherwise, that are available for an action by default. Like the -5 penalty for pocketed or protected items when Stealing.


src/module/actor/actions/single-check.ts
```diff
@@ -137,13 +145,15 @@ class SingleCheckActionVariant extends BaseActionVariant {
 
 class SingleCheckAction extends BaseAction<SingleCheckActionVariantData, SingleCheckActionVariant> {
     readonly difficultyClass?: CheckDC | string;
+    readonly modifiers: RawModifier[];
```
nikolaj-a, line 148: Using a `RawModifier` instead of an actual `ModifierPF2e` instance since the localization data is not yet available when the bundled actions are initialized and the ModifierPF2e constructor tries to immediately localise the label/name.

PR #7330, opened 2023-04-09
Correct test for when flanking is applied

### LINKED ISSUE #7318: [Bug] Flanking is applying flatfooted to strikes made with ranged weapons within melee range
Actors in this configuration:

produce this result:

screenshots were taken in a high-module world, but I tested with no modules active, just dropped valeros and ezren on opposite sides of a goblin warrior, same outcome.

asuffield:
This used to accidentally work because reach wasn't defined on ranged weapons, but that has since been changed. Explicitly check the isMelee property, which is based on ranged increments instead of reach.

fixes #7318

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -778,14 +778,14 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
                   ]
                 : [null, null];
 
-        const reach = params.item?.isOfType("melee")
-            ? params.item.reach
+        const [reach, isMelee] = params.item?.isOfType("melee")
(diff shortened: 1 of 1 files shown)


stwlam:
I'm not sure I follow the explanation. Reach still isn't defined on weapon.

asuffield:
It's over here, if there's no reach attribute then this.attributes.reach.base is used: https://github.com/foundryvtt/pf2e/blob/31b2cd14d17c83ec00c0ed9eccc1c8f443b8baff/src/module/actor/creature/document.ts#L134

That means any weapon will get a number for reach.

stwlam:
You're not making any sense, but I see you copied the code from when it did work. That'll drop the job, thanks!

PR #7334, opened 2023-04-09
korean translation started

Leafsan:
Korean translation started for fvtt


stwlam:
Thank you, but the system doesn't host other languages: for that you'll want to make a module. Please join us at https://discord.gg/Dv8YkbHZqM so we can help get you started.

PR #7345, opened 2023-04-10
Fix initiative rolls predicated on certain checks

CarlosFdez:
Fixes Elven Instincts and Battlefield Surveyor


stwlam:
Pretty gross that this just a top-level roll option, but something to think about addressing later.

PR #7409, opened 2023-04-15
Add Aid action implementation

nikolaj-a:
Add Aid action implementation, and expand the Cooperative Nature ancestry feat with a flat modifier rule element giving a circumstance bonus to Aid checks.

Also add support for expanding the list of traits when an action is used.


packs/data/feats.db/cooperative-nature.json
```diff
@@ -19,7 +19,17 @@
         "prerequisites": {
             "value": []
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "selector": "all",
```
stwlam, line 25: This is going to appear in damage rolls as well

nikolaj-a: We have a selector for everything but damage rolls?

nikolaj-a: Updated

PR #7504, opened 2023-04-22
Remove `LocalizePF2e` from `@module/apps`, `@module/canvas`, `@module/chat-message`, `@module/encounter`

In3luki:
I'm using an import of `en.json` for the world clock app.  Not sure if that's a bad idea. Seems to work as a drop in replacement for `LocalizePF2e` in that case though.


stwlam:
Using en.json directly would use english localization for everything

In3luki:
Oh right. I'll expand `CONFIG.PF2E.worldClock` then.

In3luki:
Updated. I've added a new utility function to help with creating a `CONFIG` object from `en.json`.

In3luki:
Updated

PR #7507, opened 2023-04-22
Search `CreaturePF2e#spellcasting` in `CreaturePF2e#getStatistic`

stwlam:


src/module/actor/character/document.ts
```diff
@@ -730,7 +730,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
             entry.system.ability.value = baseStat.ability ?? entry.system.ability.value;
             entry.system.proficiency.value = Math.max(entry.rank, baseStat.rank ?? 0) as ZeroToFour;
             entry.statistic = baseStat.extend({
-                slug: entry.slug ?? sluggify(entry.name),
+                slug: entry.slug ?? sluggify(`${entry.name}-spellcasting`),
```
CarlosFdez, line 733: arcane-spontaneous-spells-spellcasting?

stwlam: or athletics-spellcasting depending on what they name it

CarlosFdez: Outta have a neater way to handle it I imagine, but I can't imagine it being trivial to make it check last for a middle-of-the-road one.

PR #7508, opened 2023-04-22
Remove remaining uses of `LocalizePF2e`

In3luki:


src/module/system/localize.ts
```diff
@@ -8,20 +7,4 @@ type TranslationsPF2e = Record<string, TranslationDictionaryValue> &
...
-        }
-        if (this._translations === undefined) {
-            this._translations = mergeObject(game.i18n._fallback, game.i18n.translations, {
-                enforceTypes: true,
-            }) as TranslationsPF2e;
-        }
-        return this._translations;
-    }
-}
+export { TranslationsPF2e };
```
stwlam, line 10: Could this be moved to an existing file?

In3luki: It's not really used anywhere but eslint complains if one of the json files isn't imported. 

image

stwlam: If `TranslationsPF2e` isn't used anywhere, can be kill the whole file?

In3luki: Yeah, if I can figure out how to appease eslint,

stwlam: Push it up with eslint complaining

In3luki: Done

PR #7509, opened 2023-04-22
Support negative modifiers (and dice) in spell combine terms

CarlosFdez:
Negative dice, when combined, will now reduce dice, but it will never go below 0 (once it hits 0, the dice are removed from the formula). Removing dice wasn't the goal, but it was a neat side effect.

### DIFF
--- a/src/scripts/dice.ts
+++ b/src/scripts/dice.ts
@@ -1,6 +1,6 @@
 import { ActorPF2e } from "@actor";
 import { ItemPF2e } from "@item";
-import { ErrorPF2e } from "@util";
+import { ErrorPF2e, groupBy, sortBy, sum } from "@util";
 
 /**
  * @category Other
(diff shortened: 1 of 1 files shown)


src/scripts/dice.ts
```diff
@@ -200,23 +200,52 @@ function combineTerms(formula: string): string {
...
+                if (termToModify.modifier) termToModify.modifier *= -1;
+                if (termToModify.diceNumber) termToModify.diceNumber *= -1;
+            }
+            return result;
+        }
+
+        result.unshift({
+            modifier: term instanceof NumericTerm ? term.number : 0,
+            diceFaces: term instanceof Die ? term.faces : 0,
+            diceNumber: term instanceof Die ? term.number : 0,
```
stwlam, line 228: How about putting this in its own object instead of inventing the 10th new way to express dice number and faces? ;)

CarlosFdez: The entire function is temporary. To even work on afflictions, I'll need to standardize this and melee items to use the formula.ts pipeline. That's a major refactor though, and it'd be nice to get your elite/weak thing in sooner rather than later.

PR #7527, opened 2023-04-24
Conditionally count peachwood as darkwood for IWR purposes

stwlam:


stwlam:
> Peachwood is only considered darkwood for undead creatures. As there any non-undead with darkwood weakness?

I changed it to account for homebrew, but there are no creatures at all with a weakness to darkwood. The Provincial Jiang-Shi is the only creature with a resistance exception to it.

Issue #6588, opened 2023-02-24
Compendium Browser Improvements


Zullock:
A proposal for a number of changes and the reason, if there are some that are not needed or can be done already in some way they can be marked as done.

General:

- [ ] Instead of source, show the Pack, once one is chosen, show the different sources in that pack. This would maintain the functionality while lessening the current bloat.
- [ ] Trait Search bar: Add Clear Filter button
- [ ] Society: Add filter and icon showing if something is PFS Standard, Limited or Restricted. Would require data entry.

Action Tab:

- [ ] Number of actions: Filtering by actionType (Action, Reaction, Free, and Passive) and if the option “Action” is picked, by 1, 2 or 3 actions.

Bestiary Tab:

- [ ] Spellcaster: Whether a creature has spells or not.
- [ ] (Dev only) Ability to search for individual stats: This could help find errors in data entry. 

Equipment Tab:

- [ ] Usage: Ability to filter by usage (Held, Worn or Affixed/Etched). If Held also the number of hands, if Worn also where they are worn, if affixed/etched in what kind of equipment.

  Weapons
  - [ ] Specific Magic Weapon: For searching those specifically. Right now it can be done with the trait Magical and picking in Weapon Filters the wanted (or all) of the category, but maybe an easier method is also welcomed.

  Armor:
  - [ ] Specific Magic Armor: Same as Specific Magic Weapon but for armors.

Feats Tab:

- [ ] Skill Feats: Indicate somehow (maybe text, like the one for rarity) if you need trained, expert, master or legendary
- [ ] Prerequisite: Indicate with some icon if the feat has prerequisites or not

  Feat Types:
  - [ ] Curses: Right now it's empty, should it be deleted?
  - [ ] Deity Boons: Right now it's empty, either delete or rename and include here the Divine Intercessions
  - [ ] Society Boons: Right now it's empty, should it be deleted or are we missing the data? What is expected to go here?

Hazard Tab:

- [ ] Source Filter is currently broken.

Spell Tab:

- [ ] Cast time: Filter by cast time of the spell, for categories maybe something like: reaction, 1 action, 2 actions, 3 actions, variable (for cases like Heal), 1+ minute.
- [ ] Saving Throw: For which saving throw they target if any, right now there is no good way to find this.
- [ ] Heightenable: If the spell can be heightened or not.

TikaelSol:
- [ ] Add filter for spell range, defense #15561, #11511
- [ ] Allow for preset filters #14252
[...]

PR #7536, opened 2023-04-24
Fix Hazard Source in Compendium Browser

glont:
The field for the source filter of hazards in the compendium browser seemed to point to the wrong field, which resulted in the filter not being populated. It did use `system.details.source.value`, but it seems that the source is actually stored in `system.source.value`. 

Also mentioned in #6588


stwlam:
This would be better handled by migrating this bit of hazard data to bring it in line with NPC data.

PR #7547, opened 2023-04-24
Refactor spells to use regular damage pipeline

CarlosFdez:
Enables the ability to add damage dice to spells. Also sneaky fix so that elite only applies to the first instance of damage.

### DIFF
--- a/src/module/system/damage/types.ts
+++ b/src/module/system/damage/types.ts
@@ -55,10 +55,14 @@ interface DamageRollContext extends BaseRollContext {
 }
 
 interface DamageFormulaData {
-    base: BasicDamageData;
+    base: BaseDamageData | BaseDamageData[];
     dice: DamageDicePF2e[];
     modifiers: ModifierPF2e[];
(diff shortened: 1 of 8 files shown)


CarlosFdez:
DynamicBaseDamageData needs renaming. Spell isn't quite right. Its not a string either and its not from a formula. BaseTermsDamageData?

src/module/system/damage/types.ts
```diff
@@ -55,18 +55,29 @@ interface DamageRollContext extends BaseRollContext {
 }
 
 interface DamageFormulaData {
-    base: BasicDamageData;
+    base: BaseDamageData | BaseDamageData[];
```
stwlam, line 58: Just make it an array from everywhere

CarlosFdez: Sure. It'll blow up the lines of code but we can do that.

CarlosFdez:
In retrospect I'm considering renaming combineTerms() to simplifyFormula(). That way it'll be more distinguished from combineSimpleTerms() which just combines terms into a new list of combined terms.

PR #7562, opened 2023-04-26
Add backwards compat for attributes.initiative.roll, and pass along rollInitiative() args

CarlosFdez:
They outta be using actor.initiative.roll() if they wanna pass params but sure we can be exhaustive.


CarlosFdez:
Added backwards compat to this one, but I"m heavily regretting calling it extraRollOptions instead of options, because its confusing everyone.

CarlosFdez:
Any issues with this? If we wanna drop it then sure, so long as someone else takes rightful ownership of the macroslayer title.

PR #7576, opened 2023-04-28
Render parties like folders in sidebar

CarlosFdez:
Several ideas floated around for how to handle the case for where there exist multiple parties, but until we get more input from people who run such games this is currently the way I'd like to handle it.

So a new party list is created and injected during rendering into the actor list. This avoids us having to replace the sidebar template. The following scenarios are covered (unless I messed up somewhere):
1) Drag into party adds as member into party
2) Drag away from party removes as member
3) Party can be resorted (members are always alphabetical).
4) Right click party does what's intended
5) Right click a member of a party adds Remove Member to the options
6) Clicking on party icon opens the sheet. Clicking on the name collapses/hides the "folder".
7) Clicking on add actor on the party opens up an actor creation dialogue with reduced choices (creature types only)
8) Dragging the party's header into the scene creates a token for that party.

image

Eventually I would like a more distinct header for party, but I'd rather cross that bridge later.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -456,6 +456,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         context?: {
             parent?: TDocument["parent"];
             pack?: Collection<TDocument> | null;
+            types?: (ActorType | "creature")[];
         } & Partial<FormApplicationOptions>
     ): Promise<TDocument | null>;
     static override async createDialog(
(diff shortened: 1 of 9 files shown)


stwlam:
icon suggestions:
https://fontawesome.com/icons/people-group?f=classic&s=solid
https://fontawesome.com/icons/users?f=sharp&s=solid

CarlosFdez:
For what? Creating actors as members or removing members?

Using a bus icon for remove member since they were put on a bus!

stwlam:
The sidebar

CarlosFdez:
So the icon on the right to add new members? I tried using a similar one to the folder's add actor, but I guess we can do something wildly different. I'll change that later then.

stwlam:
I mean in place of the game-icons svg

CarlosFdez:
Game icons SVG is the default actor image. But after discussion since the image is smaller, we can remove the thumbnail. We can always bring back the actor image if the styling changes by enough that the icon becomes chunky.

CarlosFdez:
Changed to fa icon instead of using the actor image, and made the whole thing smaller to fit the folder size. I don't think it'll stay that way but it pushes the discussion on the actual appearance to a future date.

PR #7577, opened 2023-04-28
Add some Firebrands and Knights of Lastwall feats to archetypes journals

pedrogrullada:
add Firebrands feats to Acrobat, Alter Ego and Archaeologist archetype journals, correct some listed levels

add Knights of Lastwall feats to Archer, Assassin and Bastion archetype journals


TikaelSol:
This removed the explicit newlines, but some were missing already anyways, so I guess we'll just clean that up later.

PR #7585, opened 2023-04-28
Add support for "carried" usage, disable rule elements for dropped items

stwlam:
- An item with a "carried" usage is considered equipped so long as it's not dropped
- A dropped item is never counted as equipped or invested
- All REs are disabled for dropped items, even if they don't require equipping

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -2045,7 +2045,8 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
 
         // If investing and unequipped, equip first
         if (!invested && !item.isEquipped) {
-            await this.adjustCarryType(item, item.system.usage.type, item.system.usage.hands, true);
+            const newCarryType = item.system.usage.type === "carried" ? "worn" : item.system.usage.type;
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Quick rules check, but how do cursed items work when dropped?

stwlam:
They're supposed to no be removable in the first place, but we don't have any enforcement of it.

CarlosFdez:
How much do you expect the logic would simplify if the carry type options were created in prepareInventory()? The template is quite hairy.

stwlam:
> How much do you expect the logic would simplify if the carry type options were created in prepareInventory()? The template is quite hairy.

I have no idea. The template doesn't strike me as _too_ bad, especially in its own file.

CarlosFdez:
Took me a bit to actually muster up some time and will to stare at it but I think its just the selected logic that scares me a bit. Seems fine though.

CarlosFdez (approved):

stwlam:
I'm not sure what you mustered up the will to stare at. None of that is touched here.

CarlosFdez:
Because I was a fool and forgot to turn on "hide whitespace"

PR #7589, opened 2023-04-28
Fix scrolling through long descriptions when GM notes are present

stwlam:
We can sit on this one until after 4.11 patching is done


CarlosFdez:
If you're confident in it I think it should be fine to merge since its a bug fix.

stwlam:
I'm not, though: there are always surprises with these changes. I'd like people to have a chance to look at different item sheets and confirm that nothing unexpected happened.

Issue #4534, opened 2022-10-23
There should be a visual indication of flanking arrangements


ThingsCouldGetDicey:
I would like an indication for players to see when flanking is in effect. I think it would help alot of gms and players to see that when players are flanking a creature or vice versa that a satus icon appears telling them that the creature is flanked. and also when they decide to attack that it shows the flanking bonus in the roll dialogue. 

it would be a nice ui improvement.

cepvep:
As discussed on the #pf2e-system channel, while it shouldn't put flat-footed on the token, or even claim the target is flat-footed in order to avoid leaking info for all around vision etc,I think some visual indicator for the person targeting would be nice, just to verify that the position is valid for flanking.. 

In my mind, a small box that appears on the targeting indicator when flanking, that has hover text stating some variant of "You are in position to flank this creature". High quality Pic attached as an example


![ss (2022-10-23 at 02 35 05)](https://user-images.githubusercontent.com/116499012/197395875-d141d344-7453-4399-bc60-55cef7d6d4af.png)

ThingsCouldGetDicey:
yes i like the idea for it only being visible for the gm and the players who get into a flanking position.

other pcs who are not in flanking range should not see if its possible to flank or not to prevent metagaming.

more accuratley to my claim id like to have a visual indication more than anything.

JulieWinchester:
Maybe this issue should be closed as completed?

PR #7591, opened 2023-04-29
Add visual indication of token flanking position when targeting

JulieWinchester:
Addresses https://github.com/foundryvtt/pf2e/issues/4534.

This PR adds a scrolling floaty text "Flanking Position!" indicator on a token when that token targets another enemy token while possibly in flanking position. It's too big to embed, but I've uploaded a [video demonstrating this feature](https://imgur.com/a/cMrVIMZ). The indicator does not reveal whether or not the enemy token is actually flanked or is even flankable, in order to not give up possible metagame information (as discussed in the issue). Rather, this indicator shows that the targeting token _could_ be in flanking position, if it is possible to flank the enemy token. The indicator respects reach and ally/enemy relationships. 

The solution implemented is a bit different from that proposed in the issue. The issue suggested drawing a different target icon on the targeted token, with an additional hover-text note. Instead, this PR implements the indicator as temporary floating text on the targeting token. My thinking here is that while it's easy to figure out if a token is in a flanking position when the targeting is initiated, it's much more complicated to know whether that flanking position is maintained, since other tokens may move or die or experience conditions that prevent the original token from continuing to be in flanking position. Because of this, changing the targeting icon to represent flanking position either requires continually checking flanking position against many different changing circumstances or risking that the flanking-position targeting icon is displayed when no longer valid. Also, since it's the targeting token that is in position against the target, having the indicator appear over the targeting token seems to make more sense to me. 

For any questions or anything, I'm on discord as JellyfishJail#3956.

### DIFF
--- a/src/module/canvas/token/object.ts
+++ b/src/module/canvas/token/object.ts
@@ -62,21 +62,22 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
 
     /**
      * Determine whether this token can flank another—given that they have a flanking buddy on the opposite side
-     * @param flankee       The potentially flanked token
-     * @param context.reach An optional reach distance specific to this measurement */
-    canFlank(flankee: TokenPF2e, context: { reach?: number } = {}): boolean {
+     * @param flankee                  The potentially flanked token
+     * @param context.reach           An optional reach distance specific to this measurement
+     * @param context.ignoreFlankable Optionally ignore flankable (for targeting indicator) */
+    canFlank(flankee: TokenPF2e, context: { reach?: number, ignoreFlankable?: boolean } = {}): boolean {
         if (this === flankee || !game.settings.get("pf2e", "automation.flankingDetection")) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


CarlosFdez:
I don't know if shark has made contact with you on discord, but I'll try to look into it later after the next release is in. I kinda feel bad letting this one sit for so long without feedback.

I'm unsure about using a token notification to do this, but its probably the simplest way. The only other thing that comes to mind is flashing a yellow glowy circle under your token and the flanking buddy, and a red glowy circle under the enemy circle, but that's very likely more involved to implement.

MrVauxs:
When it comes to visuals, I dislike the floaty text as that's rather vague. I don't think putting "Flanked" above the flanked token is bad if it happens over every token, even if it's not flankable.

However more elaborate solutions (like animated arrows or popups) feel a bit too much for something to be in the system by default. It would be great to have this sort of thing as a Hook though, so modules like PF2e Animations can make more elaborate and customizable animations for such notifications.

JulieWinchester:
I am closing this PR because I believe it should be superseded by https://github.com/foundryvtt/pf2e/pull/9991, which addresses the same issue but using an on-demand highlight visual indicator.

PR #7595, opened 2023-04-29
Fix Keen Recollection using old EnsureProficiencyOptions rollOptions

Cerapter:
As the title says. `Statisic` uses the various `proficiency:XYZ` rollOptions instead.


packs/data/iconics.db/quinn-level-5.json
```diff
@@ -3949,4 +3949,4 @@
         }
     },
     "type": "character"
-}
+}
```
stwlam, line 3952: Restore this line break at once!

Cerapter: Put that shit in the Prettier config file. :wink:

PR #7601, opened 2023-04-29
Update One Inch Punch to use RollOption suboptions

LebombJames:
Also ternaries while we're at it

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -1979,8 +1979,8 @@
             },
             "Monk": {
                 "OneInchPunch": {
-                    "ThreeActionLabel": "One Inch Punch (Three Actions)",
-                    "TwoActionLabel": "One Inch Punch (Two Actions)"
+                    "ThreeActionLabel": "Three Actions",
+                    "TwoActionLabel": "Two Actions"
                 },
                 "SacredKi": {
                     "Prompt": "Choose your aligned damage."
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -1157,6 +1157,10 @@
             "Woodworking": "Woodworking"
         },
         "SpecificRule": {
+            "Actions": {
+                "ThreeActions": "Three Actions",
+                "TwoActions": "Two Actions"
+            },
             "AdvancedGeneralTraining": {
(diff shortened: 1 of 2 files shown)


static/lang/re-en.json
```diff
@@ -1979,8 +1979,8 @@
             },
             "Monk": {
                 "OneInchPunch": {
-                    "ThreeActionLabel": "One Inch Punch (Three Actions)",
-                    "TwoActionLabel": "One Inch Punch (Two Actions)"
```
TikaelSol, line 1983: Problem with this change is it will suddenly look weird on actors with the old version of the feat. Also `One Action`, `Two Actions`, and `Three Actions` should be generic enough to go in a generic area (and probably exist already).

LebombJames: There doesn't seem to be "One/Two/Three Actions" in en.json. Worth adding them?

TikaelSol: I think so, yeah. Add them in re-en.json under SpecificRule.

static/lang/re-en.json
```diff
@@ -1980,7 +1980,11 @@
             "Monk": {
                 "OneInchPunch": {
                     "ThreeActionLabel": "One Inch Punch (Three Actions)",
-                    "TwoActionLabel": "One Inch Punch (Two Actions)"
+                    "TwoActionLabel": "One Inch Punch (Two Actions)",
+                    "Suboptions": {
+                        "TwoActions": "Two Actions",
+                        "ThreeActions": "Three Actions"
+                    }
```
TikaelSol, line 1987: Let's move these up a few levels so they are just under SpecificRule to be more generic, since we can use them in other places.

static/lang/re-en.json
```diff
@@ -2320,6 +2320,7 @@
                 },
                 "TalismanEsoterica": "Talisman Esoterica"
             },
+            "ThreeActions": "Three Actions",
```
TikaelSol, line 2323: I meant the whole thing under the same structure. So that it's `PF2E.SpecificRule.Actions.TwoActions`.

LebombJames: Oops, my bad

PR #7602, opened 2023-04-29
Use cached check data in `CheckPF2e.rerollFromMessage` instead of recreating the check from the message flag

In3luki:
Reroll by keeping an in-memory cache of the original `CheckModifier` and `CheckRollContext`. I'm not 100% that this is viable but it didn't take long to implement and actually fixes all the reroll problems.

Pros:
- No need to special case every additional `CheckPF2e` property.
- No need to make `Statistic`, `RollNotePF2e`, etc. fully serializeable.

Cons:
- Reloading wipes the cache.
- GMs can't reroll for players anymore.

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -45,6 +45,8 @@ type CheckRollCallback = (
 ) => Promise<void> | void;
 
 class CheckPF2e {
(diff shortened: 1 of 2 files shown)


stwlam:
We have the flag precisely because we can't rely on the user not reloading. I think any solution is still going to have to work with the flag.

PR #7644, opened 2023-05-02
Expose property rune slots on unowned armor when ABP is enabled

stwlam:
(given property runes are allowed)


CarlosFdez:
What would be the downside of making getPropertySlots() wise to null actors?

stwlam:
> What would be the downside of making getPropertySlots() wise to null actors?

Updated to utilize

PR #7666, opened 2023-05-03
Remove the death overlay if present upon hit points being increased

stwlam:


CarlosFdez:
I was going to suggest adding a parameter to force a defeated state to toggleDefeated(), but toggleDefeated only applies to a single token and this unsets it from all, so not quite the same.

CarlosFdez (dismissed):

PR #7687, opened 2023-05-05
Ensure inline type parameter is sluggified

TikaelSol:
This is not really a bugfix, more making it slightly more user friendly. It's expanding the system to accept `@Check[type:Athletics]` or `@Check[type:Engineering Lore]` without causing issues where the skill isn't found. This fixes some problems with the AV module journals as they have capitalized skill names, but enabling lore skills without needing an explicit label is actually useful for the system.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -370,7 +370,7 @@ class TextEditorPF2e extends TextEditor {
                           })
                           .join(" ");
                 html.innerHTML = inlineLabel ?? skillLabel;
-                html.setAttribute("data-pf2-check", params.type);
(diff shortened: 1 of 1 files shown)


src/module/system/text-editor.ts
```diff
@@ -370,7 +370,7 @@ class TextEditorPF2e extends TextEditor {
                           })
                           .join(" ");
                 html.innerHTML = inlineLabel ?? skillLabel;
-                html.setAttribute("data-pf2-check", params.type);
+                html.setAttribute("data-pf2-check", sluggify(params.type));
```
stwlam, line 373:
```suggestion
                html.dataset.pf2Check = sluggify(params.type);
```

PR #7690, opened 2023-05-05
Condense Divine Vessel effects

### LINKED ISSUE #7046: Divine Vessel does not grant the Enlarge spell effect
Each version of Divine Vessel should also grant the effects of Enlarge, but this is not automated as part of the item within the compendium. I suggest adding a Grant Item rules element to each version of Divine Vessel.

TikaelSol:
Divine Vessel is likely slated for huge changes in the future but for now this cleans it up and closes #7046


TikaelSol:
This shouldn't need a migration since I'm just renaming one of the effects, but we can always write one to actually remove the 6 now defunct effects.

PR #7702, opened 2023-05-05
Suppress rune effects when the greater version of the rune is also present

### LINKED ISSUE #7654: Multiple etchings of the same rune should not stack

TikaelSol:
Closes #7654

`If a suit of armor or a weapon has multiple etchings of the same rune, only the highest-level one applies.`

We could suppress it another way but if an effect grants the flaming rune, such as from Flametongue's activation we need it to not stack with an existing greater rune.


stwlam:
Would this leave multiples of the exact same rune stacking?

TikaelSol:
No, the system doesn't stack multiple runes of the same type.

stwlam:
image

TikaelSol:
Is that a weapon you set flaming to 3 times in the menu? That stacking would need a different fix and really that's more a weird user behavior. This is mostly to prevent effects adding a flaming rune with REs to a weapon that already has greater flaming. We should probably nix being able to select the same rune type if you have one of them set already.

stwlam:
Would prefer a more-global fix for this

PR #7704, opened 2023-05-05
Prune duplicate and lesser versions of property runes on each weapon

### LINKED ISSUE #7654: Multiple etchings of the same rune should not stack

stwlam:
Closes #7654


CarlosFdez:
I suspect once the actual rune system exists, this will be swapped for a base rune slug or type.

PR #7705, opened 2023-05-05
Extend Check inline syntax to actions

TikaelSol:
This is much cleaner than the last implementation since I'm giving up on having it read into the traits to find actions.

image


src/module/system/text-editor.ts
```diff
@@ -370,7 +385,14 @@ class TextEditorPF2e extends TextEditor {
                           })
                           .join(" ");
                 html.innerHTML = inlineLabel ?? skillLabel;
-                html.setAttribute("data-pf2-check", params.type);
+                if (params.action) {
+                    const actionSlug = sluggify(params.action, { camel: "bactrian" });
+                    html.setAttribute("data-pf2-action", actionSlug);
```
stwlam, line 390: Pretty assigning `dataset` properties over setting attributes

TikaelSol: Yup, working on it, cleaning up the others in there while I'm at it.

src/module/system/text-editor.ts
```diff
@@ -322,22 +322,37 @@ class TextEditorPF2e extends TextEditor {
 
         // Build the inline link
         const html = document.createElement("span");
-        html.setAttribute("data-pf2-traits", `${allTraits}`);
+        html.dataset.pf2Traits = `${allTraits}`;
```
stwlam, line 325:
```suggestion
        html.dataset.pf2Traits = allTraits.toString();
```

PR #7724, opened 2023-05-07
Fix stacking modifiers in spell damage

CarlosFdez:
This originally started as a refactor, but I noticed and fixed an existing bug.


src/module/item/spell/document.ts
```diff
@@ -312,46 +320,38 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
                     dice.label += ` ${dice.dieSize}`;
                 }
             }
+        }
+
+        try {
```
stwlam, line 325: ?

CarlosFdez: I guess damage roll for spells can't really explode anymore (fingers crossed) so I removed the try. Turns out invalid formula entries were being skipped silently.

PR #7727, opened 2023-05-07
Link Enfeebled condition in descriptions where missing

xdy:
… @UUID[Compendium.pf2e.conditionitems.Enfeebled]{Enfeebled 2}, 3 and 4.


xdy:
If this PR is wanted I'll do links to the other numbered conditions in equipment.db as well.

TikaelSol:
Some of these are fine but we only link the first instance of a condition value normally to avoid filling the text with lots of needless links.

xdy:
Oh, I thought I caught those. I'll fix that so that each numerical value for enfeebled occurs only once as a link in each item.

TikaelSol:
Go ahead with the other conditions in one PR if you want.

PR #7733, opened 2023-05-07
Include compendia from modules beginning with `pf2e` in early migrations

esheyw:
Took me slightly longer than I said, but, first PR (to anything, ever, so if I did a bad let me know): fixed the dot


stwlam:
:white_circle:

PR #7735, opened 2023-05-08
Add tab for orphaned party items

CarlosFdez:
image


stwlam:
What makes an item orphaned? Is the tab visible when empty?

PR #7743, opened 2023-05-08
Show result formula for damage roll dialog

CarlosFdez:
image
image


stwlam:
The precision damage looks like it's inclusive of the entire formula

CarlosFdez:
I'll look to see if I can do a subtle styling later.

PR #7756, opened 2023-05-08
Create a `SimpleStatistic` class for use with non-check statistics

stwlam:


src/module/system/statistic/index.ts
```diff
@@ -35,51 +35,46 @@ import { RollNotePF2e, RollNoteSource } from "@module/notes.ts";
...
+        this.label = game.i18n.localize(data.label);
+        this.modifiers = [...(data.modifiers ??= [])];
+        this.domains = [...(data.domains ??= [])];
+        this.data = { ...data };
 
-/** Object used to perform checks or get dcs, or both. These are created from StatisticData which drives its behavior. */
-class Statistic {
-    /** Source of truth of all statistic data and the params used to create it. Necessary for cloning. */
-    #data: StatisticData;
+        // Add from synthetics, but only after check/dc are created so that modifiers aren't duplicated
```
CarlosFdez, line 61: But this is no longer the case!

stwlam: Fixed

PR #7757, opened 2023-05-09
Fix breakdown tag for weapon persistent damage

CarlosFdez:
I'm a bit disgusted I had to hardcode. I could hardcode against the weapon-persistent slug instead.

image

### DIFF
--- a/src/module/system/damage/formula.ts
+++ b/src/module/system/damage/formula.ts
@@ -1,6 +1,6 @@
-import { DamageDicePF2e, ModifierPF2e } from "@actor/modifiers.ts";
+import { DamageDicePF2e } from "@actor/modifiers.ts";
 import { DegreeOfSuccessIndex, DEGREE_OF_SUCCESS } from "@system/degree-of-success.ts";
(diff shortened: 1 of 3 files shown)


src/module/system/damage/formula.ts
```diff
@@ -79,15 +79,18 @@ function createDamageFormula(
     // Hold onto base terms for matching reasons (such as adding extra base dice)
     const baseTerms = [...typeMap.values()].flat();
 
-    // Dice always stack
+    // Sometimes a weapon may add base damage as bonus modifiers or dice. We need to auto-generate these
+    const BONUS_BASE_LABELS = ["PF2E.ConditionTypePersistent"].map((l) => game.i18n.localize(l));
```
stwlam, line 83: Does this really need to be run through a `map`?

CarlosFdez: I probably did that with the assumption that it might need to be expanded, and doing a map is less code for when that happens.

It doesn't seem like splash has the same issue since we don't add splash to the damage label, so I guess we can cross that bridge later.

PR #7763, opened 2023-05-09
Fix modification of NPC speeds and their display on sheets

### LINKED ISSUE #6364: Flat Modifier cannot modify speed on NPCs
You can replicate this by copy pasting Fleet's Rule Element into an NPC and see that their movement speed does not change.

stwlam:
Closes #6364, #6365


src/module/actor/npc/sheet.ts
```diff
@@ -161,12 +162,36 @@ class NPCSheetPF2e<TActor extends NPCPF2e> extends CreatureSheetPF2e<TActor> {
...
         hp.adjustedLower = hp.max < Number(hp.base);
         perception.adjustedHigher = perception.totalModifier > Number(perception.base);
         perception.adjustedLower = perception.totalModifier < Number(perception.base);
+        sheetData.speeds = {
+            land: {
+                label: speedData.label ?? "",
+                value: speedData.value,
+                details: speedData.details,
+                adjustedHigher: speedData.total > this.actor._source.system.attributes.speed.value,
+                adjustedLower: speedData.total < this.actor._source.system.attributes.speed.value,
```
CarlosFdez, line 178: This happens so often that I think we should make this a sheet util. I kinda made a util in the vehicle sheet that could probably be fleshed out more for such things.

stwlam: This is dissimilar to all other "adjustable" fields.

CarlosFdez: If by this you mean the vehicle, its the same, I just replaced the adjustedHigher/adjustedHigher if/else in the sheets with a single adjusted css class we can mindless throw in. If by this you mean this npc sheet thing, maybe, but easily rectifiable with a splat operator.

That said if we clean up all that adjustedHigher/lower stuff they should likely be done all at once anyways.

src/module/actor/sheet/base.ts
```diff
@@ -323,13 +326,11 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
...
                 input.removeAttribute("name");
-                input.removeAttribute("style");
                 const propertyPath = input.dataset.property ?? "";
                 const preparedValue = getProperty(this.actor, propertyPath);
                 if (input instanceof HTMLInputElement) {
-                    const isModifier = input.classList.contains("modifier") && Number(preparedValue) >= 0;
-                    const value = isModifier ? `+${preparedValue}` : preparedValue;
-                    input.value = String(value);
+                    const isPositiveModifier = input.classList.contains("modifier") && Number(preparedValue) >= 0;
+                    input.value = isPositiveModifier ? `+${preparedValue}` : String(preparedValue);
```
CarlosFdez, line 333: I think we have an addSign() util to handle this.

stwlam: PRs welcome

CarlosFdez: Fine. NPC sheets are already too ugly.

PR #7769, opened 2023-05-10
Add variants to Lay on Hands and Touch of Corruption

### LINKED ISSUE #7719: Lay on Hands doesn't have an option to roll damage
It'll be great to make Lay on Hands work as Heal where you can choose the kind (1, 2 (vs living), 2 (vs undead), 3).

So Lay on Hands would have the same thing for two kinds of usage:

- Healing - Roll 6*Level positive;
- Damaging Undead - Roll 1d6*Level positive and Fortitude save;

In the heat of battle, it's very easy to forget about the damage change, and it'll be just more convenient to have a button for damage in the same spell instead of manually rolling with /r 3d6[positive] each time or casting different spell for damage only.

TikaelSol:
Closes #7719
Supersedes #7260

### DIFF
--- a/packs/data/spells.db/lay-on-hands.json
+++ b/packs/data/spells.db/lay-on-hands.json
@@ -66,6 +66,9 @@
                     },
                     "save": {
                         "basic": "basic"
+                    },
+                    "spellType": {
+                        "value": "save"
                     }
                 }
             },
--- a/packs/data/spells.db/lay-on-hands.json
+++ b/packs/data/spells.db/lay-on-hands.json
@@ -65,7 +65,8 @@
                         "type": "interval"
                     },
                     "save": {
-                        "basic": "basic"
+                        "basic": "basic",
+                        "value": "fortitude"
                     },
                     "spellType": {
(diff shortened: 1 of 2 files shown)


packs/data/spells.db/lay-on-hands.json
```diff
@@ -38,19 +31,76 @@
...
+                    "heightening": {
+                        "damage": {
+                            "37YW4ZGhxx7Y2mdI": "1d6"
+                        },
+                        "interval": 1,
+                        "type": "interval"
+                    },
+                    "save": {
+                        "basic": "basic"
+                    }
```
SpartanCPA, line 69:
```suggestion
                    "save": {
                        "basic": "basic",
                        "value": "fortitude"
                    }
```

TikaelSol: Weird, it's definitely there in Foundry.

TikaelSol: Ah, found it. It was still on the parent.

PR #7771, opened 2023-05-10
Refactor skills to use Statistic

CarlosFdez:
This should preserve compatibility even with the silly macros that ignore deprecation messages.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -30,7 +30,6 @@ import {
     SAVE_TYPES,
     SKILL_ABBREVIATIONS,
     SKILL_DICTIONARY,
-    SKILL_DICTIONARY_REVERSE,
     SKILL_EXPANDED,
 } from "@actor/values.ts";
 import {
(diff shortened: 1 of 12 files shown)


src/module/system/statistic/index.ts
```diff
@@ -249,6 +275,18 @@ class Statistic extends SimpleStatistic {
             breakdown: check.breakdown ?? "",
             _modifiers: check.modifiers.map((m) => m.toObject()),
         };
+
+        if (options.rollable) {
+            trace.roll = (args: RollParameters = {}): Promise<unknown> => {
+                console.warn(`Rolling ${this.label} (${this.slug}) via actor system data is deprecated`);
```
stwlam, line 281: Can we use the Foundry deprecation API here?

src/module/actor/character/document.ts
```diff
@@ -2109,6 +1984,9 @@ interface CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocument
     flags: CharacterFlags;
     readonly _source: CharacterSource;
     system: CharacterSystemData;
+
+    /** Skills for the character, built during data prep */
+    skills: CharacterSkills;
```
stwlam, line 1989: Send 'er up top

CarlosFdez: I thought we wanted to put the ones that are extended down there since I saw something recently added in the actor base case. Sure, I'll move everything on up.

PR #7781, opened 2023-05-10
Slightly narrow class-feat compendium browser prefilters for PC sheet

stwlam:


CarlosFdez:
Definitely an improvement, though I still don't know what we should be doing with class archetypes.

PR #7817, opened 2023-05-13
Make item image styling and HTML more consistent

CarlosFdez:
This does change the background color in some places. I can swap it to the secondary color if that's the preferred one.


stwlam:
Can we simply preserve the original appearance?

PR #7837, opened 2023-05-15
Fix startup failure in run migration script

CarlosFdez:


stwlam:
Frivolous helper :(

CarlosFdez:
It type inferences better than the silly casting. I have other uses for it but if you prefer casting I can remove it.

CarlosFdez:
Superceded (and snuck into) #7805 using the Object.reduce() casting approach.

PR #7852, opened 2023-05-16
Use `Statistic` for AC

stwlam:


src/module/actor/familiar/document.ts
```diff
@@ -146,23 +147,12 @@ class FamiliarPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
-                    stat.modifiers
-                        .filter((m) => m.enabled)
-                        .map((m) => `${m.label} ${m.modifier < 0 ? "" : "+"}${m.modifier}`)
-                )
-                .join(", ");
-            systemData.attributes.ac = stat;
+            const masterModifiers = master.armorClass.modifiers
+                .filter((m) => !["status", "circumstance"].includes(m.type))
+                .map((m) => m.clone());
+            const statistic = new ArmorStatistic(this, { modifiers: masterModifiers });
```
CarlosFdez, line 153: If I recall, we can't pass the master modifiers directly, we need to get the result of said modifiers because any predication that passed for the master might no longer apply when evaluated from the perspective of the familiar.

src/module/actor/npc/document.ts
```diff
@@ -240,40 +241,17 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
...
-            const base = system.attributes.ac.value;
-            const domains = ["ac", "dex-based", "all"];
-            const modifiers = [
+        const armorStatistic = new ArmorStatistic(this, {
+            modifiers: [
                 new ModifierPF2e({
                     slug: "base",
                     label: "PF2E.ModifierTitle",
-                    modifier: base,
-                    adjustments: extractModifierAdjustments(modifierAdjustments, domains, "base"),
```
CarlosFdez, line 251: Are there any active modifier adjustments for the base value for NPCs?

src/module/system/statistic/index.ts
```diff
@@ -503,8 +504,10 @@ class StatisticDifficultyClass {
             data.dc?.domains ? extractModifiers(parent.actor.synthetics, data.dc.domains) : [],
         ].flat();
         this.modifiers = allDCModifiers.map((modifier) => modifier.clone({ test: this.options }));
+    }
 
-        this.value = (data.dc?.base ?? 10) + new StatisticModifier("", this.modifiers, this.options).totalModifier;
+    get value(): number {
+        return 10 + new StatisticModifier("", this.modifiers, this.options).totalModifier;
```
CarlosFdez, line 510: I fully agree with killing data.dc?.base (If we keep base, it should be to add a base modifier, not to simply add 10) but if I recall you prefer to keep these as values computed during initialization, not as getters.

CarlosFdez: Ah wait no, we can't remove dc.base yet. That's used by spellcasting to set the base to 0 because we don't yet have a solution for a hardcoded base dcs.

PR #7853, opened 2023-05-16
Create a chat card if a consumable item was used from the inventory and automatically deleted

In3luki:
Recording 2023-05-16 at 23 25 23


stwlam:
Can we just make it the same card?

In3luki:
Roll with description. It looks a bit different because I had to append the description to the `flavor` instead of `content`. Is it possible to add custom `content` to a roll message?
image

Without a roll:
image

In3luki:
Updated. Instead of adding the item description to the roll flavor I'm now adding the rendered roll to the item message.
image

PR #7862, opened 2023-05-17
Fix initiative using skills overriden by battle forms

CarlosFdez:


src/module/system/statistic/index.ts
```diff
@@ -481,6 +379,119 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
...
+        const rollable = options.rollable;
+        if (rollable) {
+            trace.roll = (args: RollParameters = {}): Promise<unknown> => {
+                const deprecationLabel = `Rolling ${this.label} (${this.slug}) via actor system data is deprecated`;
+                foundry.utils.logCompatibilityWarning(deprecationLabel, {
+                    since: rollable[0],
+                    until: rollable[1],
+                });
+
+                console.warn();
```
stwlam, line 432:
```suggestion
```

PR #7863, opened 2023-05-18
Add effects to pistolero's challenge

bcronin90:
https://2e.aonprd.com/Feats.aspx?ID=3174

Adds effects to Pistolero's Challenge to automate extra damage. I made this for my GM, but was wondering if I can get it to a state in which it would be mergable.
Two limitations:
- Cannot automatically detect the proficiency of the relevant skill, as it's not necessarily the highest.
- Cannot differentiate between multiple Pistoleros issuing multiple challenges.


packs/data/feat-effects.db/effect-pistoleros-challenge-legendary.json
```diff
@@ -0,0 +1,45 @@
+{
+    "_id": "1q4VIzG7mtYykEsf",
+    "img": "systems/pf2e/icons/equipment/weapons/flintlock-pistol.webp",
+    "name": "Effect: Pistolero's Challenge (Legendary)",
```
TikaelSol, line 4: Consolidate all three effects instead of making 3 effects.

packs/data/feat-effects.db/effect-pistoleros-challenge-legendary.json
```diff
@@ -0,0 +1,45 @@
...
+            "value": -1
+        },
+        "level": {
+            "value": 1
+        },
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "target:effect:pistoleros-challenge-legendary"
```
TikaelSol, line 22: This will need to be a toggle on the sheet, not something keyed off the effect being present. That's to prevent party's with more than one person with this feat. Honestly this is probably more a job for the eventually finished MarkTarget RE.

packs/data/feat-effects.db/effect-pistoleros-challenge-legendary.json
```diff
@@ -0,0 +1,45 @@
...
+                ],
+                "selector": "strike-damage",
+                "type": "status",
+                "value": 4
+            }
+        ],
+        "source": {
+            "value": ""
```
TikaelSol, line 30:
```suggestion
            "value": "Pathfinder Guns & Gears"
```

packs/data/feat-effects.db/effect-pistoleros-challenge-legendary.json
TikaelSol: We can at the very least ask which skill was used in a ChoiceSet and then detect it.

bcronin90: Can I evaluate a proficiency of the owner of the feat when applying an effect to another actor? Cause that's what would be needed here, right?

bcronin90: I could use a Choice Set to select Proficiency instead. Not as pretty though.

packs/data/feat-effects.db/effect-pistoleros-challenge-legendary.json
bcronin90: Yeah, that's what I expected. I'll keep using it in our game, but if the other comment can't be resolved, it's probably best to postpone this.

bcronin90: Gotcha. I'll get back to it once MarkTarget exists.

PR #7865, opened 2023-05-18
Fix editability inconsistencies on hazard sheet

stwlam:
plus a couple localization issues


CarlosFdez:
Editing toggle as intended is more for making sections modular rather than a true read only mode. But if it has caused confusion by doing so, we can have a stricter lock..

PR #7866, opened 2023-05-18
Automatically add object immunities to hazards

stwlam:


CarlosFdez:
Approved on the code side

TikaelSol:
Not all hazards have this I don't think, and I think we used to auto add these and removed it because of that.

stwlam:
They're removable via rule element

PR #7872, opened 2023-05-18
Add frequency for round and turn.

### LINKED ISSUE #4542: NPC feature frequency is missing "per round" and "per turn"
I was building an NPC and I expected to have "round" (and "turn") in the dropdown here:

image

xdy:
Fixes #4542

Not sure this one is worth doing, but given I've done it, I might as well submit it. :)

### DIFF
--- a/src/scripts/macros/rest-for-the-night.ts
+++ b/src/scripts/macros/rest-for-the-night.ts
@@ -4,7 +4,7 @@ import { ItemPF2e } from "@item";
 import { ChatMessageSourcePF2e } from "@module/chat-message/data.ts";
 import { ChatMessagePF2e } from "@module/chat-message/index.ts";
 import { ActionDefaultOptions } from "@system/action-macros/index.ts";
-import { localizer } from "@util";
+import { localizer, tupleHasValue } from "@util";
 import { Duration } from "luxon";
 
 /** A macro for the Rest for the Night quasi-action */
@@ -113,7 +113,8 @@ export async function restForTheNight(options: ActionDefaultOptions): Promise<Ch
         const withFrequency = actionsAndFeats.filter(
             (a) =>
                 a.frequency &&
-                (a.frequency.per === "day" || Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
+                (tupleHasValue(["turn", "round", "day"], a.frequency.per) ||
+                    Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
                 a.frequency.value < a.frequency.max
         );
         if (withFrequency.length) {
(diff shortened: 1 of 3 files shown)


src/scripts/macros/rest-for-the-night.ts
```diff
@@ -113,7 +113,10 @@ export async function restForTheNight(options: ActionDefaultOptions): Promise<Ch
         const withFrequency = actionsAndFeats.filter(
             (a) =>
                 a.frequency &&
-                (a.frequency.per === "day" || Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
+                (a.frequency.per === "turn" ||
+                    a.frequency.per === "round" ||
+                    a.frequency.per === "day" ||
```
stwlam, line 118: Use `tupleHasValue`

src/scripts/macros/rest-for-the-night.ts
```diff
@@ -113,7 +113,10 @@ export async function restForTheNight(options: ActionDefaultOptions): Promise<Ch
         const withFrequency = actionsAndFeats.filter(
             (a) =>
                 a.frequency &&
-                (a.frequency.per === "day" || Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
+                tupleHasValue(
+                    ["turn", "round", "day", Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")],
```
CarlosFdez, line 117: the duration check should be separate from the tupleHasValue, tupleHasValue is basically an array includes() but in a way that satisfies typescript.

The check that's occurring here, since its not documented, is that we want to reset all frequencies that are either daily prep or faster than nap-time. But turn/round are encounter mode timespans and don't correspond to fix time exactly (outside of the 6 second nonesense) so you have to explicitly check for those alongside day.

src/scripts/macros/rest-for-the-night.ts
```diff
@@ -113,7 +113,9 @@ export async function restForTheNight(options: ActionDefaultOptions): Promise<Ch
         const withFrequency = actionsAndFeats.filter(
             (a) =>
                 a.frequency &&
-                (a.frequency.per === "day" || Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
+                (tupleHasValue(["turn", "round", "day"], a.frequency.per) ||
+                    Duration.fromISO(a.frequency.per) <= Duration.fromISO("PT8H")) &&
+                a.frequency.per &&
```
CarlosFdez, line 118: if the tupleHasValue and less than check succeeded, why do you have to check for this? Those computations rely on per.

CarlosFdez (approved):

PR #7876, opened 2023-05-18
Condense Enlarge effects

cjneidhart:
This merges the normal (level 2) and heightened (level 4) effects for the spell Enlarge. It also changes the flat modifier on dex-based checks to actually grant the clumsy condition. Partially addresses #4144.


stwlam:
This shouldn't have a choice set except for level 6+, and it doesn't warrant using a badge.

PR #7890, opened 2023-05-19
Add origin automation to Overdrive

TikaelSol:
Adding it to the pile of condensed effects to migrate away (Maybe we push and doo all that with v11?).

This is slightly kludgy in the sense that it relies on actor level instead of the features existing but I don't see any way for inventor dedication to get overdrive, so unless I missed that this should work. If not then I can fix that to work properly but it'll require a migration.


packs/data/actions.db/overdrive.json
```diff
@@ -3,6 +3,7 @@
     "img": "systems/pf2e/icons/actions/OneAction.webp",
     "name": "Overdrive",
     "system": {
+        },
```
TikaelSol, line 6:
```suggestion
```

PR #7895, opened 2023-05-19
Set slug for all NPC tail attacks

reyzor1991:
Change slug from null to "tail" for all missed bestiary


stwlam:
Any particular reason to do this for tails in particular?

reyzor1991:
> Any particular reason to do this for tails in particular? Does some special ability look for it?

Yes, when I use some macros/scripts for optimization I read domains/options properties like "item:tail" or "tail-attack"
but with translation module it's working incorrect because it's use "item:{translated tail}" or "{translated tail}-attack"
Example:
"хвіст-attack" - uk lang
"queue-attack" - fr lang
"schwanz-attack" - de lang

PR #7904, opened 2023-05-20
Add a method to fully revert applied actor damage

In3luki:
The Crucible damage reversion looked neat so I implemented it for the system.
This records all applied updates in a flag on the damage chat message and adds a new context menu entry to revert everything stored in the flag.
It handles hp, temp hp, stamina, etc. and also shield hp and all persistent damage effects that were created.

[revert-damage.webm](https://github.com/foundryvtt/pf2e/assets/41452412/a0e072e5-4227-4fb4-857b-b21b313faabe)

[revert-healing.webm](https://github.com/foundryvtt/pf2e/assets/41452412/eeb9f80e-e6b4-458c-b78f-b682962ecc2f)

### DIFF
--- a/src/module/apps/sidebar/chat-log.ts
+++ b/src/module/apps/sidebar/chat-log.ts
@@ -74,7 +74,7 @@ export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
                 }
             }
 
-            const revertDamageButton = htmlClosest<HTMLButtonElement>(target, "button.revert-damage");
+            const revertDamageButton = htmlClosest<HTMLButtonElement>(target, "button[data-action=revert-damage]");
             if (revertDamageButton) {
                 const appliedDamageFlag = message?.flags.pf2e.appliedDamage;
                 if (appliedDamageFlag) {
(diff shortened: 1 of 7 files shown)


stwlam:
Can we get an undo button near where the IWR info tooltip is?

In3luki:
This good?
image

Is `undo` a better word? I can change revert to undo if you want.

stwlam:
> This good? image
> 
> Is `undo` a better word? I can change revert to undo if you want.

Just "Undo" is good. Try one of these icons:
https://fontawesome.com/icons/arrow-turn-down-left?f=classic&s=solid (needs to be flipped with `fa-flip-vertical`)
https://fontawesome.com/icons/rotate-left?f=classic&s=solid

In3luki:
I think `rotate-left` looks better.

image

image

In3luki:
image

Recording 2023-05-20 at 22 53 10

src/module/apps/sidebar/chat-log.ts
```diff
@@ -72,6 +73,22 @@ export class ChatLogPF2e extends ChatLog<ChatMessagePF2e> {
                     }
                 }
             }
+
+            const revertDamageButton = htmlClosest<HTMLButtonElement>(target, "button.revert-damage");
```
stwlam, line 77: Use a `data-action` selector for this: we try to keep styling and recent listeners on separate selectors.

In3luki: Done

Issue #4882, opened 2022-11-19
Tracking Issue: Spell system improvements


TikaelSol:
Condensing older issues into one place. Not all will be implemented, just trying to cut down on issues to a manageable number.

- [ ] Stave spellcasting #413
- [ ] Critical Rolls on spells, for spells with different critical scaling like Phantasmal Killer #547
- [ ] Create drag and drop interchangeable spell preparation lists #110
- [ ] Allow spell lists to be sorted alphabetically #485
- [ ] Some spells have 3D or other odd areas #1386
- [ ] Create spellcasting entries with features / feats
- [ ] Grant spell items from other items #2008
- [ ] Allow for spells with multiple spell types #1905
- [ ] Allow for alternate target DCs (Shadow Signet)
- [ ] Add heightened info to chat message data #2122 (May be already done)
- [ ] Fix healing spells getting damage bonuses #2071
- [ ] Allow spellcasting entries to be collapsed #2202
- [ ] Allow individual spell levels to be collapsed #2304
- [ ] Spell variants are orphaned when cast from a consumable #3382

ThingsCouldGetDicey:
may i suggest to have spells that are of different origin i.e from an archetype or different magic orgiin i.e occultism etc etc. to be tabbed as a miniture tab. rather than one long list?

MrVauxs:
you can already do that by making multiple spell lists, and for archetypes and different magic origins you frankly should

ThingsCouldGetDicey:
> you can already do that by making multiple spell lists, and for archetypes and different magic origins you frankly should

what i meant was to have it tabbed as so 
![image](https://user-images.githubusercontent.com/62528192/203167758-dab3e3cd-2bde-4b86-9fd2-cb1935ba8776.png)
or something close to that. 
so you can click from different sources.

ThingsCouldGetDicey:
![image](https://user-images.githubusercontent.com/62528192/203187758-2c56ae10-b006-483d-93a9-5c7f62360b77.png)
 a better mock up.

TikaelSol:
On potential way to handle organization by tab is by allowing custom tabs. It's not feasible to add a tab for each source as that becomes a bigger mess fairly quickly due to the number of sources.

![image](https://user-images.githubusercontent.com/80183198/203377163-eb8d1b2b-3be9-47ed-85c3-389cd6e7dd5e.png)

[...]

PR #7906, opened 2023-05-20
Move cast level and spell variant info to origin flag

### LINKED ISSUE #2122: Add heightened spell level to chat-card and attack roll message flags.
In GitLab by @william.ring.91 on Mar 1, 2022, 10:26

For use in parsing messages for data, the heightened spell level will allow for automation of damage rolls without relying on looking at the flavor text. Use case: in xdy-workbench

oWave:
There are currently 3 different versions of `flags.pf2e.origin` for storing/retrieving an item:
- ChatMessage contains the full item data, correctly applying spell overlays (variant/heighten) when retrieving the item
- Check/Damage only stores the item uuid, making it impossible to fetch the full item with that message alone
- `SpellPF2e.createTemplate` has its own completely different format

This PR is my attempt at unifying the creation of (and retrieval from) `flags.pf2e.origin`. This would fix #2122 (to check another box in #4882).

### DIFF
--- a/src/module/chat-message/data.ts
+++ b/src/module/chat-message/data.ts
@@ -15,7 +15,7 @@ export interface ItemOriginFlag {
     type: ItemType;
     uuid: string;
     castLevel?: number;
-    spellVariantOverlayIds?: string[];
+    variant?: { overlays: string[] };
 }
 
 type ChatMessageFlagsPF2e = foundry.documents.ChatMessageFlags & {
--- a/src/module/chat-message/data.ts
+++ b/src/module/chat-message/data.ts
@@ -11,11 +11,18 @@ interface ChatMessageSourcePF2e extends foundry.documents.ChatMessageSource {
     flags: ChatMessageFlagsPF2e;
 }
 
+export interface ItemOriginFlag {
+    type: ItemType;
+    uuid: string;
+    castLevel?: number;
+    spellVariantOverlayIds?: string[];
+}
+
 type ChatMessageFlagsPF2e = foundry.documents.ChatMessageFlags & {
     pf2e: {
         damageRoll?: DamageRollFlag;
         context?: ChatContextFlag;
-        origin?: { type: ItemType; uuid: string; castLevel?: number; spellVariantOverlayIds?: string[] } | null;
+        origin?: ItemOriginFlag | null;
         casting?: { id: string; tradition: MagicTradition } | null;
         modifierName?: string;
         modifiers?: BaseRawModifier[];
--- a/src/module/chat-message/data.ts
+++ b/src/module/chat-message/data.ts
@@ -15,14 +15,13 @@ type ChatMessageFlagsPF2e = foundry.documents.ChatMessageFlags & {
     pf2e: {
         damageRoll?: DamageRollFlag;
         context?: ChatContextFlag;
-        origin?: { type: ItemType; uuid: string } | null;
-        casting?: { id: string; level: number; tradition: MagicTradition } | null;
+        origin?: { type: ItemType; uuid: string; castLevel?: number; spellVariantOverlayIds?: string[] } | null;
+        casting?: { id: string; tradition: MagicTradition } | null;
         modifierName?: string;
         modifiers?: BaseRawModifier[];
         preformatted?: "flavor" | "content" | "both";
         isFromConsumable?: boolean;
         journalEntry?: DocumentUUID;
-        spellVariant?: { overlayIds: string[] };
         strike?: StrikeLookupData | null;
         appliedDamage?: AppliedDamageFlag | null;
         [key: string]: unknown;
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -120,7 +120,7 @@ class ChatMessagePF2e extends ChatMessage {
         }
 
         if (item?.isOfType("spell")) {
-            const overlayIds = this.flags.pf2e.origin?.spellVariantOverlayIds;
+            const overlayIds = this.flags.pf2e.origin?.variant?.overlays;
             const castLevel = this.flags.pf2e.origin?.castLevel ?? item.level;
             const modifiedSpell = item.loadVariant({ overlayIds, castLevel });
             return modifiedSpell ?? item;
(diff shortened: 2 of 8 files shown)


stwlam:
Where is #2122 not satisfied?

CarlosFdez:
I'll have to check, but spell templates using their own data is actually a module compat thing for the prettifying modules. I think it'd be wise to extend it for the templates to add additional stuff like traits (or break everything and tell people to change things, that's always an option). 

What's the compatibility with spells casted from the final scroll? The reason items are sometimes full embedded is because consuming a scroll gets rid of it, thus making the item lookup fail. Same for templates for spells made from a scroll that has been consumed, is it still possible to get the traits and name of the spell for that template?

CarlosFdez:
Good timing honestly, since the best time to break all this is 5.0 which is almost around the corner. I'll take a look with a more fine toothed comb as time permits.

CarlosFdez:
Sorry for the late response. I needed to think for a while on this one, and I realize now why I was uncertain.

Doing it this way assumes that spells will always know how to recreate themselves *and* their spellcasting entry even into the forseeable future. In the future however, a stave will be able to load a spell via a uuid and cast it directly. Shoving casting data as well means that we'll have to ensure the spell knows how it was cast, perhaps via clone and flag injection shenaniganry.

Origin is for retrieving the item but casting is for retrieving the spellcasting entry. The one area where this distinction gets kinda fuzzy is casting#level (aka castLevel), but is absolutely required for the spell to recreate itself (whereas the rest of casting just figures out the spellcasting entry to use).

Also we don't need those item static helpers, keeping those to chat message might be best.

This may be more longer term than you're willing to commit to pf2e as a volunteer, but for now I'd recommend moving spellVariant and casting.level to origin.variant and origin.level). Then later, having damage create the casting flags directly if the item.isOfType("spell").

oWave:
I have simplified the changes. I'll leave touching spell casting to somebody else (or future me)

> a stave will be able to load a spell via a uuid and cast it directly  

That would also solve the issue with casting from consumables. I'd like to try that next (in another PR)

Should the creation of/from origin flags (at least for spell) be done in helper functions to reduce duplication? I've realized that an item getter on MeasuredTemplate would be helpful, but not a fan of having that code around twice.

CarlosFdez:
I think the item non-static method that you subclass to get origin flags was fine, its the other 2 static ones (that were only called by chat message) I had issues with. I'll take a look soon.

I'll likely want a change for spell variant but let me think about it. castLevel will need to change for remaster soon anyways so we can keep that one (otherwise I'd say just level is fine, we could determine what level means based on context).

CarlosFdez:
@stwlam I'm thinking of merging this but renaming `spellVariantOverlayIds?: string[];` to `variant?: { overlays: string[] }` after I do. Thoughts?

src/module/chat-message/data.ts
```diff
@@ -11,18 +11,24 @@ interface ChatMessageSourcePF2e extends foundry.documents.ChatMessageSource {
     flags: ChatMessageFlagsPF2e;
 }
 
+export interface ItemOriginFlag {
+    type: ItemType;
+    uuid: string;
+    castLevel?: number;
+    spellVariantOverlayIds?: string[];
```
CarlosFdez, line 18:
```suggestion
    variant?: { overlays: string[] };
```

src/module/chat-message/document.ts
```diff
@@ -120,9 +120,9 @@ class ChatMessagePF2e extends ChatMessage {
         }
 
         if (item?.isOfType("spell")) {
-            const spellVariant = this.flags.pf2e.spellVariant;
-            const castLevel = this.flags.pf2e.casting?.level ?? item.level;
-            const modifiedSpell = item.loadVariant({ overlayIds: spellVariant?.overlayIds, castLevel });
+            const overlayIds = this.flags.pf2e.origin?.spellVariantOverlayIds;
```
CarlosFdez, line 123:
```suggestion
            const overlayIds = this.flags.pf2e.origin?.variant?.overlays;
```

src/module/item/spell/document.ts
```diff
@@ -865,6 +857,13 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
         );
     }
 
+    override getOriginData(): ItemOriginFlag {
+        const flag = super.getOriginData();
+        flag.castLevel = this.level;
+        if (this.isVariant) flag.spellVariantOverlayIds = [...this.appliedOverlays!.values()];
```
CarlosFdez, line 863:
```suggestion
        if (this.isVariant && this.appliedOverlays) {
            flag.variant = { overlays: [...this.appliedOverlays.values()] };
        }

```

src/module/scene/measured-template-document.ts
```diff
@@ -1,9 +1,28 @@
...
+> extends MeasuredTemplateDocument<TParent> {
+    get item(): ItemPF2e<ActorPF2e> | null {
+        const origin = this.flags.pf2e?.origin as Record<string, unknown>;
+        const uuid = origin?.uuid;
+        if (!uuid) return null;
+        const item = fromUuidSync(uuid as string);
+        if (!(item instanceof ItemPF2e)) return null;
+
+        if (item?.isOfType("spell")) {
+            const overlayIds = origin?.spellVariantOverlayIds as string[];
```
CarlosFdez, line 17:
```suggestion
            const overlayIds = origin?.variant?.overlays;
```

src/module/scene/measured-template-document.ts
```diff
@@ -1,9 +1,28 @@
...
 import { ScenePF2e } from "./document.ts";
+import { ItemPF2e } from "@item";
+import { ActorPF2e } from "@actor";
 
 export class MeasuredTemplateDocumentPF2e<
     TParent extends ScenePF2e | null = ScenePF2e | null
-> extends MeasuredTemplateDocument<TParent> {}
+> extends MeasuredTemplateDocument<TParent> {
+    get item(): ItemPF2e<ActorPF2e> | null {
+        const origin = this.flags.pf2e?.origin as Record<string, unknown>;
```
CarlosFdez, line 10:
```suggestion
        const origin = this.flags.pf2e?.origin;
```

PR #7918, opened 2023-05-21
Refactor rolling perception checks from the character sheet to remove deprecation warning

In3luki:
The `.attribute-name` selector is only used for attribute checks in the NPC sheet now, and that has its own listener.

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -263,6 +263,14 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
             this.actor.skills[key]?.check.roll(rollParams);
         });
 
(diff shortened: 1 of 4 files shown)


src/module/actor/creature/sheet.ts
```diff
@@ -263,6 +263,14 @@ export abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends Act
             this.actor.skills[key]?.check.roll(rollParams);
         });
 
+        // Roll perception checks
+        for (const element of htmlQueryAll<HTMLAnchorElement>(html, "a[data-action=perception-check]")) {
```
stwlam, line 267:
```suggestion
        for (const element of htmlQueryAll(html, "a[data-action=perception-check]")) {
```

PR #7921, opened 2023-05-21
Make initiative rollable from the hazard sheet

In3luki:
Before;
before-error
before

After:
after

### DIFF
--- a/static/templates/actors/hazard/partials/sidebar.hbs
+++ b/static/templates/actors/hazard/partials/sidebar.hbs
@@ -36,7 +36,10 @@
                     <div class="valued-icon">
                         <input type="text" value="{{numberFormat data.attributes.stealth.value decimals=0 sign=true}}" data-property="system.attributes.stealth.value" data-dtype="Number" placeholder="N/A" />
                     </div>
-                    <h4 class="attribute-name">{{localize "PF2E.StealthLabel"}}</h4>
+                    <h2 class="roll-init{{#if @root.options.editable}} rollable{{/if}}" data-action="roll-initiative">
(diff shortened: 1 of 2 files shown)


CarlosFdez:
I dont think you were able to roll from the sheet before, this is a new feature. One that became possible after my initiative refactor but i forgot to wire it up oopsy 🙃

CarlosFdez:
I think the main headers labels are <h2> correct? We should probably not use <h2> for the contents, but id need to double check the html and see if i need to clean it up

In3luki:
Yeah, it was rolling based on the `.attribute-name` selector so probably unintended as `data-attribute` was missing.
I've used h2 because the saves also use h2.

In3luki:
After #7918 removed the listener for `.attribute-name` this isn't even a bugfix anymore.

static/templates/actors/hazard/partials/sidebar.hbs
```diff
@@ -36,7 +36,10 @@
                     <div class="valued-icon">
                         <input type="text" value="{{numberFormat data.attributes.stealth.value decimals=0 sign=true}}" data-property="system.attributes.stealth.value" data-dtype="Number" placeholder="N/A" />
                     </div>
-                    <h4 class="attribute-name">{{localize "PF2E.StealthLabel"}}</h4>
+                    <h2 class="roll-init{{#if @root.options.editable}} rollable{{/if}}" data-action="roll-initiative">
+                        <i class="fas fa-dice-d20"></i>
```
stwlam, line 40:
```suggestion
                        <i class="fa-solid fa-dice-d20"></i>
```

PR #7928, opened 2023-05-22
Add shorthand syntax for multiple rollable check buttons

dmarcuse:
(Per request of @TMunsie)

This allows using syntax like `@Check[type:athletics,acrobatics]` to create two rollable buttons at once. When this syntax is used, there are a few changes:

 - The rollable buttons are created in a new block of text (div) rather  than inline, with each button occupying a single line.
 - The rollable buttons are "grouped", and posting one to chat will post  all of the buttons to chat.
 - The post to chat button uses a different icon (`fa-comment-alt-dots`) to  indicate the buttons are grouped. 

Otherwise, everything else works the same. Parameters and display text are applied to all rollable buttons, with the exception of the `adjustment` parameter, which has also been enhanced in this PR. The `adjustment` parameter can now be applied to fixed DCs (rather than just `@self.level`), and if specified multiple times, allows adjusting each individual button when using the shorthand syntax for multiple buttons, like so:

`@Check[type:religion,fiend-lore,barbazu-lore|dc:20|adjustment:0,-2,-5]`

Here's a screenshot of what the new functionality looks like in practice. The checks from the journal entry have been posted to chat in the order they appear.

Screenshot of new rollable check functionality

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -256,91 +256,6 @@ class TextEditorPF2e extends TextEditor {
         return null;
     }
 
-    static #createSingleCheck({
-        params,
-        allTraits,
-        item,
-        actor,
-        inlineLabel,
-    }: {
-        params: { type: string; dc: string } & Record<string, string>;
-        allTraits: string[];
-        item?: ItemPF2e | null;
-        actor?: ActorPF2e | null;
-        inlineLabel?: string;
-    }): HTMLSpanElement | null {
-        // Build the inline link
-        const html = document.createElement("span");
-        html.setAttribute("data-pf2-traits", `${allTraits}`);
-        const name = params.name ?? item?.name ?? params.type;
-        html.setAttribute("data-pf2-label", game.i18n.format("PF2E.InlineCheck.DCWithName", { name }));
-        html.setAttribute("data-pf2-repost-flavor", name);
-        const role = params.showDC ?? "owner";
-        html.setAttribute("data-pf2-show-dc", params.showDC ?? role);
-        html.setAttribute("data-pf2-adjustment", params.adjustment ?? "");
-
-        switch (params.type) {
-            case "flat":
-                html.innerHTML = inlineLabel ?? game.i18n.localize("PF2E.FlatCheck");
-                html.setAttribute("data-pf2-check", "flat");
-                break;
-            case "perception":
-                html.innerHTML = inlineLabel ?? game.i18n.localize("PF2E.PerceptionLabel");
-                html.setAttribute("data-pf2-check", "perception");
-                break;
-            case "fortitude":
-            case "reflex":
-            case "will": {
-                const saveName = game.i18n.localize(CONFIG.PF2E.saves[params.type]);
-                const saveLabel =
-                    params.basic === "true"
-                        ? game.i18n.format("PF2E.InlineCheck.BasicWithSave", { save: saveName })
-                        : saveName;
-                html.innerHTML = inlineLabel ?? saveLabel;
-                html.setAttribute("data-pf2-check", params.type);
-                break;
-            }
-            default: {
-                // Skill or Lore
-                const shortForm = (() => {
-                    if (objectHasKey(SKILL_EXPANDED, params.type)) {
-                        return SKILL_EXPANDED[params.type].shortform;
-                    } else if (objectHasKey(SKILL_DICTIONARY, params.type)) {
-                        return params.type;
-                    }
-                    return;
-                })();
-                const skillLabel = shortForm
-                    ? game.i18n.localize(CONFIG.PF2E.skills[shortForm])
-                    : params.type
-                          .split("-")
-                          .map((word) => {
-                              return word.slice(0, 1).toUpperCase() + word.slice(1);
-                          })
-                          .join(" ");
-                html.innerHTML = inlineLabel ?? skillLabel;
-                html.dataset.pf2Check = sluggify(params.type);
-            }
-        }
-
-        if (params.type && params.dc) {
-            // Let the inline roll function handle level base DCs
-            const checkDC = params.dc === "@self.level" ? params.dc : getCheckDC({ name, params, item, actor });
-            html.setAttribute("data-pf2-dc", checkDC);
-
-            let displayedDC = checkDC;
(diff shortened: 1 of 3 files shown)


dmarcuse:
For some extra context, this is particularly helpful for Pathfinder Society play. PFS scenarios often list a wide variety of skills for skill challenges because of the wide variance in parties, and being able to quickly write these and post them all to chat at once saves a considerable amount of time.

TikaelSol:
This is a neat idea, maybe would be nice to have a way to not have a line break through.

static/lang/en.json
```diff
@@ -5488,4 +5490,4 @@
     "SIDEBAR": {
         "TabCombat": "Encounter Tracker"
     }
-}
+}
```
TikaelSol, line 5493:
```suggestion
}

```

dmarcuse:
> This is a neat idea, maybe would be nice to have a way to not have a line break through.

I'm open to suggestions - I'm not great with the UI/UX side of things, but I'm happy to implement changes if someone comes up with a good idea of what that should look like.

src/scripts/ui/inline-roll-links.ts
```diff
@@ -210,9 +216,18 @@ export const InlineRollLinks = {
             return;
         }
 
-        const flavor = target.attributes.getNamedItem("data-pf2-repost-flavor")?.value ?? "";
-        const showDC =
-            target.attributes.getNamedItem("data-pf2-show-dc")?.value ?? (document?.hasPlayerOwner ? "all" : "gm");
+        const defaultVisibility = document?.hasPlayerOwner ? "all" : "gm";
+        let content;
```
CarlosFdez, line 220: In this project we're kinda highly alergic to let. Use an IIFE if you need to.

Also lower down I think you don't need to check against explicitly undefined, checking if its truthy should be enough.

src/scripts/ui/inline-roll-links.ts
```diff
@@ -29,7 +29,9 @@ export const InlineRollLinks = {
             if (foundryDoc && !foundryDoc.isOwner) continue;
 
             const newButton = document.createElement("i");
-            newButton.classList.add("fas", "fa-comment-alt");
+            const icon =
+                link.parentElement?.dataset?.pf2Checkgroup !== undefined ? "fa-comment-alt-dots" : "fa-comment-alt";
```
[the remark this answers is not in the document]
dmarcuse: Empty strings are falsy, and a data attribute with no value results in an empty string in the dataset

src/scripts/ui/inline-roll-links.ts
dmarcuse: Sounds good, I've refactored the PR to remove `let` bindings. The comparison to undefined used here is also to check between an empty string and `undefined`, so switching it to a simple truthiness check breaks the logic.

src/scripts/ui/inline-roll-links.ts
```diff
@@ -66,7 +67,7 @@ export const InlineRollLinks = {
                 : null;
         };
 
-        $repostLinks.filter("i[data-pf2-repost]").on("click", (event) => {
+        $html.find("i[data-pf2-repost]").on("click", (event) => {
```
CarlosFdez, line 70: We're getting off of JQuery. Try to find where we have htmlQuery() and htmlQueryAll() and do it as it says there.

It will be more verbose but the less stuff we have to change later on once the JQuery death knell is sent out the better.

src/scripts/ui/inline-roll-links.ts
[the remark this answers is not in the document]
CarlosFdez: I see. I almost feel like the property should be set to "true" or given a randomID() shared amongst the entire group, but its not a dealbreaker.

src/module/system/text-editor.ts
```diff
@@ -320,6 +320,62 @@ class TextEditorPF2e extends TextEditor {
...
+        );
+        if (buttons.length === 1) {
+            return buttons[0];
+        } else {
+            const checkGroup = document.createElement("div");
+            checkGroup.setAttribute("data-pf2-checkgroup", "");
+            for (const button of buttons) {
+                if (button === null) {
+                    // Warning should have been displayed already by #createSingleCheck
+                    return null;
```
CarlosFdez, line 355: Does it make the typecheck tricky if you move all this to a 

`else if (buttons.some((b) => !b) { 
    return null; 
}``` 

case earlier on? If it does don't worry about it, if it doesn't, it might make it easier to read for the next person if you do that.

stwlam:
Would a single link that makes a roll with the highest modifier be better?

CarlosFdez:
Looking at the adjustments section of the code, it seems they want rolls with alternative dcs to also be shown, and those might be harder to display as a single button (though subgroups would be doable, like arcana or occultism 22 as one button and society or library lore 20 as a second button all within the same group).

I can imagine it looking pretty weird for a button to get large though.

dmarcuse:
> Would a single link that makes a roll with the highest modifier be better?

That's an interesting idea, but I think in practice it's likely to be annoying when you e.g. need to add an extra bonus that's disabled by default (e.g. the +1 to visual perception checks to from goggles of night) or when the checks are for different things (e.g. you have the opportunity to recall knowledge with two different skills to learn different things, and can attempt both if you'd like). Maybe this kind of functionality could be supported using a different syntax or a parameter?

src/module/system/text-editor.ts
dmarcuse: Yeah, it currently serves as a type guard. I could use the suggestion and add a non-null assertion, but personally I like to avoid those wherever possible to get the most out of the type system.

src/scripts/ui/inline-roll-links.ts
dmarcuse: Done!

dmarcuse:
Any further thoughts on this? I used it for my tables this weekend at Paizocon Online (by rebasing the work onto the latest releases and manually building the system) and haven't run into any issues with it. I'm happy to make design changes if necessary.

src/module/system/text-editor.ts
CarlosFdez: I suspect remeda might have a way to avoid such an assertion, but I won't force it for this.

dmarcuse:
Thank you!

PR #7944, opened 2023-05-22
Copy trick item spellcasting entry to variant spell

oWave:
Selecting a variant of a spell cast with trick magic item doesn't work, it just results in a message without buttons:
image


oWave:
Apparently I was 5 hours to slow, lmao (#7942)

PR #7971, opened 2023-05-23
Batch retrieval of documents in `UUIDUtils.fromUUIDs`

In3luki:
For an actor with 40+ crafting formulas, network speed throttled to slow 3G:

Before:
Before
After:
After

I made the query compatible with v10 and v11 (untested) but it can probably wait until we start with v11.


In3luki:
Closing in favor of #8006

PR #7975, opened 2023-05-23
Fix application of dexterity modifier cap

### LINKED ISSUE #7973: Dex Cap not applying to AC
Dex Cap not working, resulting in armors granting full Dex modifier to AC.

Testing:
Fresh World
Lvl 1 Fighter with 16 Str and 18 Dex
Equipping any armor on the character increases their AC by the full Item Bonus and Dex Modifier with no Dex Cap.

Example: Breastplate (1 Dex Cap, 4 Item Bonus) is giving an AC of 21 (10 + 4 (Dex) + 3 (Prof) + 4 (Item)). Intended value should be 18 (only 1 from Dex).

stwlam:
Closes #7973

### DIFF
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -102,7 +102,8 @@ class Statistic extends SimpleStatistic {
         // If this is a character with an ability, add/set the ability modifier
         const abilityModifier =
             actor.isOfType("character") && data.ability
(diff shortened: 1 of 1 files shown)


CarlosFdez:
If this fails again I'll need to look more careful at creation of default parameters. Either more stringent deduping or moving the creation of these things to protected methods.

PR #7980, opened 2023-05-23
Hide undo damage button if no damage was dealt

In3luki:


src/module/actor/base.ts
```diff
@@ -1328,43 +1328,49 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length > 0 ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
+        const canUndoDamage = hpDamage > 0 || shieldDamage || persistentCreated.length;
```
stwlam, line 1331: Would this still allow for undoing healing?

In3luki: Oh true, that wouldn't work

src/module/actor/base.ts
```diff
@@ -1328,43 +1328,49 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length > 0 ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
+        const canUndoDamage = !!hpDamage || !!shieldDamage || !!persistentCreated.length;
```
stwlam, line 1331:
```suggestion
        const canUndoDamage = !!(hpDamage || shieldDamage || persistentCreated.length);
```

PR #7995, opened 2023-05-24
Fix acquisition of AC modifier adjustments

stwlam:


src/module/system/statistic/index.ts
```diff
@@ -515,11 +515,23 @@ class StatisticDifficultyClass<TParent extends Statistic = Statistic> {
...
     }
 
     get value(): number {
-        return 10 + new StatisticModifier("", this.modifiers, this.options).totalModifier;
+        return (
+            10 +
+            R.sumBy(
+                this.modifiers.filter((m) => m.enabled),
+                (m) => m.value
+            )
```
CarlosFdez, line 533: Wouldn't this change make it no longer resolve stacking rules?

PR #7998, opened 2023-05-24
Consolidate some styling between PC-sheet spellcasting and crafting tabs

stwlam:
Also enlarged some text in places

### Before:

image

image

### After:

image

image


CarlosFdez (approved):

CarlosFdez:
Merge when you think its safe!

PR #8001, opened 2023-05-24
Fix breakdowns when base damage is 0

CarlosFdez:


CarlosFdez:
I had to be quick and need to shut off due to a florida thunderstorm, please test more thoroughly if you can.

PR #8006, opened 2023-05-25
Speed up `fromUUIDs`

stwlam:


CarlosFdez:
I believe this supercedes #7971?

stwlam:
aw jeez I didn't even see that PR

In3luki:
This looks like a better way to determine the ids of cache misses. Parsing the UUID would probably break one way or antoher in v11.

CarlosFdez:
Does any caller rely on the returned results being in the same order as the ids that come in? 

Other than that I'll defer to in3luki's judgement, I have no idea what is good here.

stwlam:
I'm not aware of anything that depends on the order of the passed UUIDs.

PR #8041, opened 2023-05-28
Conditionally remove @UUID link labels on extract

stwlam:
Labels are removed when they are identical to their respective document names


CarlosFdez:
I cannot pretend to know what this is doing, so this is a blind merge. Regex be tricky.

PR #8042, opened 2023-05-28
Add initial kingdom model and extended campaign data support

CarlosFdez:
This PR creates a method to define and initialize extended party campaign data. The idea is that when kingmaker is enabled, the party will initialize a data model during data preparation. This object can be used by the sidebar to inject additional buttons into the side.

image

The campaign data is stored in system.campaign. This sort of structure makes it so that a party can only have a single campaign data extension. To prevent corruption, it checks for type mismatches during data preparation, and shows a warning (with an opportunity to delete data) if said corruption occurs.

image
image

The ability to register additional campaign types will be done at a later date. I was more interested in solving the data structure than the needs for campaigns we may or may not eventually support. The crown party button will do something once the sheet is being merged in.

## Alternative

If we want to reserve the ability to have multiple party data extensions active simultaneously, we could store the data in `system.kingdom` or `system.campaign.kingdom` (though the nesting slay me). This also removes the need to check for data mismatches.

### DIFF
--- a/src/module/actor/party/document.ts
+++ b/src/module/actor/party/document.ts
@@ -7,6 +7,8 @@ import { CombatantPF2e, EncounterPF2e } from "@module/encounter/index.ts";
 import { PartySheetRenderOptions } from "./sheet.ts";
 import { UserPF2e } from "@module/documents.ts";
 import { PartyCampaign, PartyUpdateContext } from "./types.ts";
+import { KingdomModel } from "./kingdom/index.ts";
+import { InvalidCampaign } from "./invalid-campaign.ts";
 
 class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | null> extends ActorPF2e<TParent> {
     override armorClass = null;
@@ -20,6 +22,11 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
 
     override get allowedItemTypes(): (ItemType | "physical")[] {
         return [...this.baseAllowedItemTypes, ...(this.campaign?.extraItemTypes ?? [])];
+    }
+
+    /** Parties use the campaign's level as their own level, as they otherwise don't have a level */
+    override get level(): number {
+        return this.campaign?.level ?? super.level;
     }
 
     /** Friendship lives in our hearts */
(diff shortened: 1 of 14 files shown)


src/module/actor/party/kingdom/model.ts
```diff
@@ -0,0 +1,127 @@
...
+        return KINGDOM_SCHEMA;
+    }
+
+    get extraItemTypes(): ItemType[] {
+        return ["action", "feat"];
+    }
+
+    get level(): number {
+        return this.details.level;
+    }
```
stwlam, line 30: Since this already isn't mirroring top-level system data, can we just dispense with the `details` object?

CarlosFdez: We can though I am concerned having a large number of top level properties in a data model subclass. 

I think level is perfect to move to the top level as it should never have been in details to begin with, but I am hesitant on the rest. I'll do so anyways though since data model schema isn't the core concern of this PR and its guaranteed this will be revisited later.

stwlam: Level aside, `details` is for stuff that is more-or-less mechanically insignificant. I believe level being in there for actors is from the original system fork.

src/module/actor/party/document.ts
```diff
@@ -22,6 +24,11 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         return [...this.baseAllowedItemTypes, ...(this.campaign?.extraItemTypes ?? [])];
     }
 
+    /** Parties use the campaign's level as their own level, as they otherwise don't have a level */
+    override get level(): number {
+        return this.campaign?.level ?? super.level;
```
stwlam, line 29: ok but isn't the level a kingdom level?

CarlosFdez: If i recall, the reason is that parties dont have a level in any sense and it came in handy for certain things (such as for the feat group which takes an actor).

Is there something you had in mind regarding parties having a level. I can get around it if I allow actor property overrides to places. I might need to be clever for statistics "anyways".

CarlosFdez: Yeah, campaign.kingdom was just an alternative if we wanted to support multiple campaigns for a party. May or may not be a good thing given modules will eventually be able to extend it (though its not on my high priority list, but I did design it so that its possible later).

PR #8049, opened 2023-05-29
implement Buckler Expertise feat

intrand:


stwlam:
This isn't going to be a viable approach.

intrand:
That's unhelpful. There isn't a trait for bucklers, and you cannot predicate upon a "contains" substring or similar. What would you recommend instead?

stwlam:
That's the problem: there isn't currently a way to isolate bucklers. Adding a toggle to every buckler isn't going to work for us, though. We'll have better approaches available when shield items are added to the system.

PR #8053, opened 2023-05-29
Fix item create behavior for exploration and downtime actions

CarlosFdez:
Also moves us away from ramming dataset into item source and hoping foundry can figure it out.

### DIFF
--- a/src/module/actor/sheet/helpers.ts
+++ b/src/module/actor/sheet/helpers.ts
@@ -1,29 +1,40 @@
 import type { ActorPF2e } from "@actor";
+import { SpellSource } from "@item/spell/index.ts";
+import { ZeroToTen } from "@module/data.ts";
 import { ErrorPF2e } from "@util";
+import * as R from "remeda";
 
-function onClickCreateSpell(actor: ActorPF2e, data: Record<string, string | string[] | number | undefined>): void {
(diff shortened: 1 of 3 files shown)


src/module/actor/sheet/helpers.ts
```diff
@@ -1,29 +1,40 @@
...
+        type: "spell",
+        name: R.compact([newLabel, levelLabel, spellLabel]).join(" "),
+        system: {
+            level: { value: level || 1 },
+            location: { value: String(data.location) },
+            traits: {
+                value: level === 0 ? ["cantrip"] : [],
+            },
+        },
+    } satisfies DeepPartial<SpellSource>;
```
stwlam, line 31: Any reason to not just do a regular type annotation here?

CarlosFdez: the mergeObject() later doesn't work if I do a regular type annotation, it doesn't know that "system" as a property exists.

stwlam: but that's exactly what's happened below

CarlosFdez: The `?? {}` is the part that's different.

stwlam: Ohh, I see.

stwlam: buenooooo

PR #8056, opened 2023-05-30
Fix creating new cantrips

CarlosFdez:


CarlosFdez:
Turns out this just fixes a regression in #8053 that I didn't realize so I'll move it over there.

PR #8063, opened 2023-05-30
Run UUIDs in embedded items through name-to-id conversions

stwlam:


TikaelSol:
Data changes approved

CarlosFdez:
uncertain about why we need to convert the choice set selection as well but extract is beyond my purview so I'll trust in your judgement. The rest looks fine anyways.

Issue #8064, opened 2023-05-30
[Feature Request] - Thaumaturge Weapon Implement - Choice option for implement


JonDemand:
Currently,  the weapon implement feature has no idea which of your available weapons is your weapon implement,  so the 5th level part of it - Critical Specialisation - Isn't applying.

Is it possible to make this implement option pop up a choice dialog where you can choose which weapon is your implement from among those on the character that you're proficient in?  this chosen weapon then at level 5 would get Critical Specialisation if you don't gain it from another source,  such as Ancestry weapons etc (Kobold and Tricky pick etc)

DocSchlock:
This request was completed and is available in the system.

PR #8065, opened 2023-05-30
Add CritSpec for Thaumaturge Weapon Implement

cjneidhart:
Users are prompted to choose a weapon after choosing the Weapon Implement. At level 5, that weapon will gain its critical specialization effect. Addresses #8064.

### DIFF
--- a/packs/data/classfeatures.db/weapon.json
+++ b/packs/data/classfeatures.db/weapon.json
@@ -21,8 +21,33 @@
         },
         "rules": [
             {
+                "allowDuplicate": false,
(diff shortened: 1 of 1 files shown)


cjneidhart:
By RAW, the player is only allowed to choose one-handed weapons. I couldn't find a way to list only one-handed weapons in the ChoiceSet, but I could be mistaken.

stwlam:
We don't really have sufficient system support for a pure rule-element solution to this.

TikaelSol:
Let's hold off on this one until we can do it properly.

PR #8071, opened 2023-05-31
Fix class feat prefilter arrays

stwlam:
It doesn't look like this ultimately made a difference, but good to keep it tidy anyway :blush:


CarlosFdez:
I think this is actually housekeeping, not a real bugfix.

image

PR #8087, opened 2023-06-02
Fix base damagetype and diesize lookup for 0 base damage

CarlosFdez:
Theoretically we could make the base damage loop not an if/else but I'm a bit concerned about changing that at this time.

### DIFF
--- a/src/module/system/damage/formula.ts
+++ b/src/module/system/damage/formula.ts
@@ -52,10 +52,11 @@ function createDamageFormula(
         const list = typeMap.get(baseEntry.damageType) ?? [];
         typeMap.set(baseEntry.damageType, list);
 
-        if ("terms" in baseEntry) {
+        if (baseEntry.terms) {
             list.push(...baseEntry.terms.map((t) => ({ ...baseEntry, ...t, label: null, critical: null })));
         } else if ((baseEntry.diceNumber && baseEntry.dieSize) || baseEntry.modifier) {
-            const { diceNumber, dieSize, modifier, damageType } = baseEntry;
+            const { diceNumber, dieSize, damageType } = baseEntry;
+            const modifier = baseEntry.modifier ?? 0;
             const label = (() => {
                 const diceSection = diceNumber ? `${diceNumber}${dieSize}` : null;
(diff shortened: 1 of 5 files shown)


src/module/system/damage/formula.ts
```diff
@@ -52,10 +52,11 @@ function createDamageFormula(
         const list = typeMap.get(baseEntry.damageType) ?? [];
         typeMap.set(baseEntry.damageType, list);
 
-        if ("terms" in baseEntry) {
+        if (baseEntry.terms) {
             list.push(...baseEntry.terms.map((t) => ({ ...baseEntry, ...t, label: null, critical: null })));
         } else if ((baseEntry.diceNumber && baseEntry.dieSize) || baseEntry.modifier) {
```
stwlam, line 57: Won't this still make it so the 0-dice override prevents the base damage from being added to `list`?

src/module/system/damage/formula.ts
```diff
@@ -107,7 +106,7 @@ function createDamageFormula(
 
     // Add modifiers
     for (const modifier of damage.modifiers.filter((m) => m.enabled && outcomeMatches(m))) {
-        const matchingBase = baseTerms.find((b) => b.damageType === modifier.damageType) ?? baseTerms[0];
+        const matchingBase = damage.base.find((b) => b.damageType === modifier.damageType) ?? damage.base[0];
```
stwlam, line 109:
```suggestion
        const matchingBase = damage.base.find((b) => b.damageType === modifier.damageType) ?? damage.base.at(0);
        if (!matchingBase) continue;
```

CarlosFdez: damage.base is explicitly checked to have a length at the start of the function, but typescript is too dumb to realize that.

If we even could though, I'm not sure if this is the correct behavior if we don't have a base at all. Perhaps it'd be better to fall back to untyped.

src/module/system/damage/formula.ts
CarlosFdez: Yup, but it continues to work just fine. Base only matters for fallbacks, and that's preserved.
image

PR #8089, opened 2023-06-02
Fix strikes for weapons lacking "base" damage

stwlam:


src/module/system/damage/weapon.ts
```diff
@@ -512,6 +512,12 @@ class WeaponDamagePF2e {
             ignoredResistances,
         };
 
+        // If a weapon deals no base damage, remove all bonuses, penalties, and modifiers to it.
```
CarlosFdez, line 515: remove all basic dice and modifiers? Its not quite removing all.

stwlam: it = base damage

CarlosFdez: I see, I interpret it as "weapon". Ambiguous but sure.

PR #8127, opened 2023-06-04
Update origin flags for spell messages

CarlosFdez:
Sorry I broke oWave's issue when I made some extra changes


CarlosFdez:
Actually nevermind, what happened is that the default github ordering is p.bad and it was old. It got through properly!

PR #8131, opened 2023-06-04
Fix showing world clock

xdy:


stwlam:
This setting is already a number.

xdy:
I should have added some context. :)
It works fine on a new world, but on a converted one I get this error (no modules enabled):
image
After I added Number to the setting, it worked on that world and continues to work.
I just noticed that configuring the clock (changing from 24h to 12h) also fixes it.

xdy:
Might be better as a migration I guess.

xdy:
Though, if I 'fix' it by just configuring the clock (i.e. on master, with no Number around the setting) the above error returns after a refresh, so something isn't being saved.
With Number it always works (for me.)

stwlam:
This doesn't look like a system issue. We'll check upstream for something, but if this lone setting problem then any other also potentially is.

PR #8142, opened 2023-06-05
Remove global style overrides for chat log

stwlam:
This has always been problematic and is no longer tenable with V11's "Jump to Bottom" feature


CarlosFdez:
How do spells currently after this change?

stwlam:
Re-removed border-bottom

image

PR #8146, opened 2023-06-05
Eldritch Researcher Dedication - Rules Added

cadowtin:


TikaelSol:
We are currently on a data entry freeze for the v11 transition.

PR #8157, opened 2023-06-06
Remove deprecated system features

stwlam:


src/module/item/spellcasting-entry/document.ts
```diff
@@ -329,18 +329,6 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
...
-    /*  Deprecations and Compatibility              */
-    /* -------------------------------------------- */
-
-    /**
-     * To prevent (or delay) console spam, will send out a deprecation notice in a later release
-     * @deprecated
-     */
-    getSpellData(): Promise<SpellcastingSheetData> {
-        return this.getSheetData();
-    }
```
CarlosFdez, line 343: I've actually been unsure about this one (I've mentioned it before). Not only did it never have a deprecation warning, but I think this signature is actually more accurate than getSheetData(). This function is used by modules to determine what spells are available to cast at what levels.

I think we should keep this one around, and then consider having a way to acquire "active spells" from a collection that getSheetData() uses. Then we can probably kill this function without remorse.

src/module/system/statistic/index.ts
```diff
@@ -271,24 +270,6 @@ class Statistic extends SimpleStatistic {
...
-                });
-
-                return this.check.roll({
-                    extraRollOptions: args.options ? [...args.options] : [],
-                    ...options,
-                });
-            };
-        }
-
-        return trace;
```
CarlosFdez, line 291: Thoughts on keeping this around for when strikes get converted? This function was created as a way for us to convert things to statistic without breakage for a bit. We can always dig into git history to bring it back, so I'm not hardset either way.

src/module/item/spellcasting-entry/document.ts
stwlam: No belly aching out of you!

CarlosFdez: This is currently the only way to acquire certain mechanically relevant information, and its known public api that never had a warning or any communication that its going away.

We need to continue work on collections anyways for future features.

stwlam: ok fine!

src/module/system/statistic/index.ts
stwlam: a shame you added a deprecation notice!

PR #8181, opened 2023-06-08
Add HTML template and functional sheet javascript for a "simple NPC" sheet

silvative:


stwlam:
:pridebrow:

silvative:
what even is a tactor anyway

silvative:
I can fix the squigglies by changing this line but I don't know what to change it to

image

Eg adding `class SimpleNPCSheet<TActor extends NPCPF2e> extends CreatureSheetPF2e<NPCPF2e>`  works but it's some cargo cult thinking, this is too typescript for me to be comfortable with

silvative:
HA
WHO'S PRETTIER NOW YOU BITCH

PR #8186, opened 2023-06-08
Add Homebrew Versatile Trait Support

MrVauxs:
This is a quick patch that I noticed could be implemented.


MrVauxs:
Welp, no clue how to fix *that* one.

cadowtin:
Probably need to look into damage typing in the system and see the allowed strinbgs for damage types and either updatre it, or change yours to fit that type

stwlam:
Closing in favor of #11373

PR #8206, opened 2023-06-10
Replace ChoiceSet pack queries with index lookups and predicate testing

stwlam:
Sample implementation with the cleric's deity class feature. We'll need to convert the rest, but the `allowedDrops` predicates should be helpful (before getting removed in some cases).

### DIFF
--- a/packs/classfeatures/deity.json
+++ b/packs/classfeatures/deity.json
@@ -21,28 +21,23 @@
         },
         "rules": [
             {
-                "allowedDrops": {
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Checked it out locally to verify. I like the change in the RE format, but what's the plan for the ol "drag thingy here"?

stwlam:
I made `pack` optional. If it's omitted, it'll getting every pack with matching item types.

CarlosFdez:
I see, so the idea is that it shouldn't need it?

CarlosFdez (dismissed):

Issue #7605, opened 2023-04-29
Advanced Weapon Training needs rule elements


stwlam:
It may need some expansion of the MartialProficiency rule element.

PR #8209, opened 2023-06-10
Add rule elements to Advanced Weapon Training

cjneidhart:
Addresses #7605.
This change adds the field `sameAsGroup` to the MartialProficiency rule element and the proficiency object in `actor.system.martial`, in order to implement the functionality required by Advanced Weapon Training. It correctly applies all proficiencies to Strikes, but it doesn't display correctly on the proficiencies tab when Advanced Weapon Training uses the same group as the level 5 feature Fighter Weapon Mastery, which is likely how most players will use it. If that's necessary, I can change `prepareMartialProficiencies` to fix it. I haven't done it for now since that would probably require a migration.


cjneidhart:
An example of what happens when Sword is chosen for Fighter Weapon Mastery and Advanced Weapon Training, on a level 6 Fighter.
image

cjneidhart:
This change should correctly fix all issues. The system can now properly calculate the rank of Advanced Weapon Training even when it uses the same group as another rule element, such as Fighter Weapon Mastery. Here's the same fighter as above now:
image

PR #8211, opened 2023-06-10
Add rule elements to Exemplary Finisher

### LINKED ISSUE #7957: Create Exemplary Finisher for Gymnast
If the foe is Grabbed, Restrained or Prone, you gain a circumstance bonus to the damage roll equal to twice the number of weapon damage dice.

cjneidhart:
Fixes #7957. Adds extra damage for the Gymnast style, and adds Notes for the other styles.


TikaelSol:
Instead of adding extra damage and notes on this item we should instead add those to the style items themselves then predicate them on having this class feature.

PR #8216, opened 2023-06-11
Fix npm run watch crashes when updates are made while a world is running

CarlosFdez:
static/packs will need to be deleted from people's machines for builds to work properly after this. Perhaps I should delete it just in case?

After this static files are not deleted (only overwritten) between refreshes in npm run watch, but they are deleted when `npm run watch` or `npm run build` is ran. Perhaps that is fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@ foundryconfig.json
 node_modules/
 package/
 packs-temp/
-static/packs/
 types/foundry/node_modules/
 types/foundry/package-lock.json
 
--- a/build/clean.ts
+++ b/build/clean.ts
@@ -1,0 +1,17 @@
+import fs from "fs";
+import path from "path";
+
+// Clean output directory, or create build directory
+const outDir = path.resolve(process.cwd(), "dist");
+if (fs.existsSync(outDir)) {
+    const filesToClean = fs.readdirSync(outDir).map((dirName) => path.resolve(outDir, dirName));
+    for (const file of filesToClean) {
+        fs.rmSync(file, { recursive: true });
+    }
+} else {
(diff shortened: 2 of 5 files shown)


stwlam:
Yeah could we just delete the static/packs directory and recreate it every time? It should be possible to do it blindly without throwing an error if nothing is there

CarlosFdez:
The problem is that vite doesn't seem configurable enough to selectively copy only *some* static files.

1) Vite currently wipes everything in dist everytime it rebuilds. This explodes due to file locks. 
2) Vite copies everything in static over to dist on every rebuild. If one of those folders is "packs", this explodes due to file locks.

CarlosFdez:
I guess as a reminder, this is the current vite build flow you set up

1) build packs writes packs to static/packs
2) vite runs, doing these subordinate steps on the first and every rebuild
    1) wipe dist (due to emptyOutDir)
    2) copy everything in static/ to dist/  (due publicDir)
    3) Do the rest of the build this part doesn't matter to foundry

To avoid the crash due to wiping dist/: we need to make sure dist/ is cleaned only on the first run. That's npm run clean. 
To avoid the crash due to overwriting pack files: packs must not exist in static/ during rebuilds.

In other words, I don't see a need to avoid a separate pre-build step, but if we can configure a pre initial build and a post build step within vite itself, we can do the clean there and delete static/packs after the first run. Is this what you were getting at?

CarlosFdez:
For npm run hot at least, it seems that the vite config is reloaded when you change it. I think attempting to do the clean there is going to end up being a bit bug prone, especially if we make packs not go into dist/ after the first build somehow.

PR #8239, opened 2023-06-12
Show a progress bar when fetching pack indices for choice set rule elements

In3luki:
Currently loading all indices for e.g. the Versatile Heritage can take more than 20 seconds on slower connections with no indicator why the prompt isn't opening.
Also removes `JQuery` from the `Progress` class.

image


stwlam:
Maybe we should use the core progress bar for this? I know Crucible is using it when loading up the talent tree.

In3luki:
Crucible is utilizing the `TextureLoader` class  that uses `SceneNavigation.displayProgressBar` which does most of what our `Progress` class was doing. I've removed all unnecessary code from `Progess`.

PR #8241, opened 2023-06-12
add pronoun field in player config

Drental:
image

image

tested locally


Drental:
yup, straight copy paste.
I'll check the template to be sure.

Drental:
there is nothing else we are missing!

PR #8243, opened 2023-06-12
Add (placeholder) compendium pack banner images

silvative:
This adds them via the system.json, but we might want a higher tech solution (eg replacing the default) instead. Regardless, here are the files. We can change the images later if I get feedback.


silvative:
Updated assets with a darker background colour & less cartoonish art style. Replaced diagonal banner with vertical to lessen "optical illusion". Note that due to a CSS rule on locked compendium packs they appear very transparent. We could always override it to .6 (default) if we want.

stwlam (dismissed):

PR #8252, opened 2023-06-13
Minor visual updates to character skills

CarlosFdez:
image

The one other source of proficiency dropdowns, spells, seems to continue working fine. Let me know if I should test anything else
image


stwlam:
I demand before/after

CarlosFdez:
BEFORE:
image

CarlosFdez:
This is a slight alternative where we make the text center again and reduce the gap, if you'd rather go that route
image

PR #8262, opened 2023-06-13
Add JSON pack files to release artifacts

In3luki:
Sample zip file:
https://github.com/In3luki/pf2e/releases/latest/download/packs-json.zip

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -25,6 +25,9 @@ jobs:
           npm ci
           npm run build
 
+      - name: Build JSON packs
+        run: npm run build:packs:json
+
       - name: Update Manifest
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/cd.yml
```diff
@@ -43,5 +50,5 @@ jobs:
           draft: false
           prerelease: false
           token: ${{ secrets.GITHUB_TOKEN }}
-          artifacts: './dist/system.json,./dist/pf2e.zip'
+          artifacts: "./dist/system.json,./dist/pf2e.zip,./packs-json/packs-json.zip"
```
stwlam, line 53: I seem to remember github actions being picky about quote style. Did you run your test before or after changing this?

In3luki: I've set up prettier to run on save so that would have been changed for all my tests but I can change it back to single quotes to be safe.

stwlam: eh, let's try and see

PR #8267, opened 2023-06-13
use ActionDefaultOptions for Earn Income Action

Drental:
I am not sure if we want to change this. It's basically a fix for my keybind menagerie, but I can make a special case there to handle this as well. I don't really want to do a beraking change here.


stwlam:
https://github.com/foundryvtt/pf2e/blob/master/packs/macros/earn-income.json

PR #8272, opened 2023-06-14
Move most inventory functions to context menu

CarlosFdez:
Also moves the backpack toggle to the left of the item name. Note that Foundry context menus don't put themselves in the document body and thus do weird things to scrolling

image

image


stwlam:
:monocle_face: 

Screenshot from 2023-06-15 00-52-54

CarlosFdez:
Bag contents are indented on both sides fsr. Old think. I could attempt to see if i can get around it later.

CarlosFdez:
That will be fixed by #8305. I'll rebase this once that's in and make more context menu tweaks.

PR #8299, opened 2023-06-17
Set `item` as `RuleElementPF2e#parent`

stwlam:


CarlosFdez (approved):

CarlosFdez:
Merge when you think its safe.

PR #8308, opened 2023-06-17
Add `pages` array to all `JournalEntry` sources in the pack builder

In3luki:
Same deal as #8195


stwlam:
Which journals are lacking them?

In3luki:
> Which journals are lacking them?

Every entry in the Domains compendium. But looking at other journals it looks like the whole pack might need a migration.

stwlam:
I recently killed off the hero point deck compendium by converting it into a single journal entry. Maybe we could do the same with the domains compendium, even though it has the additional complication of being linked from several other documents.

PR #8325, opened 2023-06-18
Add missing rarity trait to AncestryRollOptions

reyzor1991:
Need to fix Adopted Ancestry rule element

### DIFF
--- a/src/module/item/ancestry/document.ts
+++ b/src/module/item/ancestry/document.ts
@@ -150,8 +150,7 @@ class AncestryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends 
     /** Generate a list of strings for use in predication */
     override getRollOptions(): string[] {
         return [
-            ...super.getRollOptions(),
-            `item:rarity:${this.rarity}`
+            ...super.getRollOptions(), `item:rarity:${this.rarity}`
         ];
     }
 }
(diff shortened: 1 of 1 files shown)


reyzor1991:
squash commits please

src/module/item/ancestry/document.ts
```diff
@@ -143,6 +147,11 @@ class AncestryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
             actor.rollOptions.all[`self:trait:${trait}`] = true;
         }
     }
+
+    /** Generate a list of strings for use in predication */
+    override getRollOptions(): string[] {
+        return [...super.getRollOptions(), `item:rarity:${this.rarity}`];
+    }
```
stwlam, line 154: This needs to accept the standard parameter

reyzor1991: Prefix was added

stwlam: the roll option must then use that prefix

reyzor1991: new version of code 
`return [...super.getRollOptions(prefix), `item:rarity:${this.rarity}`];`

src/module/item/ancestry/document.ts
```diff
@@ -143,6 +147,11 @@ class AncestryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
             actor.rollOptions.all[`self:trait:${trait}`] = true;
         }
     }
+
+    /** Generate a list of strings for use in predication */
+    override getRollOptions(prefix = this.type): string[] {
+        return [...super.getRollOptions(prefix), `item:rarity:${this.rarity}`];
```
stwlam, line 153:
```suggestion
        return [...super.getRollOptions(prefix), `${prefix}:rarity:${this.rarity}`];
```

Issue #8326, opened 2023-06-18
Undefined item.system.description.gm on migrated items causes errors in item sheet


Nightara:
item/sheet/base.ts:getData() doesn't check item.system.description.gm for existence, causing an undefined error on item.system.description.gm.trim() in line 95 when used with items imported from a previous version (4.X).

stwlam:
It's in the template.json file. Do you have a reproduction example?

Nightara:
I only saw the issue on data imported from 4.6.X (And potentially other versions, too), items created in 5.0.1 don't have the issue, so it will probably be hard to reproduce.

I tried to reproduce the error (After running Foundry with the workaround I suggested in #8327, then reverting to the unmodified pf2e module), but it doesn't seem to occur anymore.
I'm not very familiar with the code of this module, maybe running it on the modified version once (Which allowed everything to run without crashes) caused the system to rewrite the items?

PR #8327, opened 2023-06-18
Added workaround for Issue #8326

Nightara:
Items imported from earlier versions of this module do not have the item.system.description.gm property, causing errors when trying to open their respective item sheets.

Checking whether that item exists before calling .trim() should fix the issue.

### DIFF
--- a/src/module/item/sheet/base.ts
+++ b/src/module/item/sheet/base.ts
@@ -92,7 +92,7 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
             rollData,
             async: true,
         });
-        enrichedContent.gmNotes = await TextEditor.enrichHTML(item.system.description.gm.trim(), {
+        enrichedContent.gmNotes = await TextEditor.enrichHTML(item.system.description.gm ? item.system.description.gm.trim() : "", {
             rollData,
             async: true,
         });


src/module/item/sheet/base.ts
```diff
@@ -92,7 +92,7 @@ export class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem> {
             rollData,
             async: true,
         });
-        enrichedContent.gmNotes = await TextEditor.enrichHTML(item.system.description.gm.trim(), {
+        enrichedContent.gmNotes = await TextEditor.enrichHTML(item.system.description.gm ? item.system.description.gm.trim() : "", {
```
In3luki, line 95:
```suggestion
        enrichedContent.gmNotes = await TextEditor.enrichHTML(item.system.description.gm, {
```
Linting would fail for this line because it is too long. Also I don't see a reason for that `trim` to be there in the first place so removing it should be a better fix.

In3luki:
PR titles should be in imperative mood and state clearly and concisely what is being changed.

stwlam:
Which module are you talking about? And `description.gm` is in the template.json, so it is filled if omitted.

Nightara:
I was referring to the entire pf2e module. After migrating from 4.6.X to 5.0.1, trying to open any item sheet imported from 4.6.X caused an undefined error for `description.gm`.
Since I assume the template is only used for items created in 5.0.1, the field being filled probably doesn't affect any imported data.

PR #8336, opened 2023-06-19
Filter out damage dice with invalid outcomes

CarlosFdez:


src/module/system/damage/formula.ts
```diff
@@ -81,8 +81,11 @@ function createDamageFormula(
     // Sometimes a weapon may add base damage as bonus modifiers or dice. We need to auto-generate these
     const BONUS_BASE_LABELS = ["PF2E.ConditionTypePersistent"].map((l) => game.i18n.localize(l));
 
+    // Test that a damage modifier or dice partial is compatible with the prior check result
+    const outcomeMatches = (m: { critical: boolean | null }): boolean => critical || m.critical !== true;
+
     // Add damage dice. Dice always stack
-    for (const dice of damage.dice.filter((d) => d.enabled)) {
+    for (const dice of damage.dice.filter((d) => d.enabled && outcomeMatches(d))) {
```
stwlam, line 88: I'm wondering what the `enabled` property was for if it wasn't prefiltered at one point.

CarlosFdez: Enabled did prefilter tho, i just added outcomeMatches() to match modifiers

PR #8356, opened 2023-06-19
Add schema to `CraftingEntryRuleElement`

In3luki:


src/module/actor/character/crafting/entry.ts
```diff
@@ -231,6 +232,16 @@ class CraftingEntry implements Omit<CraftingEntryData, "parentItem"> {
...
 
+interface CraftingEntryRuleData extends RuleElementData {
+    isAlchemical?: boolean;
+    isDailyPrep?: boolean;
+    isPrepared?: boolean;
+    maxItemLevel?: number;
+    maxSlots?: number;
+    craftableItems?: RawPredicate;
+    preparedFormulas?: PreparedFormulaData[];
+}
```
stwlam, line 243: Shouldn't it be possible to generate this from the schema?

stwlam: You'd want `SourceFromSchema`

In3luki: oh, didn't see that. but it has the same error 🙁
image

stwlam: make it `Partial`, I guess

In3luki: yeah that works, but it would make `itemUUID` not required anymore which would technically be wrong.

stwlam: partial with an intersection on the picked property

In3luki: should be good now

PR #8379, opened 2023-06-21
Silence warnings from `DamageDiceRuleElement#resolveInjectedProperties`

stwlam:


src/module/rules/rule-element/damage-dice.ts
```diff
@@ -54,41 +54,53 @@ class DamageDiceRuleElement extends RuleElementPF2e<DamageDiceRuleSchema> {
...
 
-            const resolvedBrackets = this.resolveValue(this.brackets, {}, { resolvables: params.resolvables });
+            // If this rule element's predicate would have passed without all fields being resolvable, send out a
+            // warning.
+            const testPassed =
+                this.predicate.length === 0
+                    ? true
+                    : this.resolveInjectedProperties(this.predicate).test(
+                          params.test ?? this.actor.getRollOptions(["damage"])
+                      );
```
CarlosFdez, line 65: Can't do a suggestion cause no prettier here to fix it for me, but 
```
const testPassed = !this.predicate.length || this.resolveInjectedProperties(this.predicate).test(
    params.test ?? this.actor.getRollOptions(["damage"])
);```

stwlam: That is less readable, but ok

PR #8387, opened 2023-06-21
Make inline Template/Check type optional and introduce roller:self

CarlosFdez:
This enables both `@Check[athletics]` syntax and also introduces a roller:self property so that the owning actor is the one that makes the roll instead of the selected actor. Added to the marshal stance skill checks as a proof of concept of the latter (but not the former, though I'd be perfectly for changing that too).

The driving purpose was actually kingmaker related. Unfortunately there is no way to make a check roll using an actor for which there is no existing token. The only alternative I could think of was taking the action macros approach with HTML strings and custom listeners, but this seemed like a cleaner solution that would also have alternative uses in the system.

We can change the property name to actor if you think that'll work better. I took the first suggestion I received.

I also would prefer making parseInlineParams a external function but I have no idea where I would put it 🙃


stwlam:
What happens if the user doesn't have ownership of the owning actor?

CarlosFdez:
Currently it rolls, but it shouldn't. I kinda want to make it contingent on actor type, so I'll revisit once https://github.com/foundryvtt/pf2e/pull/8393 is in.

CarlosFdez:
If users lack permission over the actor, it'll now render just the text. If an already existing link exists after a permission change and the user clicks it it'll fail silently.

image

PR #8393, opened 2023-06-22
Prepare kingdom skills and control dc

CarlosFdez:


src/module/actor/party/document.ts
```diff
@@ -132,6 +133,17 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
 
+    /** Include campaign statistics in party statistics */
+    override getStatistic(slug: string): Statistic | null {
+        const statistic = super.getStatistic(slug);
+        if (statistic) return statistic;
+
+        const campaignStat = this.campaign?.getStatistic?.(slug);
+        if (campaignStat) return campaignStat;
+
+        return null;
```
stwlam, line 144:
```suggestion

        return campaignStat ?? null;
```

CarlosFdez: I wrote the way I did with the anticipation of returning possibly alternative statistics, but for party I can't think of any right now.

PR #8401, opened 2023-06-22
Switch to use pnpm for builds instead of npm.

xdy:
Why? Saves a second or so when building on my machine, halves node_modules size (which makes it several times faster to clean out). And, I was in a really boring meeting... :)

Requires that developers have pnpm installed:
```
npm install -g pnpm
```

Test script:
```
git switch pnpm && rm -rf node_modules/* && time pnpm install && time pnpm run build && du -hs node_modules
```

Results with everything but times and sizes stripped out:
```
real    0m0.756s
user    0m0.015s
sys     0m0.030s

real    0m16.573s
user    0m0.000s
sys     0m0.015s

237M    node_modules
```

Using npm:
```
git switch master && rm -rf node_modules/* && time npm install && time npm run build && du -hs node_modules
```

Results with everything but times and sizes stripped out:
```
real    0m4.852s
user    0m0.015s
sys     0m0.030s

real    0m17.966s
user    0m0.031s
sys     0m0.030s

468M    node_modules
```

Reference: https://pnpm.io/pnpm-vs-npm


stwlam:
Thanks, but saving five seconds isn't really worth making sure everyone is using a new workflow.

xdy:
Yeah, I figured. What can I say, I was bored. :)

PR #8402, opened 2023-06-22
Set prototype tokens' default images in same way as token documents

stwlam:
@MrVauxs wants :shrug:


src/module/scene/token-document/document.ts
```diff
@@ -312,8 +290,33 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         }
     }
 
+    /** Synchronize the token image with the actor image if the token does not currently have an image */
+    static setDefaultImage(token: TokenDocumentPF2e | PrototypeTokenPF2e<ActorPF2e>): void {
```
CarlosFdez, line 294: assignDefaultImage()? set implies a setter method where the parameter is what you're setting it to.

PR #8415, opened 2023-06-24
send item to container from carry-type menu

reonZ:
Sometimes, the number of items on an actor can make drag & drop across the sheet's inventory a bit daunting.

I added the containers inside the carry-type menu making switching to different containers more convenient.

send


stwlam:
I'm pretty sure something like this already exists and just isn't showing in the menu.

reonZ:
Classic me !

reonZ:
If you don't scrap this PR, there will still need to be some work done to make sure we don't send a container inside a container that is inside this one, because it breaks everything.

PR #8417, opened 2023-06-24
Fix pair of bugs with carry-type menu

stwlam:
- Fix the "stowed" option never appearing
- Prevent removing an item from its (non-stowing) container if selecting the "worn" option

Maybe supersedes #8415, but the PR author may want to build on the now-fixed bug.

### DIFF
--- a/src/styles/ui/_hover.scss
+++ b/src/styles/ui/_hover.scss
@@ -92,10 +92,8 @@
                             cursor: default;
                             text-shadow: none;
                         }
-
                     }
                 }
-
 
                 ul {
                     list-style: none;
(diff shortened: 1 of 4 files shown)


src/styles/ui/_hover.scss
```diff
@@ -79,15 +80,23 @@
...
+
+                        &:hover {
+                            cursor: default;
+                            text-shadow: none;
+                        }
+
+                    }
                 }
 
+
```
CarlosFdez, line 99:
```suggestion
```

src/styles/ui/_hover.scss
```diff
@@ -79,15 +80,23 @@
...
-                a.item-location-option.selected {
-                    color: white;
+                    &.selected {
+                        color: white;
+
+                        &:hover {
+                            cursor: default;
+                            text-shadow: none;
+                        }
+
```
CarlosFdez, line 95:
```suggestion
```

PR #8419, opened 2023-06-24
Add earned income to actor when using the Earn Income macro.

### LINKED ISSUE #7328: Earn Income Macro should add currency to PC
It would be nice if the earn income macro from the system could automatically add the result of the currency to the pcs sheet after the result is produced. or a button to press to add it.

xdy:
Fixes #7328


stwlam:
Would rather have a confirmation step via chat message button

xdy:
I couldn't get the chat message button to work, then I forgot about it. :(

PR #8427, opened 2023-06-25
Fix bugs with roller:self

CarlosFdez:
plus un-jquerying

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -447,8 +447,8 @@ class TextEditorPF2e extends TextEditor {
         }
 
         // If the roller is self, don't create an inline roll if the user has no control over it
-        if (params.roller === "self" && !actor?.canUserModify(game.user, "update")) {
-            return createHTMLElement("span", { children: [html.innerHTML] });
+        if (params.roller === "self" && actor && !actor.canUserModify(game.user, "update")) {
+            return createHTMLElement("span", { children: [html.innerText] });
         }
 
         return html;
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -37,6 +37,15 @@ export const InlineRollLinks = {
             newButton.setAttribute("data-pf2-repost", "");
             newButton.setAttribute("title", game.i18n.localize("PF2E.Repost"));
             link.appendChild(newButton);
+
+            newButton.addEventListener("click", (event) => {
+                event.stopPropagation();
+                const target = event.target;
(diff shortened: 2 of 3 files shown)


src/module/system/text-editor.ts
```diff
@@ -447,8 +447,8 @@ class TextEditorPF2e extends TextEditor {
         }
 
         // If the roller is self, don't create an inline roll if the user has no control over it
-        if (params.roller === "self" && !actor?.canUserModify(game.user, "update")) {
-            return createHTMLElement("span", { children: [html.innerHTML] });
+        if (params.roller === "self" && actor && !actor?.canUserModify(game.user, "update")) {
```
stwlam, line 450:
```suggestion
        if (params.roller === "self" && actor && !actor.canUserModify(game.user, "update")) {
```

src/scripts/ui/inline-roll-links.ts
```diff
@@ -98,42 +98,44 @@ export const InlineRollLinks = {
...
-        $links.filter("[data-pf2-check]").on("click", (event) => {
-            const { pf2Check, pf2Dc, pf2Traits, pf2Label, pf2Adjustment, pf2Roller } = event.currentTarget.dataset;
-            const parent = documentFromDOM(event.currentTarget);
-            const actors = (() => {
-                if (pf2Roller === "self") {
-                    const document = documentFromDOM(event.currentTarget);
-                    const validActor = document instanceof ActorPF2e && document.canUserModify(game.user, "update");
-                    return validActor ? [document] : [];
-                }
+        for (const link of htmlQueryAll(html, "span[data-pf2-check]")) {
```
stwlam, line 101: The links are already available: you shouldn't need to requery

CarlosFdez: The links are only used for jquery to filter and the injectRepostElement. The stuff already converted from jquery requeries. Probably because its cleaner than doing an html filter for has attribute.

If I replace for a filter I'll need to convert the other one as well, and I suspect the base way to do would be separate htmlQueries for each type of link, and then combine them together for the repost elements, because querying for every single "span" isn't quite right.

src/module/system/text-editor.ts
CarlosFdez: dammit, I knew I had forgotten something when I ported it over!

src/scripts/ui/inline-roll-links.ts
CarlosFdez: in the interest of laziness I'll convert both, though it does make the code harder to skim now that I'm trying it.

PR #8430, opened 2023-06-25
Render damage taken message flavor and content separately

In3luki:
image


stwlam:
Starting to think this wasn't worth it

PR #8431, opened 2023-06-25
Propagate check message flags to chat messages created by the treat wounds macro

In3luki:


stwlam:
Is this addressing a particular problem?

In3luki:
I was testing the `healing-received` selector and noticed that it doesn't work with the macro rolls because they have no flags.

src/scripts/macros/treat-wounds.ts
```diff
@@ -156,6 +156,7 @@ async function treat(
 
             if (riskySurgery) {
                 ChatMessagePF2e.create({
+                    flags: message.flags,
```
stwlam, line 159: Is this a message document or source? We'd want the source flags.

In3luki: It's the source now.

PR #8443, opened 2023-06-26
Allow setting fixed attack modifier and damage for RE-based NPC strikes

stwlam:


CarlosFdez (approved):

CarlosFdez:
I think we can assume the damage is fixed if the attack is fixed and its an npc, but if that's not a safe assumption free to merge as is.

PR #8444, opened 2023-06-26
Fix counting of maximum background boosts in ability builder

stwlam:
plus kill the jquery


CarlosFdez:
How well does this handle Magical Experiment? Does it do locked in boosts fine?

stwlam:
It fixes magical experiment

CarlosFdez:
So it visually displays the boost? Hell yeah

PR #8450, opened 2023-06-26
Add Exploit Vulnerability effect

reyzor1991:
Need to check Thaumaturge reactions


stwlam:
We need the effect to do something, and there currently isn't sufficient system support for it.

PR #8453, opened 2023-06-26
Add a `RecordField` that handles `Record<string, DataField>` data

In3luki:
Includes a sample implementation in the battle form rule element.

Validation error:

validation-error

`keys` option set to `["test"]`:

key-error

### DIFF
--- a/src/module/rules/rule-element/battle-form/rule-element.ts
+++ b/src/module/rules/rule-element/battle-form/rule-element.ts
@@ -1,7 +1,7 @@
 import { CharacterPF2e } from "@actor";
 import { CharacterStrike } from "@actor/character/data.ts";
 import { CharacterSkill } from "@actor/character/types.ts";
-import { SENSE_TYPES } from "@actor/creature/sense.ts";
+import { SENSE_ACUITIES, SENSE_TYPES } from "@actor/creature/sense.ts";
 import { ActorType } from "@actor/data/index.ts";
 import { ActorInitiative } from "@actor/initiative.ts";
 import { DiceModifierPF2e, ModifierPF2e, StatisticModifier } from "@actor/modifiers.ts";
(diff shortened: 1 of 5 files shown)


stwlam:
We'd want this to extend `ObjectField` and validate the full index signature (keys of `StringField` or `NumberField` and values of whatever).

src/module/rules/rule-element/battle-form/rule-element.ts
```diff
@@ -57,7 +58,19 @@ class BattleFormRuleElement extends RuleElementPF2e<BattleFormRuleSchema> {
                         { required: false, initial: undefined }
                     ),
                     tempHP: new ResolvableValueField({ required: false, nullable: true, initial: null }),
-                    senses: new fields.ObjectField({ required: false, initial: undefined }),
+                    senses: new RecordField(
+                        new fields.StringField({ required: true, blank: false }),
```
stwlam, line 62: okok but then give this a restricted set of choices: the resulting object should then have an index signature with optional keys

In3luki: So no error on choices validation and extract the choices keys as optional keys for the resulting object?

In3luki: I think I've got it working now. If the `StringField` props are a specific type those keys are in the resulting object index in addition to allowing any `string` keys.

src/module/rules/rule-element/battle-form/schema.ts
```diff
@@ -27,7 +35,13 @@ type BattleFormRuleOverrideSchema = {
         false
     >;
     tempHP: ResolvableValueField<false, true, true>;
-    senses: ObjectField<BattleFormSenses, BattleFormSenses, false, false, false>;
+    senses: RecordField<
+        StringField<string, string, true, false, false>,
```
stwlam, line 39: This should then be the set of of the permitted senses rather than a mere `string`

In3luki: Done. A normal `Record` in `RecordFieldModelProp` was enough after all.

In3luki: ok it had to be a partial `Record` after all.

stwlam: Right, these will globally be `Record<string, something | undefined>`

PR #8488, opened 2023-06-29
Add metagame setting to restrict party ally visibility

CarlosFdez:
I'll address the misalignment on the inventory boxes once exploration is in. The core styling is getting some tweaks and shuffling.

image


stwlam:
I expect people will want finer control than that, but can we get the reduced denounce? :)

CarlosFdez:
Already in there, its in the party document.

For now I'm happy to do "my gm" and "your gm" as the two possible settings and then see what people actually want. The one area I think there might be disagreement is on showing the HP. I can imagine people wanting to at least show HP for all actors in the exploration tab but otherwise hiding the overview.

PR #8503, opened 2023-06-30
Add schemas to `BaseSpeedRuleElement` and `CreatureSizeRuleElement`

In3luki:


src/module/rules/rule-element/base-speed.ts
```diff
@@ -4,28 +4,31 @@ import { MovementType } from "@actor/types.ts";
...
     protected static override validActorTypes: ActorType[] = ["character", "familiar", "npc"];
 
-    private selector: string;
-
-    private value: number | string | BracketedValue = 0;
+    static override defineSchema(): BaseSpeedRuleSchema {
+        const { fields } = foundry.data;
+        return {
+            ...super.defineSchema(),
+            selector: new fields.StringField({ required: true, nullable: false }),
```
stwlam, line 21:
```suggestion
            selector: new fields.StringField({ required: true, nullable: false, empty: false }),
```

PR #8504, opened 2023-06-30
Add schemas to `DexterityModifierCapRuleElement` and `FastHealingRuleElement`

In3luki:


CarlosFdez:
fast healing might be one of those that requires joint validation. details and deactivated by are for fast healing and regeneration respectively.

PR #8507, opened 2023-06-30
Limit item retrieval for inline roll links to spells

### LINKED ISSUE #8502: Stunning Fist is calculating Incapacitation Wrong.
Just tested on Version 5.0.2. No matter what level I put an enemy on, every saving throw against stunning fist is being upgraded by Incapacitation.

stwlam:
Closes #8502


CarlosFdez:
It would be more correct to resolve the item to the feat somehow (statistic is able to identify that a feat does not contribute to effect level....) but we can cross that bridge once rune incapacitation is implemented anyways.

PR #8520, opened 2023-07-01
Give template-preview event handlers better names

stwlam:


CarlosFdez:
Is this.document working within those methods?

PR #8525, opened 2023-07-01
Prefer highest value of same type when deduping modifiers

CarlosFdez:
Allows implementing kingmaker vacancy penalties, but also adhoc "custom modifier types" without adding to the global modifier type list.


stwlam:
Is this not also done in the constructor?

CarlosFdez:
You are correct, I certainly missed a spot. Give me a bit.

PR #8526, opened 2023-07-01
Include actor roll data in item roll data

CarlosFdez:
We could allow `@details` without deprecation warning if you think that'd be better.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1618,7 +1618,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
     /** This allows @actor.level and such to work for macros and inline rolls */
     override getRollData(): NonNullable<EnrichHTMLOptionsPF2e["rollData"]> {
         const rollData = { actor: this };
-        for (const prop of ["details", "attributes", "skills", "saves"] as const) {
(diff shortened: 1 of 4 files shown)


src/module/actor/base.ts
```diff
@@ -1615,9 +1615,22 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         return Array.from(toReturn).sort();
     }
 
-    /** This allows @actor.level and such to work for roll macros */
-    override getRollData(): Record<string, unknown> {
-        return { ...duplicate(super.getRollData()), actor: this };
+    /** This allows @actor.level and such to work for macros and inline rolls */
+    override getRollData(): NonNullable<EnrichHTMLOptionsPF2e["rollData"]> {
+        const rollData = { actor: this };
+        for (const prop of ["details", "attributes", "skills", "saves"] as const) {
```
stwlam, line 1621:
```suggestion
        for (const prop of ["abilities", "attributes", "details", "skills", "saves"] as const) {
```

stwlam:
What's special about `@details` among the rest?

CarlosFdez:
it seems to come up often. I see it in the wild, so I'm assuming foundry discord likes to share stuff using it fsr.

PR #8527, opened 2023-07-01
Add getter placeholder for active party

CarlosFdez:
This allows game.actors.party to retrieve the active party.

### DIFF
--- a/src/module/collection/actors.ts
+++ b/src/module/collection/actors.ts
@@ -1,7 +1,13 @@
-import { ActorPF2e } from "@actor";
+import { ActorPF2e, PartyPF2e } from "@actor";
 import { ActorUpdateContext } from "@actor/base.ts";
 
 export class ActorsPF2e<TActor extends ActorPF2e<null>> extends Actors<TActor> {
+    /** The world's active party, if one exists */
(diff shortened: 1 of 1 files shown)


src/module/collection/actors.ts
```diff
@@ -1,7 +1,13 @@
-import { ActorPF2e } from "@actor";
+import { ActorPF2e, PartyPF2e } from "@actor";
 import { ActorUpdateContext } from "@actor/base.ts";
 
 export class ActorsPF2e<TActor extends ActorPF2e<null>> extends Actors<TActor> {
+    /** The world's active party, if one exists */
+    get party(): PartyPF2e<null> | null {
+        // Eventually we'll need to implement active parties, but this is stubbed for development of other features
+        return (this as Actors<ActorPF2e<null>>).find<PartyPF2e<null>>((a) => a.isOfType("party")) ?? null;
+    }
```
stwlam, line 9: Why an actor instance getter that doesn't check whether it's the party the actor is in?

CarlosFdez: Its not an actor instance getter, its a game.actors instance getter. As per the description.

PR #8551, opened 2023-07-03
fix use of old manifests to install old pf2e versions

Drental:
In rare cases I have reason to not update Foundry to the latest version, this allows me to install and migrate the system with just a manifest link in those cases.


stwlam:
This appears to be copied from a module somewhere. Where is the working example?

Drental:
Absolutely correct 

https://github.com/Drental/fvtt-tokenactionhud

Drental:
*nudge*

stwlam:
The URL is still wrong, and the github action used is deprecated. I think there's a simpler solution we can use.

PR #8554, opened 2023-07-03
Create kingmaker features compendium

CarlosFdez:
Hidden from production builds


CarlosFdez:
I ran these through the data entry script, with a minor tweak to handle "basic kingdom checks" since those are just kingmaker checks against a control dc. Let me know if additional changes are required like converting that unicode - to an actual -.

stwlam:
These look to all be (proper) actions. Calling them features is confusing.

CarlosFdez:
The compendium will also include kingdom feats, and I'm assuming based on @silvative's branch, army actions?

CarlosFdez:
I need to figure out why extract packs crashes when there are folders at some point, but stubbed out kingdom features and feats into new folders. Cora's army branch as an army folder going into kingmaker-features as well, so that can probably be rebased onto this one.

CarlosFdez:
This should be ready now. Its the new campaignFeature type and has more items in there now. Still hidden in production builds.

PR #8557, opened 2023-07-04
Allow setting active exploration activities + show in party sheet

CarlosFdez:
Rest will need to display a new dialog to possibly advance the clock in future revisions.

image

image


CarlosFdez:
I am highly considering changing exploration: string[] to a list of { id: string } to allow more future flexibility such as bonus exploration actions and exploration actions that are always active, but I'd like to know if others agree before I actually do this.

stwlam:
If not the object, what is the plan for multiple activities?

CarlosFdez:
Without the object it'd amount to a legendary sneak ranger saying they are avoiding notice, covering tracks, and making a sandwich. Still useful, but I'm wondering if there are any avenues to improve this if we make it an array of objects instead.

PR #8563, opened 2023-07-05
Remove pre-caching attempt at load reduction when calling `itemTypes`

stwlam:


CarlosFdez:
What do you mean by reducing load? Seems like you're still calling it in the method, and the _itemTypes cache still exists.

stwlam:
Better?

PR #8564, opened 2023-07-05
Add Item Alteration rule element

stwlam:


src/module/item/helpers.ts
```diff
@@ -0,0 +1,30 @@
+import { ActorPF2e } from "@actor";
+import { setHasElement } from "@util";
+import { ItemPF2e } from "./base.ts";
+import { ItemSourcePF2e, ItemType } from "./data/index.ts";
+import { PhysicalItemPF2e } from "./physical/document.ts";
+import { PHYSICAL_ITEM_TYPES } from "./physical/values.ts";
+import { ItemInstances } from "./types.ts";
+
+/** Determine in a type-safe way whether an object */
```
CarlosFdez, line 9: ???

src/module/item/physical/document.ts
```diff
@@ -575,6 +577,33 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
+        type MaybeItemAlteration = { key: string; property?: unknown; itemType?: unknown };
+        const hasHpChangeRules =
+            this.actor?.rules.some(
+                (r: MaybeItemAlteration) =>
+                    r.key === "ItemAlteration" && r.property === "hp-max" && r.itemType === this.type
+            ) ?? false;
+        if (hasHpChangeRules && changed.system?.equipped?.carryType) {
+            // Simulate the update to determine whether max hit points will change
+            const postUpdateHPMax = ((): number => {
+                const actorSource = this.actor!.toObject();
```
CarlosFdez, line 591: I think we can avoid the ! if actor is assigned to a variable first. Actor returning form a getter seems to be the reason TS doesn't understand that hasHpChangeRules also validates that there's an actor.

src/module/item/physical/document.ts
```diff
@@ -575,6 +577,33 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
+                const itemSource = actorSource.items.find((i) => i._id === this.id) ?? { system: {} };
+                itemSource.system = mergeObject(itemSource.system, deepClone(changed.system));
+                const actorClone = new ActorProxyPF2e(actorSource);
+                const itemClone = actorClone.items.get(this.id, { strict: true });
+                return itemClone.isOfType("armor") ? itemClone.system.hp.max : this.system.hp.max;
+            })();
+            if (postUpdateHPMax !== this.system.hp.max) {
```
CarlosFdez, line 598:
```suggestion

            // Apply the value change, and avoid "broken" changing by rounding directionally
            if (postUpdateHPMax !== this.system.hp.max) {
```

Wasn't obvious to me that this is trying to make sure that broken is the same with the floorOrCeil ternary, and I only realized that because I remember you brought it up in the discord. The if statement might benefit from a comment explaining the intent here.

src/module/item/helpers.ts
CarlosFdez:
```suggestion
/** Check if the item or source matches the type (or whether it's one among multiple types) */
```

src/module/item/physical/document.ts
stwlam: image

CarlosFdez: silly TS

PR #8565, opened 2023-07-05
Additional work on party sheet

CarlosFdez:
* Adds ancestry, class, size, and speeds to overview
* Adds padding around inventory to line up with other elements
* Centers PC name label when wealth is missing (such as when viewing in a restricted mode)
* Add HP bar to overview
* Tweak styling of PC languages
* Increase default sheet height to fit 4 PCs
* Remove Lore from best skills. They'll come back later in an RK aggregate alternative view.

Adapted from Mats design

image
image
image


CarlosFdez:
Most of this has already split off into other PRs or has changed in design to such a point where this PR is no longer necessary.

PR #8585, opened 2023-07-06
Remove kingmaker traits from action and add more

CarlosFdez:
These will become accessible in the new item type we're going to need to work on.


static/lang/en.json
```diff
@@ -3483,6 +3484,7 @@
         "TraitCapacity4": "Capacity 4",
         "TraitCapacity5": "Capacity 5",
         "TraitCarried": "Carried",
+        "TraitCavalry": "Cavalry",
```
stwlam, line 3487: So close!

CarlosFdez: I am so bad with alphabetical order you have no idea, I have to sing the childhood song each and every time.

PR #8596, opened 2023-07-07
Add elemental ammunition variants

JDCalvert:
Add variants for each of the damage types that elemental ammunition can deal, including rule elements which will pass to the firing weapon.


JDCalvert:
Closing as this kind of change is waiting for system changes to ammunition.

PR #8603, opened 2023-07-08
Fix whitespace and alignment issues in party inventory

CarlosFdez:
This primarily affects the restricted view

image
image


stwlam:
What's the `--border` variable?

CarlosFdez:
A color css variable for just the party sheet. I was considering calling it border-color but the convention seems to be that we don't append color to color variables.

PR #8610, opened 2023-07-08
Add pack's source package after its name in brackets on the Compendium Browser's Settings tab

Cerapter:
Screenshot_20230708_181501

Because of the new ability to add folders for compendia, and to give each compendium a distinct banner, peeps have started to name their compendia simpler and more to-the-point.

This PR just makes it so that the origin package where the pack comes from is added  to the end of the checkbox label on the Settings tab, so you know which of the several dozen "Feats" you're actually toggling on.

### DIFF
--- a/src/module/apps/compendium-browser/data.ts
+++ b/src/module/apps/compendium-browser/data.ts
@@ -3,6 +3,7 @@ import * as browserTabs from "./tabs/index.ts";
 interface PackInfo {
     load: boolean;
     name: string;
+    package: string;
 }
 
 interface SourceInfo {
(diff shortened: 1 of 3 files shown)


stwlam:
How does it look with the package name instead of the id?

TikaelSol:
I was going to circle back to fix those compendiums in the modules, but this works much better. Having the full module name in the compendium name looks terrible since they are under folders with that name on them.

Cerapter:
kép

Full module names are overly verbose, so it's probably best that the package names are kept.
A CSS wizard should be able to do something with this though, I think. I'm not one, however.

PR #8616, opened 2023-07-08
Add schema fields to `StrikingRuleElement`, `SubstituteRollRuleElement` and `TempHPRuleElement`

In3luki:
Includes a migration for `TempHPRuleElement`

Tested with:
`StrikingRuleElement` -> `Spell Effect: Magic Stone`,
`SubstituteRollRuleElement` -> `Assurance`
`TempHPRuleElement` -> `Effect: Numbing Tonic (Greater)` and `Effect: Belt of the Five Kings (Wearer)`


src/module/rules/rule-element/substitute-roll.ts
```diff
@@ -1,70 +1,54 @@
...
+class SubstituteRollRuleElement extends RuleElementPF2e<SubstituteRollSchema> {
+    static override defineSchema(): SubstituteRollSchema {
+        const { fields } = foundry.data;
+        return {
+            ...super.defineSchema(),
+            selector: new fields.StringField({ required: true, blank: false, initial: "check" }),
+            value: new ResolvableValueField({ required: true, nullable: false }),
+            required: new fields.BooleanField({ required: false, nullable: false, initial: false }),
+            effectType: new fields.StringField({
+                required: false,
```
stwlam, line 17:
```suggestion
                required: true,
```

src/module/rules/rule-element/temp-hp.ts
```diff
@@ -1,29 +1,56 @@
...
-        this.data.onCreate = !!(this.data.onCreate ?? true);
-        /** Whether the temporary hit points renew each round */
-        this.data.onTurnStart = !!this.data.onTurnStart;
+        // Set default values if events was omitted
+        if (!this.events) {
+            this.events = {
+                onCreate: true,
+                onTurnStart: false,
+            };
+        }
```
stwlam, line 46: Handle all this from the schema. If it doesn't validate until the migration, then it'll fail, and that's fine.

In3luki: I had some problems getting the schema field to behave like I wanted. I had it set up with default values at one point but setting the `events` value to a string will not trigger a validation error because the field just converts that to the default object.

src/module/rules/rule-element/substitute-roll.ts
In3luki: There are many existing REs that have omitted the `effectType` because it defaults to `fortune`.

stwlam: Yeah, but it has an `initial` just below, so it's ok!

stwlam:
ok to push up migration result

In3luki:
That were way less than I expected

PR #8639, opened 2023-07-10
Fix refresh of death icon on setting change

CarlosFdez:
....also fixes a regression


CarlosFdez:
A more involved fix would have us refactor by merging migrateStatusEffectUrls and detect what is changing or not. But the more involved the riskier.

PR #8645, opened 2023-07-10
Add `MeasuredTemplateDocumentPF2e#getTokens` to find all tokens inside a measured template

In3luki:
I'm not sure if this should live on the document or the object. Also the elevation calculation is probably wrong.

Recording 2023-07-10 at 22 29 55

### DIFF
--- a/src/module/canvas/helpers.ts
+++ b/src/module/canvas/helpers.ts
@@ -14,10 +14,15 @@ function measureDistanceCuboid(
         reach = null,
         token = null,
         target = null,
+        source,
     }: {
         reach?: number | null;
         token?: TokenPF2e | null;
         target?: TokenPF2e | null;
+        source?: {
+            elevation: number;
+            height: number;
+        };
     } = {}
 ): number {
     if (!canvas.dimensions) return NaN;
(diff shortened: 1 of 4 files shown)


CarlosFdez:
I think the concept is clean enough, though I'm wondering what the goal is.

stwlam:
Automatic targeting workflow

In3luki:
The immediate goal would be to make it possible to automatically target all tokens in a spell template. Combined with #8673 the template creation can be awaited and the conatined tokens can be retrieved for further filtering based on predidactes, elevation, etc.

In3luki:
Exclusion based on distance now works for all common template shapes.

stwlam:
What happens if a token is at the cone origin but elevated five feet?

In3luki:
> What happens if a token is at the cone origin but elevated five feet?

image

In this situation the smaller token is excluded as outside the cone. The top left square of the bigger token is outside the cone but the top right is inside so it is included.
The cone starts at elevation 0. Maybe half grid size would be better?

In3luki:
Cone calculations are inaccurate

PR #8673, opened 2023-07-12
Have `MeasuredTemplatePF2e#drawPreview` return the placed template after it was drawn

In3luki:
Resolving the Promise directly after the new template was created can lead to erros as, for instance, the `GridHighlight` of a newly created template is only available after `MeasuredTemplate#applyRenderFlags()` was called.


CarlosFdez:
Thoughts on making the resolve function a function that wraps the actual resolve function and clears the variable? Then you can do `MeasuredTemplatePF2e.currentPreview?.resolve(null);` everywhere instead.

In3luki:
Updated. Wrapping the resolve is much cleaner.

CarlosFdez (approved):

PR #8679, opened 2023-07-12
Update `Document#_pre(Create|Update|Delete)` return types for V11

stwlam:
They can now return `Promise<boolean | void>`.


CarlosFdez:
HELL YEAH

src/module/actor/vehicle/document.ts
```diff
@@ -188,6 +190,8 @@ class VehiclePF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e |
                 await TokenDocumentPF2e.updateDocuments(updates, { parent: canvas.scene });
             }
         }
+
+        return result;
```
CarlosFdez, line 194: going by the others, I think this might not be necessary?

PR #8682, opened 2023-07-12
Remove `RuleElementPF2e#data`

stwlam:


CarlosFdez:
I believe this is currently in use for some of the forms. flat-modifier.hbs refers to object.data.label for example.

PR #8690, opened 2023-07-13
Create campaign feature item for kingmaker

CarlosFdez:
This adds the new item type as was suggested in the discord.

To be honest, with a single new property, this looks general enough that I could probably call it campaignFeature, so that if we use it again in the future we can save ourselves the naming headache. But even if we don't, I don't know if "kingmaker" is a good name for the item type.


src/module/item/kingmaker-feature/values.ts
```diff
@@ -0,0 +1,17 @@
+import * as R from "remeda";
+import { BehaviorType, KingmakerCategory } from "./types.ts";
+
+const KINGDOM_CATEGORY_DATA = {
+    kingdomFeat: { behavior: "feat" },
+    kingdomFeature: { behavior: "feature" },
+    kingdomActivity: { behavior: "activity" },
```
stwlam, line 7: These should be hyphenated since their main use is as values

static/template.json
```diff
@@ -1250,6 +1251,24 @@
...
             "weapons": [],
             "spells": {}
+        },
+        "kingmaker": {
+            "templates": [
+                "common",
+                "traits",
+                "rarity"
+            ],
+            "category": "kingdomActivity",
```
stwlam, line 1261: hyphenx plx

static/templates/sidebar/item-document-partial.hbs
```diff
@@ -11,7 +11,7 @@
                 <span class="item-level">{{localize "PF2E.Item.Spell.LevelN" level=this.level}}</span>
             {{else if (eq this.type "effect")}}
                 <span class="item-level">{{localize "PF2E.LevelN" level=this.level}}</span>
-            {{else}}
+            {{else if this.level includeZero=true}}
```
stwlam, line 14: obscene

CarlosFdez: apparently null is greater than or equal to 0 :(

stwlam: there's an `isNumber` helper

src/module/item/kingmaker-feature/values.ts
CarlosFdez: Interestingly enough, feat categories, which this kinda mimics, are lowercase with no hyphens, but I can do hyphens.

static/templates/sidebar/item-document-partial.hbs
CarlosFdez: I should probably replace the main if with this, since greater than or equal to zero matching null is problematic in the first place, and negative levels aren't a thing we do to remove them.

src/module/item/kingmaker-feature/values.ts
stwlam: That's not interesting. They're all single words, so no hyphens would appear.

static/templates/sidebar/item-document-partial.hbs
CarlosFdez: includeZero is part of handlebars, and is used when you want an if statement to check that something has a value rather than just if its falsey. Regardless, main if.

src/module/item/kingmaker-feature/values.ts
CarlosFdez: Are you sure?
image

static/templates/sidebar/item-document-partial.hbs
```diff
@@ -4,7 +4,7 @@
     {{/if}}
     <h4 class="entry-name document-name">
         <a>{{this.name}}</a>
-        {{#if (gte this.level 0)}}
+        {{#if this.level includeZero=true}}
```
stwlam, line 7: still obscene

stwlam, line 7:
```suggestion
        {{#if (isNumber this.level)}}
```

PR #8700, opened 2023-07-14
Collect note and DoS adjustment synthetics for recovery checks

stwlam:
(dying and persistent damage)


CarlosFdez:
Do specified domains not do anything at all for flat checks period? if so, trying to specify them should error, or better yet it should error then set the domains.

PR #8701, opened 2023-07-14
Double range increment of ranged weapons with Impossible rune

stwlam:


CarlosFdez:
I don't know if these boosts are additive or multiplicative, but I'm ok with assuming additive.

PR #8706, opened 2023-07-14
Fix option acquisition and roll inspector for flat checks

CarlosFdez:
🙃 

image

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -634,6 +634,7 @@
         "CharacterTraitsHeader": "Traits & Other Details",
         "ChatRollDetails": {
             "ModifiersEmpty": "No Modifiers",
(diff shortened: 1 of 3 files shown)


static/lang/en.json
```diff
@@ -634,6 +634,7 @@
         "CharacterTraitsHeader": "Traits & Other Details",
         "ChatRollDetails": {
             "ModifiersEmpty": "No Modifiers",
+            "FlatCheckNoModifiers": "Flat Checks never includes any modifiers, bonuses, or penalties",
```
stwlam, line 637:
```suggestion
            "FlatCheckNoModifiers": "Flat checks never include any modifiers, bonuses, or penalties.",
```

PR #8707, opened 2023-07-14
Add isDestroyed check to the raiseAShield action

reonZ:
While the `raiseAShield` action prevents the use of the shield when it is `broken`, it doesn't prevent it when it is `destroyed`, so i just added the condition and warning into the action.

The `isBroken` getter specifically avoid returning `true` when the shield is destroyed.

### DIFF
--- a/src/scripts/macros/raise-a-shield.ts
+++ b/src/scripts/macros/raise-a-shield.ts
@@ -34,7 +34,10 @@ export async function raiseAShield(options: ActionDefaultOptions): Promise<void>
             return false;
         }
 
-        if (shield?.isBroken === false) {
+        if (shield?.isDestroyed) {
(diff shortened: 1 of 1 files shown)


src/scripts/macros/raise-a-shield.ts
```diff
@@ -34,7 +34,10 @@ export async function raiseAShield(options: ActionDefaultOptions): Promise<void>
             return false;
         }
 
-        if (shield?.isBroken === false) {
+        if (shield?.isDestroyed) {
+            ui.notifications.warn(localize("ShieldIsDestroyed", { actor: speaker.alias, shield: shield.name }));
```
stwlam, line 38: we don't have that localization entry

reonZ: image

stwlam: Ah, sorry, I forgot it was moved to a different file.

PR #8715, opened 2023-07-14
Add support for targeted inline checks against specified defenses

stwlam:
image


src/module/system/statistic/index.ts
```diff
@@ -375,7 +375,7 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
             const isTargetedCheck =
                 (this.type === "spell-attack-roll" && item?.isOfType("spell")) ||
                 (["perception-check", "skill-check"].includes(this.type) &&
-                    !!args.dc?.slug &&
+                    !!(args.dc?.statistic ?? args.dc?.slug) &&
```
CarlosFdez, line 378: that ?? should probably be ||, given this is boolean. I was so confused for a while here what your goal was.

stwlam: nah

PR #8767, opened 2023-07-17
Enable item option acquisition in item summaries

CarlosFdez:
Attempts to tweak it to a get document and a get actor phase, making it more predictable and also work better with items as parameters.

### DIFF
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -14,7 +14,7 @@ import { getSelectedOrOwnActors } from "@util/token-actor-utils.ts";
 const inlineSelector = ["action", "check", "effect-area"].map((keyword) => `[data-pf2-${keyword}]`).join(",");
 
 export const InlineRollLinks = {
-    injectRepostElement: (links: HTMLElement[], foundryDoc?: ClientDocument): void => {
+    injectRepostElement: (links: HTMLElement[], foundryDoc: ClientDocument | null): void => {
         for (const link of links) {
             if (!foundryDoc || foundryDoc.isOwner) link.classList.add("with-repost");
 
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I kinda realized as I was putting this up that maybe now that Kineticist is getting released and we don't have time for actions that act, maybe we outta consider opening the item retrieval to more things. Was there something we *didn't* want it to grab? I believe it was just weapon roll notes.

src/scripts/ui/inline-roll-links.ts
```diff
@@ -256,13 +257,9 @@ export const InlineRollLinks = {
...
 
-        const speaker =
-            foundryDoc instanceof ActorPF2e
-                ? ChatMessagePF2e.getSpeaker({
-                      actor: foundryDoc,
-                      token: foundryDoc.token ?? foundryDoc.getActiveTokens(false, true).shift(),
-                  })
-                : ChatMessagePF2e.getSpeaker();
+        const speaker = actor
+            ? ChatMessagePF2e.getSpeaker({ actor, token: actor.token ?? actor.getActiveTokens(false, true).shift() })
```
stwlam, line 261:
```suggestion
            ? ChatMessagePF2e.getSpeaker({ actor, token: actor.getActiveTokens(false, true).shift() })
```

src/scripts/ui/inline-roll-links.ts
```diff
@@ -289,23 +286,22 @@ export const InlineRollLinks = {
...
-        const sheet: { id?: string; document?: unknown; actor?: unknown; journalEntry?: unknown } | null =
-            ui.windows[Number(html.closest<HTMLElement>(".app.sheet")?.dataset.appid)];
+    const sheet: { id?: string; document?: unknown } | null =
+        ui.windows[Number(html.closest<HTMLElement>(".app.sheet")?.dataset.appid)] ?? null;
 
-        return sheet.document instanceof ActorPF2e || sheet.document instanceof JournalEntry ? sheet.document : null;
-    },
-};
+    const document = sheet?.document;
+    return document && (document instanceof ActorPF2e || document instanceof JournalEntry) ? document : null;
```
stwlam, line 299:
```suggestion
    return document instanceof ActorPF2e || document instanceof JournalEntry ? document : null;
```

CarlosFdez: I should have scrutinized the original logic more closely than I did whoopsie

PR #8799, opened 2023-07-19
Ensure `CoinsPF2e` objects have positive integer values

stwlam:
Also update add/remove coins popups to enforce the same


CarlosFdez:
I can understand enforcing it at the dialog level, but currently that's the only way to subtract coins using that utility (and it might even be a good refactor to convert removeCoins into doing coin math then updating the actor's state to match). I'll have to double check if ItemPiles uses it soon.

PR #8818, opened 2023-07-20
Don't conflate temporary physical items with in-memory-only ones

stwlam:


src/module/item/physical/document.ts
```diff
@@ -99,10 +99,6 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         return this.traits.has("cursed");
     }
 
-    override get isTemporary(): boolean {
-        return this.system.temporary;
-    }
-
```
CarlosFdez, line 105: There are templates that use this, and I assume some modules do to. Lets keep this around but not have it be an override.

PR #8834, opened 2023-07-20
Implement @Damage inline rolls, which support modifiers and damage dice 

CarlosFdez:


types/foundry/client/core/text-editor.d.ts
```diff
@@ -140,7 +140,7 @@ declare global {
          * Handle left-mouse clicks on an inline roll, dispatching the formula or displaying the tooltip
          * @param event The initiating click event
          */
-        static _onClickInlineRoll(event: MouseEvent): Promise<ChatMessage>;
```
stwlam, line 143: scanning the source, it looks like this should be `Promise<ChatMessage | void>`

CarlosFdez: In terms of what it does maybe, but given this is just a listener the return type doesnt actually matter. Foundry only returns chat message so thst its async i imagine.

stwlam: It's async merely by having an `async` modifier (which isn't seen in the type)

src/module/system/damage/helpers.ts
```diff
@@ -34,6 +44,99 @@ const DamageCategorization = {
     },
 } as const;
 
+const faces = [4, 6, 8, 10, 12];
```
stwlam, line 47: All caps? `faces` is shadowed immediately below, which is confusing.

CarlosFdez: Good catch, I hadn't noticed that.

PR #8843, opened 2023-07-21
Add types for `foundry.abstract.DatabaseBackend` and `ClientDatabaseBackend`

In3luki:


types/foundry/common/abstract/backend.d.ts
```diff
@@ -0,0 +1,238 @@
...
+        request: DatabaseBackendGetContext,
+        user: User
+    ): Promise<(DeepPartial<TDocument["_source"][]> & CompendiumIndexData) | TDocument[]>;
+
+    /**
+     * Get the parent Document (if any) associated with a request context.
+     * @param context    The requested operation
+     * @return           The parent Document, or null
+     */
+    private _getParent(
```
stwlam, line 43: ?

In3luki:
```js
  /**
   * Get the parent Document (if any) associated with a request context.
   * @param {object} context                The requested operation
   * @return {Promise<Document|null>}       The parent Document, or null
   * @private
   */
  async _getParent(context) {
    if ( !context.parent ) return null;
    if ( !(context.parent instanceof Document) ) {
      throw new Error("A parent Document provided to the database operation must be a Document instance");
    }
    return context.parent;
  }
  ```

stwlam: If it's truly private, it doesn't need to be there since it's not part of the actual interface.

In3luki: Ok, I wasn't sure what to do with the `@private` methods. I'll remove them when I'm home.

In3luki: They are gone

PR #8844, opened 2023-07-21
Automates cover

reonZ:
This is a base implementation of cover automation, should it be added to the system and if so, should it be implemented like that ?

- how to handle the creature cover for diagonals ? right now i am forcing a minimum of 5 ft. between the attacker and the target
- there should probably be an easy way to ignore automated covers during a specific attack
- there need to be checks for feats/features that work with cover in pf2e


stwlam:
I think we need to pass on wall cover until walls can be configured for whether and how much cover they provide.

PR #8879, opened 2023-07-23
Add 'prepare the processional' downtime activity from Ghoul's Hunger

TMunsie:


stwlam:
Can this be put in adventure-specific actions?

TMunsie:
Yeah. Going to do it via a new PR to keep it clean. Should have been there in the first place. I got lost with the pack re-org.

PR #8907, opened 2023-07-24
Add separate `CharacterSystemSource` type

stwlam:
As part of the split from `CharacterSystemData`, refactor some runtime code (mostly old migrations) and prune some unneeded properties from template.json.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -83,7 +83,6 @@ import {
     CharacterAttributes,
     CharacterFlags,
     CharacterProficiency,
-    CharacterSaves,
     CharacterSource,
     CharacterStrike,
     CharacterSystemData,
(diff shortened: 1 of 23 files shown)


src/module/actor/character/document.ts
```diff
@@ -315,10 +320,11 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
             },
         };
 
-        // Base ability scores
-        for (const abbrev of ABILITY_ABBREVIATIONS) {
-            systemData.abilities[abbrev] = mergeObject({ value: 10 }, systemData.abilities[abbrev] ?? {});
-        }
+        // Base saves structure
+        systemData.saves = mergeObject(
```
CarlosFdez, line 324: Seems saves are set up lower down in 376. Would it not make sense to move some of that up here so that we're not double reassigning base save objects?

src/module/actor/character/document.ts
```diff
@@ -361,10 +367,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         attributes.ancestryhp = 0;
         attributes.classhp = 0;
 
-        // Familiar abilities
-        attributes.familiarAbilities = { value: 0 };
-
-        // Saves and skills
+        // Skills
```
CarlosFdez, line 370: Nuh uh

PR #8913, opened 2023-07-24
Brushup rule elements on Acrobat dedication & Tumble Behind

ChasarooniZ:


packs/actions/confident-finisher.json
```diff
@@ -16,7 +16,40 @@
         "requirements": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
+            {
+              "key": "Note",
+              "selector": "attack",
+              "text": "You deal half your [[/r (2 + floor((@actor.level - 1)/4))d6/2]]{Precise Strike} damage to the target. This damage type is that of the weapon or unarmed attack you used for the Strike.",
```
TikaelSol, line 23: We should localize this, but also with the new suboptions I want to revisit how to handle the finishers I think. Confident finisher is only applied if you use that action, it doesn't apply on every finisher.

TikaelSol:
I changed the title to be in imperative mood. PR titles should be written to finish the sentence `If accepted this PR will <TITLE>`

ChasarooniZ:
Removed swashbuckler only acrobat + tumble behind

ChasarooniZ:
This now only affects 2 minor elements, and should have no issues with translation as it just takes the item text.

PR #8914, opened 2023-07-24
Automate advanced multiclass feats

JDCalvert:
(Resubmitted because of significant changes)

Contributes to completing #1956.

Automate advanced multiclass feats (e.g. Advanced Arcana, Advanced Maneuver etc) to allow a choice of feats from that class up to half your level (or rather, half the level of the feat slot being used).

### New Feat Roll Options
I've introduced two new roll option for feats: `feat:location:category` and `feat:location:level`. These take their values from the feat's `location` field. So, for example, a feat with location `class-6` would have the roll options:
- `feat:location:category:class`
- `feat:location:level:6`

### Predication Mathematical Operations
I've added the ability to use mathematical operations within the `BinaryOperation` predicates. The available operations are `add` (addition), `sub` (subtraction), `mult` (multiplication), and `div` (division). They work similarly to the existing `BinaryOperation` elements, taking the form of an object which describes the operation to take place, and a two-length array where the first element is the "left" value and the second element is the "right" value.

For example, the following predicate will return true if the value of `item:level` is less than or equal to half the value of `self:level`:
```json
{
    "lte": [
        "item:level",
        {
            "div": [
                "self:level",
                2
            ]
        }
    ]
}
```

The implementation is fairly simple at the moment (no nested maths operations, only the right operator can be a maths operation) but the way I've written it should allow it to be easily extendable.

#### Bonus Fix
In the binary operation code, if the `right` value was a number, then the `left` value would be compared against this same number once for every roll option in the `domains` array. I've moved this check to outside the `flatMap` operation, so the comparison will only happen once.

## Feat Updates
Using the features introduced above, each of the "advanced" multiclass feats has had rule elements applied to give a choice of the feats, exactly like the "basic" multiclass feats, but replacing the `2` in the max level with a maths operation dividing the feat's slot by 2.

For example, taking `Advanced Dogma` in the 6th-level class feat slot will allow a choice of all the 1st- and 2nd-level cleric class feats, whereas taking it in the 8th-level class feat slot will allow a choice of all the 1st-4th-level cleric class feats.

For the case of unlevelled feat slots (e.g. bonus feats) the character's level is used to determine the level of feat that can be taken.

### DIFF
--- a/packs/feats/advanced-arcana.json
+++ b/packs/feats/advanced-arcana.json
@@ -24,7 +24,35 @@
                 }
             ]
         },
-        "rules": [],
+        "rules": [
+            {
+                "adjustName": false,
+                "choices": {
+                    "filter": [
+                        {
+                            "lte": [
+                                "item:level",
+                                {
+                                    "div": [
+                                        "parent:location:level",
+                                        2
+                                    ]
+                                }
+                            ]
+                        },
+                        "item:category:class",
+                        "item:trait:wizard"
+                    ],
+                    "itemType": "feat"
+                },
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Prompt.AdvancedMulticlassFeat"
+            },
+            {
+                "key": "GrantItem",
+                "uuid": "{item|flags.pf2e.rulesSelections.advancedArcana}"
+            }
+        ],
         "source": {
             "value": "Pathfinder Core Rulebook"
         },
--- a/packs/feats/advanced-arcana.json
+++ b/packs/feats/advanced-arcana.json
@@ -40,6 +40,17 @@
                                 }
                             ]
                         },
+                        {
+                            "lte": [
+                                "item:level",
+                                {
+                                    "div": [
+                                        "self:level",
+                                        2
+                                    ]
+                                }
+                            ]
+                        },
                         "item:category:class",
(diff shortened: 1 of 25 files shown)


CarlosFdez:
I'll leave it to someone else to decide on more of the particulars, but why you not using the feat group instead of the location field?

JDCalvert:
> I'll leave it to someone else to decide on more of the particulars, but why you not using the feat group instead of the location field?

The `location:category` part will change based on which slot (if any) the spell is added to. So if you're using the Free Archetype variant and take the feat in a free archetype slot, `feat:location:category` will be `archetype` but `feat:category` will still be `class`.

`location:category` isn't strictly needed for this change. I added it for completeness since I was adding `feat:location:level`.

CarlosFdez:
`feat.group?.id`

JDCalvert:
> `feat.group?.id` and `feat.group?.slots[feat.system.location].level` (roughly) are better ways to get it than string splicing.

Ah, I didn't know about that. Yes that's better, I'll change it.

JDCalvert:
@CarlosFdez I tried it, but it looks like the group data doesn't exist at that point in the feat's creation, so I can't use it. I think that's all populated after the feat's initialisation, which includes resolving the rule elements.

I'll have to leave the regex for now.

PR #8932, opened 2023-07-25
Use action header in messages for some inline check vs defenses

stwlam:
The action header will be used if the check is against a defense and the originating item is an encounter action.

image


src/scripts/ui/inline-roll-links.ts
```diff
@@ -176,31 +176,55 @@ export const InlineRollLinks = {
                                 return null;
                             })();
 
-                            // Retrieve the item if it is not a weapon and this is a saving throw
+                            const isSavingThrow = tupleHasValue(SAVE_TYPES, pf2Check);
+
+                            // Retrieve the item if:
+                            // (2) The item is an action or,
+                            // (1) The check is a saving throw and the item is not a weapon.
```
CarlosFdez, line 183: I think we can always retrieve the item if its not a weapon. This was probably being too cautious.

stwlam: can we reassess after this PR?

PR #8946, opened 2023-07-26
Fix elite/weak adjustment for cantrips

CarlosFdez:
> Increase the damage of its Strikes and other offensive abilities by 2. If the creature has limits on how many times or how often it can use an ability (such as a spellcaster’s spells or a dragon’s Breath Weapon), increase the damage by 4 instead. 

While it says "spells" universally, the context is for limited use abilities.

### DIFF
--- a/src/module/item/spell/document.ts
+++ b/src/module/item/spell/document.ts
@@ -236,9 +236,10 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
                 }
             }
 
-            // Increase or decrease the first instance of damage by 4 if elite or weak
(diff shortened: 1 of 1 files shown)


src/module/item/spell/document.ts
```diff
@@ -236,9 +236,10 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
                 }
             }
 
-            // Increase or decrease the first instance of damage by 4 if elite or weak
+            // Increase or decrease the first instance of damage by 2/4 if elite or weak
```
CarlosFdez, line 239:
```suggestion
            // Increase or decrease the first instance of damage by 2 or 4 if elite or weak
```

PR #8956, opened 2023-07-26
Change compendiums to new inline damage syntax (Part 3)

TikaelSol:
Some more complex damage rolls are in this one. This should be the bulk of non-player facing content done. There's some stragglers with weirder formats that we can chase down later.


TikaelSol:
This will also conflict with part two, I'll fix the conflicts when that is merged.

PR #8961, opened 2023-07-27
Rename all uses/mentions of "flat-footed" to "off-guard"

stwlam:
Can't push up migration sample since the build script will throw on broken links


TikaelSol:
Right now I am doing new entry for RoE and swapping off guard to flat footed, this going to be an issue merging this in before the 3rd when we can put the actors in?

CarlosFdez (approved):

PR #8964, opened 2023-07-27
Parse @Damage with intermediate expressions + fallback in some cases

CarlosFdez:
There might be a better way to do the intermediate term stuff, I'm all ears.

This falls back to regular formula handling if there's an error or if there is an`@actor` term anywhere and the rolldata doesn't provide one (which would just render 0 damage otherwise).


CarlosFdez:
How would we even turn those into base damage objects? Those will fall back to the regular formula but really outta use [[/r]].

PR #8978, opened 2023-07-28
Add "Clear Placed Templates" button to area-effect spell cards

stwlam:
image

### DIFF
--- a/src/module/chat-message/helpers.ts
+++ b/src/module/chat-message/helpers.ts
@@ -202,8 +202,7 @@ function toggleClearTemplatesButton(message: ChatMessagePF2e | null): void {
         const clearTemplatesButton = htmlQuery(chatLogDOM, selector);
         if (!clearTemplatesButton) continue;
         const hasMeasuredTemplates = !!canvas.scene?.templates.some((t) => t.message === message && t.isOwner);
-        const method = hasMeasuredTemplates ? "remove" : "add";
-        clearTemplatesButton.classList[method]("hidden");
+        clearTemplatesButton.classList.toggle("hidden", !hasMeasuredTemplate);
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/module/chat-message/helpers.ts
```diff
@@ -190,4 +190,21 @@ function toggleOffShieldBlock(messageId: string): void {
...
+function toggleClearTemplatesButton(message: ChatMessagePF2e | null): void {
+    if (!message || !canvas.ready) return;
+
+    const selector = `li[data-message-id="${message.id}"] button[data-action=spell-template-clear]`;
+    for (const chatLogDOM of htmlQueryAll(document.body, "#chat-log, #chat-popout")) {
+        const clearTemplatesButton = htmlQuery(chatLogDOM, selector);
+        if (!clearTemplatesButton) continue;
+        const hasMeasuredTemplates = !!canvas.scene?.templates.some((t) => t.message === message && t.isOwner);
+        const method = hasMeasuredTemplates ? "remove" : "add";
+        clearTemplatesButton.classList[method]("hidden");
```
CarlosFdez, line 206:
```suggestion
        clearTemplatesButton.classList.toggle("hidden", !hasMeasuredTemplate);
```

Though there is also the `hidden` attribute you could set which is part of the web start, but unfortunately it loses in priority to all `display` css properties so you need to style it specifically if you go that route most of the time.

stwlam: The `hidden` class is from core (just has `display: none`).

src/module/chat-message/helpers.ts
```diff
@@ -190,4 +190,20 @@ function toggleOffShieldBlock(messageId: string): void {
...
+ */
+function toggleClearTemplatesButton(message: ChatMessagePF2e | null): void {
+    if (!message || !canvas.ready) return;
+
+    const selector = `li[data-message-id="${message.id}"] button[data-action=spell-template-clear]`;
+    for (const chatLogDOM of htmlQueryAll(document.body, "#chat-log, #chat-popout")) {
+        const clearTemplatesButton = htmlQuery(chatLogDOM, selector);
+        if (!clearTemplatesButton) continue;
+        const hasMeasuredTemplates = !!canvas.scene?.templates.some((t) => t.message === message && t.isOwner);
+        clearTemplatesButton.classList.toggle("hidden", !hasMeasuredTemplate);
```
CarlosFdez, line 205:
```suggestion
        clearTemplatesButton.classList.toggle("hidden", !hasMeasuredTemplates);
```

PR #8984, opened 2023-07-28
Remove penalty for amphibious and aquatic unarmed strikes underwater

TikaelSol:
They don't take those penalties on their unarmed strikes per their traits.


stwlam:
This would be limited to NPCs

TikaelSol:
I'm not sure that's the case actually. You don't gain the benefits of actor traits from your ancestry, like construct, but this isn't a benefit of the trait itself, it's a rules interaction.

stwlam:
I would agree except the benefit comes from the trait description itself

packs/other-effects/effect-aquatic-combat.json
```diff
@@ -32,6 +32,18 @@
...
+                    },
+                    {
+                        "nand": [
+                            {
+                                "or": [
+                                    "self:trait:amphibious",
+                                    "self:trait:aquatic"
+                                ]
+                            },
+                            "unarmed",
```
stwlam, line 44:
```suggestion
                            "item:category:unarmed",
```

PR #9055, opened 2023-08-01
Add Godless Healing Feat to Treat Wounds macro

reyzor1991:
Godless Healing

You recover an additional 5 Hit Points from a successful attempt to Treat your Wounds or use Battle Medicine on you. After you or an ally use Battle Medicine on you, you become temporarily immune to that Battle Medicine for only 1 hour, instead of 1 day.

Screenshot 2023-08-01 at 06 24 09


stwlam:
Godless Healing affects how much healing the one with the feat receives, not dispenses.

PR #9056, opened 2023-08-01
Fix more party sheet bugs

CarlosFdez:
Fixes RK button not working for players, party folder closing when members are updated, and kingmaker stale folder bugs.

Casting `this` is a crime, but without that typescript complains that the type is infinitely recursive.

### DIFF
--- a/static/templates/sidebar/party-document-partial.hbs
+++ b/static/templates/sidebar/party-document-partial.hbs
@@ -1,11 +1,11 @@
 <ol class="party-list">
-{{#each parties}}
-    <li class="party flexcol {{#unless (lookup @root.extraFolders this.id)}}collapsed{{/unless}}" data-document-id="{{this.id}}">
-        <header class="directory-item party-header flexrow" data-entry-id="{{this.id}}" data-document-id="{{this.id}}">
+{{#each parties as |party|}}
+    <li class="party flexcol {{#unless (lookup @root.extraFolders party.id)}}collapsed{{/unless}}" data-entry-id="{{party.id}}" data-document-id="{{party.id}}">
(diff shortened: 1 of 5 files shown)


static/templates/sidebar/party-document-partial.hbs
```diff
@@ -1,6 +1,6 @@
 <ol class="party-list">
 {{#each parties}}
-    <li class="party flexcol {{#unless (lookup @root.extraFolders this.id)}}collapsed{{/unless}}" data-document-id="{{this.id}}">
+    <li class="party flexcol {{#unless (lookup @root.extraFolders this.id)}}collapsed{{/unless}}" data-entry-id="{{this.id}}" data-document-id="{{this.id}}">
```
stwlam, line 3: Here is the actual `this` crime!

CarlosFdez: hah! Sure I'll try to replace it all. The thing was copied from the partials foundry has, and they use this! Perhaps its deviated enough to where I can replace the whole thing.

stwlam:
I though RK not working for players was a feature :)

CarlosFdez:
100% not. It doesn't RK creatures, it just shows you the RK-worthy skills in your party. Note that the overview already shows everyone's best skills.

If the RK tab shouldn't be shown, the skill summary shouldn't be shown either. At that point, is there a reason to show the interview? There's a setting the GM can toggle to completely hide the overview if not, but if there still is, perhaps we need to make the setting more granular?

CarlosFdez:
I'll start working on more visibility settings, but I'm going to handle such things as another PR.

PR #9059, opened 2023-08-01
Add crit stage logic to affliction

reyzor1991:
Increment/Decrement 2 stages if crit fail/success

Skip createStageMessage on create if onset present


reyzor1991:
https://github.com/foundryvtt/pf2e/assets/6688366/42b4e817-9e1a-40f3-bd92-454f7bacbeda

PR #9066, opened 2023-08-01
Add alchemical resistance

Drental:
known issue: same as radiation resistance, the alchemical resistance will trigger on every damage instance of a damage pool (tested with Schoma Lyzerius' bottled Lightning + Firebug ability...)


stwlam:
Could you reduce this PR to the actual changes you're attempting to make?

stwlam:
Additionally, the description of the resistance has to do with action traits rather than item traits.

Drental:
damn, sorry about that, rebased and it's now the proper changes only

PR #9071, opened 2023-08-02
Properly observe the scope of RollOption rule elements in beforeRoll

nikolaj-a:


CarlosFdez:
Why does this need to be filtered out at rule element level? Shouldn't placement only determine where it is rendered in the character sheet, and otherwise have no impact?

PR #9099, opened 2023-08-03
Fix broken links in remaster journal

TikaelSol:


stwlam:
The UUIDs are still malformed

PR #9108, opened 2023-08-03
Add affliction user stage handling

reyzor1991:
Create save dc button in chat message
Handle onset of Affliction
Handle stage duration of Affliction
Handle max duration of Affliction
Increment/Decrement 2 stages if crit fail/success

### DIFF
--- a/src/module/apps/effects-panel.ts
+++ b/src/module/apps/effects-panel.ts
@@ -1,10 +1,11 @@
 import { AbstractEffectPF2e, AfflictionPF2e, ConditionPF2e, EffectPF2e } from "@item";
-import { EffectExpiryType } from "@item/effect/data.ts";
+import { AbstractEffectExpiryType } from "@item/abstract-effect/data.ts";
 import { ActorPF2e } from "@actor";
 import { InlineRollLinks } from "@scripts/ui/inline-roll-links.ts";
 import { htmlQuery, htmlQueryAll } from "@util";
 import type { TokenDocumentPF2e } from "@scene/token-document/document.ts";
 import { PersistentDialog } from "@item/condition/persistent-damage-dialog.ts";
+import { isRelevantEvent } from "@scripts/sheet-util.ts";
 
 export class EffectsPanel extends Application {
     private get token(): TokenDocumentPF2e | null {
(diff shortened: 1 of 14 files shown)


src/module/chat-message/listeners/cards.ts
```diff
@@ -76,6 +77,8 @@ class ChatCards {
                     return;
                 case "spell-save":
                     return this.#rollActorSaves({ event, button, actor, item });
+                case "affliction-save":
```
CarlosFdez, line 80:
```suggestion
                case "affliction-save":
```

CarlosFdez:
Thanks for this. I'll be looking on and off in the next few days. I don't know about the increment/decrement taking a double stage thing, but I don't know what to recommend yet (there's several ways I'd go, and I'll need to look at this more to narrow it down)

src/module/actor/base.ts
```diff
@@ -359,10 +359,14 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
     getSelfRollOptions(prefix: "self" | "target" | "origin" = "self"): string[] {
-        const { rollOptions } = this;
-        return Object.keys(rollOptions.all).flatMap((o) =>
-            o.startsWith("self:") && rollOptions.all[o] ? o.replace(/^self/, prefix) : []
-        );
+        const { rollOptions, itemTypes } = this;
+        return Object.keys(rollOptions.all)
+            .flatMap((o) => (o.startsWith("self:") && rollOptions.all[o] ? o.replace(/^self/, prefix) : []))
+            .concat(
+                ...itemTypes.affliction
```
CarlosFdez, line 366: Instead of adding item stuff to the actor roll options here, do it in the item itself. For example, check out AbstractEffectPF2e#prepareActorData()

src/module/item/affliction/document.ts
```diff
@@ -1,26 +1,32 @@
...
-import { AfflictionFlags, AfflictionSource, AfflictionSystemData } from "./data.ts";
 
 class AfflictionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends AbstractEffectPF2e<TParent> {
-    constructor(source: object, context?: DocumentConstructionContext<TParent>) {
-        super(source, context);
-        if (BUILD_MODE === "production") {
-            throw ErrorPF2e("Affliction items are not available in production builds");
-        }
-    }
+    static DURATION_UNITS: Readonly<Record<string, number>> = {
```
CarlosFdez, line 17: Given this seems to be a copy of the effect one, I think we outta move this to the abstract effect superclass.

src/module/apps/effects-panel.ts
```diff
@@ -99,21 +100,25 @@ export class EffectsPanel extends Application {
 
             const iconElem = effectEl.querySelector(":scope > .icon");
             // Increase or render persistent-damage dialog on left click
-            iconElem?.addEventListener("click", async () => {
+            iconElem?.addEventListener("click", async (event?: JQuery.TriggeredEvent | Event | null) => {
```
CarlosFdez, line 103:
```suggestion
            iconElem?.addEventListener("click", async (event) => {
```

Same for the other one.

CarlosFdez:
If you could find a way to break this up sub-components, it'd be much easier to review, but I'll try my best

src/module/apps/effects-panel.ts
reyzor1991: Thank you, if you have any idea how to better change crit increasing logic it would be great

CarlosFdez:
I need to look more at this, but my immediate thought is that we need to clean up that duplicate code between this and EffectPF2e. AbstractEffect was created for a reason. I just don't have any actionable advice as to how and where yet. I'll need to pull this and dig locally (probably this monday).

If you can find standalone aspects that are simple (like representing onset), making a new PR with just those aspects will make it a lot quicker to review and merge it. Unfortunately for major projects like this its easier when its one aspect at a time.

reyzor1991:
I will try to create small PRs

CarlosFdez:
Awesome! The one bit I'm unsure about feature wise that I'll need to think throughly about is the left/right click with modifiers for 2 stages bit. Everything else seems fine at least user wise.

reyzor1991:
Static files PR - https://github.com/foundryvtt/pf2e/pull/9162

reyzor1991:
affliction traits PR- https://github.com/foundryvtt/pf2e/pull/9163

reyzor1991:
save button PR - https://github.com/foundryvtt/pf2e/pull/9172

PR #9114, opened 2023-08-03
Add selectors: parameter for rollable checks

dmarcuse:
This works similarly to the `traits:` parameter, supporting multiple values. This PR also includes some minor data updates to take advantage of this functionality, adding the `attack`, `attack-roll`, and `impulse-attack-roll` selectors to the impulse attack roll button in Elemental Blast, and adding rule elements to Gate Attenuators to automate the item bonuses and constitution increase.

### DIFF
--- a/packs/actions/elemental-blast.json
+++ b/packs/actions/elemental-blast.json
@@ -11,7 +11,7 @@
         },
         "category": "offensive",
         "description": {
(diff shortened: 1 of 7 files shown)


stwlam:
I don't think this is a good way to go. A saving throw shouldn't be able to declare itself a skill check.

dmarcuse:
Any ideas for alternatives? The main scenarios I was hoping to address with this PR are for attack rolls (for kineticist impulses, deviant abilities, spellguns, etc), so maybe a new `@Attack` inline button would work?

CarlosFdez:
It looks like https://github.com/foundryvtt/pf2e/pull/9137 is an attempt to handle this scenario.

I'm not fully opposed to this idea and I have considered it. A saving throw declaring itself a skill check is silly but it doesn't really concern me: its a goof localized to a single button that amounts to garbage in / garbage out. But anything involving these tend to end up being a bit sticky, we can't do them willy nilly (I'm already kinda sad we had to re-introduce overrideTraits).

One idea I had was for domain prefixes, so if you put "impulse" it'll complete it to "impulse-attack", but I'm not quite sure about that one either.

PR #9123, opened 2023-08-04
Assume actor is level 1 in @Damage instead of skipping

CarlosFdez:
This avoids the need for the max(1, formula) hack and also makes most formulas prettier in unowned items.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -613,13 +613,9 @@ async function augmentInlineDamageRoll(
 ): Promise<{ template: InlineDamageTemplate; context: DamageRollContext } | null> {
     const { name, actor, item, traits } = args;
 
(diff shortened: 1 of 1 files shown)


src/module/system/text-editor.ts
```diff
@@ -613,13 +613,9 @@ async function augmentInlineDamageRoll(
...
 
-    // Special case: this is probably an unowned item. Show the normal formula in such cases.
-    if (!actor && baseFormula.includes("@actor")) {
-        return null;
-    }
-
     try {
-        const rollData = item?.getRollData() ?? actor?.getRollData();
+        const rollData: Record<string, unknown> = item?.getRollData() ?? actor?.getRollData() ?? {};
+        if (!rollData.actor) rollData.actor = { level: 1 };
```
stwlam, line 618:
```suggestion
        rollData.actor ??= { level: 1 };
```

PR #9128, opened 2023-08-04
Handle top level math terms and subtraction in @Damage

CarlosFdez:
The if/else wall is starting to intimidate me.


stwlam:
Perils of writing a roll parser on top of a roll parser

PR #9129, opened 2023-08-04
Add leveled damage/healing to kineticists impulses.

Zullock:


packs/feats/aerial-boomerang.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p>A blade of shearing wind races away from you in a @Template[type:line|distance:60]. Each creature in the area takes @Damage[(floor((-1+@actor.level)/2 +2))d4[slashing]]{leveled slashing damage} damage with a basic Reflex save against your class DC. In the final square of the line, the boomerang whirls in place until the end of your next turn. Any creature that ends its turn in that square has to save against the boomerang.</p>\n<p>On your next turn, you can use a single action, which has the concentrate trait, to return the boomerang to you. It returns in a line from its square to your current location, with the same effect as the initial line, then the impulse ends. You must have line of effect to the boomerang and be within 60 feet of it.</p>\n<hr />\n<p><strong>Level (+2)</strong> The damage increases by [[/r 1d4]].</p>"
```
TikaelSol, line 14:
```suggestion
            "value": "<p>A blade of shearing wind races away from you in a @Template[type:line|distance:60]. Each creature in the area takes @Damage[(floor((-1+@actor.level)/2 +2))d4[slashing]] damage with a basic Reflex save against your class DC. In the final square of the line, the boomerang whirls in place until the end of your next turn. Any creature that ends its turn in that square has to save against the boomerang.</p>\n<p>On your next turn, you can use a single action, which has the concentrate trait, to return the boomerang to you. It returns in a line from its square to your current location, with the same effect as the initial line, then the impulse ends. You must have line of effect to the boomerang and be within 60 feet of it.</p>\n<hr />\n<p><strong>Level (+2)</strong> The damage increases by 1d4.</p>"
```
This PR will need rebased to resolve the conflict, and we are going to go through and remove all the "Leveled Damage" buttons now that `@Damage` can just show the correct amount of dice.

Zullock: Closed the PR on accident, created new one with the changes (https://github.com/foundryvtt/pf2e/pull/9157).

PR #9139, opened 2023-08-04
Respect `overrideTraits` (and its absence) in @Damage expressions

### LINKED ISSUE #9138: @Damage inline roll with "traits" field always overrides containing item traits
When I create a `@Damage` link with no "traits" field, e.g.:

`@Damage[1d6]`

It inherits the traits of the containing item as expected.

However, when I create a `@Damage` link with the field:

`@Damage[1d6|traits:area-damage]`

It overrides all of the containing item's traits with the set of new traits.

I expected it to augment the existing traits, similarly to a `@Check`.

https://github.com/foundryvtt/pf2e/blob/2b954fefbc4a1f78ac61f1d7b1408554721a3fcd/src/module/system/text-editor.ts#L515C14-L515C14

stwlam:
Closes #9138

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -512,8 +512,17 @@ class TextEditorPF2e extends TextEditor {
         const item = args.rollData?.item instanceof ItemPF2e ? args.rollData?.item : null;
         const actor = (args.rollData?.actor instanceof ActorPF2e ? args.rollData?.actor : null) ?? item?.actor ?? null;
 
-        const traits = params.traits?.split(",") ?? item?.system.traits?.value;
-        const result = await augmentInlineDamageRoll(params.formula, { skipDialog: true, actor, item, traits });
+        const rollOptions = ((): string[] => {
+            const fromParams = params.traits?.split(",").flatMap((t) => t.trim() || []) ?? [];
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I was hoping to figure out how to represent frequency for dragon breath to do the most usual case, or failing that add a roll options thing since override traits is a awful, but we are indeed out of time on this, and the convention has been locked in :(

dcoffee:
> I was hoping to figure out how to represent frequency for dragon breath to do the most usual case, or failing that add a roll options thing since override traits is a awful, but we are indeed out of time on this, and the convention has been locked in :(

Out of curiosity, would you mind elaborating a bit more on the dragon breath frequency part, and why overrideTraits is awful?

CarlosFdez:
Roll options aren't traits, and having to define them via traits is a hack (one we've inherited through `@Check`). A better structure would have an "options:" or some-such param explicitly built for that purpose. "area-damage" isn't a trait, its a roll option.

I mistakenly had assumed this was for dragon breath, which needs to deal +4 damage when elite instead of +2 Someone can fix that via adding the "item:frequency:limited" roll option but a better solution is to be able to add a frequency to it.

PR #9163, opened 2023-08-06
Add affliction traits

reyzor1991:


CarlosFdez:
I don't think this is right. Actions already have those traits, I don't think there is a spell with the drug trait, and its possible for an affliction to be magical (they inherit the traits from the granting action if they don't have one already, such as "divine").

reyzor1991:
Here I try to add some traits of affliction to use in next PRs

I will check all packs and try to find all possible traits for afflictions

CarlosFdez:
If that's the reason, actionTraits already has them, as does spell (except for drug).

image

reyzor1991:
Small question, affliction trait could be alchemical? Or it only venom alchemical?
Screenshot 2023-08-06 at 06 29 07

CarlosFdez:
Personally, I think you don't have to change the traits for afflictions, and action traits suffices. But perhaps @stwlam can shed more light on this.

CarlosFdez:
> Small question, affliction trait could be alchemical? Or it only venom alchemical? Screenshot 2023-08-06 at 06 29 07

Yeah I think affliction traits are action traits. They're often put together as part of the same item. I think you can close this issue and remove changing traits from the main PR.

reyzor1991:
Ok, than I close PR and delete changes from main PR

PR #9171, opened 2023-08-06
Add missing Shory language

chrpow:
Add Shory Language from Stolen Fate Player's Guide


TikaelSol:
Make sure to `git checkout master` before starting new branches so you don't have old changes floating around on your PRs.

Also we need to add the language to the languages list in `src/scripts/config/index.ts` as well as the localization key.

TikaelSol:
Renamed to be in imperative case. PR titles should finish the sentence "If merged this PR will <PR Title>", this makes it easier to quickly write the changelog during updates.

PR #9172, opened 2023-08-06
Handling save button for affliction

reyzor1991:
Screenshot 2023-08-06 at 11 19 54

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -76,6 +76,11 @@ class ChatCards {
                     return;
                 case "spell-save":
                     return this.#rollActorSaves({ event, button, actor, item });
+                case "affliction-save":
+                    if (item?.isOfType("affliction")) {
+                        item.rollRecovery();
+                    }
+                    return;
                 case "spell-counteract":
                     spell?.rollCounteract(event);
                     return;
--- a/src/module/item/affliction/document.ts
+++ b/src/module/item/affliction/document.ts
@@ -6,6 +6,7 @@ import { AfflictionDamageTemplate, BaseDamageData, DamagePF2e, DamageRollContext
 import { DamageRoll } from "@system/damage/roll.ts";
 import { ErrorPF2e } from "@util";
 import { AfflictionFlags, AfflictionSource, AfflictionSystemData } from "./data.ts";
+import { DegreeOfSuccess } from "@system/degree-of-success.ts";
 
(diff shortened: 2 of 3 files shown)


src/module/chat-message/listeners/cards.ts
```diff
@@ -233,14 +235,24 @@ class ChatCards {
...
+            const dc = ((): CheckDC | null => {
+                const v = Number(button.dataset.dc ?? "NaN");
+                if (Number.isInteger(v)) {
+                    const name = item?.name;
+                    const label = localize("DCWithName", { name });
+                    return { label, value: v };
+                }
+                return null;
+            })();
+
```
CarlosFdez, line 248: This entire block looks like a separate feature that should probably be supported in Statistic#roll(), should we choose to handle it. This affects spells as well, and should probably be for everything if we want it.

If you think it is indeed better, perhaps separating this into a new thing handled by Statistic's roll function might be best.

CarlosFdez:
Checks for isSave should instead check if item.actor is falsey. Afflictions always have a save, even if the DC is zero, but if there's no actor then there is no reason to do a save.

If possible, try to avoid overriding getChatData(), I think it might be possible to refer to `item` in the template directly for all the info you need. Spells are actually pretty old code and thus aren't the greatest example for how to write it. If you find afterwards that you still need it though feel free to keep it around it.

src/module/item/affliction/document.ts
```diff
@@ -4,17 +4,11 @@ import { UserPF2e } from "@module/user/index.ts";
...
+import { DegreeOfSuccess } from "@system/degree-of-success.ts";
 
 class AfflictionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends AbstractEffectPF2e<TParent> {
-    constructor(source: object, context?: DocumentConstructionContext<TParent>) {
-        super(source, context);
-        if (BUILD_MODE === "production") {
-            throw ErrorPF2e("Affliction items are not available in production builds");
-        }
-    }
-
```
CarlosFdez, line 17: shark isn't gonna be too happy about this one, though I personally think showing a ui notification error is fine instead. You'll have to ask him for this one :(

reyzor1991: @stwlam Is it ok if I delete this one check? I can change error to ui notification.

CarlosFdez: I'm guessing not. The rest of this looks fine, and I pushed up some condition granting stuff elsewhere. Lets at least try to finish this as soon as possible. If we're fast enough it can be either the next release or the one after.

reyzor1991: So I will roll back this change

PR #9173, opened 2023-08-06
Support options: param for inline @Damage and @Check

dcoffee:
Add support for `options:` syntax to `@Damage`.

Add roll option data to the inline link element created by `@Damage`.

Account for roll options (as well as traits) in `augmentInlineDamageRoll`.


src/module/system/text-editor.ts
```diff
@@ -513,16 +515,21 @@ class TextEditorPF2e extends TextEditor {
...
         const actor = (args.rollData?.actor instanceof ActorPF2e ? args.rollData?.actor : null) ?? item?.actor ?? null;
 
-        const rollOptions = ((): string[] => {
+        const traits = ((): string[] => {
             const fromParams = params.traits?.split(",").flatMap((t) => t.trim() || []) ?? [];
             const fromItem = item?.system.traits?.value ?? [];
             return params.overrideTraits === "true" ? fromParams : R.uniq([...fromParams, ...fromItem]);
         })().sort();
+
+        const extraRollOptions = ((): string[] => params.options?.split(",").flatMap((t) => t.trim() || []) ?? [])()
```
CarlosFdez, line 524: I can't write it as a suggestion without verifying but `R.compact(params.options?.split(",").map((t) => t.trim() ?? []).sort()` should be cleaner here.

stwlam:
Damage roll options are already supported. This would be a breaking change and require a migration.

dcoffee:
> Damage roll options are already supported. This would be a breaking change and require a migration.

Subject to the name change- care to elaborate?  This shouldn't break anything as it's only adding a new param: `option:`, which shouldn't be used yet anywhere.

src/module/system/text-editor.ts
dcoffee: Should be updated now- thanks for the suggestion!

CarlosFdez:
I think this is fine if it also does `@Check` since we wouldn't want an inconsistency. We'll also need to inform data entry to phase out using traits to inject roll options.

dcoffee:
> I think this is fine if it also does `@Check` since we wouldn't want an inconsistency. We'll also need to inform data entry to phase out using traits to inject roll options.

I noticed that there are some domains for inline damage rolls, what about doing an RE to get this behavior instead of this pull request?

```json
{
    "key":"RollOption",
    "selector":"{item|_id}-inline-damage",
    "option(s)":"put roll option(s) here"}
```

I tested this and got equivalent behavior for my use case (dragon breath weapon damage on a PC Eidolon).

CarlosFdez:
That's a question for @TikaelSol and @SpartanCPA but my immediate guess is that it may not work in situations with multiple rolls, and its also more cumbersome.

dcoffee:
> That's a question for @TikaelSol and @SpartanCPA but my immediate guess is that it may not work in situations with multiple rolls, and its also more cumbersome.

Got it- working on the `@Check` part now just to see how it goes.

CarlosFdez:
Wunderbar! Except the = for the destructuring going to the new line, but prettier gets what it wants.

Later far in the future we can remove traits from extra roll options and life will be good.

dcoffee:
@stwlam @CarlosFdez Is there no CD for this? I was kinda expecting these changes to be in the next release after it was merged.

CarlosFdez:
Except for hotfixes, releases are *usually* on a two week cycle,

dcoffee:
So 2 weeks after commit or just everything up to that point?

TMunsie:
> So 2 weeks after commit or just everything up to that point? I thought the merge was prior to 5.3.2 was cut for release.
> 
> Ah well, was hoping to use them in my game.

You can run your game off of the development branch. If you know how to commit I’m pretty sure you know how to build :)

dcoffee:
I don't run the server I play on, unfortunately, but they do keep their
pf2e system up to date.  I had been using the master build locally in
foundry which I was surprised that the changes weren't in 5.3.2.

TMunsie:
> I don't run the server I play on, unfortunately, but they do keep their pf2e system up to date. I had been using the master build locally in foundry, which is why I was surprised that the changes weren't in 5.3.2. Also, master is the development branch?

Well, it is more of a feature than a bug fix so that’s why it didn’t get into 5.3.2. And yes, the master branch is dev. Release is release and they’re version-tagged.

dcoffee:
> > I don't run the server I play on, unfortunately, but they do keep their pf2e system up to date. I had been using the master build locally in foundry, which is why I was surprised that the changes weren't in 5.3.2. Also, master is the development branch?
> 
> Well, it is more of a feature than a bug fix so that’s why it didn’t get into 5.3.2. And yes, the master branch is dev. Release is release and they’re version-tagged.

Gotcha- I think I was confused by the CD.yml, assuming that the delivery was automated.  Thanks!

PR #9194, opened 2023-08-07
Refactor strike button lables and fix display of strike modifiers for strikes modified by rule elements

In3luki:
Breaking (?) changes:
- Strike variant labels are no longer present in the actor data, only in sheet data. This could be changed to an async getter but that would be a breaking change too.
- The `getFormula` option is now `getLabel`. I could add a deprecation warning for `getFormula` if that is in use by modules at all.

Strikes from `Spell Effect: Dinosaur Form (Tyrannosaurus)` heightened to rank 8.
image

NPC strikes:
image


stwlam:
I think we can settle for a simpler solution until all this gets killed off.

PR #9203, opened 2023-08-07
Use `Statistic` for familiar attack-roll statistic

stwlam:
plus kill the jquery

### DIFF
--- a/static/templates/actors/familiar-sheet.hbs
+++ b/static/templates/actors/familiar-sheet.hbs
@@ -111,7 +111,7 @@
                             </div>
                             <div class="rollable skills-attack" title="{{data.attack.breakdown}}" data-attack-roll="">
                                 <strong>{{localize "PF2E.Familiar.AttackRoll"}}</strong>
-                                <span>{{numberFormat data.attack.value decimals=0 sign=true}}</span>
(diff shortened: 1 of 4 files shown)


static/templates/actors/familiar-sheet.hbs
```diff
@@ -111,7 +111,7 @@
                             </div>
                             <div class="rollable skills-attack" title="{{data.attack.breakdown}}" data-attack-roll="">
                                 <strong>{{localize "PF2E.Familiar.AttackRoll"}}</strong>
-                                <span>{{numberFormat data.attack.value decimals=0 sign=true}}</span>
+                                <span>{{numberFormat data.attack.check.mod decimals=0 sign=true}}</span>
```
CarlosFdez, line 114: If I recall, we can do data.attack.mod directly correct? Its dc that returns an object.

CarlosFdez:
Conflicts :^)

static/templates/actors/familiar-sheet.hbs
stwlam, line 114:
```suggestion
                                <span>{{numberFormat data.attack.mod decimals=0 sign=true}}</span>
```

PR #9205, opened 2023-08-08
Nudge `FilePicker` launched from `TokenConfig` to redirect to root

stwlam:


CarlosFdez:
Redirect to root is an array? Not a boolean?

stwlam:
Yep.

PR #9210, opened 2023-08-08
Fix bugged description for Composite trait

### LINKED ISSUE #9102: Bug:  Composite trait text is missing
image

chrpow:
Fix the bug that caused "Composite" trait to not show a proper description.
Bug was caused by omission of "PF2E." in the map in traits.ts.

Fixes #9102


chrpow:
image

PR #9214, opened 2023-08-08
Add onset to affliction

reyzor1991:


src/module/item/affliction/document.ts
```diff
@@ -65,6 +71,22 @@ class AfflictionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
         await this.update({ system: { stage } });
     }
 
+    get onsetDuration(): number {
+        const value = this.totalDuration(this.system.onset);
+        return value === Infinity ? 0 : value;
+    }
+
+    totalDuration(duration: AfflictionDuration | AfflictionOnset | undefined): number {
```
CarlosFdez, line 79: This needs to be a getter and match effect's signature as best as possible.

src/module/item/abstract-effect/document.ts
```diff
@@ -9,6 +9,14 @@ import { EffectBadge } from "./data.ts";
 
 /** Base effect type for all PF2e effects including conditions and afflictions */
 abstract class AbstractEffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends ItemPF2e<TParent> {
+    /** Duration units for affliction and effect */
+    static DURATION_UNITS: Readonly<Record<string, number>> = {
```
CarlosFdez, line 13: What if this was in a newly created `abstract-effect/values.ts`?

src/module/item/affliction/document.ts
reyzor1991: in this case we will have 2 similar getters, 1 for duration of affliction(max duration) and 1 for stage duration

CarlosFdez: That sounds fine to me. They're different concepts, and totalDuration is more or less the same as the effect one.

src/module/item/affliction/document.ts
```diff
@@ -196,8 +204,11 @@ class AfflictionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
             }))
         );
 
-        // Create the message in the chat
-        await this.createStageMessage();
+        // Skip create message if there is onset
+        if (!this.system.onset) {
+            // Create the message in the chat
```
CarlosFdez, line 209:
```suggestion
        // Show create message if there is no onset
        if (!this.system.onset) {
```

Issue #9182, opened 2023-08-06
Send to chat function in crafting menu not working


GuitarGuyNick:
![image](https://github.com/foundryvtt/pf2e/assets/139762764/8027cb3f-9c2e-4f13-b70e-028914ed1772)
if you click these icons they dont do anything when my player was trying to send the item to chat that he crafted.

dellatorreadrian:
Having a look at this

PR #9221, opened 2023-08-08
Fix Known Formulas not being prompted to chat from Character Sheet.

dellatorreadrian:
Issue: https://github.com/foundryvtt/pf2e/issues/9182

Notes: this was the easiest solution I had to solve this problem, but I don't think it is the best solution.

In my analysis, the problem here is the ActorSheetPF2e class referencing a property of one of its subclasses, CharacterSheetPF2e. Using "private", this implementation offers no problem, since it can be accessed even from the superclass.

However, using this great article as support (https://jamie.build/javascripts-new-private-class-fields.html), I found out that the new private "#" doesn't allow this type of access, even from a superclass.

This was the best workaround I could get without having too much work for the purpose intended, but I'm all ears to a better implementation.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -63,7 +63,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
     protected readonly actorConfigClass = CharacterConfig;
 
     /** A cache of this PC's known formulas, for use by sheet callbacks */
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I see the issue and that is silly! That said I don't know if setting this to private is the solution, the entire thing is flawed from the getgo.

dcoffee:
I also tend to agree that the current implementation of the base class ActorSheetPF2e magically "knowing" to check itself for possible extended properties seems like a flawed implementation.

PR #9230, opened 2023-08-09
Add rules to deviant feat effect

reyzor1991:


packs/feat-effects/effect-strained-metabolism.json
```diff
@@ -24,7 +24,41 @@
...
+                "alterations": [
+                    {
+                        "mode": "override",
+                        "property": "badge-value",
+                        "value": 2
+                    }
+                ],
+                "key": "GrantItem",
+                "uuid": "Compendium.pf2e.conditionitems.Item.Sickened"
+            }
```
TikaelSol, line 60: When you use `inMemoryOnly` it locks the effect value at that state, with no way to remove it until the granting effect is removed. In the case of the sickened here or the frightened on the other one those can be removed as normal. The frightened is fine to just leave unpredicated and without the in memory grant since that fires off when you get the mild backlash every time, but the sickened here won't be doable properly.

reyzor1991: But without ```inMemoryOnly``` sickened will not work correctly, when you apply effect with 3 stage (Severe) all is ok, sickened is added, but if you apply effect with 1 stage (Mild) and after it update to 3 condition not added

reyzor1991: https://github.com/foundryvtt/pf2e/assets/6688366/b8162aff-5c67-4d59-b161-a4349a7b5081

packs/feat-effects/effect-encroaching-presence.json
```diff
@@ -24,7 +24,58 @@
         "level": {
             "value": 1
         },
-        "rules": [],
+        "rules": [
+            {
+                "predicate": [
+                    "parent:badge:value:1"
+                ],
+                "alterations": [
```
TikaelSol, line 32:
```suggestion
            {
                "alterations": [
```

TikaelSol:
For strained metabolism I think there's no way around doing it like we had Oracle curses previously, with a ChoiceSet for the current stage. All of the penalties can be removed without leaving the current stage. For example in the first stage you can eat a snack to remove the penalties, but you are still at the minor backlash stage for the day even after the penalties are gone.

reyzor1991:
Then I will delete ```inMemoryOnly``` param if we can delete these condition

reyzor1991:
@TikaelSol I deleted `inMemoryOnly` flag and rebase PR

PR #9243, opened 2023-08-10
Adding the Reposition Athletic skill action from Pathfinder Player Core

Vesselchuck:


Vesselchuck:
It seems that something has failed in here, but I don't understand what.

TikaelSol:
We won't add things that only exist in previews from streams. The pipeline is failing because you are importing from `./basic/` instead of `./athletics/`

Vesselchuck:
> We won't add things that only exist in previews from streams. The pipeline is failing because you are importing from `./basic/` instead of `./athletics/`

I had to try my luck, cuz the action is too awesome! Anyway, hope this would be of any use to you in the future. Thanks for telling about my mistake, corrected that!

PR #9268, opened 2023-08-12
Migrate instances of "action:*" to "self:action:*"/"origin:action:*"

stwlam:
Includes sample migration run


TikaelSol:
We'll want to include inlines. `@Check[type:thievery|dc:29|traits:action:disable-a-device]` and the like are used in compendiums. I'd have to check if any premium content uses that too. For ones with proper actions we could push them to actual actions now, but a lot of these were done before the action macros for those actions were added. For ones that don't have existing action macros it should be self every time at least.

PR #9275, opened 2023-08-12
Add support for including self-applied effects in ability/feat items

stwlam:
This is just the data and sheet part: will follow up with inclusion in chat messages.

image

### DIFF
--- a/src/module/item/ability/document.ts
+++ b/src/module/item/ability/document.ts
@@ -29,6 +29,12 @@ class AbilityItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> exten
         if (this.actor && this.system.frequency) {
             this.system.frequency.value ??= this.system.frequency.max;
         }
+
+        this.system.selfEffect ??= null;
+        // Self effects are only usable with actions
+        if (this.system.actionType.value === "passive") {
+            this.system.selfEffect = null;
+        }
     }
 
     override getRollOptions(prefix = this.type): string[] {
@@ -70,7 +76,11 @@ class AbilityItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> exten
         options: DocumentModificationContext<TParent>,
         user: UserPF2e
     ): Promise<boolean | void> {
+        if (typeof changed.system?.category === "string") {
+            changed.system.category ||= null;
+        }
         normalizeActionChangeData(this, changed);
+
         return super._preUpdate(changed, options, user);
     }
 }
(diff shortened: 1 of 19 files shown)


src/module/item/ability/document.ts
```diff
@@ -70,7 +76,11 @@ class AbilityItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> exten
         options: DocumentModificationContext<TParent>,
         user: UserPF2e
     ): Promise<boolean | void> {
+        if (typeof changed.system?.category === "string") {
+            changed.system.category ||= null;
+        }
```
CarlosFdez, line 81:
```suggestion
        changed.system.category ||= null;
```

I believe that'll work?

src/module/item/ability/helpers.ts
```diff
@@ -40,4 +42,39 @@ function addSheetFrequencyListeners(item: ItemPF2e & SourceWithFrequencyData, ht
...
 }
 
-export { addSheetFrequencyListeners, normalizeActionChangeData };
+/** Create data for the "self-applied effect" drop zone on an ability or feat sheet. */
+function createSelfEffectSheetData(
+    data: SelfEffectReference | null
+): SelfEffectReference | { name: string; empty: true } {
+    if (data && !data.img) {
+        type MaybeIndexData = ((ClientDocument | CompendiumIndexData) & { img?: unknown }) | null;
+        const indexEntry: MaybeIndexData = fromUuidSync(data.uuid);
```
CarlosFdez, line 51: Since sheet data is async, couldn't you use the async version of this function?

src/module/item/feat/sheet.ts
```diff
@@ -44,10 +53,20 @@ class FeatSheetPF2e extends ItemSheetPF2e<FeatPF2e> {
             });
         }
 
+        htmlQuery(html, "a[data-action=delete-effect")?.addEventListener("click", () => {
+            if (this.item._source.system.selfEffect) {
+                this.item.update({ "system.-=selfEffect": null });
+            }
+        });
```
CarlosFdez, line 60: Stuff like these should probably be in the ability helpers, since they're going to be fully shared I imagine.

src/module/item/ability/document.ts
stwlam: It wouldn't because `system` is possibly undefined

src/module/item/ability/helpers.ts
stwlam: The one that makes a server call, in order to get the same data already indexed by default?

src/module/item/feat/sheet.ts
stwlam: You are a crazy person

PR #9339, opened 2023-08-15
Import kingdom features and fix feat data prep

CarlosFdez:


stwlam:
Almost feel like the this should be a hard-coded (still with the ability add homebrew feature items) since the core list is finite and will never grow.

CarlosFdez:
The feats themselves or which ones will get added? Given how easy it is to pull from compendium, its good either way, but some of these are going to want rule elements (there's one that adjusts the status bonus from investing in a skill).

PR #9341, opened 2023-08-16
Use item level as default actor level in @Damage

CarlosFdez:
Not correct for spells but actor level shouldn't be a param in spells anyways. This is mostly for action/feat/physical/campaignFeature


stwlam:
Why would the level of a feat affect @Damage?

CarlosFdez:
A level 6 feat should display the damage it would have on a level 6 actor, but it was showing the damage for a level 1 actor with that feat.

This code is the fallback for when the item is unowned, basically.

PR #9347, opened 2023-08-16
Speed up `TextEditor._enrichContentLinks`

stwlam:
Upstream retrieves documents from UUID links sequentially, which has a noticable load time with text containing many links: retrieve everything at once beforehand with the faster `UUIDUtils.fromUUIDs` system helper.


CarlosFdez:
This relies on caching in order to function?

PR #9349, opened 2023-08-16
Manage apex items without rule elements

stwlam:
image
image


src/module/item/equipment/document.ts
```diff
@@ -13,16 +14,48 @@ class EquipmentPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
...
+            for (const item of otherApexItems) {
+                item.system.apex!.selected = false;
+            }
+        } else if (
+            !this.system.apex.selected &&
+            this.isInvested &&
+            otherApexItems.every((e) => !e.system.apex!.selected)
+        ) {
+            this.system.apex.selected = true;
+        }
```
CarlosFdez, line 45: I think you can simplify some of this by checking for invested earlier. Perhaps you can early return if its not invested?

PR #9400, opened 2023-08-18
Add campaign effect for Malevolence

TikaelSol:
Technically called a condition, but I don't think we need a bespoke condition for a single adventure.


stwlam:
So make a condition :grin:

PR #9401, opened 2023-08-19
Convert weapon specialization to rule elements

CarlosFdez:
Also adds weapon specialization to psychic, which wasn't working before.

### DIFF
--- a/packs/classfeatures/greater-weapon-specialization.json
+++ b/packs/classfeatures/greater-weapon-specialization.json
@@ -19,7 +19,16 @@
         "prerequisites": {
             "value": []
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "AdjustModifier",
+                "mode": "multiply",
+                "relabel": "PF2E.GreaterWeaponSpecialization",
+                "selector": "strike-damage",
+                "slug": "weapon-specialization",
+                "value": 2
+            }
+        ],
         "source": {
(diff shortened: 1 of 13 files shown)


packs/classfeatures/greater-weapon-specialization.json
```diff
@@ -19,7 +19,16 @@
...
-        "rules": [],
+        "rules": [
+            {
+                "key": "AdjustModifier",
+                "mode": "multiply",
+                "relabel": "PF2E.GreaterWeaponSpecialization",
+                "selector": "strike-damage",
+                "slug": "weapon-specialization",
+                "value": 2
+            }
```
stwlam, line 30: Wouldn't this adjust damage of weapons regardless of proficiency?

src/module/migration/migrations/850-flat-footed-to-off-guard.ts
```diff
@@ -65,6 +65,8 @@ export class Migration850FlatFootedToOffGuard extends MigrationBase {
     }
 
     override async updateJournalEntry(source: JournalEntrySource): Promise<void> {
+        if (source._id === "6L2eweJuM8W7OCf2" && source.name === "Remaster Changes") return;
```
stwlam, line 68:
```suggestion
        if (source.name === "Remaster Changes") return;
```

PR #9427, opened 2023-08-20
Use ItemAlteration for Sparkling Targe

TikaelSol:
Doesn't automatically increase the hardness when defending against a spell, but it does work with the toggle on. Also adds the missing feat grant.

### DIFF
--- a/packs/classfeatures/sparkling-targe.json
+++ b/packs/classfeatures/sparkling-targe.json
@@ -31,42 +31,72 @@
                 "toggleable": true
             },
             {
-                "key": "ActiveEffectLike",
+                "itemType": "armor",
+                "key": "ItemAlteration",
                 "mode": "add",
-                "path": "system.attributes.shield.hardness",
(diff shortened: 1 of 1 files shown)


packs/classfeatures/sparkling-targe.json
```diff
@@ -31,42 +31,72 @@
...
-                "path": "system.attributes.shield.hardness",
-                "phase": "afterDerived",
                 "predicate": [
-                    "defend-against-magic",
+                    "item:category:shield",
+                    "item:equipped",
+                    {
+                        "or": [
+                            "magical",
+                            "spell",
```
stwlam, line 43:
```suggestion
                            "item:type:spell",
```

packs/classfeatures/sparkling-targe.json
```diff
@@ -31,42 +31,72 @@
...
```
stwlam, line 66:
```suggestion
                            "item:type:spell",
```

PR #9459, opened 2023-08-21
Add way to track partial boost in manual mode

CarlosFdez:
I noticed this when poking around, but its not possible to use manual ability boosts right now in any reasonable way without tracking your boosts elsewhere. This rectifies that.

This also fixes a bug where selected key score wasn't selected in manual.

image

### DIFF
--- a/src/module/actor/character/attribute-builder.ts
+++ b/src/module/actor/character/attribute-builder.ts
@@ -304,7 +304,7 @@ class AttributeBuilder extends Application {
         for (const input of htmlQueryAll<HTMLInputElement>(html, "input[type=text], input[type=number]")) {
             input.addEventListener("focus", () => {
                 if (input.type === "text" && input.dataset.dtype === "Number") {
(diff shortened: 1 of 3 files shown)


src/module/actor/character/attribute-builder.ts
```diff
@@ -294,10 +300,11 @@ class AttributeBuilder extends Application {
 
         $html.find("div.tooltip").tooltipster();
 
+        // Input handling for manual attribute score entry
         for (const input of htmlQueryAll<HTMLInputElement>(html, "input[type=text], input[type=number]")) {
             input.addEventListener("focus", () => {
                 if (input.type === "text" && input.dataset.dtype === "Number") {
-                    input.value = input.value.replace(/\D/g, "");
+                    input.value = input.value.replace(/[+-]/g, "");
```
stwlam, line 307:
```suggestion
                    input.value = input.value.replace(/[^-+.0-9]/g, "");
```

PR #9460, opened 2023-08-21
Wait for adjustCarryType to complete before considering execute complete

JDCalvert:
`WeaponAuxiliaryAction.execute` calls `Actor.adjustCarryType` (which is async) but doesn't wait for it to complete. This means that if we call `WeaponAuxiliaryAction.execute` programmatically, even with an `await`, then `Actor.adjustCarryType` will not have completed by the time `execute`'s returned Promise is resolved, and there is no way to determine when the `adjustCarryType` operation has actually completed.

This change makes `WeaponAuxiliaryAction.execute` await the completion of `Actor.adjustCarryType` before continuing.

### DIFF
--- a/src/module/actor/character/helpers.ts
+++ b/src/module/actor/character/helpers.ts
@@ -167,7 +167,6 @@ class WeaponAuxiliaryAction {
 
     async execute({ selection = null }: { selection?: string | null } = {}): Promise<void> {
         const { actor, weapon } = this;
-
         if (typeof this.carryType === "string") {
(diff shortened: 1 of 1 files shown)


stwlam:
Is there a concrete issue being addressed?

JDCalvert:
I'm trying to write a macro which, will draw a weapon (if it's not already) posting to chat just as if the player had clicked the "Draw" button, and then do things with the weapon that require the weapon be in a "held" state.

At the moment I've no way of waiting until the weapon is drawn before continuing with my operations.

PR #9474, opened 2023-08-22
Fix regression in display of actions on vehicle sheet

stwlam:
This required a bit of a refactor due to shared styling. The action markup is now in a separate, shared partial template.


CarlosFdez:
borken

PR #9491, opened 2023-08-23
Conditionally hide some chat buttons using data-visibility

stwlam:
plus kill the jquery


CarlosFdez:
I have a slight personal preference to continuing to show the variant buttons (but disabled), but its not worth the extra effort to do that.

PR #9498, opened 2023-08-23
Fix detection of animal companions

CarlosFdez:


src/module/canvas/token/object.ts
```diff
@@ -137,10 +136,7 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
             this.isAdjacentTo(flankee) &&
             flanking.canGangUp.includes("animal-companion") &&
             flankingBuddies.some(
-                (b) =>
-                    b.actor?.isOfType("character") &&
-                    ANIMAL_COMPANION_SOURCE_IDS.includes(b.actor.class?.sourceId ?? "") &&
-                    b.isAdjacentTo(flankee)
+                (b) => b.actor?.isOfType("character") && b.actor.traits.has("minion") && b.isAdjacentTo(flankee)
```
stwlam, line 139: This one is actually does need to specifically be an animal companion.

CarlosFdez: In practice what I had is fine (you can't have an AC and a construct companion) but more checks no hurty

PR #9503, opened 2023-08-23
Properly extract synthetics during damage application

stwlam:


src/module/chat-message/helpers.ts
```diff
@@ -115,28 +124,32 @@ async function applyDamageFromMessage({
             ...contextClone.getSelfRollOptions(),
         ]);
 
-        // target-specific healing adjustments
+        // Target-specific damage/healing adjustments
         const outcome = message.flags.pf2e.context?.outcome;
         const breakdown: string[] = [];
         const rolls: Rolled<Roll>[] = [];
-        if (typeof damage === "number" && damage < 0) {
+        if (typeof damage === "number" && damage !== 0) {
```
CarlosFdez, line 131:
```suggestion
        if (typeof damage === "number" && damage < 0) {
```

PR #9533, opened 2023-08-24
Fill out check data to be usable by damaging actions

stwlam:
This is a step towards being able to use the same roll class for strikes. It adds a dedicate template for `CheckRoll`s that includes action, damage, and other information.


static/templates/chat/check/roll.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="dice-roll {{type}}{{#if action}} {{action}}{{/if}}">
+    {{#if flavor}}<div class="dice-flavor">
+        {{flavor}}
+    </div>{{/if}}
```
CarlosFdez, line 4: Formatting went a bit weird here

PR #9633, opened 2023-08-28
Create an effect for Cinder Gaze

LebombJames:


packs/spells/cinder-gaze.json
```diff
@@ -20,17 +20,34 @@
...
+                "hC5GKInrfS2NyRMF": {
+                    "applyMod": true,
+                    "type": {
+                        "categories": [],
+                        "subtype": "",
+                        "value": ""
+                    },
+                    "value": "1"
+                }
+            }
```
TikaelSol, line 33: This doesn't deal damage

packs/spell-effects/effect-cinder-gaze.json
```diff
@@ -0,0 +1,77 @@
...
+                "predicate": [
+                    "action:recall-knowledge",
+                    "self:effect:cinder-gaze:3"
+                ],
+                "removeAfterRoll": true,
+                "selector": "skill-check"
+            }
+        ],
+        "source": {
+            "value": ""
```
TikaelSol, line 63:
```suggestion
            "value": "Pathfinder Dark Archive"
```

packs/spells/cinder-gaze.json
TikaelSol: A TempHP rule element.

TikaelSol: The issue with setting damage on the spell is bonuses to damage that apply to spells, like inspire courage, would affect this too.

LebombJames: That makes sense, but I don't know of a way to do "1 + spellcasting ability modifier" in a formula. Does an `@attribute` exist for the value of your spellcasting ability mod?

LebombJames: Well, it works, if only moderately hacky

PR #9663, opened 2023-08-29
Fixed Weapon Infusion Properties

aclethos:
Added missing Damage Types for Weapon Infusion, while the effect is on an actor
"You can choose to change the blast’s damage type to **bludgeoning, piercing, or slashing**—whichever suits the weapon shape"
 (RoE pg. 21)

### DIFF
--- a/packs/feats/weapon-infusion.json
+++ b/packs/feats/weapon-infusion.json
@@ -78,161 +78,161 @@
                 ],
                 "toggleable": true
             },
-			 {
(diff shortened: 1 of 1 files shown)


stwlam:
looks like you introduced a formatting issue

stwlam (changes requested):
Fix what is probably line breaks?

stwlam:
This needs to be done on the effect

PR #9670, opened 2023-08-29
Fixed Effect Weapon Infusion Properties

aclethos:
Added missing Damage Types for Weapon Infusion, while the effect is on an actor "You can choose to change the blast’s damage type to bludgeoning, piercing, or slashing—whichever suits the weapon shape" (RoE pg. 21)


stwlam:
Same problem as before—formatting issues were introduced

stwlam:
Expanded elemental blast code to handle this with less data

PR #9673, opened 2023-08-30
Save active party folder state and default to expanded

CarlosFdez:
And makes the No members display a bit smaller.

### DIFF
--- a/src/module/system/settings/index.ts
+++ b/src/module/system/settings/index.ts
@@ -335,12 +335,22 @@ function registerTrackingSettings(): void {
 
     game.settings.register("pf2e", "activeParty", {
         name: "Active Party",
+        scope: "world",
         config: false,
         type: String,
         default: "",
         onChange: () => {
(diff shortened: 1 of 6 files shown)


src/module/system/settings/index.ts
```diff
@@ -343,6 +343,14 @@ function registerTrackingSettings(): void {
         },
     });
 
+    // Tracks the last party folder state for next launch. Defaults to true so that "No Members" shows on initial creation.
+    game.settings.register("pf2e", "activePartyFolderState", {
+        name: "Active Party Opened or closed",
+        config: false,
```
stwlam, line 349:
```suggestion
        scope: client,
        config: false,
```

It's the default, but good to be explicit.

CarlosFdez: Indeed it is, because I forgot it for active party, and its where I copied from. I suppose this is now also a bug fix

PR #9721, opened 2023-08-31
Allow custom domains for inline damage rolls

CarlosFdez:
Restricted to sluggified that ends with -damage.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -138,11 +138,13 @@ class TextEditorPF2e extends TextEditor {
         if (baseFormula) {
             const item = rollData.item instanceof ItemPF2e ? rollData.item : null;
             const traits = anchor.dataset.pf2Traits?.split(",") ?? [];
+            const domains = anchor.dataset.pf2Domains?.split(",");
             const extraRollOptions = anchor.dataset.pf2RollOptions?.split(",") ?? [];
             const result = await augmentInlineDamageRoll(baseFormula, {
                 ...eventToRollParams(event),
                 actor,
                 item,
+                domains,
                 traits,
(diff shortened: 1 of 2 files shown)


src/module/system/text-editor.ts
```diff
@@ -676,7 +682,8 @@ async function augmentInlineDamageRoll(
             "inline-damage",
             item ? `${item.id}-inline-damage` : null,
             item ? `${sluggify(item.slug ?? item.name)}-inline-damage` : null,
-        ]);
+            args.domains,
```
stwlam, line 685: Can this be vetted?

CarlosFdez: Its vetted in the params, but we could move the point of validation to augmentDamageRoll() if you prefer.

src/module/system/text-editor.ts
```diff
@@ -534,6 +536,7 @@ class TextEditorPF2e extends TextEditor {
 
         const item = args.rollData?.item instanceof ItemPF2e ? args.rollData?.item : null;
         const actor = (args.rollData?.actor instanceof ActorPF2e ? args.rollData?.actor : null) ?? item?.actor ?? null;
+        const domains = params.domains?.split(",").filter((d) => sluggify(d) === d && d.endsWith("-damage"));
```
stwlam, line 539:
```suggestion
        const domains = params.domains?.split(",")..map((d) => sluggify(d)).filter((d) =>  /^[a-z][-a-z0-9]+-damage$/.test(d));
```

CarlosFdez: We can do the regex but I think if its not a valid slug, it should fail rather than surprise anyone.

PR #9722, opened 2023-08-31
Move physical-item material data to single property

stwlam:


src/module/item/physical/data.ts
```diff
@@ -42,11 +42,9 @@ interface PhysicalSystemSource extends ItemSystemSource {
         value: string;
     };
     containerId: string | null;
-    preciousMaterial: {
-        value: Exclude<PreciousMaterialType, "dragonhide" | "grisantian-pelt"> | null;
-    };
-    preciousMaterialGrade: {
-        value: PreciousMaterialGrade | null;
+    material: {
```
CarlosFdez, line 45: Why is this not ItemMaterialData? The type seems to match completely.

src/module/item/physical/sheet.ts
```diff
@@ -105,7 +105,7 @@ class PhysicalItemSheetPF2e<TItem extends PhysicalItemPF2e> extends ItemSheetPF2
             return result;
         }, {} as MaterialSheetData["materials"]);
 
-        const value = material.precious ? `${material.precious.type}|${material.precious.grade}` : "";
+        const value = material.type && material.grade ? `${material.type}|${material.grade}` : "";
```
CarlosFdez, line 108: I see that adamantine earlier in the PR defaults grade to standard if its unset. Should this do the same?

CarlosFdez:
Neato!

src/module/item/physical/data.ts
stwlam: I think I had some idea that one or both properties would be optional in source but decided to keep them both present but possibly `null`.

src/module/item/physical/sheet.ts
stwlam: That was to handle adamantine coming from a special ability, like Adamantine Strikes, which doesn't specify a grade.

CarlosFdez (approved):

PR #9740, opened 2023-08-31
Allow any item type to be passed to `ActorPF2e#getRollContext`

stwlam:


src/module/system/damage/damage.ts
```diff
@@ -86,7 +86,7 @@ export class DamagePF2e {
                 : "";
 
             const properties = ((): string => {
-                const range = item?.range ?? null;
+                const range = item?.isOfType("action","melee","weapon") ? item.range : null;
```
CarlosFdez, line 89:
```suggestion
                const range = item?.isOfType("action", "melee", "weapon") ? item.range : null;
```

PR #9748, opened 2023-09-01
Fix broken source ids for sheath item

MSAbaddon:


TikaelSol:
We should just delete these items.

PR #9749, opened 2023-09-01
Remove scroll case and sheath from actors

MSAbaddon:


TikaelSol:
Can delete these too

MSAbaddon:
> Can delete these too

Fine for me, done. I included the removing sheaths and will close the other pr.
I kept the scroll cases for the tax collector and the navigator, since those have other names and only had the scroll case as a base item.

PR #9790, opened 2023-09-02
Include roll options passed to strike methods when rolling

reyzor1991:
Possibility to add own option for damage/attack

`_token.actor.system.actions[0].roll({event, options: ['test']})`

`_token.actor.system.actions[0].damage({event, options: ['test']})`

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -1635,7 +1635,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                     target: { token: targetToken },
                     domains,
                     outcome: method === "damage" ? "success" : "criticalSuccess",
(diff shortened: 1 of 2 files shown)


src/module/actor/character/document.ts
```diff
@@ -1635,7 +1635,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                     target: { token: targetToken },
                     domains,
                     outcome: method === "damage" ? "success" : "criticalSuccess",
-                    options: baseOptions,
+                    options: new Set([...params.options, ...baseOptions]),
```
stwlam, line 1638:
```suggestion
                    options: new Set([...baseOptions, ...params.options]),
```

stwlam:
You'll want to do the same for NPC/hazard attacks (`@actor/helpers.ts`)

PR #9837, opened 2023-09-03
Add additional kingdom features and tweaks

CarlosFdez:
* Add hover for vacancy penalties
* Add support for effects
* Visual Tweaks to ability scores
image

### DIFF
--- a/src/styles/actor/party/kingdom/_activities.scss
+++ b/src/styles/actor/party/kingdom/_activities.scss
@@ -84,8 +84,7 @@
 
     h3.header .filters {
         float: right;
-        margin: 0;
-        margin-left: auto;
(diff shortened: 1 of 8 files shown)


src/styles/actor/party/kingdom/_activities.scss
```diff
@@ -86,6 +86,7 @@
         float: right;
         margin: 0;
         margin-left: auto;
+        margin-top: -0.125rem;
```
stwlam, line 89: Might be time to have a single `margin` line

CarlosFdez: It also turns out margin 0 is no longer required due to global tag changes but better safe than sorry.

PR #9877, opened 2023-09-05
Add rules-based vision override for individual scenes

stwlam:
image


src/module/scene/sheet.ts
```diff
@@ -1,58 +1,100 @@
...
+            if (darknessInput) {
+                darknessInput.disabled = true;
+                darknessInput.nextElementSibling?.classList.add("disabled");
+                const managedBy = WorldClock.createSyncedMessage();
+                darknessInput.closest(".form-group")?.querySelector("p.notes")?.append(managedBy);
             }
         }
     }
+
+    /** Nu */
```
CarlosFdez, line 94: ??????

stwlam: Nu

CarlosFdez: Fully understandable, valid even.

PR #9886, opened 2023-09-06
Add hearing range to system scene settings

stwlam:
Screenshot from 2023-09-06 04-27-55

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -2834,7 +2834,7 @@
                     "Name": "Remove Expired Effects"
                 },
                 "RulesBasedVision": {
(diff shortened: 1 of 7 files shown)


static/lang/en.json
```diff
@@ -3183,7 +3183,12 @@
         "SavesWillShort": "Will",
         "SavingThrowWithName": "{saveName} Saving Throw",
         "Scene": {
-            "Plural": "Scenes"
+            "HearingRange": {
+                "Caveat": "*For the performance purposes, the radius is a real measurement offset from a token's center rather than a grid-based one from its edges.",
```
CarlosFdez, line 3187:
```suggestion
                "Caveat": "*For performance purposes, the radius is a real measurement offset from a token's center rather than a grid-based one from its edges.",
```
real measurement offset?

PR #9891, opened 2023-09-06
Rename `Statistic#ability` to `Statistic#attribute`

stwlam:
Also:

- `Statistic#data#ability` to `Statistic#data#attribute`
- `createAbilityModifier` to `createAttributeModifier`
- `SKILL_EXPANDED#ability` to `SKILL_EXPANDED#attribute`

### DIFF
--- a/src/module/system/statistic/index.ts
+++ b/src/module/system/statistic/index.ts
@@ -3,7 +3,7 @@ import { TraitViewData } from "@actor/data/base.ts";
 import { calculateMAPs } from "@actor/helpers.ts";
 import {
     CheckModifier,
-    createAbilityModifier,
+    createAttributeModifier,
     createProficiencyModifier,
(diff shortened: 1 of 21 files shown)


src/module/system/statistic/index.ts
```diff
@@ -172,6 +172,15 @@ class Statistic extends BaseStatistic {
         return this.check.mod;
     }
 
+    /** @deprecated */
+    get ability(): AttributeString | null {
+        foundry.utils.logCompatibilityWarning(
+            "`SpellcastingEntryPF2e#ability` is deprecated. Use `SpellcastingEntryPF2e#attribute` instead.",
```
CarlosFdez, line 178:
```suggestion
            "`Statistic#ability` is deprecated. Use `Statistic#attribute` instead.",
```

PR #9939, opened 2023-09-08
Restore "flat-check" domain to flat-check `Statistic`s

stwlam:


src/module/system/statistic/statistic.ts
```diff
@@ -307,26 +307,30 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
...
+                extractModifiers(parent.actor.synthetics, data.check?.domains ?? []),
+            ].flat()
+        );
         const rollOptions = parent.createRollOptions(this.domains, config);
         this.modifiers = [
             ...parentModifiers,
             ...checkOnlyModifiers.map((modifier) => modifier.clone({ test: rollOptions })),
         ];
+        if (this.type === "flat-check" && this.modifiers.length > 0) {
+            throw ErrorPF2e("Flat checks cannot have modifiers");
```
CarlosFdez, line 331: I'd filter out the modifiers and do a warning instead. Best not to crash actor data preparation because someone decided to get frisky.

PR #9991, opened 2023-09-10
Add token-to-token visual flanking indicator when highlighting objects

JulieWinchester:
Addresses https://github.com/foundryvtt/pf2e/issues/4534. This PR supersedes a previous now-closed PR https://github.com/foundryvtt/pf2e/pull/7591 that approached this issue differently.

Screen Shot 2023-09-09 at 5 22 42 PM

Based on a suggestion from @stwlam in Discord, this PR adds token-to-token flanking indicator line(s) that appear when objects are highlighted by holding down ALT or otherwise. Lines highlight when a user's selected or own token is in flanking position with an allied token against a targeted enemy token. Highlight lines have an accompanying "Flanking" text label.

Screen Shot 2023-09-10 at 2 24 21 PM

Flanking highlight line(s) will appear only when: 1) the highlight objects key is pressed, 2) a token is selected or owned by a player, 3) the player has targeted one or more tokens, and 4) the selected token is potentially flanking one or more target tokens with allied tokens on opposite sides of the target enemy token(s). Multiple highlight lines can be drawn if the token is flanking multiple targets, or flanking a single target with multiple allies. Flanking calculation does not reveal whether the enemy token is actually flanked or is even flankable, in order to not reveal possible metagame information. It also does not account for Gang Up or similar features. Specifically, the flanking highlight line indicates that the enemy token _could_ be flanked by two or more tokens on opposite sides of the enemy token. 

Flanking highlight lines are drawn when objects are highlighted and get cleared when objects are no longer highlighted. Highlight lines will update on the fly if the user holds down the highlight object key and moves the selected token, selects a new token, or targets a new token. Lines **do not** get updated if the user holds down the highlight object key and another enemy or ally token moves "out from under" the rendered highlight line. This PR also **does not** show any indicator when a selected token is itself flanked by multiple enemy tokens. 

I've put some thought into optimization and keeping this change from affecting much of anything until the user is highlighting objects while selecting and targeting tokens. The code tries to not recalculate flanking buddies for a token unless necessary. Draw calls are associated with highlightObject and targetToken hooks,  token refreshPosition render flag, and token onControl and onRelease events. A slightly simpler alternative might be to associate draw calls with  token refereshVisibility, but this would trigger on token hover-in and hover-out, which doesn't affect flanking highlight line rendering. The code also tries to wait to add any flanking highlight PIXI Graphics object to a canvas layer as a child until it's explicitly needed. I've checked out performance snappiness on large maps from the Abomination Vaults module with many tokens on the field, and have not noticed any impacts from this change. 
 
For questions or anything else, I'm on the Discord as JellyfishJail.


stwlam:
Could this be drawn straight to a highlight layer? If it could, it would remove the need to rely on hooks or manually deallocate memory.

JulieWinchester:
@stwlam I don't think that graphics on highlight layers automatically become visible or invisible based on whether objects are highlighted (sorry if I'm misunderstanding you or missing something!). Drawing to a highlight layer seems to produce an always-on graphic, unless something is specifically handling visibility and refreshing, which would need a hook or a token event. Also, it seems like canvas highlight layers draw underneath tokens rather than on top of them, which isn't ideal for this. Drawing to a layer could help simplify clearing graphics, though, in some cases. 

One way to implement this that could reduce or eliminate the need to rely on hooks would be to have highlight lines draw and refresh on the token refreshVisibility event. I'm not sure if it gets called when target changes, so you might still need to rely on the targetToken hook, but it should take care of catching when objects are highlighted or when tokens move at least. 

Using refreshVisibility would make this change more self-contained within the token object. But it would lead to more unnecessary checks on whether or not to draw flanking highlight lines, including some more unnecessary calls to calculate flanking buddies for tokens. The refreshVisibility event gets called for every token on the board when objects are highlighted, and it gets called whenever the user hovers in or out of the token. Most of those checks would get stopped by one of the pre-conditions (e.g., that highlight objects key must be pressed, that token must be selected or owned, etc.), but some scenarios would lead to multiple repeated unnecessary flanking buddy checks, like repeatedly hovering in and out of a token while holding down the highlight objects key. If there's no optimization concern with running the flanking checks repeatedly, then that could be a way to simplify this down some and partially if not completely reduce dependency on hooks.

stwlam:
Compare with aura highlights: it is not a measured template but draws similar coloration directly to a highlight layer, without a `highlightObjects` hook. The hook is more for modules that don't have more direct access to client API internals.

JulieWinchester:
I’ll try and reimplement without using hooks, though like I said, that will probably lead to slightly less optimized use of flanking buddy checks, since it will have to key off of the refreshVisibility event (same as how auras draw highlight coloration). I’m not sure if there is a within-token way to detect change of target, but I’ll look again.

stwlam:
Auras ride along in `drawEffects`, and depending on the size of the aura, grid highlighting can be a relative expensive operation.

JulieWinchester:
Okay, I've revamped this to be simpler and to better align with token auras. HighlightHook is no longer used, and almost everything happens within the scope of a token. The shouldRender conditions were also updated to avoid running `getSelectedOrOwnActors`, since multiple tokens will likely be checking that at the same time, and now each individual token checks itself to see if it is selected, owned, etc. as appropriate. Also, instead of running around destroying multiple Graphics objects, the renderer keeps track of a main layer graphics object that it destroys and recreates as needed. This is similar to how auras use a highlight layer, except this renderer creates a layer child off of TokenPF2eLayer due to needing to display graphics above existing tokens. 

This code does still make use of the targetToken hook. I don't think there is a way for a token to know when other tokens are targeted, so you can't update the flanking highlight line originating from the first token when another token is targeted unless you use the system-wide hook.

src/scripts/hooks/target-token.ts
```diff
@@ -1,9 +1,21 @@
...
+            ui.combat.refreshTargetDisplay(tokenDocument);
+
+            // Draw flanking highlights if applicable
+            const actors = getSelectedOrOwnActors();
+            if (actors.length === 1 && tokenDocument.object) {
+                const actor = actors[0];
+                const actorToken = actor.getActiveTokens(false, true).shift();
+                if (canvas.tokens.highlightObjects && user === game.user && actorToken) {
+                    actorToken.object?.flankingHighlight.draw();
+                }
```
stwlam, line 17:
```suggestion
                const actorToken = actor.getActiveTokens(false, false).shift();
                if (canvas.tokens.highlightObjects && user === game.user) {
                    actorToken?.flankingHighlight.draw();
                }
```

src/module/canvas/token/object.ts
```diff
@@ -2,15 +2,20 @@ import { EffectPF2e } from "@item";
 import type { UserPF2e } from "@module/user/document.ts";
 import type { TokenDocumentPF2e } from "@scene/index.ts";
 import type { Renderer } from "pixi.js";
+import { pick } from "@util";
```
stwlam, line 5:
```suggestion
```

src/module/canvas/token/object.ts
```diff
@@ -105,27 +113,40 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
...
+     * @param context.reach           An optional reach distance specific to this measurement
+     * @param context.ignoreFlankable Optionally ignore flankable (for flanking position indicator) */
+    isFlanking(flankee: TokenPF2e, context: { reach?: number; ignoreFlankable?: boolean } = {}): boolean {
+        if (!(this.actor && this.canFlank(flankee, context))) return false;
+
+        // Return true if a flanking buddy is found
         const { flanking } = this.actor.attributes;
-        const flankingBuddies = canvas.tokens.placeables.filter((t) => t !== this && t.canFlank(flankee));
+        const flankingBuddies = canvas.tokens.placeables.filter(
+            (t) => t !== this && t.canFlank(flankee, pick(context, ["ignoreFlankable"]))
```
stwlam, line 148:
```suggestion
            (t) => t !== this && t.canFlank(flankee, R.pick(context, ["ignoreFlankable"]))
```

src/module/canvas/token/object.ts
```diff
@@ -144,13 +165,28 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
...
+     * Find other tokens that are in fact flanking a flankee with this token.
+     * Only detects tokens on opposite sides of flankee, does not support Gang Up or Side By Side.
+     * @param flankee                  The potentially flanked token
+     * @param context.reach           An optional reach distance specific to this measurement
+     * @param context.ignoreFlankable Optionally ignore flankable (for flanking position indicator) */
+    buddiesFlanking(flankee: TokenPF2e, context: { reach?: number; ignoreFlankable?: boolean } = {}): TokenPF2e[] {
+        if (!this.actor || !this.canFlank(flankee, context)) return [];
+
+        return canvas.tokens.placeables
+            .filter((t) => t !== this && t.canFlank(flankee, pick(context, ["ignoreFlankable"])))
```
stwlam, line 181:
```suggestion
            .filter((t) => t !== this && t.canFlank(flankee, R.pick(context, ["ignoreFlankable"])))
```

src/module/canvas/token/flanking-highlight/renderer.ts
```diff
@@ -0,0 +1,174 @@
...
+
+        // Styling
+        const style = CONFIG.canvasTextStyle.clone();
+        style.fontSize = 24;
+        if (canvas.dimensions?.size && canvas.dimensions.size >= 200) style.fontSize = 28;
+        else if (canvas.dimensions?.size && canvas.dimensions.size < 50) style.fontSize = 20;
+        style.fill = this.lineColor;
+        style.stroke = 0x000000;
+
+        const text = new PreciseText("Flanking", style);
```
simonward, line 140: Does this need to be localizable?

JulieWinchester:
Just pushed up another small update to fix a bug I noticed where it was possible to re-draw highlight lines and text without them being cleared, resulting in the line and text borders getting bolder and bolder and bolder. I also made the highlight line text localizable (image shows proof of concept, actual PR only includes text value for English).

Screen Shot 2023-09-18 at 10 43 29 AM

src/module/canvas/token/flanking-highlight/renderer.ts
JulieWinchester: This is done now, this text is localizable.

PR #10001, opened 2023-09-11
Support deletions in kingdom data

CarlosFdez:
DataModel update side only. Doesn't include the RecordField changes or any of the stuff relying on it.

### DIFF
--- a/src/module/actor/party/document.ts
+++ b/src/module/actor/party/document.ts
@@ -65,6 +65,21 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         return true;
     }
 
+    override updateSource(
+        data?: DocumentUpdateData<PartyPF2e>,
+        options?: DocumentSourceUpdateContext
+    ): DeepPartial<this["_source"]> {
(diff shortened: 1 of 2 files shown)


src/module/actor/party/document.ts
```diff
@@ -65,6 +65,19 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
+        options?: DocumentSourceUpdateContext
+    ): DeepPartial<this["_source"]> {
+        const expanded: DeepPartial<PartySource> = expandObject(data ?? {});
+        const campaignDiff =
+            expanded?.system?.campaign && this.campaign
+                ? this.campaign.updateSource(expanded.system.campaign, options)
+                : {};
+        const diff = super.updateSource(data, options);
+        return mergeObject(diff, campaignDiff);
+    }
```
stwlam, line 79: Can this super up and return early if there is no campaign data?

CarlosFdez: I could replace the final bit for a ternary and do an early check. But I found some bugs with this, so once I resolve that I'll make the change.

PR #10013, opened 2023-09-11
Include fuller blurb for each member in party sheet overview

stwlam:
- member name
- gender/pronouns
- level ancestry class

Before:
image

After:
image

### DIFF
--- a/src/styles/actor/party/_overview.scss
+++ b/src/styles/actor/party/_overview.scss
@@ -78,6 +78,7 @@
     font-family: var(--sans-serif);
     padding: 0.5rem 1rem;
     position: relative;
+    height: 7.375rem;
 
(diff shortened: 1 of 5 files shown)


src/styles/actor/party/_overview.scss
```diff
@@ -136,139 +136,155 @@
...
-            font-weight: 500;
-            font-variant: all-small-caps;
-            margin-left: 0.5rem;
+        header {
+            border-left: 1px solid #c9bfa9;
+            display: grid;
+            grid:
+                "name hero-points" 3fr
+                "gender-pronouns hero-points" auto
+                "blurb hero-points" auto
```
CarlosFdez, line 147: Doesn't seem like the styling matches the images.

PR #10031, opened 2023-09-12
Fix item summary / spell buttons firing when pressing enter elsewhere in the form

### LINKED ISSUE #10007: Pressing enter in a character sheet text field triggers use/consume buttons in unrelated item summaries
This is my own followup research from [a comment by /u/440Music on the PF2e subreddit](https://www.reddit.com/r/Pathfinder2e/comments/16fqd8i/comment/k04wlaz/?utm_source=reddit&utm_medium=web2x&context=3):

> When typing +X and hitting Enter, in the XP space, Foundry will auto-use items. It will burn scrolls, consumables, etc., and it entirely removes the object from your inventory (rather than setting its count to 0) so you have to look them up again. I have tried making sure that non-inventory tabs are open when adding XP only for it to not matter and burn random items anyway.

From my own testing I can confirm this happens for various text input fields such as the Hit Points entries in the sidebar, and can even trigger buttons after their summaries are collapsed or the overall sheet tab has changed. (I assume they are still attached to the DOM once they've been opened once?)

---

Since the character sheet overall is a `<form>`, pressing Enter in any control is likely triggering the [implicit submission behavior](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission) for that form, which fires a synthetic click event on the first submit button in tree order, in turn triggering `ConsumablePF2e.consume`.

It would be sufficient to change these buttons from the default submit to `type="button"` but I'm not sure if that submission behavior is desirable elsewhere. You might also be able to detect these synthetic events but I don't know if there's a consistent or standards compliant way to do so -- certainly they seem to have some bogus data though like `clientX === 0`.  Probably other options too like adding extra forms to the tree?

I'm happy to make a PR but was a little cautious about the moving parts and preferences here for my first contribution! 😆

tdhsmith:
Several `<button>` elements in the Item Summary and Spell Details interfaces were not given a `type` attribute, defaulting them to type `submit`. Because these interfaces are included inside of `<form>` elements, it made them viable targets for _implicit submission_ of the form when certain actions are taken, most notably when the Enter key was pressed in another input. This creates a synthetic click event and triggers the button's corresponding handlers, that are seemingly unrelated to the field the user has focused.

These buttons were changed to `type="button"`:

- triggered from anywhere on the character sheet:
  - use consumable
  - item summary spell attack and spell damage
- triggered from anywhere in a spell details window:
  - add heightening (fixed & interval)
  - all of the add-absent-property buttons in the fixed heighten section

Closes #10007

### DIFF
--- a/static/templates/actors/partials/item-summary.hbs
+++ b/static/templates/actors/partials/item-summary.hbs
@@ -34,15 +34,15 @@
 
         {{#unless item.hasVariants}}
             {{#if chatData.isAttack}}
-                <span><button class="spell_attack" data-action="spellAttack">{{localize "PF2E.AttackLabel"}}</button></span>
+                <span><button type="button" class="spell_attack" data-action="spellAttack">{{localize "PF2E.AttackLabel"}}</button></span>
             {{/if}}
             {{#if chatData.hasDamage}}
-                <span><button class="spell_damage" data-action="spellDamage">{{chatData.damageLabel}}: {{chatData.formula}}</button></span>
+                <span><button type="button" class="spell_damage" data-action="spellDamage">{{chatData.damageLabel}}: {{chatData.formula}}</button></span>
             {{/if}}
         {{/unless}}
 
         {{#if (and (eq item.type "consumable") item.uses.max)}}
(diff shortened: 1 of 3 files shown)


CarlosFdez:
Good catch!

PR #10032, opened 2023-09-12
Allow divvying of party stash using inventory members sidebar

tdhsmith:
To improve the usability of the stash screen for sorting loot, it seemed simpler to allow the party member blocks in the inventory sidebar to function as special drop targets for dragged items:

https://github.com/foundryvtt/pf2e/assets/1791252/c97af133-dfb1-4995-bd8c-ca859b2c65a4

This is a pretty simple reuse of `ActorSheetPF2e.moveItemBetweenActors` and thus has the standard restrictions of requiring a GM active for Player-initiated transfers.

### DIFF
--- a/src/module/actor/party/sheet.ts
+++ b/src/module/actor/party/sheet.ts
@@ -451,7 +451,7 @@ class PartySheetPF2e extends ActorSheetPF2e<PartyPF2e> {
                 await this.moveItemBetweenActors(
                     event,
                     item.actor.id,
-                    item.actor?.token?.id ?? null,
+                    item.actor.token?.id ?? null,
                     actorUuid,
                     null,
                     item.id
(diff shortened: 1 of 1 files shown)


stwlam:
This is a good idea. Can permissions be checked so that players can only drop on their own actors?

tdhsmith:
> This is a good idea. Can permissions be checked so that players can only drop on their own actors?

Good call, will investigate this

tdhsmith:
Ok so I confirmed through testing and code walkthrough the default is behavior is that players cannot drop items onto characters without update permission (via pre-existing `transferItemToActor` code).
https://github.com/foundryvtt/pf2e/blob/ec7d2c32e35caec1604c309939aeb3566a89430e/src/module/actor/base.ts#L1475-L1478 

Screen Shot 2023-09-13 at 7 43 47 AM

----

Did you just want that verified or were you suggesting making this into a setting of some sort?

PR #10041, opened 2023-09-13
Add hooked rune, run damage roll strike adjustments set by property runes

### LINKED ISSUE #9519: Unable to apply hooked property rune to melee weapons
I'm unable to apply the hooked property rune to melee weapons.

image

tdhsmith:
The goal here was to close #9519, but the challenge was that **Hooked** has different behaviors based on whether it added a trip trait and I couldn't identify a way for a static roll note to distinguish if a trait had been present on the source weapon or not. Certainly one might be able to abuse predicates, but even introducing a new roll option from a rune wasn't straightforward without uncomfortable levels of indirection.

So then I observed that strike adjustments have an `adjustDamageRoll` function -- _sure_ this currently only tweaks the damage _materials_, but it seemed by its name, placement, and 2nd-argument genericness that the intent was to provide a hook for any necessary data modifications. Thus this PR does as follows:

1. Updated `adjustDamageRoll` to also pass the current `RollNotePF2e` array
2. Fixed API omission where `adjustDamageRoll` on runes would never be called (it was only applied to adjustments from synthetics)
3. Updated typing on `WeaponPropertyRuneData.strikeAdjustments` to exclude `StrikeAdjustment.adjustTraits` and clarify that it is not currently supported
4. Introduced hooked weapon property rune, which adds the trip trait, and a critical damage note when the source weapon already had the trip trait

I'm sure it's very plausible I overlooked a better approach though.

### DIFF
--- a/src/module/item/physical/runes.ts
+++ b/src/module/item/physical/runes.ts
@@ -1652,17 +1652,19 @@ const WEAPON_PROPERTY_RUNES: { [T in WeaponPropertyRuneType]: WeaponPropertyRune
                 },
                 adjustDamageRoll: (weapon: WeaponPF2e | MeleePF2e, { notes }: { notes?: RollNotePF2e[] }): void => {
                     if (weapon._source.system.traits.value.includes("trip")) {
-                        notes?.push(new RollNotePF2e({
-                            selector: "strike-damage",
-                            outcome: ["criticalSuccess"],
-                            predicate: ["item:trait:trip"],
-                            title: "PF2E.WeaponPropertyRune.hooked.Name",
-                            text: "PF2E.WeaponPropertyRune.hooked.Note.criticalSuccess",
-                        }));
+                        notes?.push(
+                            new RollNotePF2e({
+                                selector: "strike-damage",
+                                outcome: ["criticalSuccess"],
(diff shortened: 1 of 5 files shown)


stwlam:
Using strike adjustments to add notes is a no-go, but I think we already have a place for it?

tdhsmith:
No prob.

Runes can add roll notes easily, but there aren't any roll option predicates that would correctly distinguish which weapons _originally_ had the trip trait.  It seemed like adding a new roll option just for this rune was a messy solution. 

Was there another place you had in mind?

tdhsmith:
I acknowledge that might be trying to over-automate, but the note risks being pretty confusing if it isn't conditional. When the player sees "If a hooked weapon normally has the trip trait [...]" they immediately have to ask themselves if the trip badge was added automatically by the rune or not...

Screen Shot 2023-09-13 at 10 14 05 AM

stwlam:
I would skip the note for now. Rune automation is getting moved to rule elements, so something that can't currently be replicated by them generally shouldn't be hard-coded in.

tdhsmith:
Well that future work makes sense to just wait for, sounds like a lovely improvement. I'll just start a fresh PR rather than force pushing over the top of this.

PR #10044, opened 2023-09-13
Fix ancestry feats without an ancestry trait being filtered in the compendium browser feats tab

### LINKED ISSUE #9803: Ancestry Feats open to all ancestries are filtered out when browsing for ancestry feats.
When I open the compendium browser from an empty ancestry feat slot on the feats tab of my character sheet, Ancestry Feats that are open to all ancestries are filtered from view. This applies particularly to a number of related ancestry feats from the _Impossible Lands_ guide, but there may be others. See attached screenshots. The affected feats can be found on the [_All Ancestries_ page of Archives of Nethys](https://2e.aonprd.com/Traits.aspx?ID=467).

Ancestry feats that do not have any specific ancestry trait should not be excluded from the Ancestry Feat list when ancestry traits are included in the filter.

Screenshot 2023-09-02 at 3 41 30 PM
Screenshot 2023-09-02 at 3 44 13 PM

In3luki:
When the `Ancestry Feat` category is selcted Ancestry Feats without an ancestry trait are now always shown regardless of the selected traits.

image

Closes #9803

### DIFF
--- a/src/module/apps/compendium-browser/tabs/feat.ts
+++ b/src/module/apps/compendium-browser/tabs/feat.ts
@@ -2,7 +2,7 @@ import { isObject, sluggify } from "@util";
 import { CompendiumBrowser } from "../index.ts";
 import { ContentTabName } from "../data.ts";
 import { CompendiumBrowserTab } from "./base.ts";
-import { CompendiumBrowserIndexData, FeatFilters } from "./data.ts";
+import { CompendiumBrowserIndexData, FeatFilters, MultiselectData } from "./data.ts";
 
 export class CompendiumBrowserFeatTab extends CompendiumBrowserTab {
     tabName: ContentTabName = "feat";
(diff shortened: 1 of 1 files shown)


src/module/apps/compendium-browser/tabs/feat.ts
```diff
@@ -133,6 +137,21 @@ export class CompendiumBrowserFeatTab extends CompendiumBrowserTab {
...
+        traits: string[],
+        selected: MultiselectData["selected"],
+        condition: MultiselectData["conjunction"]
+    ): boolean {
+        // Always return true for ancestry feats that have no ancestry trait
+        if (
+            this.filterData.checkboxes.category.selected.includes("ancestry") &&
+            !this.arrayIncludes(traits, this.#ancestryTraits)
+        ) {
+            return true;
```
tdhsmith, line 150: This works for the case where the selected traits are _all_ ancestry traits, but returns excess results if the user adds other non-ancestry traits because it returns early before checking any additional logic. 

For example, filtering on `["kobold", "mental"]` should return **Cringe** (matching kobold feat) and **Drain Emotion** (matching all-ancestries feat) but not other all-ancestry feats.

I'd suggest that instead of returning true here, you pass the selected trait list _with the ancestry traits removed_ to the super call so that you don't have to rebuild any logic here.

In3luki: Good catch. I agree that filtering the ancestry traits first should yield the expected result. I'll push up a new fix for my fix when I'm home.

PR #10057, opened 2023-09-14
Add automation for Weapon Siphon

dmarcuse:
This PR adds an effect for the Weapon Siphon from treasure vault. The toggleable roll option can be used to select the energy damage type (or disable damage if no bomb is currently loaded), and the increased MAP applies as long as the effect is on the sheet since even an unloaded weapon siphon affects the weapon's balance.
image

### DIFF
--- a/packs/equipment-effects/effect-weapon-siphon.json
+++ b/packs/equipment-effects/effect-weapon-siphon.json
@@ -1,0 +1,108 @@
+{
+    "_id": "mPb9Bhq8T1nsyWug",
+    "img": "systems/pf2e/icons/default-icons/effect.svg",
+    "name": "Effect: Weapon Siphon",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Weapon Siphon]</p>"
+        },
+        "duration": {
+            "expiry": "turn-start",
+            "sustained": false,
+            "unit": "unlimited",
+            "value": -1
+        },
+        "level": {
+            "value": 1
+        },
+        "rules": [
+            {
+                "choices": {
+                    "ownedItems": true,
+                    "predicate": [
+                        "item:melee"
+                    ],
+                    "types": [
+                        "weapon"
+                    ]
+                },
+                "flag": "effectWeaponSiphonWeapon",
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Prompt.Weapon"
+            },
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.WeaponSiphonDamage",
+                "option": "effect:weapon-siphon:damage",
+                "suboptions": [
+                    {
+                        "label": "PF2E.TraitAcid",
+                        "value": "acid"
+                    },
+                    {
+                        "label": "PF2E.TraitCold",
+                        "value": "cold"
+                    },
+                    {
+                        "label": "PF2E.TraitElectricity",
+                        "value": "electricity"
+                    },
+                    {
+                        "label": "PF2E.TraitFire",
+                        "value": "fire"
+                    },
+                    {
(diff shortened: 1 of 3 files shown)


packs/equipment-effects/effect-weapon-siphon.json
```diff
@@ -0,0 +1,108 @@
...
+                "mode": "subtract",
+                "selectors": [
+                    "{item|flags.pf2e.rulesSelections.effectWeaponSiphonWeapon}-attack"
+                ],
+                "slug": "multiple-attack-penalty",
+                "value": 1
+            }
+        ],
+        "source": {
+            "value": ""
```
TikaelSol, line 94:
```suggestion
            "value": "Pathfinder Treasure Vault"
```

packs/equipment-effects/effect-weapon-siphon.json
```diff
@@ -0,0 +1,108 @@
...
+                },
+                "flag": "effectWeaponSiphonWeapon",
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Prompt.Weapon"
+            },
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.WeaponSiphonDamage",
+                "option": "effect:weapon-siphon:damage",
+                "suboptions": [
```
TikaelSol, line 37:
```suggestion
                "option": "effect:weapon-siphon:damage",
                "alwaysActive": true,
                "suboptions": [
```

packs/equipment-effects/effect-weapon-siphon.json
```diff
@@ -0,0 +1,108 @@
+{
+    "_id": "mPb9Bhq8T1nsyWug",
+    "img": "systems/pf2e/icons/default-icons/effect.svg",
+    "name": "Effect: Weapon Siphon",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Weapon Siphon]</p>"
```
TikaelSol, line 7:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Weapon Siphon]</p>\n<p>The selected weapon deals an additional 1d4 damage of the selected damage type and the multiple attack penalty is one greater than usual.</p>"
```

TikaelSol:
Add a counter with an initial value of 3 to track the three strikes.

packs/equipment-effects/effect-weapon-siphon.json
dmarcuse: This is intentionally left false because there are three states the item can be in:

- Not attached to a weapon - nothing happens, don't apply the effect 
- Attached to a weapon, but no bomb loaded: you don't deal bonus damage, but the weapon siphon still worsens MAP, so you apply the effect but leave the toggle disabled. 
- Attached to a weapon, with a bomb loaded: you deal bonus damage and have worse MAP, so you apply the effect and toggle on the damage

dmarcuse:
@TikaelSol I chose not to use a counter for this effect because the effect is meant to be applied permanently (similar to e.g. Effect: Flaming Star - Armor) because the MAP increase applies regardless of whether a bomb is loaded. I've updated added a description (along the lines of your earlier comment) as well as a usage note explaining this.

packs/equipment-effects/effect-weapon-siphon.json
dmarcuse: Interesting, this isn't assignable from the UI so I had missed this. Fixed!

TikaelSol:
In which case we should make a labeled counter (`unloaded`, `1`, `2`, `3`) and make the damage contingent on `not` being at `unloaded`.

TikaelSol:
Let's hold off on this for a bit and see if we can't get more clever rather than doing this now and cleaning up after ourselves later.

Issue #6962, opened 2023-03-14
Treasure Vault runes not listed in weapon and armor rune selection dropdowns


AFigureOfBlue:
When applying runes to a weapon or armor, at least some (possibly all?) of the new runes from _Treasure Vault_ seem to be missing. The "merciful" and "authorized" weapon runes and the "portable" armor rune are ones that my players have encountered this with so far, and all the other ones I've spot-checked have been similarly missing from the dropdowns.

stwlam:
The entire runes framework is undergoing a refactor, but newer ones will make it in soon.

gourinator:
Can we get an ETA for the undergoing? And a workaround?

TMunsie:
ETA: when someone volunteers to do the work. The system is updated and improved by volunteer contribution.

Workaround: Since most runes are labels just change the label. If it does something specific add an item-specific rule element that does that thing or posts a note about it.

PR #10077, opened 2023-09-14
Add property runes missing from armor and weapon dropdowns

### LINKED ISSUE #9519: Unable to apply hooked property rune to melee weapons

tdhsmith:
Added some missing runes to their respective item dropdown menus. Closes #9519, #6962

- **Giant-Killing** (incl. damage die, resistance, and notes with save effect)
- **Hooked** (incl. addition of `trip` trait, but _no_ note, see objections in #10041) 
- **Malleable** (entry only)
- **Rooting** (incl. crit note with conditions/escape action link)
- **Spellbreaking** (entry only)

As part of this process, also addressed two minor bugs in property rune handling:

1. Clearing the dropdown for a weapon property rune slot that non-empty _later_ slots would incorrectly hide those slots but still keep the runes selected. Fixed by now keeping slots displayed up to one after the _last_ non-empty slot.
2. The string search used to check when to suppress lesser versions of runes on the same item could only ever suppress base versions. (e.g. "Major Fanged" could suppress "Fanged" but never "Greater Fanged") Fairly niche issue but more relevant now that there's a rune with 4 levels.

Lastly, removed redundant labels in the note strings of 5 property runes that repeated the rune name. (For example, a returning weapon would render to the roll chat as "Returning Returning Rune")


stwlam:
We intentionally excluded Spellbreaking on account of its imminent irrelevance

tdhsmith:
Spellbreaking has failed its save against School of Deletion magic

PR #10092, opened 2023-09-15
Stack item quantities of items looted with the `LootNPCsPopup`

### LINKED ISSUE #8796: Feature Request:  combine quantities when using "Loot selected tokens" on loot actors
Last week my players went through a large encounter with more than a dozen enemies, and I wanted looting to be quick, so I tried using the fancy button to combine all loot onto a single actor.  However, I ended up seeing this:

<details>
  <summary>Screenshot (minor Abomination Vaults spoilers)</summary>
  
image

</details>

I think the system should combine all identical instances of items when using this button.

To determine "identical", I think relying on the item slug or name should be enough.  Even though sometimes GMs alter items in NPC inventories, I think it'll be incredibly rare for a GM to only alter some but not all items of the same name in a single encounter, so combining based on that property should be fine in 99.99% of cases.

Here's what I think it should look like after combining items:

<details>
  <summary>Screenshot (minor Abomination Vaults spoilers)</summary>
  
image

</details>

(sorry about the discrepancies, I changed some amounts to include more enemies, so the sums don't add up;  I hope you still get what I'm aiming for)

In3luki:
This has the same logic as `PhysicalItemPF2e#isStackableWith`.

Recording 2023-09-15 at 13 53 27

I've also fixed the popup UI after recording the video:

image

Closes #8796

### DIFF
--- a/src/module/actor/sheet/loot/loot-npcs-popup.ts
+++ b/src/module/actor/sheet/loot/loot-npcs-popup.ts
@@ -1,5 +1,6 @@
 import { ActorPF2e } from "@actor/base.ts";
-import { PhysicalItemSource } from "@item/data/index.ts";
+import { PhysicalItemPF2e } from "@item";
+import { ScenePF2e, TokenDocumentPF2e } from "@scene";
 import { ErrorPF2e } from "@util";
 
 interface PopupData extends FormApplicationData<ActorPF2e> {
@@ -10,7 +11,7 @@ interface PopupData extends FormApplicationData<ActorPF2e> {
     }[];
 }
 
-export class LootNPCsPopup extends FormApplication<ActorPF2e> {
+class LootNPCsPopup extends FormApplication<ActorPF2e> {
     static override get defaultOptions(): FormApplicationOptions {
         const options = super.defaultOptions;
         options.id = "loot-NPCs";
@@ -25,8 +26,12 @@ export class LootNPCsPopup extends FormApplication<ActorPF2e> {
         _event: Event,
         formData: Record<string, unknown> & { selection?: boolean }
     ): Promise<void> {
-        const itemData: PhysicalItemSource[] = [];
+        const lootActor = this.object;
+        const newItems: PhysicalItemPF2e[] = [];
+        const itemUpdates = new Map<string, number>();
+        const itemsToDelete = new Map<ActorPF2e, string[]>();
         const selectionData = Array.isArray(formData.selection) ? formData.selection : [formData.selection];
+
         for (let i = 0; i < selectionData.length; i++) {
             const token = canvas.tokens.placeables.find((token) => token.actor && token.id === this.form[i]?.id);
             if (!token) {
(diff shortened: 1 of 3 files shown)


stwlam:
Can this reuse `ActorPF2e#findStackableItem`?

In3luki:
hmm, it might be possible to build a list of stackable items first and then only turn one of them to source and add the quantity of the rest.
I'll think about it,

In3luki:
Updated to use existing methods to determine stackability. Also now updates existing items on the loot actor if a stackable item is looted.

src/module/actor/sheet/loot/loot-npcs-popup.ts
```diff
@@ -34,14 +36,51 @@ export class LootNPCsPopup extends FormApplication<ActorPF2e> {
             }
             const currentSource = token.actor;
             if (selectionData[i] && currentSource) {
-                const currentSourceItemData = currentSource.inventory.map((item) => item.toObject());
-                itemData.push(...currentSourceItemData);
-                const idsToDelete = currentSourceItemData.map((item) => item._id);
+                for (const item of currentSource.inventory) {
+                    const stackableItem = currentSource.findStackableItem(this.object, item);
```
CarlosFdez, line 40: Should probably be this.object.findStackableItem() to keep it consistent with other uses of that same function.

After this I'm probably going to move findStackableItem() to actor.inventory and remove the first parameter, and add a deprecation warning for the existing function.

src/module/actor/sheet/loot/loot-npcs-popup.ts
```diff
@@ -34,14 +36,51 @@ export class LootNPCsPopup extends FormApplication<ActorPF2e> {
...
+                const idsToDelete = currentSource.inventory.map((item) => item.id);
                 currentSource.deleteEmbeddedDocuments("Item", idsToDelete);
             }
         }
-        if (itemData.length > 0) {
-            await this.object.createEmbeddedDocuments("Item", itemData);
+        // Create new items in the loot actor
+        if (newItems.length > 0) {
+            const stacked: PhysicalItemPF2e[] = [];
+            for (const item of newItems) {
```
CarlosFdez, line 55: This entire for looks like it could be a reduce. With a comment saying to merge stackable items within the same incoming batch.

src/module/actor/sheet/loot/loot-npcs-popup.ts
In3luki: yeah, I was wondering why wasn't at least a static method. I'll change it to this.object for now.

src/module/actor/sheet/loot/loot-npcs-popup.ts
In3luki: will do

In3luki:
Updated with applied suggestions. I've also moved the deletion of the source items to the end of the method to make it easier to recover from one of the other steps failing for whatever reason.

PR #10096, opened 2023-09-15
Fix `Select Other Variant` button being show on heightened spell chat cards

In3luki:
I'm not sure what exactly changed here but a heightened spell is technically a variant so it shoudn't have worked from the start.
Checking for applied overlays instead should only catch the correct instances.
The naming of Variant/Overlay/Override is a bit confusing.

image

### DIFF
--- a/static/templates/chat/spell-card.hbs
+++ b/static/templates/chat/spell-card.hbs
@@ -72,7 +72,7 @@
                             </button>
                         </div>
                     {{/if}}
-                    {{#if item.isVariant}}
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Using isVariant to define both might have indeed been a mistake. Most uses of isVariant should probably have been a check for the original, leaving isVariant to define an actual variant.

Heightened spells are a variation of the original spell, and use the same pipeline, but they're not a "variant" in the way we've coined.

PR #10158, opened 2023-09-18
Move character height/weight to biography

CarlosFdez:
image
image

### DIFF
--- a/static/templates/actors/character/tabs/biography.hbs
+++ b/static/templates/actors/character/tabs/biography.hbs
@@ -8,12 +8,20 @@
                 </a>
             {{/if}}
         </h3>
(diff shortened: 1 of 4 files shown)


static/templates/actors/character/tabs/biography.hbs
```diff
@@ -15,6 +15,17 @@
                 </div>
             </div>
         </section>
+
+        <section class="appearance-details{{#if editable}} editable{{/if}}">
```
stwlam, line 19: This shouldn't be an independent `section` but under the appearance one

CarlosFdez: Converted to grid then so that it fits nicely.

PR #10173, opened 2023-09-18
Fix aura rendering and detection for tiny tokens

CarlosFdez:
image

### DIFF
--- a/src/module/canvas/token/aura/renderer.ts
+++ b/src/module/canvas/token/aura/renderer.ts
@@ -39,17 +39,19 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
         this.token = params.token;
         this.appearance = params.appearance;
         this.radius = params.radius;
-        this.radiusPixels = 0.5 * this.token.w + (this.radius / (canvas.dimensions?.distance ?? 0)) * canvas.grid.size;
+        this.radiusPixels =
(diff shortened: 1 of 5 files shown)


src/module/canvas/token/aura/renderer.ts
```diff
@@ -72,6 +74,13 @@ class AuraRenderer extends PIXI.Graphics implements TokenAuraData {
 
     /** Draw the aura's border and texture */
     async draw(showBorder: boolean): Promise<void> {
+        // Offset if its a tiny token that isn't at the topleft position
+        if (this.token.actor?.size === "tiny") {
```
stwlam, line 78:
```suggestion
        if (this.token.document.width < 1) {
```

src/module/canvas/token/object.ts
```diff
@@ -70,6 +70,22 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
         return `Token.${this.id}`;
     }
 
+    /** Bounds used for mechanics, such as flanking and drawing auras */
+    get mechanicalBounds(): PIXI.Rectangle {
+        const bounds = super.bounds;
+        if (this.actor?.size === "tiny") {
```
stwlam, line 76:
```suggestion
        if (this.document.width < 1) {
```

PR #10177, opened 2023-09-18
Allow editing effects from effect panel tooltip

CarlosFdez:
A certain module (and previous PR by someone) did it as shift+click, but an actual clickable button is hopefully less controversial. It doesn't show for conditions unless the condition is persistent damage (which opens up the persistent damage dialog and highlights the relevant formula).

image


stwlam:
Swap for consistency with how they are on sheets

CarlosFdez:
That makes sense, though the reason I broke consistency here is that because there's no gap between text and icons, the icons on the right most side end up being the most visible ones, and send to chat is a more common operation.

But perhaps that visibility isn't by enough to matter.

stwlam:
Chat is also the one that will have public consequences on actual click.

CarlosFdez:
Done, and swapped to fixed width icons and changed the h1 to flex to remove whitespace

PR #10195, opened 2023-09-19
Build config objects with Remeda functions and eliminate instances of accumulator spreading in reduces

In3luki:
`deityDomains` only works with `mapToObject` because we don't need to return the original key.

### DIFF
--- a/src/scripts/config/index.ts
+++ b/src/scripts/config/index.ts
@@ -69,6 +69,7 @@ import {
     vehicleTraits,
     weaponTraits,
 } from "./traits.ts";
+import * as R from "remeda";
 
 export type StatusEffectIconTheme = "default" | "blackWhite";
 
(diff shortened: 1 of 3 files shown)


src/scripts/config/index.ts
```diff
@@ -260,17 +249,17 @@ const alignments: Record<Alignment, string> = {
     CE: "PF2E.AlignmentCE",
 };
 
-const deityDomains = Object.keys(enJSON.PF2E.Item.Deity.Domain).reduce((domains, key) => {
-    const slug = sluggify(key);
+const deityDomains = R.mapToObj(R.keys.strict(enJSON.PF2E.Item.Deity.Domain), (key) => {
+    const slug = sluggify(key) as DeityDomain;
     const casedKey = sluggify(key, { camel: "bactrian" });
```
stwlam, line 254:
```suggestion
const deityDomains = R.mapToObj(Object.keys(enJSON.PF2E.Item.Deity.Domain), (key) => {
    const label = `PF2E.Item.Deity.Domain.${key}.Label`;
    const description = `PF2E.Item.Deity.Domain.${key}.Description`;
    return [sluggify(key) as DeityDomain, { label, description }];
});
```

src/scripts/config/index.ts
```diff
@@ -260,17 +249,17 @@ const alignments: Record<Alignment, string> = {
...
-    return {
-        ...domains,
-        [slug]: {
+    return [
+        slug,
+        {
             label: `PF2E.Item.Deity.Domain.${casedKey}.Label`,
             description: `PF2E.Item.Deity.Domain.${casedKey}.Description`,
         },
```
stwlam, line 260:
```suggestion
```

src/scripts/config/index.ts
```diff
@@ -260,17 +249,17 @@ const alignments: Record<Alignment, string> = {
...
-    };
-}, {} as Record<DeityDomain, { label: string; description: string }>);
+    ];
+});
```
stwlam, line 262:
```suggestion
```
github made me split it up :)

In3luki:
Updated. Makes sense to use `Object.keys` if the slug is asserted anyway. Also much cleaner like this.

PR #10226, opened 2023-09-20
Add form for MultipleAttackPenalty RE

stwlam:


src/module/item/sheet/rule-element-form/base.ts
```diff
@@ -46,7 +46,7 @@ class RuleElementForm<
         this.index = options.index;
         this.rule = options.rule;
         this.object = options.object;
-        this.schema = RuleElements.all[String(this.rule.key)]?.schema ?? null;
+        this.schema = this.object?.constructor.schema ?? null;
```
CarlosFdez, line 49: object doesn't exist for unowned items :(

PR #10268, opened 2023-09-22
Display carried bulk and encumbrance on NPC sheet

### LINKED ISSUE #6427: NPC sheets do not display bulk values
image

NPCs do not currently track bulk, as seen above (Troubadour from the NPC Gallery).

When using the PF2e Workbench module, there is an option to apply encumbrance; I assume this implies that bulk can be tracked on NPCs sheets.

image
image

stwlam:
Closes #6427


src/styles/actor/npc/_index.scss
```diff
@@ -317,6 +313,11 @@ $faded-color: #7a7971;
                 }
             }
         }
+
+        @import "inventory";
+        @import "spells";
+        @import "effects";
+        @import "../red-action-boxes";
```
CarlosFdez, line 320: Was this moved to a deeper level of nesting for a technical reason?

stwlam: I'm not aware of other potential reasons, but yes: in this case it was because legacy styling was at higher specificity.

PR #10270, opened 2023-09-22
Finish TokenMarker RE, implement in Devise a Stratagem

stwlam:


CarlosFdez:
Is it possible for strategic strike to work if you remove the roll option? Unfortunately the effect is set to auto remove, which means its gone when damage occurs.

src/module/rules/rule-element/token-marker/rule-element.ts
```diff
@@ -0,0 +1,68 @@
...
+                : await new MarkTargetPrompt({ prompt: null, requirements: null }).resolveTarget());
+        if (!(token instanceof TokenDocumentPF2e)) {
+            // No token was targeted: abort creating item
+            pendingItems.splice(pendingItems.indexOf(itemSource), 1);
+            return;
+        }
+
+        this.#checkRuleSource(ruleSource);
+        this.uuid = ruleSource.uuid = token.uuid;
+        itemSource.name = `${itemSource.name} (${token.name})`;
```
In3luki, line 38: Does `token.name` respect the `tokenSetsNameVisibility` setting?

stwlam: I removed it and will follow up with some visual indication on tokens.

src/module/rules/rule-element/token-mark/rule-element.ts
```diff
@@ -0,0 +1,72 @@
...
+            (game.user.targets.size === 1
+                ? Array.from(game.user.targets)[0]!.document
+                : await new MarkTargetPrompt({ prompt: null, requirements: null }).resolveTarget());
+        if (!(token instanceof TokenDocumentPF2e)) {
+            // No token was targeted: abort creating item
+            pendingItems.splice(pendingItems.indexOf(itemSource), 1);
+            return;
+        }
+
+        this.#checkRuleSource(ruleSource);
```
CarlosFdez, line 41: Why is this check necessary? 👀

stwlam: Because otherwise the object has no `uuid` property.

CarlosFdez: That can cause a rule source to not have the key set? I'd assume this wouldn''t be executed if the key wasn't TokenMark already.

stwlam: It can't do that, no.

PR #10289, opened 2023-09-23
Fix parties being token size linked

CarlosFdez:
Its a bit odd migrating for this when we usually handle this by updating the form such that it resolves the field, but it be what it be.

### DIFF
--- a/src/module/migration/migrations/866-link-to-actor-size-again.ts
+++ b/src/module/migration/migrations/866-link-to-actor-size-again.ts
@@ -1,0 +1,30 @@
+import type { ActorPF2e } from "@actor";
+import type { ActorSourcePF2e } from "@actor/data/index.ts";
+import { MigrationBase } from "../base.ts";
+import { SIZE_LINKABLE_ACTOR_TYPES } from "@actor/values.ts";
+
+export class Migration866LinkToActorSizeAgain extends MigrationBase {
+    static override version = 0.866;
+
+    override async updateActor(actorSource: ActorSourcePF2e): Promise<void> {
+        if (SIZE_LINKABLE_ACTOR_TYPES.has(actorSource.type)) return;
+
+        if (actorSource.prototypeToken.flags.pf2e) {
+            actorSource.prototypeToken.flags.pf2e.linkToActorSize = false;
+            actorSource.prototypeToken.flags.pf2e.autoscale = false;
+        }
+    }
+
+    override async updateToken(tokenSource: foundry.documents.TokenSource, actor: ActorPF2e | null): Promise<void> {
+        if (!actor || SIZE_LINKABLE_ACTOR_TYPES.has(actor.type)) {
+            return;
+        }
+
+        mergeObject(tokenSource.flags, {
+            pf2e: {
+                linkToActorSize: false,
+                autoscale: false,
+            },
+        });
+    }
+}
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -77,7 +77,7 @@ import { StatisticModifier } from "./modifiers.ts";
 import { ActorSheetPF2e } from "./sheet/base.ts";
 import { ActorSpellcasting } from "./spellcasting.ts";
 import { TokenEffect } from "./token-effect.ts";
-import { CREATURE_ACTOR_TYPES, SAVE_TYPES, UNAFFECTED_TYPES } from "./values.ts";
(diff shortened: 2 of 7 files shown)


src/module/migration/migrations/866-link-to-actor-size-again.ts
```diff
@@ -0,0 +1,30 @@
...
+    override async updateActor(actorSource: ActorSourcePF2e): Promise<void> {
+        if (isActorTypeSizeLinkable(actorSource.type)) return;
+
+        if (actorSource.prototypeToken.flags.pf2e) {
+            actorSource.prototypeToken.flags.pf2e.linkToActorSize = false;
+            actorSource.prototypeToken.flags.pf2e.autoscale = false;
+        }
+    }
+
+    override async updateToken(tokenSource: foundry.documents.TokenSource): Promise<void> {
```
stwlam, line 17: `updateToken` will include the actor if you need it.

CarlosFdez: True.

src/module/migration/migrations/866-link-to-actor-size-again.ts
```diff
@@ -0,0 +1,30 @@
...
+    override async updateActor(actorSource: ActorSourcePF2e): Promise<void> {
+        if (isActorTypeSizeLinkable(actorSource.type)) return;
+
+        if (actorSource.prototypeToken.flags.pf2e) {
+            actorSource.prototypeToken.flags.pf2e.linkToActorSize = false;
+            actorSource.prototypeToken.flags.pf2e.autoscale = false;
+        }
+    }
+
+    override async updateToken(tokenSource: foundry.documents.TokenSource, actor: ActorPF2e): Promise<void> {
```
stwlam, line 18:
```suggestion
    override async updateToken(tokenSource: foundry.documents.TokenSource, actor: ActorPF2e | null): Promise<void> {
```

src/module/migration/migrations/866-link-to-actor-size-again.ts
```diff
@@ -0,0 +1,30 @@
...
+        if (!actor || !("type" in actor) || isActorTypeSizeLinkable(actor.type)) {
```
stwlam, line 19:
```suggestion
        if (!actor || isActorTypeSizeLinkable(actor.type)) {
```

src/module/migration/migrations/866-link-to-actor-size-again.ts
```diff
@@ -0,0 +1,30 @@
...
+            return;
+        }
+
+        mergeObject(tokenSource.flags, {
+            pf2e: {
+                linkToActorSize: false,
+                autoscale: false,
+            },
+        });
```
stwlam, line 28: This bypasses type checking of any kind

src/module/scene/helpers.ts
```diff
@@ -42,4 +44,9 @@ const checkAuras = foundry.utils.debounce(async function (this: ScenePF2e): Prom
     lock.release();
 }, 100);
 
-export { checkAuras };
+/** Returns true if the given actor type is valid for token size linking and autoscaling */
+function isActorTypeSizeLinkable(type: string): boolean {
+    return tupleHasValue(CREATURE_ACTOR_TYPES, type) || type === "vehicle";
+}
```
stwlam, line 50: A bit overwrought as a helper: can this just be a set (even if only a `Set<string>`)?

CarlosFdez: Can be, I'll swap it over.

src/module/migration/migrations/866-link-to-actor-size-again.ts
CarlosFdez: No longer necessary, you are correct. This was from when I tried to use fromUuid() and never fixed it after.

src/module/scene/helpers.ts
CarlosFdez: Its a set now. Since its a set it can go in values, so back to actor/values.ts it goes, avoiding the cyclical reference.

src/module/migration/migrations/866-link-to-actor-size-again.ts
CarlosFdez: Token sources don't have a type for pf2e flags, only the prep data does. There would be no type checking regardless unless we do that. Wasn't sure if its worth it for this but I'll try to figure that out I suppose.

Issue #10305, opened 2023-09-24
Missing RE: Spell Effect: Inspire Heroics (Defense, +2/+3)


Symon-S:
Unless I am mistaken, Inspire Heroics effects for Defense should also give the physical damage resistance that Inspire Defense gives.

![image](https://github.com/foundryvtt/pf2e/assets/81383360/0a967051-c0d7-4b35-9255-b52d6aed9405)

Intervencion:
#10310

PR #10310, opened 2023-09-24
Brushup Inspire Defense automation

Intervencion:
Unless I'm missunderstanding the Inspire Heroics, it's as simple as just adding the regular Inspire Defense resistance RE


TikaelSol:
Change the title to be in imperative case. PR titles should finish the sentence "If accepted this PR will ___" in a way that says what they do without referencing another issue. To reference an issue this would resolve we can say Closes #10305 

As for the PR itself this should all be condensed to a single effect and add the old effects to the list to migrate away.

PR #10315, opened 2023-09-24
Fix migration - 702 

reyzor1991:
Avoid situation when link to weapon is broken

Now ```@weapon.system.damage.dice``` convert to ```@actor.system.weapon.system.damage.dice```

actor.system.weapon — is invalid link


stwlam:
That migration is older than the data path apparently being affected. Is this occuring on attempting to rerun migrations on existing data?

reyzor1991:
It happens when import Item(Effect) from module to Items Tab or apply this effect with RE to actor

stwlam:
Importing migrates from the schema version recorded in the actor or item, which means the path would theoretically never be present.

stwlam:
This doesn't make sense except for someone trying to run it on already-corrupted data.

PR #10324, opened 2023-09-25
Standardize prefix for IDs in RE form

stwlam:
Also guarantee presence of `object` for actual forms


src/module/item/sheet/rule-element-form/base.ts
```diff
@@ -141,10 +152,10 @@ class RuleElementForm<
      * Helper to update the item with the new rule data.
      * This function exists because array updates in foundry are currently clunky
      */
-    updateItem(updates: Partial<TSource> | Record<string, unknown>): void {
+    async updateItem(updates: Partial<TSource> | Record<string, unknown>): Promise<void> {
         const rules: Record<string, unknown>[] = this.item.toObject().system.rules;
         rules[this.index] = mergeObject(this.rule, updates, { performDeletions: true });
-        this.item.update({ [`system.rules`]: rules });
+        this.item.update({ "system.rules": rules });
```
CarlosFdez, line 158:
```suggestion
        await this.item.update({ "system.rules": rules });
```

PR #10329, opened 2023-09-25
Fix kingdom resource collection giving free commodities

7H3LaughingMan:
If implemented this fixes the issue with the resource collection on kingdom upkeep granting free commodities. Previously it was setting the commodities to the maximum allowed value.


CarlosFdez:
TYVM for catching it.

Issue #10317, opened 2023-09-24
Bug: Gate Junction, Skill Junction for Air adding improperly


OtterDelight:
The gate skill junction for air is adding a +2 status bonus at level 5, and a +3 at level 10 to stealth instead of +1 and +2 as it should be

PR #10330, opened 2023-09-25
Fix air gate junction granting incorrect bonus to stealth

7H3LaughingMan:
If implemented this fixes the issue with the air gate junction granting the incorrect bonus to stealth if you select skill junction.


7H3LaughingMan:
This closes #10317

PR #10362, opened 2023-09-26
Fix aura re resort and tighten form reuse logic

CarlosFdez:


CarlosFdez:
The bugfix is only in the aura deep clone, everything else is an extra fix to what inluki did that fixes the root of the problem (but what he did is still important to avoid re-prep spam).

PR #10365, opened 2023-09-27
Create a toggleable button on character sheets to roll Recall Knowledge from proficiency tab skills

mysurvive:
This creates a toggleable button on player character sheets' proficiencies tab to roll Recall Knowledge from skills. This is not viewable by the player, only the GM.

- In the roll dialog:
- options to change simple or level-based DC is provided (if there is no user target)
- If a lore skill is rolled, an option to choose whether the lore is applicable or specific
- options to change the difficulty of the check and rarity of the creature on the fly

- When targeting a creature, it rolls a level-based check taking into account rarity of the creature
- When targeting a hazard, it rolls a level-based check taking into account rarity of the hazard
- When targeting nothing, defaults to DC 10 (the default setting in the roll dialog is an untrained simple check)

- If a skill with a physical attribute is rolled, the attribute is changed to int per CRB p 239

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -921,7 +921,11 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                 armorCheckPenalty.predicate.push({ nor: ["attack", "armor:ignore-check-penalty"] });
                 if (["acrobatics", "athletics"].includes(longForm)) {
                     armorCheckPenalty.predicate.push({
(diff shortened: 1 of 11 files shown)


mysurvive:
Jest seems to be having a hard time and I have no clue how to fix it. To test, I changed the path in the build error to the relative path, and that worked but it just threw another error that foundry isn't defined  in src/module/system/schema-data-fields.ts... but it's defined globally so I'm not sure what I've done to make this mess.

PR #10375, opened 2023-09-27
Treat 0 opacity auras as having null borders

CarlosFdez:
image


CarlosFdez:
This is a bit weird but this will need a change to the aura schema to make consistent
image

PR #10380, opened 2023-09-27
Add variants to Magic Missile

Maxi351:
Make it so Magic Missle spell allows you to choose 1,2 or 3-action variant rather than 1-action being the only one avaliable.
It also works with feats like dangerous sorcery or sorcer's blood magic
Zrzut ekranu 2023-09-27 155725

### DIFF
--- a/packs/spells/magic-missile.json
+++ b/packs/spells/magic-missile.json
@@ -45,11 +45,75 @@
         "materials": {
             "value": ""
         },
-        "overlays": {},
-        "prepared": {
+        "overlays": {
+            "MqVPEdlq857SaX8o": {
+                "_id": "MqVPEdlq857SaX8o",
+                "overlayType": "override",
+                "sort":  1,
+                "system": {
+                    "time": {
+                        "value": "1"
+                    },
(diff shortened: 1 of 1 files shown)


TikaelSol:
Issue is that magic missile can be split amongst targets. The PF2e Workbench has a macro that handles it, dangerous sorcery included.

Maxi351:
While it's true, the fact that a module makes it work, doesn't mean it should not be fixed, and most of the time all the missles go on the same target anyways. Maybe it isn't perfect, but it's still way better than what's right now at the core.

Maxi351:
> Issue is that magic missile can be split amongst targets. The PF2e Workbench has a macro that handles it, dangerous sorcery included.

While it's true, the fact that a module makes it work, doesn't mean it should not be fixed, and most of the time all the missles go on the same target anyways. Maybe it isn't perfect, but it's still way better than what's right now at the core.

TikaelSol:
We'll try to come up with something proper for this one, it's going to be such a unique case that it will likely just need hardcoding.

PR #10381, opened 2023-09-27
Add Glimpse Vulnerability Rules

7H3LaughingMan:
If implemented this adds a roll option that grants you +2 to strike damage. This is similar to how it was implemented for Implement's Empowerment.


TikaelSol:
Dealing 2 extra damage and gaining weakness 2 isn't the same thing. Implement's Empowerment isn't them gaining weakness, it's just extra damage.

PR #10453, opened 2023-10-01
Condense community repair effects

stevecambridge:
Removes spell effects `Community Repair (Critical Success)` and `Community Repair (Critical Failure)` and replaces them with a single ChoiceSet-based `Community Repair` spell effect.


TikaelSol:
We don't want to remove the old effects until we're ready to migrate away the links.

stevecambridge:
I've put them back. When and how does that migration happen?

Is the modified spell description okay to point toward the new effect, or should it be left as it was referencing the old effects?

TikaelSol:
A migration is something we write. With these condensed effects I want to do a bunch all at once then I'll write a migration to remove the existing links to the items from actors in people's worlds, replacing them with just the condensed effect if necessary. I just don't want to write a single migration for a single effect since we have so many of them to do. Better to just let the old effects sit until then.

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                    "community-repair-proximity",
+                    "community-repair-speed:-10"
+                ],
+                "selector": "saving-throw",
+                "text": "You cannot benefit from emotion effects.",
+                "title": "Community Repair"
+            }
+        ],
+        "source": {
+            "value": ""
```
TikaelSol, line 86:
```suggestion
            "value": "Pathfinder #170: Spoken on the Song Wind"
```

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                ],
+                "flag": "outcome",
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Prompt.DegreeOfSuccess",
+                "rollOption": "community-repair-speed"
+            },
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "community-repair-proximity"
```
TikaelSol, line 38:
```suggestion
                    "terrain:community-repair"
```
We prefer this kind of roll option, so while we're at the condensing we may as well clean this up a bit.

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                "key": "Note",
+                "outcome": [],
+                "predicate": [
+                    "community-repair-proximity",
+                    "community-repair-speed:-10"
+                ],
+                "selector": "attack",
+                "text": "You cannot benefit from emotion effects.",
+                "title": "Community Repair"
+            },
```
TikaelSol, line 72:
```suggestion
            {
                "key": "Note",
                "predicate": [
                    "terrain:community-repair",
                    "community-repair-speed:-10"
                ],
                "selector": [
                    "attack-roll",
                    "skill-check",
                    "perception",
                    "saving-throw"
                ],
                "text": "You cannot benefit from emotion effects.",
                "title": "{item|name}"
            }
```
We can localize the note string later but at least it's easy enough to remove one of the note rules now.

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                "key": "Note",
+                "outcome": [],
+                "predicate": [
+                    "community-repair-proximity",
+                    "community-repair-speed:-10"
+                ],
+                "selector": "saving-throw",
+                "text": "You cannot benefit from emotion effects.",
+                "title": "Community Repair"
+            }
```
TikaelSol, line 83:
```suggestion
```

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                ],
+                "selector": "speed",
+                "type": "status",
+                "value": "@item.flags.pf2e.rulesSelections.outcome"
+            },
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "Within 60 feet of a Community Repair Ritual",
+                "option": "community-repair-proximity",
```
TikaelSol, line 48:
```suggestion
                "option": "terrain:community-repair",
```

packs/spell-effects/spell-effect-community-repair.json
```diff
@@ -0,0 +1,100 @@
...
+                    "community-repair-proximity"
+                ],
+                "selector": "speed",
+                "type": "status",
+                "value": "@item.flags.pf2e.rulesSelections.outcome"
+            },
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "Within 60 feet of a Community Repair Ritual",
```
TikaelSol, line 47: Also need localized later.

Localization is done by adding a new key to the re-en.json file in the system's language files and then referring to that key here with `PF2E.SpecificRule.CommunityRepair.ProximityLabel` (or whatever we call it). The other one that needs localized is more general so it shouldn't go under community repair, since other things may re-use that language.

TikaelSol:
If you're wondering about the localization stuff, translation modules have difficulty accessing the text inside rule elements because they are stored in an array and don't have a convenient ID to reference like items on an actor do. So to help them out we move that text to the localization files so all they have to do it translate that file and all the rule elements are translated. We are still finding old unlocalized strings like these in the system and we fix them when they crop up.

stevecambridge:
I saw a bit about that. From what I was seeing in other PRs, you only need to add the text to the EN loc file and translators will pick it up from there?

Thanks for the comprehensive review. I'll watch out for these things when I tackle some of the other issues.

TikaelSol:
Yeah, we have multiple en.json files. For organization purposes we split them out. rule element stuff normally goes in re-en.json while core system things like traits, UI strings, and languages go in en.json.

PR #10473, opened 2023-10-02
Add support for alterations of several armor properties

stwlam:


src/module/item/physical/values.ts
```diff
@@ -8,6 +8,60 @@ const PHYSICAL_ITEM_TYPES = new Set([
...
+    "41",
+    "42",
+    "43",
+    "44",
+    "45",
+    "46",
+    "47",
+    "48",
+    "49",
+    "50",
```
CarlosFdez, line 61: This can be ...R.range(1, 51).map(String)

stwlam: I GUESS!

PR #10477, opened 2023-10-02
Fix missing emdash characters in Eternal Torch

### LINKED ISSUE #9103: Data entry errors from Rage of elements (several)
**Solar Detonation** - missing the word "Vitality" from the 2nd damage embedded in the description.

image

**Channel Elements** - typo in description, "fl w".  Also, "10-foot emanation" should not be an embedded template;  it's an aura that emanates from your token and moves with it.

image

**Aerial boomerang** - not exactly an error, but I think it would be nice to include a `@Template[type:line|distance:5]` (single square template) in the description, to let users keep track of the boomerang's square during the round after the ability is used.

**Eternal Torch** - in the sentence "All your torches-even ones you already created-shed bright light...", the dashes should be spaced out.  (I think this is a common issue with text that was copied from the pdf, e.g. in Four Winds)

**Flying Flame** - the button in the description doesn't work.  It's: `[[/r (floor(@actor.level/2)+1)d6)[fire]]]{leveled fire damage}`

TikaelSol:
Closes #9103


stwlam:
Those are actually good as hyphens: some style guides may recommend an en dash for phrasal adjectives, but I'm not aware of any that recommended em dashes. Is that what the book uses?

TikaelSol:
Yes, the book uses emdash
image

stwlam:
Ohh, that. Right, right!

PR #10497, opened 2023-10-03
Include priority in header for rule element forms

CarlosFdez:
image

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -1113,7 +1113,7 @@
                 "Opacity": "Opacity",
                 "Predicate": "Predicate",
                 "Primitive": "Primitive",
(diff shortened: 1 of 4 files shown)


static/lang/re-en.json
```diff
@@ -1113,6 +1113,7 @@
                 "Opacity": "Opacity",
                 "Predicate": "Predicate",
                 "Primitive": "Primitive",
+                "PriorityHint": "Priority determines the order rule elements are executed in. Lower values are executed first, and higher values are executed last.",
```
stwlam, line 1116:
```suggestion
                "PriorityHint": "Priority determines the order in which rule elements are executed. Rule elements with lower priority values are executed first.",
```

PR #10499, opened 2023-10-03
Prepare effect rule elements last

CarlosFdez:
This makes certain merge rules more predictable. Currently it means that aura merges won't succeed during testing but fail during the actual game.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -773,7 +773,8 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
     }
 
     protected prepareRuleElements(): RuleElementPF2e[] {
-        return this.items.contents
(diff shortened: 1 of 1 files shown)


src/module/actor/base.ts
```diff
@@ -773,7 +773,8 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
     }
 
     protected prepareRuleElements(): RuleElementPF2e[] {
-        return this.items.contents
+        // Ensure "temporary" items have their rule elements go last when priority is equal
+        return R.sortBy(this.items.contents, (i) => (i instanceof AbstractEffectPF2e ? 2 : 1))
```
stwlam, line 777:
```suggestion
        return R.sortBy(this.items.contents, (i) => i instanceof AbstractEffectPF2e)
```

CarlosFdez: I'm surprised that even works.

PR #10509, opened 2023-10-03
Use params.target in attack roll

this-gavagai:
The AttackRollParams type definition includes a `target` property, but at present this property is not used when calling `roll()` from actor.system.actions objects.

This PR evaluates params.target, making the attack roll more in line with the damage roll defined later in the same file.

https://github.com/foundryvtt/pf2e/blob/e2acfc98745bc2c5d3490937485b4621c4736fcb/src/module/actor/character/document.ts#L1596


stwlam:
Do the same for hazards/NPCs in `@actor/helpers.ts`

this-gavagai:
I believe it's updated to include hazards and NPCs. If I missed anything, kindly let me know.

PR #10541, opened 2023-10-05
Add system macro to create skill and save roll prompts in chat via party sheet and hotbar 

### LINKED ISSUE #10244: Feature Request: Add method for GM to create chat roll prompts for saves and skills
GMs regularly need to ask players to make various checks. The inline roll prompt button support in journals and other sheets is fantastic but requires preparation that isn't always possible, and there will always be some need for players to roll arbitrary skill checks or saves at times. It would be great to have some method for GMs to bring up a dialog that allows for adding a message to chat that prompts players to do a skill check or save, ideally with the ability to specify title or label, difficulty class, and to flag certain types of rolls (recall knowledge, etc.) so that rules elements could act on these rolls. Being able to select multiple skills to produce multiple prompts would be a bonus. Possibly this could take the form of a macro that can be placed on the hotbar but could also be accessed through a button somewhere in the system UI, such as on the party sheet or as a button on the actor tab (similar to the compendium browser button).

JulieWinchester:
Closes #10244.

This PR: 
* Adds a system GM macro that provides a UI interface to send skill or save roll prompts to chat, specifying specific skills (including lores), saves, DC, and roll options such as action type or traits. Lores can be suggested by pulling from characters in a party or specified custom by the GM. Suggested options for actions are pulled from the PF2E Actions compendium, but custom actions can be specified as well. Suggested options for Roll Options are currently pulling from `CONFIG.PF2E.actionTraits` (could be changed, suggestions welcome), but any freeform roll option can also be specified custom just like for lores and actions. A checkbox for "Secret Check" is provided - though a user could theoretically just enter "secret" in roll options, I think it's valuable to keep the checkbox because of how commonly GMs need secret checks. DCs can be specified arbitrarily or can be determined used Simple or Level-Based DCs.
* Adds a button to the party sheet exploration tab to launch the macro. When launched via the party sheet, the macro will suggest lores from characters in that party. If the system macro is launched not from the exploration tab, then characters from the system active party (if one exists) will be used for lore suggestions. 
* Adds a script macro to the PF2E Macros compendium that launches the system macro, allowing system macro to be placed on the hotbar.

Here is a video with the changes in action. I'll also include screenshots. I'm on the discord as JellyfishJail. 

Video: https://youtu.be/wgEMAATV3gk

Screen Shot 2023-10-04 at 8 09 27 PM
Screen Shot 2023-10-04 at 8 09 56 PM
Screen Shot 2023-10-04 at 8 10 03 PM
Screen Shot 2023-10-04 at 8 10 19 PM


CarlosFdez:
UI suggestion, but I think at this point you can remove the labels, and have a bed for Rest and a Die icon for skills.

You could also combine the buttons (using the + CSS selector to control borders). The mockup for the buttons in the figma is:
image

JulieWinchester:
@CarlosFdez I've removed the button labels on the party sheet exploration header as suggested, mainly by restoring the original CSS prior to this PR but with a couple of tweaks to better match the figma and ensure that hover box shadows don't get overlapped and hidden by surrounding buttons. Here are the new buttons:

Screen Shot 2023-10-05 at 1 47 21 PM

stwlam:
Could the additional DC details somehow fit in the main dialog? A dialog with a dialog is a little unwieldy.

JulieWinchester:
@stwlam Sure. What do you think about tabs on the DC section, like the tabs gif Skills vs Saves?

stwlam:
> @stwlam Sure. What do you think about tabs on the DC section, like the tabs gif Skills vs Saves?

That sounds promising!

JulieWinchester:
@stwlam This is done. DC sub-dialog removed and replaced with tabs:

Screen Shot 2023-10-08 at 9 40 15 PM
Screen Shot 2023-10-08 at 9 40 24 PM
Screen Shot 2023-10-08 at 9 40 31 PM

CarlosFdez:
Choose or Add custom Lores seems extraneous considering that we already know what lore skills are possible.

src/scripts/macros/skill-save-prompt/skill-save-prompt.ts
```diff
@@ -0,0 +1,283 @@
...
+                            <div class="form-group">
+                                <input id="save" name="prompt.save" placeholder="Choose Saving Throws" data-dtype="JSON"></input>
+                            </div>
+                            <div class="form-group">
+                                <label for="basic-save">Basic</label>
+                                <input id="basic-save" name="basic-save" type="checkbox" />
+                            </div>
+                        </section>
+                    </section>
+                </form>
```
CarlosFdez, line 107: This should be a template

src/module/actor/party/sheet.ts
```diff
@@ -433,6 +433,10 @@ class PartySheetPF2e extends ActorSheetPF2e<PartyPF2e> {
         htmlQuery(html, "[data-action=rest]")?.addEventListener("click", (event) => {
             game.pf2e.actions.restForTheNight({ event, actors: this.actor.members });
         });
+
+        htmlQuery(html, "[data-action=prompt]")?.addEventListener("click", () => {
+            game.pf2e.gm.skillSavePrompt({ actors: this.actor.members });
```
CarlosFdez, line 438: It seems actors isn't respected by the skillSavePrompt. It should either default to `game.actors.party.members` in the skill save prompt, or allow omitting actors altogether.

src/scripts/macros/skill-save-prompt/helpers.ts
```diff
@@ -0,0 +1,71 @@
+import { ActorPF2e, CharacterPF2e } from "@actor";
+import { calculateSimpleDC, DCAdjustment } from "@module/dc.ts";
+import { CharacterSkill } from "@actor/character/types.ts";
+import { PROFICIENCY_RANKS } from "@module/data.ts";
+
+const dcAdjustmentsLabel = new Map<DCAdjustment, string>([
```
CarlosFdez, line 6: This is already in src/module/dc.ts. Either try to find a way to use the functions in there, or export the structure that's already in there. Ideally the first one, the second one is last resort.

CarlosFdez:
Before the final iteration, try to make sure the strings are localized. It could be in `PF2E.Actor.Party.CheckPrompt` or something like that.

JulieWinchester:
@CarlosFdez the following changes have been made:
* SkillSavePrompt is now an Application subclass
* View HTML is in a template
* "Choose or Add Custom Lores" has been changed to "Choose Lores". If Lores are available to be chosen, freeform lores can no longer be specified (if no lores are listed, then freeform lores can still be added). 
* Strings are localized
* Removed dcAdjustmentsLabel, now that's populated by using the existing DC module
* Hopefully made it more apparent how SkillSavePrompt uses actors. If actors aren't provided in the options, the dialog falls back on the active party members. But if launched from a party, the list of lores is tailored to the members of the current party from which the dialog is launched. This means that different parties with different members see different but appropriate lists of lores.

src/scripts/macros/skill-save-prompt/skill-save-prompt.ts
```diff
@@ -0,0 +1,236 @@
...
+            id: "generate-skill-save-prompt",
+            template: "systems/pf2e/templates/gm/skill-save-prompt/skill-save-prompt.hbs",
+            title: game.i18n.localize("PF2E.Actor.Party.SkillSavePrompt.Title"),
+            width: 400,
+        };
+    }
+
+    override async getData(): Promise<SkillSavePromptDialogData> {
+        this.#actions = await getActions();
+        this.#lores = this.options.actors ? loreSkillsFromActors(this.options.actors) : loreSkillsFromActiveParty();
```
CarlosFdez, line 46:
```suggestion
        this.#lores = loreSkillsFromActors(this.options.actors ?? game.actors.party.members ?? []);
```

You can get rid of loreSkillsFromActiveParty by doing this.

CarlosFdez (approved):

PR #10555, opened 2023-10-05
Include persistent damage in Phantom Pain spell damage

LebombJames:


TikaelSol:
The issue with this being part of the spell damage is the success state doesn't take the persistent damage at all, but does take the initial damage.

PR #10557, opened 2023-10-05
Add Recall Knowledge StatisticCheck to Statistic class

mysurvive:
Adds a Recall Knowledge StatisticCheck to the Statistic class which can be retrieved with actor.skills["skillName"].recallKnowledge


stwlam:
Actions use statistics, but they don't have them.

PR #10559, opened 2023-10-05
Change combatant comparison to constructed object

### LINKED ISSUE #10494: `self:participant:own-turn` RollOption seemingly absent when targetting on attack rolls
I can't get `self:participant:own-turn` to consistently appear on a test actor's attack rolls that target a token, despite an encounter being active and it indeed being their turn. I was getting some inconsistent results, so I gave the actor a test feat containing a Note RE predicated on the roll option in question, just to speed things up for myself.

Reproduction steps:
1. Create an encounter with two actors. I used a PC and an NPC.
2. Roll initiatives, start the encounter and set it to the PC's turn.
3. Roll an attack roll (in my case a weapon attack) against the NPC
`self:participant:own-turn` should be absent from the Roll Options upon inspection.

Here is a video demonstration

https://github.com/foundryvtt/pf2e/assets/103401186/fea4f05e-593d-4e25-80e8-8411700e2063

Foundry V11 311, PF2e 5.6.2

mysurvive:
If merged, resolves #10494

It looks like the `self:participant:own-turn` comparison in the entries array in `createEncounterRollOptions()` is comparing a constructed and not-yet-constructed actor when targeting. I quickly did a console log to see what was happening:
```js
    console.log(encounter.combatant?.actor === actor);
    console.log(encounter.combatant?.actor, actor);
```
and got
image
The first pair (two combatants) of logs are when nothing is targeted, the second pair are when an enemy is targeted.

### DIFF
--- a/src/module/actor/helpers.ts
+++ b/src/module/actor/helpers.ts
@@ -213,7 +213,7 @@ function createEncounterRollOptions(actor: ActorPF2e): Record<string, boolean> {
             [`encounter:threat:${threat}`, !!threat],
             [`encounter:round:${encounter.round}`, true],
             [`encounter:turn:${Number(encounter.turn) + 1}`, true],
-            ["self:participant:own-turn", encounter.combatant?.actor === actor],
+            ["self:participant:own-turn", encounter.combatant === actor.combatant],
             [`self:participant:initiative:roll:${initiativeRoll}`, true],
             [`self:participant:initiative:rank:${initiativeRank}`, true],
             [`self:participant:initiative:stat:${initiativeStatistic}`, !!initiativeStatistic],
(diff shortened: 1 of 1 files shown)


src/module/actor/helpers.ts
```diff
@@ -213,7 +213,7 @@ function createEncounterRollOptions(actor: ActorPF2e): Record<string, boolean> {
             [`encounter:threat:${threat}`, !!threat],
             [`encounter:round:${encounter.round}`, true],
             [`encounter:turn:${Number(encounter.turn) + 1}`, true],
-            ["self:participant:own-turn", encounter.combatant?.actor === actor],
+            ["self:participant:own-turn", encounter.combatant === actor.combatant],
```
stwlam, line 216: This would be applied when both `combatant` properties are `null`

mysurvive:
Yeah, I can reproduce the issue myself 
image

src/module/actor/helpers.ts
stwlam, line 216:
```suggestion
            ["self:participant:own-turn", encounter.combatant === participant],
```

PR #10576, opened 2023-10-06
Move stamina and resolve out of attributes, refactor settings

stwlam:


src/module/migration/migrations/874-move-stamina-stuff.ts
```diff
@@ -0,0 +1,68 @@
...
+        if (["1", '"1"'].includes(staminaVariant?._source.value ?? "")) {
+            await game.settings.set("pf2e", "staminaVariant", true);
+        } else if (staminaVariant) {
+            await game.settings.set("pf2e", "staminaVariant", false);
+        }
+
+        // Aayyy, while we're at it:
+        const pwolVariant = game.settings.storage.get("world").find((s) => s.key === "pf2e.proficiencyVariant");
+        if (pwolVariant?._source.value === '"ProficiencyWithoutLevel"') {
+            await game.settings.set("pf2e", "staminaVariant", true);
```
CarlosFdez, line 54: Different setting 😊

src/module/system/settings/variant-rules.ts
```diff
@@ -58,17 +58,16 @@ const SETTINGS: Record<string, SettingRegistration> = {
...
+        type: Boolean,
+        onChange: () => {
+            resetActors(game.actors.filter((a) => a.type === "character"));
         },
     },
     proficiencyUntrainedModifier: {
         name: "PF2E.SETTINGS.Variant.UntrainedModifier.Name",
         hint: "PF2E.SETTINGS.Variant.UntrainedModifier.Hint",
-        default: 0,
+        default: -2,
```
CarlosFdez, line 70: Is this correct? Untrained just adds level + 0. Why -2?

stwlam: https://2e.aonprd.com/Rules.aspx?ID=1370

> The initial implementation is fairly straightforward: the proficiency bonus just becomes +2 for trained, +4 for expert, +6 for master, and +8 for legendary. We recommend giving an untrained character a –2 proficiency modifier instead of a +0 proficiency bonus.

CarlosFdez: Interesting. And these fields are only used in pwol. Sure!

PR #10600, opened 2023-10-07
Allow Thaumaturge Implements to accept homebrew

Cerapter:
As the title says.
Replaces the hardcoded ChoiceSet lists in every implement-related class feature with `other-tags` based ChoiceSet filters and ItemAlterations.

All implements now have the `thaumaturge-implement` other tag. Second and Third Implement by default will not pick the already picked implements because class features by default get a `maxTakeable` of 1. 

Adept implements get the `thaumaturge-implement-adept` other tag, and Second Adept checks for its (non-)presence.

Paragon implements get the `thaumaturge-implement-paragon` other tag.

`adept:XYZ` and `paragon:XYZ` rollOptions are thrown out as literally nothing checks for them (but could _theoretically_ be readded if I add a `RollOption` RE to every implement that forces it to add the relevant rollOption when it has the `thaumaturge-implement-adept` or `thaumaturge-implement-paragon` other tag).

**There is currently a bug that I'm unsure how to solve** where because ChoiceSets are ran first, if you set your level to a high enough point where you jump over multiple such ChoiceSets, you might have erroneous options -- for example, from level 0 to level 7, where you can correctly pick your two implements, but Implement Adept will be unable to detect them.

### DIFF
--- a/packs/classfeatures/amulet.json
+++ b/packs/classfeatures/amulet.json
@@ -29,6 +29,9 @@
             "value": "Pathfinder Dark Archive"
         },
         "traits": {
+            "otherTags": [
+                "thaumaturge-implement"
+            ],
             "rarity": "common",
             "value": [
                 "thaumaturge"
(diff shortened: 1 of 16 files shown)


Cerapter:
Depends on #10845 .

PR #10604, opened 2023-10-07
Compact `ClassicLevel` databases after creation

In3luki:
This reduced the uncompressed database size by ~57 MB. I don't know why this isn't happening by default on `close()`.


In3luki:
I've made a test release on my fork and the resulting zip file is 15 MB bigger than before. The unpacked /packs folder is 61 MB  vs. 119 MB  in master. So leaving the db uncompacted is actually better? 🙃

PR #10628, opened 2023-10-09
Draft: Convert embeddd items to references in `extractPack` if possible. Restore full items on world import

In3luki:
Supported item types that have a `sourceId` flag are compared to their source item. If the data is similar enough the item is replaced with item reference data.
References are resolved in `ClientDatabaseBackend#_getDocuments` prior to creating the actor.
Currently supports spells and physical items.

```ts
interface SpellReferenceSource extends ItemReferenceBaseSource {
    system?: {
        level: SpellSystemSource["level"];
        location: SpellSystemSource["location"];
    };
    type: "spell";
}

interface PhyiscalItemReferenceSource extends ItemReferenceBaseSource {
    system: {
        containerId?: PhysicalSystemSource["containerId"];
        equipped: PhysicalSystemSource["equipped"];
        quantity: PhysicalSystemSource["quantity"];
        usage: PhysicalSystemSource["usage"];
    };
    type: PhysicalItemType;
}
```

ToDo:
- Determine which `system` values are absolutely necessary per item type. This might be easier once we have full `DataModel` support.
- Thoroughly check all physical item type references for differences in required data.
- Try to unify migrations of `ItemSource` and `ItemReferenceSource`.

### DIFF
--- a/build/extract-packs.ts
+++ b/build/extract-packs.ts
@@ -17,6 +17,11 @@ const args = argv
                 type: "boolean",
                 default: false,
             })
+            .option("convertReferences", {
+                describe: "Converts items to references if possible.",
+                type: "boolean",
+                default: false,
+            })
             .option("logWarnings", {
                 describe: "Turns on logging out warnings about extracted data.",
                 type: "boolean",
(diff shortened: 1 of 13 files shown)


stwlam:
Spells are the main concern, and we want to keep them as references even after import.

In3luki:
Off the top of my head there are two main problems with keeping them as references:

- There is still no async workflow on world load to hook into to resolve the references
- References would either need to be a new item type or live somewhere else than `ActorSource#items` to pass data validation.

This approach would at least make it less tedious to update the items of compendium actors.

Another idea I just had while writing this would be to update actor items when a non-embedded item changes during `extractPacks`. I think it would be possible to use most of the code of this PR to that effect.
Collect all changed items and then check all actors in all compendiums if they own those items and merge the owned item with the updated data if a data comparison passes.

stwlam:
Spell preparation itself is asynchronous, so there is indeed a workflow already in place.

In3luki:
Oh it is? I'll take a look.

PR #10663, opened 2023-10-11
Add dialog to perform compendium migrations

CarlosFdez:
image

image

image

Also fixes compendium migration but since its not a released feature and the bug only occurred this cycle (?), we don't really need to add the bugfix tag.

### DIFF
--- a/src/module/apps/sidebar/compendium-directory.ts
+++ b/src/module/apps/sidebar/compendium-directory.ts
@@ -1,8 +1,8 @@
 import { ActorPF2e } from "@actor";
 import { ItemPF2e } from "@item";
-import { MigrationList, MigrationRunner } from "@module/migration/index.ts";
 import { ErrorPF2e, fontAwesomeIcon, htmlQuery } from "@util";
 import MiniSearch from "minisearch";
+import { CompendiumMigrationStatus } from "../compendium-migration-status.ts";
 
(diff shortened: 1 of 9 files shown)


CarlosFdez:
I attempted to do the `compendium.apps[this.appId] = this` thing to have the dialog auto-refresh but it seems that unlike documents, compendium apps are an array. So my attempt merely caused it crash.

src/module/apps/sidebar/compendium-directory.ts
```diff
@@ -63,27 +63,24 @@ class CompendiumDirectoryPF2e extends CompendiumDirectory {
...
-                callback: async ($li) => {
-                    const compendium = game.packs.get($li.data("pack"), { strict: true }) as CompendiumCollection<
-                        ActorPF2e<null> | ItemPF2e<null>
-                    >;
-                    const runner = new MigrationRunner(MigrationList.constructFromVersion(null));
-                    runner.runCompendiumMigration(compendium);
-                },
-            });
-        }
+        options.push({
```
stwlam, line 66: Can this be restricted to GMs?

CarlosFdez: It already is in the condition, unless you mean you want to move that out of the condition. Does getting your perms updated mid session force a refresh?

PR #10667, opened 2023-10-12
Condense effects for Corrosive Body and Fiery Body

stevecambridge:
- Condenses spell effects for Corrosive Body and Fiery Body with a ChoiceSet predicated on parent level.
- Adds description of 9th level heightened effects to both spell effects' descriptions, and links to Acid Splash in Corrosive Body's description.
- Adds damage entry for Corrosive Body and adds heightened damage for Fiery Body.
- Adds predicates preventing both spells from buffing their own damage.

### DIFF
--- a/packs/spell-effects/spell-effect-corrosive-body.json
+++ b/packs/spell-effects/spell-effect-corrosive-body.json
@@ -4,7 +4,7 @@
     "name": "Spell Effect: Corrosive Body",
     "system": {
         "description": {
-            "value": "<p>Granted by <em>@UUID[Compendium.pf2e.spells-srd.Item.Corrosive Body]</em></p>\n<p>You exhale corrosive gas, and acidic secretions coat your skin as you transform into a being of living acid. You gain acid immunity. Any creature that touches you or damages you with an unarmed melee attack or non-reach melee weapon takes @Damage[3d6[acid]] damage, and your unarmed attacks deal 1d4 additional acid damage. The first time each round that you deal acid damage to a creature in this way, you gain [[/r 3d6 #Temporary Hit Points]]{3d6 temporary Hit Points} as your body digests the eroded material and transforms it into a protective acid. When the spell ends, any remaining temporary HP expires as well. Your acid spells deal one additional die of acid damage (of the same damage die the spell uses). You can cast acid splash as an innate spell; the splash damage affects all creatures within 15 feet instead of the normal 5 feet.</p>"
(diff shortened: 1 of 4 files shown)


stevecambridge:
Addresses one task in #4144

packs/spell-effects/spell-effect-fiery-body.json
```diff
@@ -55,9 +55,42 @@
                 "key": "DamageDice",
                 "label": "Fiery Body",
                 "predicate": [
-                    "fire"
+                    "fire",
+                    {
+                        "not": "item:fiery-body"
+                    }
```
TikaelSol, line 61: Is this actually needed? I don't think the effect should be getting spell damage increases, though maybe the inlines are?

packs/spell-effects/spell-effect-fiery-body.json
```diff
@@ -35,7 +35,7 @@
             },
             {
                 "damageType": "fire",
-                "diceNumber": 1,
+                "diceNumber": "{item|flags.pf2e.rulesSelections.fieryBody.unarmedDice}",
```
TikaelSol, line 38:
```suggestion
                "diceNumber": "ternary(gte(@item.level,9),2,1)",
```

packs/spells/corrosive-body.json
```diff
@@ -18,17 +18,47 @@
...
+                "Pb1eT1upkmw6q6JG": {
+                    "applyMod": false,
+                    "type": {
+                        "categories": [],
+                        "subtype": "",
+                        "value": "acid"
+                    },
+                    "value": "3d6"
+                }
+            }
```
TikaelSol, line 31: I think I would rather just remove the damage roll from fiery body than add it here.

packs/spell-effects/spell-effect-corrosive-body.json
```diff
@@ -44,9 +44,42 @@
...
+                            "parent:level:9"
+                        ],
+                        "value": {
+                            "unarmedDice": 2
+                        }
+                    }
+                ],
+                "flag": "corrosiveBody",
+                "key": "ChoiceSet"
             }
```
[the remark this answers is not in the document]
TikaelSol: To explain, we don't need the choice sets here and this would fail for a 10th level casting of the spell anyways. This is more compactly done with a ternary in the dice number field instead.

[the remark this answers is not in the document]
stevecambridge: Woof completely forgot about 10th level spells. I'll fix with the ternary.

packs/spell-effects/spell-effect-fiery-body.json
stevecambridge: Yeah I was surprised to see but as it was before, the spell effect was buffing its own inline damage. Won't be an issue if we remove the damage rolls as you suggested so I'll get rid of this.

PR #10677, opened 2023-10-12
Have slots in `FeatGroup` contain their own child slots

stwlam:


src/module/actor/character/feats.ts
```diff
@@ -208,7 +208,7 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<FeatGroup<
             const location = feat.system.location ?? "";
             const group = groupsBySlot[location] ?? this.get(location) ?? this.get(feat.category);
             if (!group?.assignFeat(feat)) {
-                this.unorganized.feats.push({ feat });
+                this.unorganized.feats.push({ feat, children: this.unorganized.getChildSlots(feat) });
```
CarlosFdez, line 211: I think this can be `this.unorganized.assignFeat(feat);` now. That's what it is for kingdom at least.

PR #10681, opened 2023-10-12
Fix weapon on palzu

MSAbaddon:


packs/outlaws-of-alkenstar-bestiary/palzu.json
```diff
@@ -453,11 +448,6 @@
         },
         {
             "_id": "KKkWifgG7SHLOwVt",
-            "flags": {
-                "pf2e": {
-                    "linkedWeapon": "zyo1tEI1ErNEoww4"
-                }
-            },
```
TikaelSol, line 460: Is there a reason to remove the linked weapon?

MSAbaddon: Don't know about that - basically I deleted the weapon and added the item "Jax" from the compendium, because the item on Palzu was a piercing wind with data modified to Jax

PR #10686, opened 2023-10-12
Show actor level on limited actors, and make collapse all button collapse parties

LebombJames:
Pardon the double feature PR

### DIFF
--- a/src/module/apps/sidebar/actor-directory.ts
+++ b/src/module/apps/sidebar/actor-directory.ts
@@ -49,7 +49,7 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
         // Strip actor level from actors we lack proper observer permission for
         for (const element of htmlQueryAll(html, "li.directory-item.actor")) {
             const actor = game.actors.get(element.dataset.documentId ?? "");
-            if (!actor?.testUserPermission(game.user, "OBSERVER")) {
+            if (!actor?.testUserPermission(game.user, "LIMITED")) {
                 element.querySelector("span.actor-level")?.remove();
             }
         }
@@ -140,6 +140,25 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
             });
         }
 
+        // Extend core collapse all button to close folder-likes
+        const collapseAll = htmlQuery(html, ".collapse-all");
+        collapseAll?.addEventListener("click", (event) => {
+            this.collapseAll();
+
+            const folderEls = htmlQueryAll(html, ".folder-like");
+            for (const folderEl of folderEls) {
+                const entryId = folderEl?.dataset.entryId ?? "";
+                if (folderEl && entryId) {
+                    event.stopPropagation();
+                    this.extraFolders[entryId] = true;
+                    folderEl.classList.toggle("collapsed", true);
+                    if (this.popOut) this.setPosition();
+
+                    this.saveActivePartyFolderState();
+                }
+            }
+        });
+
         this.#appendBrowseButton(html);
     }
 
@@ -184,7 +203,7 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
         }
     }
 
-    /** Overriden to prevent highlighting of certain types of draggeed data (such as parties) */
+    /** Overriden to prevent highlighting of certain types of dragged data (such as parties) */
     protected override _onDragHighlight(event: DragEvent): void {
         if (event.type === "dragenter" && this.#draggingParty) {
             return event.stopPropagation();


stwlam:
Mmm, I dunno. We don't show that on limited NPC sheets. This would be revealing information that we don't already.

LebombJames:
Last commit excluded NPCs

src/module/apps/sidebar/actor-directory.ts
```diff
@@ -140,6 +141,25 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
...
+        // Extend core collapse all button to close folder-likes
+        const collapseAll = htmlQuery(html, ".collapse-all");
+        collapseAll?.addEventListener("click", (event) => {
+            this.collapseAll();
+
+            const folderEls = htmlQueryAll(html, ".folder-like");
+            for (const folderEl of folderEls) {
+                const entryId = folderEl?.dataset.entryId ?? "";
+                if (folderEl && entryId) {
+                    event.stopPropagation();
```
CarlosFdez, line 153: Pretty odd for this to be here rather than at the start of the function.

src/module/apps/sidebar/actor-directory.ts
```diff
@@ -140,6 +141,25 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
...
+
+            const folderEls = htmlQueryAll(html, ".folder-like");
+            for (const folderEl of folderEls) {
+                const entryId = folderEl?.dataset.entryId ?? "";
+                if (folderEl && entryId) {
+                    this.extraFolders[entryId] = true;
+                    folderEl.classList.toggle("collapsed", true);
+                    if (this.popOut) this.setPosition();
+
+                    this.saveActivePartyFolderState();
```
CarlosFdez, line 158: This should not be happening on every loop iteration

stwlam:
> Last commit excluded NPCs

PC-build NPCs?

LebombJames:
How does this look now?

src/module/apps/sidebar/actor-directory.ts
```diff
@@ -49,9 +49,13 @@ class ActorDirectoryPF2e extends ActorDirectory<ActorPF2e<null>> {
         // Strip actor level from actors we lack proper observer permission for
         for (const element of htmlQueryAll(html, "li.directory-item.actor")) {
             const actor = game.actors.get(element.dataset.documentId ?? "");
-            if (!actor?.testUserPermission(game.user, "OBSERVER")) {
-                element.querySelector("span.actor-level")?.remove();
-            }
+            const hideLevel = actor?.isOfType("character", "familiar", "vehicle")
+                ? !actor?.testUserPermission(game.user, "LIMITED")
+                : actor?.isOfType("npc")
+                ? actor.visible
```
CarlosFdez, line 55: wouldn't this hide npcs that the player has ownership over?

LebombJames: Doh, didn't test thoroughly enough

CarlosFdez: Now it hides stuff from GMs!

CarlosFdez: Honestly I'd probably end up doing `(actor?.isOfType("character", "familiar", "vehicle") && actor?.testUserPermission(game.user, "LIMITED")) || actor?.testUserPermission(game.user, "OBSERVER"))` but I don't know if that causes an issue with not-listed npcs, If it does, then you'd do

```ts
const hasPermission = (actor?.isOfType("character", "familiar", "vehicle") && actor.testUserPermission(game.user, "LIMITED")) || actor?.testUserPermission(game.user, "OBSERVER"));
if (hasPermission && actor?.visible) {
    // blah
}
```

accounting for prettier of course.

CarlosFdez: That said I don't know if its worth getting fiddly to see teammate levels. One might argue seeing the level for a player helps makes it faster to find your own PC on a skim.

LebombJames: As it is, it works fine for me. `testUserPermission` will always return true for a GM, which will set `hideLevel` to false, so it won't hide the level. It accounts for NPCs since Limited NPCs don't show level anyway... Although now I think about it we don't need that special check for NPCs; we can default to Observer for all other actors... which is the behavior I had 4 commits ago!

CarlosFdez: As its written now, it excludes hazard levels.

LebombJames:
Coming back to this, I think this should be fine now. PCs, hazards, vehicles and familiars show their level on Limited. NPCs show their level on Observer, but not Limited (linked vs unlinked on NPCs doesn't actually make a difference here, since the limited NPC sheet doesn't show the level, so it doesn't need to be shown on the sidebar).

PR #10711, opened 2023-10-14
Add size increase for heightened Righteous Might spell effect

### LINKED ISSUE #10651: Righteous Might Size Increase
The heightened (8th rank) version of Righteous Might does not increase the character's size. This just seems to be missing from the effect.

j-bs:
Fixes #10651


TikaelSol:
We are currently in a data entry freeze for CRB/APG data, due to the remaster. This will have to wait until we see if the spell is changed in the remaster.

PR #10721, opened 2023-10-14
Reframe skill/save prompt as "check prompt", make small refactors

stwlam:
- Use "check" instead of "skill/save" language in dialog
- Add perception
- Allow DC of 0
- Make all DOM ids globally unique
- Use foundry application API to setup tabs
- Keep dialog open after posting prompt


src/scripts/macros/check-prompt/prompt.ts
```diff
@@ -26,22 +26,27 @@ interface TagifyValue {
...
-            title: game.i18n.localize("PF2E.Actor.Party.SkillSavePrompt.Title"),
+            id: "generate-check-prompt",
+            tabs: [
+                { navSelector: ".skill-save-navigation", contentSelector: ".check-prompt-content", initial: "skills" },
+                { navSelector: ".dc-navigation", contentSelector: ".dc-content", initial: "set-dc" },
+            ],
+            template: "systems/pf2e/templates/gm/check-prompt.hbs",
+            title: game.i18n.localize("PF2E.Actor.Party.CheckPrompt.Title"),
             width: 400,
+            height: "auto",
```
JulieWinchester, line 45: Supplying the dialog height as auto here unfortunately leads to the dialog not correctly resizing when roll options are hidden or shown, leading to the form over- or under-flowing the dialog window. And then the resizing gets fixed if the dialog window is moved, or if the user switches from skills/perception to saves. Changing this to `auto !important` makes it resize correctly on initial dialog window load, but if you move the dialog around, the behavior persists. Setting the height as `auto !important` through CSS fixes these issues and makes resizing behavior consistent across the board.

stwlam: sad!

PR #10737, opened 2023-10-15
Avoid Whitespace wrap in tags

Far2Casual:
Small CSS touch to avoid undesred CSS wrapping of tags in case of whitespaces. In the french translation, this happens:
image

With the proposed change:
image


stwlam:
I think this may have some undesirable results with certain tag types.

PR #10751, opened 2023-10-15
Fix compendium browser publication source loading

### LINKED ISSUE #10750: Compendium Browser filtering is bugged
I believe this is because of how sources in items, etc... have changed, GMs can be affected if you toggle "Let GMs ignore filtering and see all entries" off.

In3luki:
I'm not sure if `"system.source.value"` is needed anymore so I left it in.

Closes #10750


stwlam:
It needs to handle legacy data, like in the individual tabs

In3luki:
Added a fix for text search as well. `storeFields` are derived index fields not item data so `publication` doesn't work there.

PR #10752, opened 2023-10-15
Move Take a Breather macro to system codebase, migrate existing macros

stwlam:


src/module/migration/migrations/878-take-a-breather.ts
```diff
@@ -0,0 +1,13 @@
+import type { MacroSource } from "types/foundry/common/documents/macro.d.ts";
+import { MigrationBase } from "../base.ts";
+
+/** Migrate "Take a Breather" macro to use function exposed at `game.pf2e.actions` */
+export class Migration878TakeABreather extends MigrationBase {
+    static override version = 0.878;
+
+    override async updateMacro(source: MacroSource): Promise<void> {
+        if (source.type === "script" && source.command.includes("system.attributes.resolve.value")) {
```
In3luki, line 9: Isn't that a bit too general? Something like ```source.command.includes("title: `Take a Breather`")``` would make sure no other custom macros are hit.

stwlam: That would miss babele-localized ones

In3luki: `console.log(resolve, sp);` is in there for some reason. That's probably unique enough?

stwlam: I can look for that

PR #10776, opened 2023-10-16
Implement spell references

In3luki:
### General
All new embedded spells are created as spell references if they come from a compendium or if they have a valid `sourceId` flag.
References are resolved the first time a character or npc sheet is openend and a cache of resolved spells is kept indefinitely.
Loading the compendium spells works as efficient as possible but it can still take some time for slower connections.
The following is Chrome throttled to slow 3G:

image

References should be indistinguishable from normal embedded items in all workflows unless I've missed some edge cases.

The references store minimal data on their spellcasting entry:
```ts
interface SpellReferenceSource {
    sourceId: ItemUUID;
    sort: number;
    system?: {
        location: Partial<SpellSystemSource["location"]>;
    };
}
```

They are stored at `SpellcastingEntryPF2e#system.reference` as `Record<string, SpellReferenceSource>` where the key is the item id of the created reference.

### UI

Reference item sheets can be viewed but editing is diabled:
image
image

Pressing the `Edit Item` button will show a warning dialog and, if confirmed, convert the reference to a normal embedded item:
image

Embedded spells have a `To Reference` button that is currently only visible in development. Maybe useful for everyone?
image

### DIFF
--- a/src/module/item/spellcasting-entry/collection.ts
+++ b/src/module/item/spellcasting-entry/collection.ts
@@ -5,6 +5,7 @@ import { ErrorPF2e, groupBy, ordinalString } from "@util";
 import { SlotKey } from "./data.ts";
 import { RitualSpellcasting } from "./rituals.ts";
 import { ActiveSpell, BaseSpellcastingEntry, SpellPrepEntry, SpellcastingSlotRank } from "./types.ts";
+import { SpellReferences } from "./references.ts";
 
 class SpellCollection<TActor extends ActorPF2e, TEntry extends BaseSpellcastingEntry<TActor | null>> extends Collection<
     SpellPF2e<TActor>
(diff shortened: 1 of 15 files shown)


src/module/item/spellcasting-entry/document.ts
```diff
@@ -370,6 +375,8 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
             throw ErrorPF2e("Spellcasting entries can only exist on characters and npcs");
         }
 
+        // Resolve spell references first
+        await this.references?.resolve();
```
CarlosFdez, line 379: No rush on this due to remaster delays, but to leave a note: this should probably be moved to the collection's getSpellData().

src/module/item/spellcasting-entry/collection.ts
```diff
@@ -88,17 +89,37 @@ class SpellCollection<TActor extends ActorPF2e, TEntry extends BaseSpellcastingE
             );
         }
 
-        const heightenedUpdate =
-            canHeighten && heightenedRank >= spell.baseRank
-                ? { "system.location.heightenedLevel": heightenedRank }
-                : {};
+        const isHeigthened = canHeighten && heightenedRank >= spell.baseRank;
```
fryguy1013, line 92: This variable is spelled wrong

PR #10803, opened 2023-10-18
Move weapon persistent damage to base-damage array

stwlam:


CarlosFdez (approved):
Might be best to delay the merge but this looks fine.

PR #10804, opened 2023-10-18
Prevent party error when familiars fail data prep

CarlosFdez:


src/module/actor/creature/document.ts
```diff
@@ -262,6 +262,7 @@ abstract class CreaturePF2e<
         attributes.flanking.flankable = true;
         attributes.flanking.offGuardable = true;
         attributes.reach = { base: 0, manipulate: 0 };
+        attributes.speed = mergeObject({ total: 0, speed: 0 }, attributes.speed ?? {});
```
stwlam, line 265:
```suggestion
        attributes.speed = mergeObject({ total: 0, value: 0 }, attributes.speed ?? {});
```

PR #10806, opened 2023-10-18
Add support for alternative critical specialization damage

stwlam:
Depends on #10805

### DIFF
--- a/src/module/rules/rule-element/crit-spec.ts
+++ b/src/module/rules/rule-element/crit-spec.ts
@@ -1,11 +1,14 @@
 import { AutomaticBonusProgression as ABP } from "@actor/character/automatic-bonus-progression.ts";
 import { ActorType } from "@actor/data/index.ts";
-import { DamageDicePF2e, ModifierPF2e } from "@actor/modifiers.ts";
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/crit-spec.ts
```diff
@@ -1,29 +1,85 @@
...
 
     static override validateJoint(data: SourceFromSchema<CritSpecRuleSchema>): void {
         super.validateJoint(data);
-        if (data.alternate && !data.text) {
-            throw Error("  alternate: must also include text");
+
+        if (data.alternate && !data.text && !data.damageDice && !data.modifier) {
+            throw Error("  alternate: must also include at least one of text, damage dice, or modifier");
+        } else if (!data.alternate && (data.text || data.damageDice || data.modifier)) {
+            const badProperty = data.text ? "text" : data.damageDice ? "damageDice" : data.modifier;
```
CarlosFdez, line 81: Final ternary case probably shouldn't be data.modifier but rather "modifier".

Why not `const badProperty = ["text", "damageDice", "modifier"].find((p) => !!data[p])`?

PR #10811, opened 2023-10-18
Fix ancestry featgroup not being able to filter for homebrew ancestries

Cerapter:
Simple enough: the ancestry featgroup didn't account for the `hb_` prepended string for homebrew ancestries (and heritages), so when you tried to filter for those using the magnifying glass icons next to the ancestry feat slots, you'd get shown all ancestries' feats. 
Used the same logic as the class featgroup uses, then lifted out the now-common logic.

An alternate option would have been to modify either the slug or the filling up of `sameAs`, but neither felt like safe options. 

The previous logic would have worked if the homebrew ancestry / versatile heritage slugs started with `hb_`, but this isn't communicated anywhere, and honestly should not be expected considering no other feat or feature would need it.
So I think this is the most reasonable approach to fixing the issue.

### DIFF
--- a/src/module/actor/character/feats.ts
+++ b/src/module/actor/character/feats.ts
@@ -32,11 +32,10 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<FeatGroup<
 
         // Find every ancestry and versatile heritage the actor counts as, then get all the traits that match them,
         // falling back to homebrew
-        const ancestryTraitsFilter = (
+        const ancestryTraitsFilter =
             actor.system.details.ancestry?.countsAs
                 .map((t) => getVanillaOrHomebrewTrait(t))
-                .filter((e): e is Exclude<typeof e, null> => e !== null) ?? []
-        ).map((x) => `traits-${x}`);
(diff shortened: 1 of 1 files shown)


src/module/actor/character/feats.ts
```diff
@@ -30,22 +30,25 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<FeatGroup<
             supported: ["classfeature"],
         });
 
+        // Find every ancestry and versatile heritage the actor counts as, then get all the traits that match them,
+        // falling back to homebrew
+        const ancestryTraitsFilter = (
+            actor.system.details.ancestry?.countsAs
+                .map((t) => getVanillaOrHomebrewTrait(t))
+                .filter((e): e is Exclude<typeof e, null> => e !== null) ?? []
+        ).map((x) => `traits-${x}`);
```
stwlam, line 39:
```suggestion
        const ancestryTraitsFilter =
            actor.system.details.ancestry?.countsAs
                .map((t) => getVanillaOrHomebrewTrait(t))
                .flatMap((t) => (t ? `traits-${t}` : [])) ?? [];
```

PR #10824, opened 2023-10-19
Add round-end expiry option for effects

stwlam:


src/module/item/abstract-effect/helpers.ts
```diff
@@ -30,10 +31,17 @@ export function calculateRemainingDuration(
 
         // A familiar won't be represented in the encounter tracker: use the master in its place
         const fightyActor = effect.actor?.isOfType("familiar") ? effect.actor.master ?? effect.actor : effect.actor;
-        const isEffectTurnStart =
+        const atTurnStart = () =>
```
CarlosFdez, line 34: Is it expensive to run what's in this function? It looks fairly straightforward, but perhaps effect.origin is expensive?

PR #10836, opened 2023-10-20
Data updates for Vegetable Lamb and Falling Stars

TiloBuechsenschuss:
In my recent games I noticed that two items had data issues.

Vegetable Lamb now has correct Strikes.
Falling Stars variants now should work correctly for all variants, not just Airburst.

### DIFF
--- a/packs/spells/falling-stars.json
+++ b/packs/spells/falling-stars.json
@@ -71,11 +71,24 @@
                 "system": {
                     "damage": {
                         "value": {
-                            "i7p7fmqd0A8a64Oz": {
+                            "5hgRJg7CfV9pZSUa": {
                                 "type": {
                                     "value": "electricity"
                                 }
+                            },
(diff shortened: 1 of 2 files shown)


packs/spells/falling-stars.json
```diff
@@ -118,13 +131,20 @@
...
-                            "i7p7fmqd0A8a64Oz": {
-                                "applyMod": false,
-                                "type": {
-                                    "subtype": "",
-                                    "value": "cold"
+                            "value": {
+                                "5hgRJg7CfV9pZSUa": {
+                                    "type": {
+                                        "value": "cold"
+                                    }
```
TikaelSol, line 138: This looks malformed, it would lead to a `value.value`

TiloBuechsenschuss: You're right, fixed that. Thanks for double checking!

PR #10844, opened 2023-10-21
Use core tooltip API for pair of tooltips in attribute builder

stwlam:
Also one in PC sheet


static/templates/actors/character/attribute-builder.hbs
```diff
@@ -122,7 +119,7 @@
         <section class="row{{#if manual}} not-eligible{{/if}}">
             {{#if class}}
                 <div class="row-heading">
-                    <img class="" src="{{class.img}}" title="{{ancestry.name}}" width="32" height="32" loading="lazy" />
+                    <img src="{{class.img}}" alt="{{ancestry.name}}" loading="lazy" />
```
CarlosFdez, line 122:
```suggestion
                    <img src="{{class.img}}" alt="{{class.name}}" loading="lazy" />
```

PR #10846, opened 2023-10-21
Fix Class image `title` attribute in Attribute Boosts dialog

j-bs:
Noticed while updating my character that the class icon title was displaying the Ancestry name instead of the Class name:

image


stwlam:
Superseded by #10844

PR #10880, opened 2023-10-23
Update Kingmaker features data entry for branch parity

silvative:
This won't be the "final" update to these, but most notably it refolders them and includes traits for the various tactics. The intent of this PR is mostly to avoid divergences in the data entry by avoiding having two versions of the compendium.


silvative:
I think I did it correctly, but the build process is failing because it doesn't include the other files. Should I add the sheet to this?

CarlosFdez:
Data Entry question, but if/once everything else has folders, how deep do we want to nest them? 

If we do nest this deep, kingdom/kingdom-features should probably just be kingdom/features but I don't know if we should make the user open up two folders when the max number of folders will be around 6 anyways. Definitely a data entry question.

CarlosFdez:
I really do think it should be kept to top level folders, otherwise the PR is fine by me.

PR #10881, opened 2023-10-23
Create an Army document type (No sheet)

silvative:


silvative:
Hmm, I think Git is recognising the moved files as new files instead of updated files, which is why it says there's an ID conflict. What's the best way to resolve this?

static/lang/kingmaker-en.json
```diff
@@ -319,5 +319,141 @@
...
+                "warning": "Warning: The old statistics will be permanently overwritten.",
+                "parametersHeader": "Parameters",
+                "levelLabel": "Level: ",
+                "saveLabel": "Strong save: ",
+                "confirmButton": "Update Stats",
+                "levelUpTitle": "Level Up Army",
+                "levelUpDesc": "Increases Army level by 1, updating stats using the default values as defined in the Warfare rules. New statistics will be dependent on the level. Gear, tactics, actions, and traits will not be replaced."
+            }
+        },
+        "Traits": {
```
CarlosFdez, line 383: Move these traits to `PF2E.Kingmaker.Trait`, and create a new `PF2E.Kingmaker.TraitDescription` object. `armyDescription` would then be `PF2E.Kingmaker.TraitDescription.army` instead.

CarlosFdez:
Doing the ranks the way listed is pretty verbose and I don't think its necessary. Instead make the ranks a number property listing the number of ranks, and the rank idx to fill `PF2E.Warfare.Gear.Weapons.rank${idx}.name` in the sheet itself.

src/module/actor/army/data.ts
```diff
@@ -0,0 +1,117 @@
...
+    BaseActorSourcePF2e,
+    BaseHitPointsSource,
+} from "@actor/data/base.ts";
+import { ARMY_TYPES } from "./values.ts";
+import { ActorSizePF2e } from "@actor/data/size.ts";
+
+type ArmySource = BaseActorSourcePF2e<"army", ArmySystemSource>;
+
+interface ArmySystemSource extends ActorSystemSource {
+    attributes: ArmyAttributesSource;
```
CarlosFdez, line 21: Attribute is on the chopping block because of remaster, though I don't know what @stwlam plans to do yet.

stwlam: We still need to get everything out of there, so it's fine for now.

PR #10894, opened 2023-10-24
Kill jquery in migration summary dialog

stwlam:


src/module/apps/migration-summary.ts
```diff
@@ -66,39 +66,47 @@ export class MigrationSummary extends Application<MigrationSummaryOptions> {
...
-                item:
-                    game.items.size > 0
-                        ? Math.min(...game.items.map((a) => a.schemaVersion ?? 0))
-                        : LATEST_SCHEMA_VERSION,
-            };
-            const lowestSchemaVersion = Math.max(
-                Math.min(lowestVersions.actor, lowestVersions.item),
-                RECOMMENDED_SAFE_VERSION
-            );
+        for (const button of html.querySelectorAll("button")) {
```
CarlosFdez, line 71: Wouldn't you be able to avoid a layer of indentation if you query for `button[data-action=remigrate]`? What's the performance difference in doing it this way?

PR #10897, opened 2023-10-24
Fix a missing info in Sorcerer archetype

rectulo:


TikaelSol:
Not likely to conflict with work on Player Core 1 but archetype journals will get a lot of revisions in 5.9, even in places you may not expect.

PR #10912, opened 2023-10-25
Makes Overdrive Ally an action not a passive

Intervencion:
discord


stwlam:
It needs the action cost set as well.

PR #10942, opened 2023-10-27
Remove superfluous `PackLoader#loadedPacks` cache

### LINKED ISSUE #10928: Hazards Missing in Compendium Browser
compendium problems
In current version of PF2e (5.7.4) on latest Foundry Build (313), majority of the sources for Hazards went missing in Compendium Browser, but were still accessible in Compendium Pack search. Issue persisted in older migrated worlds (still V11) with and without modules until I created new world, which fixed Compendium Browser issue.

In3luki:
~~This is more of a workaround than a fix. The problem is that hazards and npcs live in the same packs but need different index fields and indexing npcs with added hazard fields would brick the bestiary tab checks.~~

~~This change ensures that either the hazard or the bestiary tab can be openeded first and the other one will still get fresh index data.~~

Closes #10928

### DIFF
--- a/src/module/apps/compendium-browser/loader.ts
+++ b/src/module/apps/compendium-browser/loader.ts
@@ -3,11 +3,6 @@ import { localizer, sluggify } from "@util";
 import { CompendiumBrowserSources } from "./index.ts";
 
 class PackLoader {
-    loadedPacks: {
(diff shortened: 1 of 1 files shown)


In3luki:
After further inspection it turns out the only thing the cache in `PackLoader` did was brick hazards/nps. 🙃

PR #10956, opened 2023-10-28
Migrate Devise a Stratagem and related feats to use token mark

stwlam:
This also removes the totm toggle, which doesn't really work alongside a token mark


TikaelSol (approved):

CarlosFdez:
What if the totm toggle adds `target:mark:devise-a-stratagem`?

stwlam:
> What if the totm toggle adds `target:mark:devise-a-stratagem`?

I'd be more interested in officially dropping support for such toggles.

PR #10994, opened 2023-10-30
Add support for designating damage instances as being healing

stwlam:
Just laying some groundwork; not doing much with it yet.

image

### DIFF
--- a/src/module/system/damage/roll.ts
+++ b/src/module/system/damage/roll.ts
@@ -144,6 +144,11 @@ class DamageRoll extends AbstractDamageRoll {
         }
     }
 
+    get pool(): InstancePool | null {
+        const firstTerm = this.terms.at(0);
+        return firstTerm instanceof InstancePool ? firstTerm : null;
+    }
+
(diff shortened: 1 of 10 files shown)


src/module/system/damage/roll.ts
```diff
@@ -249,6 +259,7 @@ class DamageRoll extends AbstractDamageRoll {
             total: isPrivate ? "?" : Math.floor((this.total! * 100) / 100),
             increasedFrom: this.options.increasedFrom,
             splashOnly: !!this.options.splashOnly,
+            healingOnly: !this.kinds.has("damage"),
```
CarlosFdez, line 262: Might need to test if it has healing as well, in case there are no instances somehow?

stwlam: Something would have gone so catastrophically wrong in that case that healing/damage would be the least of the problems.

CarlosFdez: There's even a guard against that earlier, huh.

PR #11026, opened 2023-10-31
Prune and reorganize several areas of spell data

stwlam:


src/module/item/spell/document.ts
```diff
@@ -789,25 +765,27 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
...
+                    : null,
+                systemData.target.value
+                    ? `${game.i18n.localize("PF2E.SpellTargetLabel")}: ${systemData.target.value}`
+                    : null,
+                area ? game.i18n.format("PF2E.SpellArea", { area: area.label }) : null,
+                systemData.time.value ? `${game.i18n.localize("PF2E.SpellTimeLabel")}: ${systemData.time.value}` : null,
+                systemData.duration.value
+                    ? `${game.i18n.localize("PF2E.SpellDurationLabel")}: ${systemData.duration.value}`
+                    : null,
+            ].filter((p): p is string => p !== null),
```
CarlosFdez, line 787: Why is this filter required? Isn't that the purpose of compact?

stwlam: That part isn't new

CarlosFdez: It wasnt, but i removed it recently. This PR adds it back.

Might have come about from resolving conflicts then.

src/module/migration/migrations/882-spell-data-reorganization.ts
```diff
@@ -0,0 +1,270 @@
...
+
+        // Flatten traditions object
+        if (isObject(system.traditions) && "value" in system.traditions && Array.isArray(system.traditions.value)) {
+            system.traits = this.#ensureTraitsPresence(system);
+            system.traits.traditions = [...system.traditions.value].sort();
+        }
+        if ("traditions" in system) system["-=traditions"] = null;
+
+        // Add focus trait if a focus spell
+        if (system.category?.value === "focus") {
```
CarlosFdez, line 37: Will need to check that its not a cantrip. This will convert any homebrew focus cantrip that don't have a class name into regular spells, but I don't think that's avoidable. Hopefully its a small number.

src/module/item/spell/document.ts
```diff
@@ -152,19 +142,19 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
...
+        return this.traits.has("attack");
     }
 
     get isCantrip(): boolean {
         return this.traits.has("cantrip") && !this.isRitual;
     }
 
     get isFocusSpell(): boolean {
-        return this.system.category.value === "focus";
+        return this.traits.has("focus");
```
CarlosFdez, line 153: Needs to also check for class traits and/or special focus spell traits.

src/module/item/spell/document.ts
```diff
@@ -152,19 +142,19 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
...
     }
 
     get isRitual(): boolean {
-        return this.system.category.value === "ritual";
+        return !!this.system.ritual;
```
CarlosFdez, line 157: Is it possible in the new system for a spell to be a focus ritual? Should we relax some of the drag/drop restrictions in that case?

src/module/apps/compendium-browser/tabs/spell.ts
```diff
@@ -41,11 +29,11 @@ export class CompendiumBrowserSpellTab extends CompendiumBrowserTab {
         const indexFields = [
             "img",
             "system.level.value",
-            "system.category.value",
-            "system.traditions.value",
```
CarlosFdez, line 45: Given that category filters are completely gone, I think the only thing you need to try to make backwards compatible here is traditions. It should be doable, but we do have compendium migrations now so I don't think its necessary anymore. You can resolve this if that's your angle.

We should probably start saving min versions for each tab so that the compendium browser shows warnings in the sources tab at some point.

CarlosFdez:
Should be fine since focus cantrips have the focus trait, and that leads to us not losing any data. Easy enough to back out on that one property (or handle it a different way).

PR #11035, opened 2023-11-01
Add "not" button to Compendium Browser filter tags

In3luki:
No functional change just better UI:

Recording 2023-11-01 at 13 29 35

### DIFF
--- a/src/module/apps/compendium-browser/index.ts
+++ b/src/module/apps/compendium-browser/index.ts
@@ -4,7 +4,19 @@ import { ActionType, ItemType } from "@item/base/data/index.ts";
 import { PHYSICAL_ITEM_TYPES } from "@item/physical/values.ts";
 import { BaseSpellcastingEntry } from "@item/spellcasting-entry/index.ts";
 import type { UserPF2e } from "@module/user/document.ts";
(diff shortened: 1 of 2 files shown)


src/module/apps/compendium-browser/index.ts
```diff
@@ -596,13 +608,15 @@ class CompendiumBrowser extends Application {
...
-                            const current = !!selected.not;
-                            selected.not = !current;
-                            this.render();
+                        const action = htmlClosest(target, "[data-action]")?.dataset?.action;
+                        if (action === "toggle-not") {
+                            const value = event.detail.data.value;
+                            const selected = data.selected.find((s) => s.value === value);
+                            if (selected) {
+                                const current = !!selected.not;
+                                selected.not = !current;
```
stwlam, line 617: what in ternation

stwlam: Is this `selected.not = !selected.not;`?

In3luki: > Is this `selected.not = !selected.not;`?

Ehhh, probably? I saw that but assumed I did it that way for a reason. I'll see if anything breaks.

In3luki: Don't know what happended there. It's gone now. You saw nothing.

PR #11043, opened 2023-11-02
Fix creating and editing spell area for fixed heightening

CarlosFdez:
This fixes the issue where creating an area for a heightened spell where the base spell doesn't have an area would be impossible to edit or remove.

### DIFF
--- a/src/module/item/spell/sheet.ts
+++ b/src/module/item/spell/sheet.ts
@@ -206,23 +206,13 @@ export class SpellSheetPF2e extends ItemSheetPF2e<SpellPF2e> {
                 const property = addProperty.dataset.property;
                 if (!overlay.system || !property || property in overlay.system) continue;
                 addProperty.addEventListener("click", () => {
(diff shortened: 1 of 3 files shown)


src/module/item/spell/sheet.ts
```diff
@@ -393,6 +381,30 @@ export class SpellSheetPF2e extends ItemSheetPF2e<SpellPF2e> {
...
+    #getDefaultProperty(property: string): unknown {
+        const scaling = this.item.getHeightenLayers().reverse();
+        const baseValue = (() => {
+            for (const entry of [...scaling, { system: this.item.system }]) {
+                if (objectHasKey(entry.system, property)) {
+                    return entry.system[property];
+                }
+            }
+
+            return undefined;
```
stwlam, line 398: Any reason this can't be `return null`?

CarlosFdez: Null is theoretically a valid value, but I had to throw an exception instead.

PR #11047, opened 2023-11-02
Shore up reliability of not applying effects from hidden tokens

stwlam:


src/module/scene/token-document/aura/index.ts
```diff
@@ -69,14 +68,18 @@ class TokenAura implements TokenAuraData {
...
     /** Does this aura overlap with (at least part of) a token? */
     containsToken(token: TokenDocumentPF2e): boolean {
-        // 1. If the token is the one emitting the aura, return true early
+        // If either token is hidden or not rendered, return false early
+        if (this.token.hidden || token.hidden || !this.token.object || !token.object) {
+            return false;
+        }
+
+        // If the token is the one emitting the aura, return true early
         if (token === this.token) return true;
```
CarlosFdez, line 77: I suspect this should probably be before the hidden checks. It could be a self buffing aura that the monster might take advantage of while sneaky sneaky.

stwlam: The entire point of this is to transition away from GM-hidden being used as a mechanical tool.

PR #11081, opened 2023-11-03
Fix broken journal links on GM Screen - ABP

Intervencion:
Also fixed (and updated) on data entry if you just want to wait for that.


TikaelSol:
Closing since this journal will get hit by the remaster heavily.

Issue #11066, opened 2023-11-02
Request to make the damage preview dialog toggleable independent of the d20 preview dialog.


Joshmvii7:
Just requesting to have these two preview dialogs be able to be turned on and off separate from one another.

PR #11089, opened 2023-11-04
Add setting to skip damage modifiers dialog

jfn4th:
Completes #11066 

- Adds a new user setting `showDamageDialogs` and renames the existing `showRollDialogs` to the more distinct `showCheckDialogs`.
- Adds a new parameter `rollType` to `eventToRollParams` in order to determine which user setting to check.  Updates all existing calls to pass in either `"Check"` or `"Damage"`.


CarlosFdez:
The concern I have, and why I wanted to hold off a bit, is for what reason do people want there to be a separate skip damage option from the check prompt. Most people seem to disable the check prompt by default as well, and the current design is with the assumption that people who show the check dialog by the default are a minority.

Personally, I'm suspecting one of the following is going on, and was waiting for more data:
1) Its new so it sucks
2) There are characters that might require a lot of tweaks to be made to checks (assurance) but not damage, and wanting check/damage dialogs might not be fully due to the personality of each people.
3) There aren't conditional things that use the damage dialog *yet*. Guidance requires the check dialog, but power attack is currently a toggle. If things like power attack require you to open the damage dialog, I suspect opinions will change.

If its 1 or 3 I'd like to hold off, but I can see a case for 2 suggesting the option exists sooner. That said the fact this is coming up is a surprise because I've been constantly told that "no one uses the check dialog".

stwlam:
Dialogs are inherently annoying, so it's not surprising people want it disabled. It doesn't take any further pondering.

CarlosFdez:
But this request is not about disabling it. Its about disabling it while keeping the check one visible. That's slightly different.

stwlam:
I don't think we're going to see a flood of toggle-less damage modifiers expected to be enabled in the dialog.

CarlosFdez:
Regardless more configuration can only be a good thing when the toggle exists on a per dialog basis, and in defined places.

The function signature should be `eventToRollParams(params.event, { type: "damage" })` though.

jfn4th:
Signature is updated, though I'm still passing `rollType` first or else we get eslint error `A required parameter cannot follow an optional parameter.`

Since in `eventToRollParams` I was dynamically accessing the user setting via the following,
```typescript
const key = `show${rollType.type}Dialogs`;
const skipDefault = !game.user.settings[key as keyof UserSettingsPF2e];
``` 
I've updated to change the first letter of the passed in `rollType` to uppercase.
```typescript
const key = `show${rollType.type[0].toUpperCase()}${rollType.type.slice(1)}Dialogs`;

```

CarlosFdez:
You can make it so that event is non-optional. I don't know why it was optional to begin with. If its because undefined is something that can be passed in, you can do | undefined if its common, otherwise ?? null should be fine.

Since there's only two valid values, and I doubt this is going to grow, I think a ternary rather than trying to concatenate strings might be better. More type safe too, you won't need to cast.

jfn4th:
Undefined does actually come through pretty often. At a quick glance, once for each damage and once for each critical button on the action tab of a character sheet whenever the sheet is opened, while `activateListeners` is running. Set as non optional and added `| undefined.`

jfn4th:
As a side note, you can count my players and I amongst the minority that _do_ keep the roll checks enabled by default, and I like the new damage prompt too! I just thought this would be a good feature for getting a bit more familiar with the code base 🙂

src/scripts/sheet-util.ts
```diff
@@ -7,8 +7,12 @@ function isRelevantEvent(
     return !!event && "ctrlKey" in event && "metaKey" in event && "shiftKey" in event;
 }
 
-function eventToRollParams(event?: JQuery.TriggeredEvent | Event | null): ParamsFromEvent {
-    const skipDefault = !game.user.settings.showRollDialogs;
+function eventToRollParams(
+    event: JQuery.TriggeredEvent | Event | null | undefined,
+    rollType: { type: string },
```
CarlosFdez, line 12:
```suggestion
    rollType: { type: "check" | "damage" },
```

src/scripts/sheet-util.ts
```diff
@@ -7,8 +7,12 @@ function isRelevantEvent(
     return !!event && "ctrlKey" in event && "metaKey" in event && "shiftKey" in event;
 }
 
-function eventToRollParams(event?: JQuery.TriggeredEvent | Event | null): ParamsFromEvent {
-    const skipDefault = !game.user.settings.showRollDialogs;
+function eventToRollParams(
+    event: JQuery.TriggeredEvent | Event | null | undefined,
```
stwlam, line 11:
```suggestion
    event: Maybe<JQuery.TriggeredEvent | Event>,
```

CarlosFdez:
This will need a migration, but we can do that similarly.

PR #11092, opened 2023-11-04
Makes nonlethal attacks cause unconsciousness instead of death if dea…

xdy:
…th at zero hp automation is enabled.

The implementation hinges on me being right that 'instant death' (whether by massive damage, death effects or being an npc undead with zero hp) takes precedence. E.g. if a level 20 fighter attacks a bunny with his (nonlethal) fists, the bunny will die due to massive damage.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1282,19 +1282,39 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
                 : null;
         })();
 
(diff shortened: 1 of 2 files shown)


stwlam:
We don't actually have support for non-lethal attacks. The trait being on a weapon doesn't make a strike with it nonlethal: the choice to make a nonlethal attack (with or without the trait) does.

xdy:
Mmm. True. Ah well. Was hoping to kill another Workbench feature. :)

PR #11093, opened 2023-11-04
Fix typo in description of retch rust feat

rectulo:


simonward:
It may be a typo, but it's in the source.
Screen Shot 2023-11-04 at 9 56 57 AM

PR #11098, opened 2023-11-05
Add Reposition action implementation, action item, and action macro

nikolaj-a:
Reposition check success

Reposition check critical success

Reposition check failure

Reposition check critical failure


TikaelSol:
Reposition is already added in the data entry world.

nikolaj-a:
> Reposition is already added in the data entry world.

Makes sense. I'll take out the action item from this PR then.

PR #11123, opened 2023-11-06
Fix NPC shield labels

jfn4th:
This fixes NPC shield stat labels, which broke sometime between 5.7.4 and 5.8.0 when the old localized text was removed.

For the moment I've pointed the text to the newer `PF2E.Actor.Character.Shield` and `PF2E.Item.Physical` strings since they're already there rather than adding `Shield` to `PF2E.Actor.NPC` and having duplicates there. I suspect it's best to have one place outside of Character and NPC for these shield specific strings but wasn't quite sure where, so let me know what/where is preferred.

Before: 
Screenshot 2023-11-05 at 8 18 58 PM
Screenshot 2023-11-05 at 8 23 14 PM
After:
Screenshot 2023-11-05 at 8 22 58 PM

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -210,18 +210,6 @@
                 },
                 "Rest": {
                     "Label": "Rest for the Night"
-                },
-                "Shield": {
-                    "ACBonus": "Shield AC Bonus (When Raised)",
-                    "BrokenThreshold": "Shield Broken Threshold",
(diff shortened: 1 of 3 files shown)


stwlam:
Could you move the localization strings to the `Creature` level?

jfn4th:
All moved.

PR #11128, opened 2023-11-06
Never delete consumed ammunition carrying rule elements

stwlam:


CarlosFdez:
I see some magazine related code in here, how does this interact with repeating ammo?

PR #11137, opened 2023-11-06
Remove spell components from actor sheets

CarlosFdez:
Also addressed a minor issue involving resize behavior and extra whitespace in the spell preparation sheet caused by the lack of a casting button.

image

### DIFF
--- a/src/module/actor/creature/spell-preparation-sheet.ts
+++ b/src/module/actor/creature/spell-preparation-sheet.ts
@@ -36,7 +36,7 @@ class SpellPreparationSheet<TActor extends CreaturePF2e> extends ActorSheet<TAct
             classes: ["default", "sheet", "spellcasting-entry", "preparation"],
             width: 480,
             height: 600,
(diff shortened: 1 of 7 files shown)


j-bs:
Thoughts on bumping the `Range` column on actor sheet to be right-aligned like the components were?

Personally I think it looks better than a whitespace void between the rightmost column and the `Cast` button.

CarlosFdez:
So there are three routes it can go after this, but should probably be relegated to a new PR since its variable:
1) We show both range and uses for innate spells. Then the space available to range becomes smaller, and its no longer a void (the reason the range area is so big is because of innate spells).
2) We show range and area in that space ex: `500 feet (20-foot burst)`. If we do that, the discrepancy in sizes will make that look strange if right aligned.
3) We show emanation if there is one, but don't show area otherwise. This makes emanations slightly bigger than range, and right aligning them may or may not look strange, I'd have to see.

Each of these I think vastly alter what we're doing with that space, and at the very least still less space than there is on live currently.
image

stwlam:
Conflicts!

PR #11149, opened 2023-11-07
Add Aasimar Trait to Purge Sins

Intervencion:


TikaelSol:
The trait will be going away soon.

PR #11159, opened 2023-11-08
Update kingmaker army actions and tactics

CarlosFdez:
Supercedes https://github.com/foundryvtt/pf2e/pull/10880


CarlosFdez:
I accidentally force pushed to the other one, and it didn't break, so I guess we can keep that one instead.

stwlam:
well are ya going to reopen it?

PR #11208, opened 2023-11-10
Add token params to applyDamageFromMessage and make it a static function on ChatMessagePF2e

### LINKED ISSUE #10927: Create applyDamageToTokenFromMessage()
This is a request to create a function that works similarly to `applyDamageFromMessage` but allows you to provide your own tokens to damage, and of course exported for outside use.

https://github.com/foundryvtt/pf2e/blob/0f09409447bb5a2a6ae285f1f73977e4ed14753c/src/module/chat-message/helpers.ts#L73C30-L73C30

Its possible it would be easier to have the existing function updated to allow for providing tokens with the default values being the same as before, instead of maintaining two functions at once.

This request is purely for module-related reasons (Target Damage) as currently the only way to process a message's damage rolls and apply them to a token is to copy the entire function, which has to be maintained and updated every time the system changes.

MrVauxs:
Fixes #10927

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -14,6 +14,7 @@ import { CriticalHitAndFumbleCards } from "./crit-fumble-cards.ts";
 import { ChatMessageFlagsPF2e, ChatMessageSourcePF2e } from "./data.ts";
 import * as Listeners from "./listeners/index.ts";
 import { RollInspector } from "./roll-inspector.ts";
+import { ApplyDamageFromMessageParams, applyDamageFromMessage } from "./helpers.ts";
 
 class ChatMessagePF2e extends ChatMessage {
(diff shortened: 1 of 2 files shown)


stwlam:
This isn't public API

MrVauxs:
It's retrievable through the ChatMessagePF2e class which is accessible. But if you have suggestions, I'll try.

CarlosFdez:
The main issue I have with this is that its too volatile. Some of that function should be in public API, but not like this.

MrVauxs:
My goal behind this suggestion is the elimination of having to copy code from the system for anything that wants to deal damage based on the actual damage rolls.

The function that the system uses still cant be 🐵🩹 as far as I am aware. Its just a new public function that calls the, still private, system one.

MrVauxs:
I would love to hear why this is volatile and unsafe, maybe I could think of a compromise.

PR #11209, opened 2023-11-10
Remove margin from Party list on Actors sidebar

MrVauxs:
## Before and After
image image

image image

### DIFF
--- a/src/styles/ui/sidebar/_actor-directory.scss
+++ b/src/styles/ui/sidebar/_actor-directory.scss
@@ -15,7 +15,7 @@
 
     ol.party-list {
         list-style-type: none;
-        margin: 0 0 0.25rem 0;
+        margin: 0;
         padding: 0;
         position: relative;
     }


CarlosFdez:
The margin's not a bug, its intentional. If anything it should be made more obvious. I'd be willing to take suggestions on that front, a solid black line might look odd.

(the appearance when its open is a bug though)

MrVauxs:
What is the intent behind the margin? Because it will always look odd, especially when having other folders which have no margin between them and actors. Its inconsistent with the rest of the UI and given that you cannot have parties anywhere _but_ the top, has very little meaning in terms of separation.

stwlam:
Because it's not actually a folder and needs _some_ way to distinguish it. Currently it's still not distinguishing enough to prevent confusion among new users.

MrVauxs:
And what is the issue with people making Parties instead of Folders like they can already?

MrVauxs:
How about small, centered, `::after` text saying where parties begin and where actors begin?

Mockup:
image

CarlosFdez:
Thanks for giving it a shot. I asked internally about it and while I think what you did was super worth a shot (I think it could maybe work if only the actors header exists and the underline was gone),t it seems people internally prefer the barrier to exist but also be hard to perceive.

MrVauxs:
Alas, "hard to perceive" often means that when you _do_ perceive it, it looks even more so out of place / like a bug. Which was my reaction to it which spurred this PR.

PR #11221, opened 2023-11-11
Unify spellcasting entries across traditions

CarlosFdez:
Also do the innate spell auto expert thingy

### DIFF
--- a/src/module/item/spellcasting-entry/document.ts
+++ b/src/module/item/spellcasting-entry/document.ts
@@ -93,7 +93,7 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
 
     override prepareBaseData(): void {
         super.prepareBaseData();
-        this.system.proficiency.slug ||= this.system.tradition.value;
+
         // Spellcasting abilities are always at least trained
         this.system.proficiency.value = Math.max(1, this.system.proficiency.value) as OneToFour;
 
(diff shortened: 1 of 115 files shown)


src/module/item/spellcasting-entry/document.ts
```diff
@@ -93,7 +93,7 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
 
     override prepareBaseData(): void {
         super.prepareBaseData();
-        this.system.proficiency.slug ||= this.system.tradition.value;
+        this.system.proficiency.slug ||= "spellcasting-base";
```
stwlam, line 96: Why base? There's only ever one spellcasting proficiency

CarlosFdez: The main thing I struggled with is that if I went spellcasting, the call would be actor.getSpellcasting("spellcasting"), which should probably return the "highest" spellcasting instead (does nothing right now).

I considered doing something like `actor.getSpellcasting("spellcasting, { type: "base" })` but since this is all prepared data anyways (spellcasting-base isn't persisted), I wanted to defer the question until spellcasting could be automated.

Still not quite sure what to do about that entire ordeal.

src/module/migration/migrations/884-unified-spellcasting.ts
```diff
@@ -0,0 +1,30 @@
...
+        // traditions was prepared data, but remove it just in case someone was naughty
+        const profs: CharacterProficienciesOld | null = source.system.proficiencies ?? null;
+        if (profs?.traditions) {
+            profs["-=traditions"] = null;
+        }
+    }
+
+    override async updateItem(source: ItemSourcePF2e): Promise<void> {
+        source.system.rules = recursiveReplaceString(source.system.rules, (value) =>
+            value.replace(/^system\.proficiencies\.traditions\.(.*)\.(\w+)$/, "system.proficiencies.spellcasting.$2"),
```
stwlam, line 22:
```suggestion
            value.replace(/^system\.proficiencies\.traditions\.[^.]+\.(\w+)$/, "system.proficiencies.spellcasting.$1"),
```

CarlosFdez: This will fail to catch the {actor|blah.blah} stuff. We have to be greedy unfortunately.

stwlam: Is "blah.blah" going to be a valid property path following "system.proficiencies.spellcasting"?

PR #11224, opened 2023-11-11
Add support for using token marks on origins

stwlam:
(instead of just targets)

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -874,7 +874,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             options: [...params.options, ...(params.item?.getRollOptions("item") ?? [])],
         });
 
-        const tokenMarkOption = ((): string | null => {
+        const targetMarkOption = ((): string | null => {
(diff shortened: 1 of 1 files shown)


src/module/actor/base.ts
```diff
@@ -991,15 +989,21 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         }
 
         // Clone the actor to recalculate its AC with contextual roll options
+        const originMarkOption = ((): string | null => {
+            const tokenMark = selfToken ? targetToken?.actor?.synthetics.tokenMarks.get(selfToken.document.uuid) : null;
+            return tokenMark ? `origin:mark:${tokenMark}` : null;
+        })();
+
```
CarlosFdez, line 996:
```suggestion
        const originMarkOption = ((): string | null => {
            const tokenMark = selfToken ? targetToken?.actor?.synthetics.tokenMarks.get(selfToken.document.uuid) : null;
            return tokenMark ? `origin:mark:${tokenMark}` : null;
        })();

        // Clone the actor to recalculate its AC with contextual roll options
```

PR #11252, opened 2023-11-12
Add remaster changes to long jump action macro

nikolaj-a:
Update the text with the default difficulty class to reflect the text in Player Core, and assign the default difficulty class for the action macro when no other DC is provided.

### DIFF
--- a/src/module/system/action-macros/athletics/long-jump.ts
+++ b/src/module/system/action-macros/athletics/long-jump.ts
@@ -14,7 +14,7 @@ export function longJump(options: SkillActionOptions): void {
         traits: ["move"],
         event: options.event,
         callback: options.callback,
-        difficultyClass: options.difficultyClass,
+        difficultyClass: options.difficultyClass ?? { value: 15 },
         extraNotes: (selector: string) => [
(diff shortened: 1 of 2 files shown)


src/module/system/action-macros/athletics/long-jump.ts
```diff
@@ -14,7 +14,7 @@ export function longJump(options: SkillActionOptions): void {
         traits: ["move"],
         event: options.event,
         callback: options.callback,
-        difficultyClass: options.difficultyClass,
+        difficultyClass: options.difficultyClass ?? { value: 15 },
```
stwlam, line 17: Why a default?

nikolaj-a: That’s how it works now. It’s like High Jump, except with DC 15 instead of DC 30.

nikolaj-a: > You Stride, then attempt a DC 15 Athletics check to make a long jump in the direction you were Striding. If you didn’t Stride at least 10 feet, you automatically fail your check. The GM might increase or decrease this DC depending on the situation.

stwlam: Lame!

PR #11307, opened 2023-11-14
Add keybinding to focus compendium search bar

dmarcuse:
By default, Ctrl+Space is used as the keybinding, but only when Quick Insert isn't enabled due to the overlap in functionality and keybinding.

### DIFF
--- a/src/scripts/register-keybindings.ts
+++ b/src/scripts/register-keybindings.ts
@@ -1,4 +1,5 @@
 import type { PartyPF2e } from "@actor";
+import { CompendiumDirectoryPF2e } from "@module/apps/sidebar/compendium-directory.ts";
 
 export function registerKeybindings(): void {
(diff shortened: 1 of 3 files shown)


stwlam:
A single action in a single application window is too narrow a use case for a keybinding. We also will never add one for Ctrl-Space: that was an unfortunate decision on part of Quick Insert's author.

dmarcuse:
> A single action in a single application window is too narrow a use case for a keybinding.
I'm not sure what you mean - isn't the shortcut for the party sheet in a similar category this this, a keybind to quickly navigate to a specific application?
> We also will never add one for Ctrl-Space: that was an unfortunate decision on part of Quick Insert's author.
I'm happy to change it. Maybe Ctrl + K?

stwlam:
Ctrl-K is Chrome's search bar focus keybind. We can talk about it on Discord.

PR #11317, opened 2023-11-14
Add `none` group to rollOptions for weapons and armour without groups

Cerapter:
As the title says.

Decided to *only* modify the rollOptions and not the logic before it so anything that depended on an actual `null` value in-code or in modules need not change.
This rollOption change is just an additional value that didn't exist before, and `group:null` should be "unnatural" enough to not fuck anything over for modules and the like.

### DIFF
--- a/src/module/item/armor/document.ts
+++ b/src/module/item/armor/document.ts
@@ -100,11 +100,13 @@ class ArmorPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Phy
 
     /** Generate a list of strings for use in predication */
     override getRollOptions(prefix = "armor"): string[] {
+        const group = this.group || "null";
+
         return [
             super.getRollOptions(prefix),
             Object.entries({
                 [`category:${this.category}`]: true,
-                [`group:${this.group}`]: !!this.group,
+                [`group:${group}`]: true,
                 [`base:${this.baseType}`]: !!this.baseType,
                 [`rune:potency`]: this.system.runes.potency > 0,
                 [`rune:resilient`]: this.system.runes.resilient > 0,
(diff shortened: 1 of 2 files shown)


src/module/item/armor/document.ts
```diff
@@ -100,11 +100,13 @@ class ArmorPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Phy
 
     /** Generate a list of strings for use in predication */
     override getRollOptions(prefix = "armor"): string[] {
+        const group = this.group || "null";
+
```
stwlam, line 104:
```suggestion
```

src/module/item/armor/document.ts
```diff
@@ -100,11 +100,13 @@ class ArmorPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Phy
...
         return [
             super.getRollOptions(prefix),
             Object.entries({
                 [`category:${this.category}`]: true,
-                [`group:${this.group}`]: !!this.group,
+                [`group:${group}`]: true,
```
stwlam, line 109:
```suggestion
                [`group:${this.group ?? "none"}`]: true,
```

src/module/item/weapon/document.ts
```diff
@@ -234,12 +234,13 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
             return unitBulk.isNegligible ? "negligible" : unitBulk.isLight ? "light" : unitBulk.toString();
         })();
         const rangeIncrement = this.range?.increment;
+        const group = this.group || "null";
```
stwlam, line 237:
```suggestion
```

src/module/item/weapon/document.ts
```diff
@@ -234,12 +234,13 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
...
 
         return [
             super.getRollOptions(prefix),
             Object.entries({
                 [`category:${this.category}`]: true,
-                [`group:${this.group}`]: !!this.group,
+                [`group:${group}`]: true,
```
stwlam, line 243:
```suggestion
                [`group:${this.group ?? "none"}`]: true,
```

stwlam (changes requested):
There's an existing precedent for using "none"

stwlam (approved):

PR #11363, opened 2023-11-15
Various spell compendium fixes for after remaster

azrazalea:
FocusPool opens with Focus trait and I made up a filter called "Type" (not the
best name but I was stumped) to be able to select if you want to see rituals or
not. Then I made it so that if you open the spell browser from the ritual entry
then the ritual filter is pre-selected.

The new options need to be moved to the proper place in config, and I don't
think we want `focus` and `ritual` hardcoded in openSpellBook but I'm out of
time for now.


azrazalea:
Another idea I had was maybe focus should be under "type" (or whatever it ends up being called) as well and we could just convert that to a trait filter inside filterIndexData without the user even knowing. With that method you could also still filter by the focus trait but it feels more intuitive to me to have "Focus", "Spell", and "Ritual" as types. 🤷‍♀️

stwlam:
Searching by the trait is all that's needed

azrazalea:
Closed by #11367

PR #11365, opened 2023-11-15
Add filter traits for runes to compendium browser equipment tab

### LINKED ISSUE #6407: Can we get a way to search for runes exclusively on the Compendium Browser?
The system sadly does not have a "Rune" or "property" tag, so this is would have to be a meta sorting option. Maybe Runestone(Name of Rune) to sort of stay canonical is a workaround solution. 

The point is: Players need to look up runes often, and they are an item that always gets bought and sold, and a lot of people are not familiar with all the property runes, especially after books like Grand Bazaar and Treasure Vault. Is it a viable goal to have a way to search for them?

In3luki:
Also enbales fuzzy search for the tagify traits dropdown.

Closes #6407

### DIFF
--- a/src/module/apps/compendium-browser/index.ts
+++ b/src/module/apps/compendium-browser/index.ts
@@ -591,7 +591,7 @@ class CompendiumBrowser extends Application {
                         tagTextProp: "label",
                         dropdown: {
                             enabled: 0,
-                            fuzzySearch: false,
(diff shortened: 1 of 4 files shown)


In3luki:
I wasn't aware that runes as items are going away soonish so this will be obsolete. I'll revisit this when the conversion is finished.

PR #11423, opened 2023-11-16
Cleanup custom damage versatile traits (and restrict deletion to damage types only)

CarlosFdez:
Technically a bugfix, though the bug has already been fixed. This is mostly future proofing to avoid the bug resurfacing again at a later date.


stwlam:
Won't this leave homebrew IWR in place?

CarlosFdez:
Is it possible to add homebrew iwr that arent new damage types?

stwlam:
No, you're about about that: the only thing I can think of is versatile weapon traits not getting removed.

CarlosFdez:
oh, that's actually possible now by collecting a list of damage types to remove. I'll do that soon.

PR #11424, opened 2023-11-16
Allow null categories in module custom damage

CarlosFdez:
It was previously allowed for world settings, this extends it to modules.

### DIFF
--- a/src/module/system/settings/homebrew/helpers.ts
+++ b/src/module/system/settings/homebrew/helpers.ts
@@ -23,7 +23,7 @@ function isHomebrewCustomDamage(value: object): value is Record<string, CustomDa
         (value) =>
             isObject<CustomDamageData>(value) &&
             typeof value.label === "string" &&
-            ["physical", "energy"].includes(value.category ?? ""),
+            (!value.category || ["physical", "energy"].includes(value.category)),
     );
 }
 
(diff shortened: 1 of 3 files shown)


src/module/system/settings/homebrew/helpers.ts
```diff
@@ -23,7 +23,7 @@ function isHomebrewCustomDamage(value: object): value is Record<string, CustomDa
         (value) =>
             isObject<CustomDamageData>(value) &&
             typeof value.label === "string" &&
-            ["physical", "energy"].includes(value.category ?? ""),
+            ["physical", "energy", null].includes(value.category ?? ""),
```
stwlam, line 26:
```suggestion
            ["physical", "energy", ""].includes(value.category ?? ""),
```

CarlosFdez: The type we are doing a typeguard for doesnt allow "" there nor undefined, only null. The way i did it makes it a required property.

?? "" is because typescript complains, but we could do tupleHasValue

stwlam: `value.category ?? ""` is never `null`

CarlosFdez: true. I opened it up to allow undefined as well.

PR #11446, opened 2023-11-16
Migrate non-boolean `adjustName` values in choice set REs

In3luki:
#11242 changed `adjustName` to a `StrictBooleanField`


stwlam:
Do those need special handling, or is Foundry coercing "false" to `false`?

PR #11543, opened 2023-11-18
Fix Everstand Stance not adding hardness when shield is raised

### LINKED ISSUE #11528: Everstand Stance not adding hardness to shield during use
**_Everstand Stance's_** effect of adding 2 to hardness to a shield hasn't worked for a few updates (it stopped working on or before 5.7.2).  Previously, it would add 2 to the shield's hardness during it's use, but now that doesn't happen.

I've tested on a fresh install with 5.9.3 and no modules, and it's not working. 

NB: The part of the automation that increases _die size_ **DOES** work, just not the _hardness_ part.

kakesu:
Fixes #11528

### DIFF
--- a/packs/feat-effects/stance-everstand-stance.json
+++ b/packs/feat-effects/stance-everstand-stance.json
@@ -28,7 +28,7 @@
                 "predicate": [
                     "item:category:shield",
                     "item:equipped",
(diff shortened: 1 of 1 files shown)


stwlam:
Superseded by #11542

PR #11602, opened 2023-11-20
Update REs on Champion's Code to not Add both Holy and Unholy when selecting a Deity that adds both

avagdumortesin:
Adds a predicate to check which variant of the Champion is selected and only add the Tenets of Evil or Tenets of Good based on that selection.


stwlam:
Have the cause items set "sanctification:holy" or "sanctification:unholy" roll options

Issue #10722, opened 2023-10-14
Elemental Blast damage rolls lack rolloptions for damage type


esheyw:
![](https://i.imgur.com/UkLHRXg.png)
Tested with both metal and fire single gates

PR #11627, opened 2023-11-20
Add melee/ranged and damage type to elemental blast roll options

CarlosFdez:
Added the melee-attack-roll and ranged-attack-roll domains will require a migration, so roll option only for now.

Closes https://github.com/foundryvtt/pf2e/issues/10722


stwlam:
Most things are going to look for "item:melee" or "item:ranged"

PR #11637, opened 2023-11-21
Remove sorcerer bloodline AE-likes enumerating spells

stwlam:


static/lang/re-en.json
```diff
@@ -2059,14 +2059,6 @@
...
                 }
             },
-            "Genie": {
-                "Faydhaan": "Faydhaan",
-                "Ifrit": "Ifrit",
-                "Jaathoom": "Jaathoom",
-                "Jabali": "Jabali",
-                "Janni": "Janni",
-                "Prompt": "Select a Genie."
-            },
```
TikaelSol, line 2069: We should keep these and the choice set on genie to select which genie.

PR #11665, opened 2023-11-22
Add shield item type, migrate armor-shields to shield-shields

stwlam:


src/module/actor/sheet/base.ts
```diff
@@ -157,18 +157,32 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
...
-            const category = item.isOfType("book") ? sections.equipment : sections[item.type];
-            category.items.push(this.prepareInventoryItem(item));
+            if (item.isInContainer) continue;
+            const section = objectHasKey(sections, item.type)
+                ? sections[item.type]
+                : item.isOfType("book")
+                  ? sections.equipment
+                  : item.isOfType("shield")
+                    ? sections.weapon
+                    : null;
```
CarlosFdez, line 184: Should this be a search for the section's types instead? Of course after adding book to the equipment types.

stwlam: That would be better, yes

src/module/actor/character/document.ts
```diff
@@ -1316,10 +1322,15 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
+            const { handsHeld } = weapon;
 
             switch (weapon.carryType) {
                 case "held": {
-                    if (weapon.handsHeld === 2) {
+                    if (weapon.shield && !weapon.shield.isRaised) {
+                        auxiliaryActions.push(new WeaponAuxiliaryAction({ weapon, action: "RaiseAShield" }));
+                    }
+
+                    if (handsHeld === 2) {
```
CarlosFdez, line 1333: I think you can keep this to weapon.handsHeld, since the var isn't used anywhere else? Should keep it clearer hopefully.

src/module/actor/creature/document.ts
```diff
@@ -406,12 +406,15 @@ abstract class CreaturePF2e<
...
-            if (isEquipped(usage, equipped) && item instanceof ArmorPF2e && item.isArmor) {
+            if (isEquipped(usage, equipped) && item.isOfType("armor")) {
                 // see if they have another set of armor equipped
-                const wornArmors = this.itemTypes.armor.filter((a) => a !== item && a.isEquipped && a.isArmor);
+                const wornArmors = this.itemTypes.armor.filter((a) => a !== item && a.isEquipped);
                 for (const armor of wornArmors) {
                     updates.push({ _id: armor.id, system: { equipped: { inSlot: false } } });
                 }
+            } else if (equipped.carryType !== "held" && item.isOfType("weapon") && item.shield?.isRaised) {
+                const raiseAShieldEffect = item.actor.itemTypes.effect.find((e) => e.slug === "effect-raise-a-shield");
```
CarlosFdez, line 416:
```suggestion
                // Stop raising the shield the weapon is attached to when it becomes stowed
                const raiseAShieldEffect = item.actor.itemTypes.effect.find((e) => e.slug === "effect-raise-a-shield");
```

Took me a bit to realize what this was for :(

pf2e.zip
CarlosFdez, line 1: 🙃

src/module/item/base/document.ts
```diff
@@ -854,7 +854,11 @@ const ItemProxyPF2e = new Proxy(ItemPF2e, {
         _target,
         args: [source: PreCreate<ItemSourcePF2e>, context?: DocumentConstructionContext<ActorPF2e | null>],
     ) {
-        const ItemClass = CONFIG.PF2E.Item.documentClasses[args[0]?.type] ?? ItemPF2e;
+        const type =
+            args[0]?.type === "armor" && (args[0].system?.category as string | undefined) === "shield"
+                ? "shield"
```
CarlosFdez, line 859: is this not handled by the migration?

stwlam: migrating doesn't reconstruct the item.

src/module/item/shield/document.ts
```diff
@@ -0,0 +1,335 @@
...
+
+        const adjustFromMaterialAndRune = (property: "hardness" | "maxHP", base: number): number => {
+            const fromMaterial =
+                material.type && material.grade
+                    ? materialData[material.type]?.[material.grade]?.[property] ?? base
+                    : base;
+            const additionalFromRune = reinforcingRune ? reinforcingRuneData[reinforcingRune]?.[property] : null;
+            const sumFromRune = fromMaterial + (additionalFromRune?.increase ?? 0);
+            return additionalFromRune && sumFromRune > additionalFromRune.max
+                ? Math.max(fromMaterial, additionalFromRune.max)
```
CarlosFdez, line 105: Should this not be min? Also, I think you might be able to skip the ternary with return `Math.min(fromMaterial, additionalFromRune?.max ?? Infinity)`

src/module/item/shield/sheet.ts
```diff
@@ -0,0 +1,89 @@
...
+        }));
+
+        return {
+            ...sheetData,
+            baseHardness: item._source.system.hardness,
+            basePrice: new CoinsPF2e(item._source.system.price.value),
+            baseTypes: sortStringRecord(CONFIG.PF2E.baseShieldTypes),
+            preciousMaterials: this.prepareMaterials(SHIELD_MATERIAL_VALUATION_DATA),
+            propertyRuneSlots,
+            reinforcing: R.mapToObj([1, 2, 3, 4, 5, 6] as const, (t) => [t, REINFORCING_RUNE_LOC_PATHS[t]]),
```
CarlosFdez, line 37: Thoughts on `R.mapToObj.indexed(REINFORCING_RUNE_LOC_PATHS.slice(1), (p, idx) => [idx + 1, p])` instead?

src/module/item/shield/document.ts
stwlam: `Math.max` is correct there: a reinforcing rune limits how much it can increase a shield's hardness and HP

CarlosFdez: But if the max is 20, and materials increase to 30, and you do Math.max(30, 20), the result is 30.

stwlam: And that would be correct

PR #11672, opened 2023-11-22
Fix errors without canvas because of canvas.tokens being undefined

OmegaRogue:
Fix various errors that happen when  `canvas.tokens.controlled` is accessed when the canvas is disabled because of `canvas.tokens` being `undefined`

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -713,7 +713,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         this.preparePrototypeToken();
         if (this.constructed && canvas.ready) {
             // Work around `t.actor` potentially being a lazy getter for a synthetic actor (viz. this one)
-            const thisTokenIsControlled = canvas.tokens.controlled.some(
+            const thisTokenIsControlled = canvas.tokens?.controlled.some(
                 (t) => t.document === this.parent || (t.document.actorLink && t.actor === this),
             );
             if (game.user.character === this || thisTokenIsControlled) {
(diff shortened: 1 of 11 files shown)


src/module/chat-message/listeners/cards.ts
```diff
@@ -322,14 +322,14 @@ class ChatCards {
...
-        if (canvas.tokens.controlled.length > 0) {
+        if (canvas.tokens?.controlled.length > 0) {
             const saveType = button.dataset.save;
             if (!tupleHasValue(SAVE_TYPES, saveType)) {
                 throw ErrorPF2e(`"${saveType}" is not a recognized save type`);
             }
 
             const dc = Number(button.dataset.dc ?? "NaN");
-            for (const token of canvas.tokens.controlled) {
+            for (const token of canvas.tokens?.controlled) {
```
shemetz, line 332: I think this needs `?? []` at the end, otherwise it'll still throw because `for (const token of undefined)` would throw

OmegaRogue: Yea, probably

stwlam:
The type needs to be updated such that it isn't guaranteed to be present.

src/module/actor/base.ts
```diff
@@ -713,7 +713,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         this.preparePrototypeToken();
         if (this.constructed && canvas.ready) {
             // Work around `t.actor` potentially being a lazy getter for a synthetic actor (viz. this one)
-            const thisTokenIsControlled = canvas.tokens.controlled.some(
+            const thisTokenIsControlled = canvas.tokens?.controlled.some(
```
stwlam, line 716: If the canvas is disabled, then it won't be `ready`.

OmegaRogue:
> The type needs to be updated such that it isn't guaranteed to be present.
As far as I can tell, `getSelectedOrOwnActors` always returns at least an empty array

stwlam:
> > The type needs to be updated such that it isn't guaranteed to be present.
> 
> As far as I can tell, `getSelectedOrOwnActors` always returns at least an empty array

Sorry, by "the type" I mean that `Canvas#tokens` needs to be made optional.

OmegaRogue:
I think this is already resolved by #11693

PR #11680, opened 2023-11-22
Update text of Recall Knowledge action

### LINKED ISSUE #11662: 5.9.4 still has old Recall Knowledge text
When viewing the Recall Knowledge action in the compendium browser, going to add it to my PCs, I see the old text of the skill instead of the new remaster text.

image

stwlam:
Closes #11662 

I decided to not include the additional panel since it's 4-5 times as long as the action text now. Is there an appropriate journal entry to place it in?


TikaelSol:
Somewhere in the GM screen journal makes sense to hold that

PR #11689, opened 2023-11-23
Change Arcane trait given from Stance: Arcane Cascade to be a weapon trait

ivycase:
Fixes an issue with damage resistances when interacting with Arcane Cascade.

Before change:

image

After change: 

image

This change is consistent with packs/feats/arcane-fists.json but it may not be best practice, given that arcane (alongside magical, the general trait) is not considered a weapon trait in [PF2E terminology](https://2e.aonprd.com/Traits.aspx). A future solution could be to change the _Immunities, Weaknesses & Resistances_ implementation to consider general traits when calculating damage for weapons.

### DIFF
--- a/packs/feat-effects/stance-arcane-cascade.json
+++ b/packs/feat-effects/stance-arcane-cascade.json
@@ -107,7 +107,7 @@
                 ],
                 "key": "AdjustStrike",
                 "mode": "add",
-                "property": "traits",
+                "property": "weapon-traits",
                 "value": "arcane"
             },
             {


stwlam:
> Any Strike that benefits from this damage gains the arcane trait, making it magical.

ivycase:
Yes. A Strike benefitting from Arcane Cascade should have the arcane trait, making it a magical Strike. Expected behavior is that the Strike's damage should be unaffected by physical resistances that except magic damage. Current behavior is that the arcane trait is displayed but the damage is still resisted.

This PR results in the expected behavior, and is consistent with the current implementation of the Arcane Fists class feat.

stwlam:
Superseded by #11695

PR #11693, opened 2023-11-23
Consolidate retrieval of controlled/tokens with assigned character fallbacks

stwlam:


src/util/token-actor-utils.ts
```diff
@@ -1,22 +1,20 @@
 import type { ActorPF2e } from "@actor";
-import { ActorType } from "@actor/data/index.ts";
+import type { ActorType } from "@actor/data/index.ts";
+import * as R from "remeda";
 
 /**
  * Collects every actor whose token is controlled on the canvas, and if none are, collects the current user's character, if it exists.
- *
  * @param types The actor types the function should take into consideration.
  * @param useOwnCharacter If true, the function will append the user's own character to the list of collected actors.
```
CarlosFdez, line 8: this comment line should probably be gone. Do users purchasing items for their own items from the compendium browser still work after this.

CarlosFdez:
loot-npc-popup seems to have unrelated changes, but they're good changes. Seems to be an attempt to fix a bug maybe.

Issue #11454, opened 2023-11-16
[Feature Request] Edict and Anathema fields


Ghostdesu:
Now that Player Core is out, every ancestry and a few classes have (suggested or required) edicts and anathema that can have a mechanical impact in the case of Champion, Cleric and Druid. Especially Champion, since the new errata refers to their Edicts in feats, not just the part that say they can lose their powers if XYZ. 
It would be very nice to have a dedicated place on the biography tab to put them for reference.

azrazalea:
@stwlam I can take this if you want, should be relatively easy

sambomb:
Bonus points if it inherited from your deity from cleric/champion, champion tenant, druidic order, barbarian instinct, etc...

So you can have all your edits and anathemas at the same place.

Maybe add them as a class feature and use the grant item RE and the edits/anathemas list shows all available.

That would make link those to the chat or journals more easily.

azrazalea:
@sambomb i'm a new contributor so assuming someone else doesn't jump in it'll definitely just be some additional basic text boxes you can paste into for the first iteration. You could definitely put links in it though too. It sounds really cool to have it be more automatic though just out of my experience range for the moment.

sambomb:
Yeah, take your time and do what you can.
I know that what I'm suggesting is way harder then just 2 text box, but it would be very nice, maybe you can expand in the idea later

PR #11696, opened 2023-11-23
Add edicts and anathema formatted text fields to actor bio

azrazalea:
Not sure about any context I might need re: existing actors, but this seems to work well now.


azrazalea:
Fixes https://github.com/foundryvtt/pf2e/issues/11454

CarlosFdez:
Thanks for this! I have a few comments before we can merge it in. Also, please share a screenshot!

First of all, its anathema, not anathemas. Anathema is already plural.

Secondly, run `npx prettier src/styles/actor/character/_biography.scss --write` before committing. There seem to be some code styling tweaks that are required.

Otherwise, good job, please share screenshots!

azrazalea:
Awesome thanks for the command @CarlosFdez! I got VS Code setup with eslint and that's how it was auto-formatting 🙃.

Renamed anathemas to anathema 😅  (I noticed when makings sure I had caught all the instances btw that Nhimbaloth's entry says anathemas in it as a typo. Wasn't sure if any migrations would be necessary if I fixed that so left it alone).

Here are the screenshots. The only problems I've found are ones all the rich text editing seems to share (no way to close the editor if you don't make a change, some of the formatting breaks the bio page, etc)

image

image

image

azrazalea:
image

New screenshot after pivot from @CarlosFdez to regular text field.

PR #11700, opened 2023-11-23
Prevent granted heritages from overriding PCs' main heritages 

### LINKED ISSUE #11675: Ancestral Paragon does not allow Earned Glory for Aiuvarin with Elf Ativism
When adding the Ancestral Paragon feat to a PC that has the _Aiuvarin_ Heritage, **Earned Glory** is not shown as an allowed option.

azrazalea:
Fixes #11675


packs/feats/earned-glory.json
```diff
@@ -42,7 +42,7 @@
                     "target:trait:elf"
                 ],
                 "selector": "performance",
-                "text": "PF2E.SpecificRule.HalfElf.EarnedGlory.Note",
+                "text": "PF2E.SpecificRule.Aiuvarin.EarnedGlory.Note",
```
stwlam, line 45: This is going to break localization for all instances of this feat out in the wild

PR #11737, opened 2023-11-24
Constrain `MeasuredTemplate` previews to spell range if possible

In3luki:
Spell range being a string (that might have translated/converted numeric values?) makes this a bit unsafe. Might have to wait for a refactor.

[spell-range-no-border.webm](https://github.com/foundryvtt/pf2e/assets/41452412/3b624093-08c7-4e84-9bdc-40234e30f363)

### DIFF
--- a/src/module/canvas/measured-template.ts
+++ b/src/module/canvas/measured-template.ts
@@ -1,5 +1,5 @@
 import { MeasuredTemplateDocumentPF2e } from "@scene/measured-template-document.ts";
-import { TemplateLayerPF2e } from "./index.ts";
+import { TemplateLayerPF2e, TokenPF2e } from "./index.ts";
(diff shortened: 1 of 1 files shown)


CarlosFdez:
The main concern with usability I see here is planning a turn. Someone for example may click on the place template in their summary, decide where to place it, and then when there turn comes, stride into position then place it there. This prevents that use case. There's also the entire "what are valid placements" thing.

This might need to instead mark if a placement is out of range in some way. Either some floating text or some special rendering in the little icon.

In3luki:
Hm, I see. Maybe holding alt or shift to move the template unconstrained could work? But changing the icon or highlight color when it's out of range instead of constraining the movement would also be an improvement I think.

PR #11750, opened 2023-11-25
Only check for aura and sense updates in visible scenes

CarlosFdez:


src/module/scene/token-document/document.ts
```diff
@@ -454,11 +454,12 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
     ): void {
         super._onRelatedUpdate(update, options);
 
+        // If this scene isn't in view, skip all later checks
+        if (!this.scene?.isView) return;
```
stwlam, line 458:
```suggestion
        if (!this.scene?.isInFocus && !this.scene?.isView) return;
```

PR #11759, opened 2023-11-25
Add more context flags and getters to `MeasuredTemplateDocumentPF2e`

In3luki:

PR #11762, opened 2023-11-25
Force emanation template creation at token center

In3luki:
Depends on #11759 

Movement can be enabled by a new emanation area property:

image

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -133,7 +133,17 @@ export class TemplateLayerPF2e<
                 preview.snapForShape();
                 const { document, position } = preview;
                 this.#deactivatePreviewListeners(initialLayer, event);
-                document.updateSource(canvas.grid.getSnappedPosition(position.x, position.y, this.gridPrecision));
+                const { x, y } = (() => {
+                    if (preview.areaType === "emanation" && !preview.document.flags.pf2e.unconstrained) {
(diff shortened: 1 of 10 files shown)


CarlosFdez:
Does this block all previewing of the ghost template if you mouse elsewhere, or does it only block placement? 

Not sure what unconstrained is meant for.

In3luki:
The preview is rendered on the token center and can't be moved. Right click still cancels and left click still places the preview.

In3luki:
Reworked to allow free movement of the preview template but the created template will be centered on the token (with a notification if it was off center) unless `unconstrained` is set.

CarlosFdez:
Sweet. Abilities like https://2e.aonprd.com/Feats.aspx?ID=4354 might be rough to use if you can't preview a good position to stride into. What purpose does unconstrained solve? Emanations that get left behind?

In3luki:
There are some spells that are emanations and specify a range like https://2e.aonprd.com/Spells.aspx?ID=613

stwlam:
> There are some spells that are emanations and specify a range like https://2e.aonprd.com/Spells.aspx?ID=613

Not sure how to make sense of that range. It's LO:G&M, though, so really anything goes.

In3luki:
Yeah, that should probably have been a burst instead of an emanation.
Maybe a better example: https://2e.aonprd.com/Spells.aspx?ID=38

PR #11774, opened 2023-11-26
Add rule elements to Halfling Luck ancestry feat

### LINKED ISSUE #11739: Data entry: Halfling luck feat could use RE
Something like below. If it's possible to predicate it on the feature being available for its 1/day use that'd be even better but not sure if that's possible.

{
  "key": "Note",
  "title": "Halfling Luck",
  "text": "If you haven't used it today, you can re-roll this check or saving throw.",
  "outcome": [
    "failure",
    "criticalFailure"
  ],
  "selector": [
    "skill-check",
    "saving-throw"
  ]
}

Joshmvii7:
resolves #11739

### DIFF
--- a/packs/feats/halfling-luck.json
+++ b/packs/feats/halfling-luck.json
@@ -1,5 +1,5 @@
 {
-    "_id": "ZbRVqf14RTJJIZXG",
+    "_id": "SBNg9Fl1HhuXZf74",
     "img": "icons/sundries/books/book-red-exclamation.webp",
     "name": "Halfling Luck",
     "system": {
@@ -28,7 +28,33 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "Note",
+                "outcome": [
+                    "failure",
+                    "criticalFailure"
+                ],
+                "predicate": [
+                    {
+                        "gt": [
+                            "halfling-luck",
+                            0
+                        ]
+                    }
+                ],
+                "selector": [
+                    "skill-check",
+                    "saving-throw"
+                ],
+                "text": "You can re-roll this check or saving throw.",
+                "title": "Halfling Luck"
+            },
+            {
+                "key": "RollOption",
+                "option": "halfling-luck:{item|frequency.value}"
+            }
+        ],
         "traits": {
             "rarity": "common",
             "value": [
--- a/packs/feats/halfling-luck.json
+++ b/packs/feats/halfling-luck.json
@@ -1,5 +1,5 @@
 {
-    "_id": "SBNg9Fl1HhuXZf74",
+    "_id": "ZbRVqf14RTJJIZXG",
     "img": "icons/sundries/books/book-red-exclamation.webp",
     "name": "Halfling Luck",
     "system": {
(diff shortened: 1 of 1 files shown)


packs/feats/halfling-luck.json
```diff
@@ -1,5 +1,5 @@
 {
-    "_id": "ZbRVqf14RTJJIZXG",
+    "_id": "SBNg9Fl1HhuXZf74",
```
stwlam, line 2:
```suggestion
    "_id": "ZbRVqf14RTJJIZXG",
```

packs/feats/halfling-luck.json
```diff
@@ -28,7 +28,33 @@
...
+                            0
+                        ]
+                    }
+                ],
+                "selector": [
+                    "skill-check",
+                    "saving-throw"
+                ],
+                "text": "You can re-roll this check or saving throw.",
+                "title": "Halfling Luck"
```
TikaelSol, line 51:
```suggestion
                "text": "{item|description}",
                "title": "<span class=\"action-glyph\">f</span> {item|name}"
```
Text in a rule element should either just be a reference back to the item description and name, or localized.

packs/feats/halfling-luck.json
```diff
@@ -28,7 +28,33 @@
...
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "Note",
+                "outcome": [
+                    "failure",
+                    "criticalFailure"
+                ],
```
TikaelSol, line 37:
```suggestion
                "outcome": [
                    "failure"
                ],
```
This triggers on failure, not critical failure.

packs/feats/halfling-luck.json
```diff
@@ -28,7 +28,33 @@
...
+                    "saving-throw"
+                ],
+                "text": "You can re-roll this check or saving throw.",
+                "title": "Halfling Luck"
+            },
+            {
+                "key": "RollOption",
+                "option": "halfling-luck:{item|frequency.value}"
+            }
+        ],
```
TikaelSol, line 57: Clever, but the incredible luck and guiding luck feats make this a much bigger pain in the ass

packs/feats/halfling-luck.json
Joshmvii7: > Trigger You fail a skill check or saving throw.

I believe the RAW is that critical failures are a subset of failure that without a distinction excepting them also apply to things like Halfling Luck. 

From page 8 of Player Core
> Once a check is rolled, the GM compares the result to a
> target number called the difficulty class (DC) to determine
> the outcome. If the result of the check is equal to or greater
> than the DC, the check is successful. If it is less, the check is
> a failure.

I would expect if critical failure were meant to be excluded it would be like Dubious Knowledge and other examples that specify _When you fail (but don’t critically fail)_

Probably the biggest indicator that this is intended are these two feats as the 2nd one makes it clear that the first one (which uses the same language as Halfling Luck) works on fail and critical fail.

image
image

packs/feats/halfling-luck.json
Joshmvii7: I didn't even think about those future feats, looking at them I can see how they will make it much more difficult, hah.

Joshmvii7:
I'm fine with just getting rid of the roll option and the predicate on it for showing up. Like you said, you wouldn't want it to cause problems later. Is that something you want me to change and re-submit it, or something you fine people will do?

TikaelSol:
Treating this as superseded by #13480

PR #11783, opened 2023-11-26
Add basic Astral and Greater Astral Rune support

avagdumortesin:
Closes https://github.com/foundryvtt/pf2e/issues/11479


avagdumortesin:
I think I'll have to modify `src/module/system/damage/damage.ts:90` to add an `or` to have it check for astral, ghost touch, or greater astral, but I'm not entirely certain how to do that. Moving to a draft for now.

PR #11787, opened 2023-11-26
Fix scroll and wand trick magic item skill selection from spell tradition

nikolaj-a:


stwlam:
Are scrolls and wands not gaining the traits of their spells?

nikolaj-a:
> Are scrolls and wands not gaining the traits of their spells?

Not the traditions apparently. Which kind of makes sense since that decision happens upon casting for spells with multiple traditions.

PR #11793, opened 2023-11-27
Update spellcasting entries to change proficiency ranks

stwlam:
Also move display of spellcasting rank in proficiencies tab to own subsection

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -25,7 +25,7 @@ import { toggleWeaponTrait } from "@item/weapon/helpers.ts";
 import { BaseWeaponType, WeaponGroup } from "@item/weapon/types.ts";
 import { WEAPON_CATEGORIES } from "@item/weapon/values.ts";
 import { DropCanvasItemDataPF2e } from "@module/canvas/drop-canvas-data.ts";
-import { PROFICIENCY_RANKS } from "@module/data.ts";
+import { ZeroToFour } from "@module/data.ts";
 import { SheetOptions, createSheetTags } from "@module/sheet/helpers.ts";
 import { eventToRollParams } from "@scripts/sheet-util.ts";
 import { craft } from "@system/action-macros/crafting/craft.ts";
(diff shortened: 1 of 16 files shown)


src/module/actor/character/sheet.ts
```diff
@@ -754,6 +754,35 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             }
         }
 
+        // Update all "normal" spellcasting entries
+        for (const select of htmlQueryAll<HTMLSelectElement>(html, "select[data-action=update-spellcasting-rank]")) {
```
CarlosFdez, line 758:
```suggestion
        for (const select of htmlQueryAll<HTMLSelectElement>(html, "select[data-action=change-spellcasting-rank]")) {
```

src/module/actor/character/sheet.ts
```diff
@@ -754,6 +754,35 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
+                        "Item",
+                        entries.map((e) => ({ _id: e.id, "system.proficiency.value": newRank })),
+                    );
+                }
+            });
+        }
+
+        // Update a single spellcasting entry that uses an alternative statistic
+        for (const select of htmlQueryAll<HTMLSelectElement>(
+            castingPanel,
```
CarlosFdez, line 773: why castingPanel instead of html like the other?

stwlam: These are only in the spellcasting panel

CarlosFdez: then should the other one also refer to castingPanel?

static/templates/actors/character/tabs/spellcasting.hbs
```diff
@@ -66,7 +66,11 @@
                         <section class="spell-ability">
                             <span class="spell-tradition">{{localize entry.statistic.label}}</span>
                             <div class="spellcasting-prof button-group">
-                                <select class="skill-proficiency pf-rank adjust-item-stat-select" data-item-property="system.proficiency.value" data-rank="{{entry.statistic.rank}}" data-dtype="Number">
+                                <select
+                                    class="skill-proficiency pf-rank"
+                                    data-action="{{#if entry.usesSpellProficiency}}update-spellcasting-rank{{else}}update-spellcasting-entry{{/if}}"
```
CarlosFdez, line 71: I think you can avoid a specific handler for the regular update if you use data-item-property

PR #11794, opened 2023-11-27
Allow setting the ability of a flatmodifer to a resolvable

esheyw:
eg `{actor|class.system.keyAbility.selected}`. Don't know what to do about the form input for that field being only a select, and I don't know where/how best to add back in the checking that its a valid ability string.


stwlam:
Not seeing a use case for this that makes it worth giving up the validation safety

Issue #11788, opened 2023-11-26
Disrupting runes haven't been renamed


Vendergast:
Additionally at the moment you cant pick Vitalizing as a property Rune.

![grafik](https://github.com/foundryvtt/pf2e/assets/134313750/fa552730-1b96-4470-ba1d-047e3f690e74)

PR #11799, opened 2023-11-27
Rename Disrupting Rune to Vitalizing Rune

avagdumortesin:
Will need a migration, but I'm not entirely sure how to make that work. Will partially fix https://github.com/foundryvtt/pf2e/issues/11788; migration will complete the fix.


stwlam:
Just change the localization for now so the merge isn't breaking

avagdumortesin:
I think that does what you asked, @stwlam.

PR #11814, opened 2023-11-27
Fix Spell Compendium action glyph sizing

j-bs:
I noticed that the latest changes to spell list in the compendium made the glyphs a bit too small.

Not sure how specific we want the CSS styles so happy for feedback there.

Before vs After fix:
pf2e-spell-compendium-action-glyph

### DIFF
--- a/src/styles/packs/_index.scss
+++ b/src/styles/packs/_index.scss
@@ -26,6 +26,10 @@
                     ul {
                         grid-column-start: 2;
                     }
+
+                    &.spell-browser > ul .action-glyph {
+                        font-size: 1.5em;
+                    }
                 }
             }
 


stwlam:
The glyph sizes are the same there as everywhere else

stwlam:
They aren't in the actions tab: those are the actual image icons for them. I think we're good with how it is now.

j-bs:
I know they're images in the Action tab; that's also what used to be in the spell list compendium which is what I'm comparing the current glyphs against.

I just think it's a shame to have tiny glyphs that can render as slightly malformed on a screen with modest DPI instead of slightly larger glyphs that have ideal clarity.

stwlam:
Are your screenshots from a monitor with modest DPI? They're not malformed at all in the left screenshot.

j-bs:
True, most of those glyphs aren't malformed, I was thinking of the similarly small-sized glyphs in the Spellcasting and Spell Preparation sheets. But the reaction glyph *is* quite pixelated in my original screenshot. To me the reaction glyph looks infinitely better at a slightly higher font size that can do the curve justice.

These are the other two sheets that have the smaller glyphs that I was comparing to but didn't show in the original (I also think that the glyphs shouldn't be bolded cos that also looks kinda shit but that's a slight tangent). The spell prep sheet has a font 1px font smaller and to me it definitely has some poor rendering.

image

PR #11831, opened 2023-11-28
Default `options.secrets` in `TextEditorPF2e#enrichHTML` to `game.user.isGM`

stwlam:


src/module/actor/sheet/item-summary-renderer.ts
```diff
@@ -66,7 +66,7 @@ export class ItemSummaryRenderer<TActor extends ActorPF2e> {
                 summary.hidden = true;
                 insertLocation?.after(summary);
 
-                const chatData = await item.getChatData({ secrets: actor.isOwner }, element.dataset);
+                const chatData = await item.getChatData({}, element.dataset);
```
CarlosFdez, line 69: Actually not certain about this one. The owner can see such secrets by editing items, and a mechanism for true secrets (gm notes) already exist.

stwlam: What is the use of showing secrets a player can see by editing, though?

PR #11840, opened 2023-11-28
Rename all instances of Negative Healing ability to "Void Healing"

### LINKED ISSUE #11838: Negative Healing ability name not changed, also possible damage resistance bug discovered
The text for the ability was changed correctly to reflect the changes in the remaster, but the name of the ability itself was not changed, despite being clearly called "void healing" inside of the text. Searching for Negative Healing in the compendium finds the ability, but searching for "Void Healing" does not.
image

Out of curiosity I tested whether a creature with Negative Healing could take Void damage and it could. I was using the Lich creature to test this, as it has a Void damage attack. It took the damage. 
image

Out of further curiosity, I tested the Lich's Cold resistance by having it take damage from it's own Cone of Cold spell. It took the full damage despite having Resistance 10 to Cold. (I healed the creature to full before testing)
image

I then searched for a creature with immunity to test out on and found the Ice Golem. I then had it take damage from the Cone of Cold. It took full damage despite being immune.
image

Initially I wondered if the trait Negative Healing didn't negate Void damage because of the renaming of the damage, because I vaguely remember creatures with this ability not being able to take Negative damage. But after testing it with resistances (which I clearly remember working before) and immunities (which I did not try before the remaster) I believe there might be an issue with the way damage interacts with resistances/immunities.

stwlam:
Closes #11838


TikaelSol:
Should also do this for the hp details

stwlam:
Already done

PR #11845, opened 2023-11-28
Fix recently adjusted container bulk localisation strings

j-bs:
This was just a minor error that was missed during the item bulk refactor: the modified localisation strings don't have the nested `Bulk` level.


stwlam:
Instead of changing the template, change en.json to match.

j-bs:
Done, wasn't sure which was preferred.

PR #11849, opened 2023-11-29
Move held (but not "equipped") weapons above worn/stowed ones

stwlam:


CarlosFdez:
Would it not be awkward if a weapon shuffles its position in the list when you draw it?

stwlam:
That happens right now

PR #11851, opened 2023-11-29
Add automation for One Shot, One Kill

pedrogrullada:
This technically isn't perfectly RAW, as someone could choose not to use the ability, or attack with a firearm other than the one they drew as part of the free action, but I figured I'd "shoot my shot", as it were.

### DIFF
--- a/packs/actions/one-shot-one-kill.json
+++ b/packs/actions/one-shot-one-kill.json
@@ -23,21 +23,23 @@
         },
         "rules": [
             {
-                "domain": "damage",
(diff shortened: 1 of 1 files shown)


stwlam:
Keep the toggle and include it in the subsequent predicate, but the rest can stay.

stwlam:
Oh, but using MAP as a proxy for number of attacks in a round. I'm afraid this is something we've decided to not do, as it yields incorrect results too often.

PR #11879, opened 2023-11-30
Fix adding kits to actors

azrazalea:
Broke by https://github.com/foundryvtt/pf2e/pull/11742


azrazalea:
I have zero idea if this is where it should be fixed but it fixes it for me so thought I'd share @stwlam

stwlam:
Place it at the creature level

PR #11964, opened 2023-12-02
Add strength modifier addition/upgrade to Assume Earth's Mantle stance

Daniel-V1:
This seems to work, but I totally understand if it needs to be cleaned up or isn't in the right place etc. Please let me know what can be fixed or where this might actually need to go if it's not here. 
Stolen/adapted from https://discord.com/channels/880968862240239708/880969253765935174/1166941805535842324 with appreciation.

### DIFF
--- a/packs/feat-effects/stance-assume-earths-mantle.json
+++ b/packs/feat-effects/stance-assume-earths-mantle.json
@@ -53,6 +53,12 @@
                 ],
                 "type": "circumstance",
                 "value": 2
+            },
+            {
+                "key":"ActiveEffectLike",
+                "path":"system.abilities.str.mod",
+                "mode":"add",
+                "value":"ternary(lt(@actor.abilities.str.mod, 4), 4-@actor.abilities.str.mod, 1)"
             }
         ],
         "start": {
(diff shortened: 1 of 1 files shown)


stwlam:
Rather than attempting to handle partial boosts, this would be better done with a pair of AE-likes that add and then upgrade.

stwlam:
Like so:
```json
{
    "key": "ActiveEffectLike",
    "path": "system.abilities.str.mod",
    "mode": "add",
    "value": 1
}
 ```
 ```json
{
    "key": "ActiveEffectLike",
    "path": "system.abilities.str.mod",
    "mode": "upgrade",
    "value": 4
}
 ```

Daniel-V1:
Oh so just skip the predicate entirely. I guess that works because if it's less than 4 then the +1 won't matter, and if it's more then the upgrade won't take it back down. The ternary seems a little more intuitive as a translation the written rules, vs the multiple rules requiring a bit more knowledge of the pf2e system and rule priorities but :shrug:. I'll change it when I'm next at my computer. 
Thanks for the pointers and for the awesome system!  

    On Sunday, December 3, 2023 at 05:21:00 PM EST, Shark that walks like a man ***@***.***> wrote:  
 
 


Like so:
{
    "key": "ActiveEffectLike",
    "path": "system.abilities.str.mod",
    "mode": "add",
    "value": 1
}{
   "key": "ActiveEffectLike",
   "path": "system.abilities.str.mod",
   "mode": "upgrade",
   "value": 4
}

PR #12000, opened 2023-12-03
Add support for PC1 Custom Mixed Heritages

SpartanCPA:
image


SpartanCPA:
The GrantItems don't properly replace, it just adds the Aiuvarin and Dromaar heritages _as well_

stwlam:
This sort of thing shouldn't be rule-element-based.

PR #12014, opened 2023-12-04
Migrate instances of "undercommon" to "sakvroth"

stwlam:


TikaelSol (approved):

src/module/migration/migrations/904-undercommon-to-sakvroth.ts
```diff
@@ -0,0 +1,25 @@
+import { ActorSourcePF2e } from "@actor/data/index.ts";
+import { ItemSourcePF2e } from "@item/base/data/index.ts";
+import { recursiveReplaceString } from "@util";
+import { MigrationBase } from "../base.ts";
+
+/** Reorganize bulk data on physical items into single object */
```
CarlosFdez, line 6:
```suggestion
```

Probably one of those that probably doesn't benefit from a comment, but if you think it do you can put what you think should go here instead

PR #12062, opened 2023-12-06
Fix description for empower ley line (replace vitality/void with positive/negative)

MSAbaddon:


stwlam:
This is a break, not a fix.

PR #12074, opened 2023-12-07
Only show healing buttons for "damage" rolls that include healing

### LINKED ISSUE #11887: Feature Request: change some Splash damage buttons in the damage chat card
Currently, when rolling damage for something like Acid Splash, the chat message that appears looks like this:

image

The "Splash" damage has three useless buttons.  The Damage button is useful (to apply splash damage to everyone in the area), but the other buttons are never going to get used:
- Damage - useful
- Half - irrelevant, there's no basic save for splash damage
- Double - irrelevant, there's no basic save for splash damage
- Block - useful (very rarely), because physical splash damage can be blocked by shields, and magical splash can be blocked by the Shield cantrip
- Heal - irrelevant, there's nothing that applies a "splash heal"

So, **1 - remove these buttons**.

I would also like to **2 - change the splash damage icon** from a cracked heart to a burst, just because the text is different so the icon may as well be different too.

image

Finally, **3 - prevent splash damage from being double-able**.  This one is a sort of a bugfix:

image

When clicking the Double button of the normal damage roll (which includes the splash damage inside it), the splash value should not be doubled.  Splash damage does not double on a crit, and the "Double" button is basically the standard way to apply critical damage for spells that deal splash damage (i.e. Acid Splash) - such spells don't have a Critical button.

stwlam:
Closes #11887

image

### DIFF
--- a/src/module/item/spell/document.ts
+++ b/src/module/item/spell/document.ts
@@ -660,7 +660,7 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
         const isAreaEffect = !!this.system.area?.value;
         if (isAreaEffect) options.add("area-effect");
 
-        if (Object.values(this.system.damage).some((d) => d.kinds.has("damage"))) {
+        if (this.damageKinds.has("damage")) {
             options.add("damaging-effect");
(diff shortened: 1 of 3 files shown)


stwlam:
It has both

PR #12102, opened 2023-12-08
Fix regression with item chat and opening backpacks

CarlosFdez:
Item summaries are still broken in the inventory. I'm ok with either bringing back the item-name css class or adding a new selector. I'll let you pick your poison for that one.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -254,7 +254,7 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
         this.activateInventoryListeners(inventoryPanel);
 
         // Item chat cards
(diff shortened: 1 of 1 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -254,7 +254,7 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
         this.activateInventoryListeners(inventoryPanel);
 
         // Item chat cards
-        for (const element of htmlQueryAll(html, ".item[data-item-id] .item-image, .item[data-item-id] .item-chat")) {
+        for (const element of htmlQueryAll(html, "[data-item-id] .item-image, [data-item-id] .item-chat")) {
```
stwlam, line 257: Attribute selectors aren't indexed in the DOM, so they should never be used in isolation.

CarlosFdez: You'll have to restore item I believe then. Good point.

PR #12165, opened 2023-12-11
Allow suppression of ui notifications in refreshFromCompendium

esheyw:
For the use case of batch calling this method on many items, where a summary is being provided. not having several tens of notifications to click through is a better user experience.

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -374,7 +374,7 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         }
 
         await this.update(updates, { diff: false, recursive: false });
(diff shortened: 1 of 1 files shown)


src/module/item/base/document.ts
```diff
@@ -373,7 +380,7 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         }
 
         await this.update(updates, { diff: false, recursive: false });
-        ui.notifications.info(`Item "${this.name}" has been refreshed.`);
+        if (options.notify) ui.notifications.info(`Item "${this.name}" has been refreshed.`);
```
stwlam, line 383: Limit it to this

PR #12171, opened 2023-12-12
Add skill feat grant to rogue dedication

CarlosFdez:


packs/feats/rogue-dedication.json
```diff
@@ -71,6 +71,22 @@
             {
                 "key": "GrantItem",
                 "uuid": "Compendium.pf2e.classfeatures.Item.Surprise Attack"
+            },
+            {
+                "adjustName": false,
+                "choices": {
+                    "filter": [
+                        "item:trait:skill"
```
stwlam, line 79: Maybe good to limit it to general skill feats so the list isn't flooded

CarlosFdez: Good call. I copied it from the Skill Mastery feat actually, so it might be wise to do it there as well.

....well actually.... are such feats limited to the general skill feats?

PR #12173, opened 2023-12-12
Fix NPC strike descriptions collapsing after the first update

CarlosFdez:


static/templates/actors/npc/partials/attack.hbs
```diff
@@ -57,4 +43,10 @@
             </div>
         {{/if}}
     </div>
+
+    {{#if action.description}}
+        <div class="description">
```
stwlam, line 48:
```suggestion
        <div class="description item-summary">
```

CarlosFdez: Unfortunately that causes the save and restore state to render it hidden.

PR #12185, opened 2023-12-12
Micro-optimize some phases of actor data preparation

stwlam:


CarlosFdez:
No longer a micro-optimization by this point. Hell yeah.

PR #12202, opened 2023-12-13
Move item summary item retrieval to method

CarlosFdez:
Preparatory work to have certain items be just sheet data.


CarlosFdez:
Sure, I'll do that tomorrow when I wake up.

CarlosFdez:
I've chosen to backtrack on adding item summary listeners to the global click listeners. I think we need to avoid collisions as much as possible. This change is all I really need.

PR #12207, opened 2023-12-13
Remove erroneous attack traits from some spells

pedrogrullada:
and remove some seemingly accidental bolded text description here and there.


packs/spells/vector-screen.json
```diff
@@ -25,12 +25,14 @@
...
                         "areaType": "square",
+                        "type": "burst",
                         "value": "10"
                     }
                 },
                 "6": {
                     "area": {
                         "areaType": "square",
+                        "type": "burst",
                         "value": "10"
```
TikaelSol, line 36: This looks like a bug in the heightening overlay. @CarlosFdez any thoughts?

stwlam: I think it's legacy data that slipped in. It's be good to remove it.

pedrogrullada: Anything I should do with this? It came out like that by itself so I assumed it was something the system did when I unlocked the compendium to edit the spell. It looked correct according to the spell's description, so I left it alone.

CarlosFdez: Value and type are the correct fields, so areaType is legacy data. I don't see any reference to it in the templates.

packs/spells/vector-screen.json
```diff
@@ -25,12 +25,14 @@
                 "5": {
                     "area": {
                         "areaType": "square",
```
stwlam, line 27:
```suggestion
```

packs/spells/vector-screen.json
```diff
@@ -25,12 +25,14 @@
...
+                        "type": "burst",
                         "value": "10"
                     }
                 },
                 "6": {
                     "area": {
                         "areaType": "square",
```
stwlam, line 34:
```suggestion
```

PR #12245, opened 2023-12-15
Move handling of "set as initiative" button click to `ChatLogPF2e`

stwlam:


CarlosFdez:
Looks alright, and the less events we are creating the better, though I don't know if that listener can really be qualified as a listener anymore.

stwlam:
Yes, that whole directory could stand to be renamed.

CarlosFdez (approved):

PR #12282, opened 2023-12-16
Infer focus cantrip status by lack of traditions

stwlam:


src/module/item/spell/document.ts
```diff
@@ -150,11 +150,12 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
...
 
     get isCantrip(): boolean {
-        return this.traits.has("cantrip") && !this.isRitual;
+        return this.system.traits.value.includes("cantrip");
     }
 
     get isFocusSpell(): boolean {
-        return this.traits.has("focus");
+        const traits = this._source.system.traits;
+        return traits.traditions.length === 0 && (this.isCantrip || traits.value.includes("focus"));
```
CarlosFdez, line 158:
```suggestion
        return (traits.traditions.length === 0 && this.isCantrip) || traits.value.includes("focus");
```
if you don't want to give tikael a heart attack in the meantime

src/module/migration/migrations/912-rm-focus-trait-focus-cantrips.ts
```diff
@@ -0,0 +1,28 @@
...
+    override async updateItem(source: ItemSourcePF2e): Promise<void> {
+        if (source.type !== "spell") return;
+
+        const magicTraditions: Set<string> = MAGIC_TRADITIONS;
+        const traits = source.system.traits;
+        const misplacedMagicTraditions = traits.value.filter((t): t is MagicTradition => magicTraditions.has(t));
+        traits.traditions.push(...misplacedMagicTraditions);
+        traits.value = traits.value.filter((t) => !magicTraditions.has(t));
+
+        if (traits.value.includes("focus")) traits.traditions = [];
```
CarlosFdez, line 20: This should maybe instead check if traditions are blank, rather than setting traditions to blank, and then remove the focus trait if so.

CarlosFdez (dismissed):

PR #12342, opened 2023-12-17
Fix "yellow" result in Rainbow Fumarole spell

manfred7168:
Wrong damage at choice 3 yellow.


stwlam:
Would you mind clarifying what is being corrected? PR titles should also be in [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood).

PR #12397, opened 2023-12-19
Add Fly Speed to Constant Levitation Feat

Intervencion:
IDK what the thoughts of the team are on this one. Feat gives you the benefits of a constant Fly spell
image


stwlam:
It should be a reevaluated grant since it can be dispelled but then come back.

PR #12422, opened 2023-12-20
Add REs for Doubling Rings

DocSchlock:
This adds REs to the Doubling Rings, both versions, using toggleable roll options.

The only thing not supported is the Greater version's transfer of Property runes, although that may be doable in the future when the rune system gets reworked.


TikaelSol:
This is not the way to go about this one, we'll have to come up with something more clever that does it all properly since there's multiple items of this type that would need done.

DocSchlock:
No problem!

I'll abandon this one for now.

PR #12470, opened 2023-12-23
Switch references to "slot levels" (or ranks) to "slot groups"

stwlam:


static/templates/actors/partials/spell-collection.hbs
```diff
@@ -1,48 +1,69 @@
...
+                    data-item-type="spell-slot-group"
+                    data-group-id="{{group.id}}"
                     data-tooltip-class="pf2e"
                 >
                     <div class="item-name">
-                        <h3>{{localize section.label}}</h3>
+                        <h3>{{localize group.label}}</h3>
 
-                        {{#if (and section.isCantrip (not section.uses))}}
+                        {{#if (and (eq group "cantrips") (not group.uses))}}
```
CarlosFdez, line 16: I don't think group ever equals "cantrips"? Seems to be an object.

static/templates/actors/partials/spell-collection.hbs
```diff
@@ -57,16 +78,21 @@
...
+                                    data-group-id="{{group.id}}"
+                                    data-location="{{entry.id}}"
+                                ><i class="fa-solid fa-fw fa-plus"></i></a>
                                 <a
                                     data-action="browse-spells"
                                     data-tooltip="PF2E.OpenSpellBrowserTitle"
-                                    {{#if (eq section.level 0)}}data-category="cantrip"{{/if}}
-                                    data-rank="{{#if (or (eq entry.category "ritual") (eq rank 0))}}10{{else}}{{rank}}{{/if}}"
+                                    {{#if (eq group.id "cantrips")}}data-category="cantrip"{{/if}}
+                                    data-rank="{{#if (or (eq entry.category "ritual") (eq group.id "cantrips"))}}{{entry.maxRank}}{{else}}{{group.id}}{{/if}}"
```
CarlosFdez, line 95: It might be cleaner to have rank or castRank as part of the spell sheet data itself to clean this up a tad. The less if statements we have here the better.

src/module/item/spellcasting-entry/collection.ts
```diff
@@ -107,11 +113,11 @@ class SpellCollection<TActor extends ActorPF2e, TEntry extends BaseSpellcastingE
     }
 
     /** Saves the prepared spell slot data to the spellcasting entry  */
-    async prepareSpell(spell: SpellPF2e, slotRank: number, spellSlot: number): Promise<TEntry | undefined> {
+    async prepareSpell(spell: SpellPF2e, groupId: SpellSlotGroupId, slotId: number): Promise<TEntry | undefined> {
         this.#assertEntryIsDocument(this.entry);
 
-        if (spell.baseRank > slotRank && !(slotRank === 0 && spell.isCantrip)) {
-            const targetRankLabel = game.i18n.format("PF2E.Item.Spell.Rank.Ordinal", { rank: ordinalString(slotRank) });
+        if (groupId !== "cantrips" && spell.baseRank > groupId) {
```
CarlosFdez, line 119: So I think before you couldn't prepare cantrips in non-cantrip slots. Is this still true after this change?

src/module/item/spellcasting-entry/collection.ts
```diff
@@ -294,24 +282,24 @@ class SpellCollection<TActor extends ActorPF2e, TEntry extends BaseSpellcastingE
         // Handle signature spells, we need to add signature spells to each spell level
         if (this.entry.isSpontaneous || isFlexible) {
             for (const spell of signatureSpells) {
-                for (const result of results) {
-                    if (spell.baseRank > result.level) continue;
```
CarlosFdez, line 298: The check here is to prevent fireball from being added to a 1st / 2nd level slot as a signature spell. Does that still work?

PR #12489, opened 2023-12-24
Fix usage of long air repeater

stwlam:
This one seems to be in the "obvious error with obvious correction" category


TikaelSol:
Probably, given the description but I'd still want Paizo to fix it.

stwlam:
https://paizo.com/threads/rzs43kh6?Long-air-repeater

TMunsie:
I think that Michael Sayre’s clarification is certainly sufficient and clear :)

PR #12503, opened 2023-12-25
Add a Fortissimo Composition spell effect

pedrogrullada:
I hope this is up to snuff. This is my attempt at consolidating Fortissimo Composition (Inspire Heroics) into a single effect. It will address #10305 by way of granting Rallying Anthem's effect itself, too. The Grant Item is not really necessary, but it does make the workflow easier for the user.


TikaelSol:
I don't know that this is how we want to go about it, will need to think if there's some cleaner way to do it.

pedrogrullada:
Yeah, it is a bit sloppy but I couldn't think of a cleaner way. I figured it'd be an improvement over the current way. Cheers!

PR #12526, opened 2023-12-27
Restore converting selector to array for fresh flat modifier forms

CarlosFdez:
This prevents empty array from getting cleared out by the form's clearing behavior. FlatModifier's without a selector property now complain about selector cannot be undefined instead of "must have at least 1".


src/module/rules/rule-element/flat-modifier.ts
```diff
@@ -195,7 +196,7 @@ interface FlatModifierRuleElement
 
 type FlatModifierSchema = RuleElementSchema & {
     /** All domains to add a modifier to */
-    selector: ArrayField<StringField<string, string, true, false, false>>;
+    selector: ArrayField<StringField<string, string, true, false, false>, string[], string[], true>;
```
stwlam, line 199:
```suggestion
    selector: ArrayField<StringField<string, string, true, false, false>, string[], string[], true, false, false>;
```

PR #12544, opened 2023-12-28
Increase visibility of show image hover icon

CarlosFdez:
and fix npc portrait borders being shown when they shouldn't.

image


stwlam:
Is that more visible? And what about up against the default NPC icon?

CarlosFdez:
image

PR #12566, opened 2023-12-28
Overlap deity icon over value

CarlosFdez:
image


src/module/actor/character/sheet.ts
```diff
@@ -88,7 +88,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         const options = super.defaultOptions;
         options.classes = [...options.classes, "character"];
         options.width = 750;
-        options.height = 800;
+        options.height = 750;
```
stwlam, line 91: !

CarlosFdez: YES

PR #12582, opened 2023-12-29
Refactor getDefaultVariant method on BaseAction to be synchronous

nikolaj-a:
Change `BaseAction#getDefaultVariant` from being asynchronous to being synchronous since the asynchronicity was not necessary and future synchronous use cases are anticipated.

### DIFF
--- a/src/module/actor/actions/base.ts
+++ b/src/module/actor/actions/base.ts
@@ -145,41 +145,44 @@ abstract class BaseAction<TData extends BaseActionVariantData, TAction extends B
         return new Collection(variants);
     }
 
-    protected async getDefaultVariant(options?: { variant?: string }): Promise<TAction> {
+    protected getDefaultVariant(options?: { variant?: string }): TAction {
         const variants = this.variants;
         if (options?.variant && !variants.size) {
-            const reason = game.i18n.format("PF2E.ActionsWarning.Variants.None", {
+            throw game.i18n.format("PF2E.ActionsWarning.Variants.None", {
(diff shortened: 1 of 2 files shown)


src/module/actor/actions/base.ts
```diff
@@ -145,41 +145,47 @@ abstract class BaseAction<TData extends BaseActionVariantData, TAction extends B
...
-    protected async getDefaultVariant(options?: { variant?: string }): Promise<TAction> {
+    protected getDefaultVariant(options?: { variant?: string }): TAction {
         const variants = this.variants;
         if (options?.variant && !variants.size) {
             const reason = game.i18n.format("PF2E.ActionsWarning.Variants.None", {
                 action: game.i18n.localize(this.name),
                 variant: options.variant,
             });
-            return Promise.reject(reason);
+            throw new Error(reason);
```
stwlam, line 155: use `ErrorPF2e`

nikolaj-a: I'm wondering if it makes more sense to just throw the string in this particular case. It's always caught bye the promise in the other methods calling it and is really meant to eventually go in a UI notification instead of the logs.

stwlam: for that we sometimes do `ui.notifications.error(ErrorPF2e("abcdefg").message)`;

nikolaj-a: The Error never makes it out of the class - it is turned into a rejected promise. With the error I have to do an instanceof check to satisfy TS before extracting the message.

PR #12584, opened 2023-12-29
Add REs for Staff of Healing bonus HP

DocSchlock:
I went very specific on the predicate - item:slug:heal may be good enough? It works fine as is though.

### DIFF
--- a/packs/equipment/staff-of-healing-greater.json
+++ b/packs/equipment/staff-of-healing-greater.json
@@ -51,7 +51,19 @@
         "reload": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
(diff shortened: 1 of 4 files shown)


packs/equipment/staff-of-healing-greater.json
```diff
@@ -51,7 +51,19 @@
         "reload": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "item:type:spell",
+                    "item:trait:healing",
```
stwlam, line 59:
```suggestion
```

PR #12586, opened 2023-12-29
Add Holy and correct duration to Amplifying Touch Effect

DocSchlock:
> In the Amplifying Touch feat, replace "deals 1 additional good damage on all their Strikes" with "deals 1 additional spirit damage on all their Strikes". Add the following sentence at the end of the feat's description, "In addition, all their Strikes are holy until the end of their next turn."

Per the remaster compatibility errata. 

I tested and adding a weapon-trait of Holy seems to work with any Strike. Not 100% if that's the correct way to do that though.

### DIFF
--- a/packs/spell-effects/spell-effect-amplifying-touch.json
+++ b/packs/spell-effects/spell-effect-amplifying-touch.json
@@ -4,10 +4,10 @@
     "name": "Spell Effect: Amplifying Touch",
     "system": {
         "description": {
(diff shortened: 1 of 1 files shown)


packs/spell-effects/spell-effect-amplifying-touch.json
```diff
@@ -38,6 +38,12 @@
                 "key": "FlatModifier",
                 "selector": "strike-damage",
                 "value": 1
+            },
+            {
+                "key": "AdjustStrike",
+                "mode": "add",
+                "property": "weapon-traits",
```
stwlam, line 45:
```suggestion
                "property": "traits",
```

PR #12592, opened 2023-12-30
Finish implementation of roll-option-toggle placement

stwlam:
image


static/templates/actors/army/sheet.hbs
```diff
@@ -202,10 +202,10 @@
             {{editor data.details.description target="system.details.description" button=true owner=owner editable=editable}}
         </fieldset>
 
-        <!-- TOGGLES -->
+        {{!-- TOGGLES --}}
         {{#if toggles}}
```
CarlosFdez, line 206:
```suggestion
        {{#if toggles.actions}}
```

since I don't think an empty object is falsy

PR #12593, opened 2023-12-30
Add `fp-cost` item alteration, consume focus points according to spell's cost

stwlam:


CarlosFdez (approved):

CarlosFdez:
Fundamentally I don't see an issue with this. actor.system.fpCost is a weird property name, and it seems clunky as a solution for the current known use cases (unless there's an action that consumes FP, then maybe). But the PR is sound.

Issue #12594, opened 2023-12-30
Scroll Thaumaturgy Feat should add Items Only Spellcasting Entry


apuma32:
When the Scroll Thaumaturgy Feat is added to a character sheet, 
The Items Only Spellcasting option should be added.
![image](https://github.com/foundryvtt/pf2e/assets/154711470/192cf5fb-2b70-4cfe-a373-9d6f768adf3f)
So that players do not receive the "$Player lacks the proficiency to use $scroll.name and does not have the trick magic item feat." message. and does not have to manually add the spellcasting entry.
-or- 
the toast message should include instruction that a player with this feet must manually add the spellcasting entry

stwlam:
Spellcasting entries in general aren't added automatically.

apuma32:
@stwlam is it possible to modify the scroll rules to allow casting when this feat is active? (as i assume it does when you have trick magic item)

stwlam:
Spellcasting entries aren't currently possible to grant.

PR #12603, opened 2023-12-30
Allow scroll thaumaturgy to cast scrolls

apuma32:
Allow a character with the scroll-thaumaturgy feat to cast spells from scrolls.
https://github.com/foundryvtt/pf2e/issues/12594


stwlam:
Sorry, but we're not adding hard-coded checks for specific feats.

PR #12630, opened 2023-12-30
Add `description` item alteration

stwlam:
An alternative to previously requested "item notes". This has a large types update, so I just added the alteration validator and data structure for it. Will follow up with rest of implementation if/when this PR is merged.


CarlosFdez:
This is pretty major and the upcoming patch is pretty massive. Unless you need this for one of your extra life incentives, any thoughts on delaying merge on this till the next cycle?

PR #12640, opened 2023-12-31
Add outcomes to the bravery note RE

DocSchlock:
Not sure "success" needs to be in there because the RE upgrades to Critical Success automatically, but left it in just in case.

### DIFF
--- a/packs/classfeatures/bravery.json
+++ b/packs/classfeatures/bravery.json
@@ -28,8 +28,7 @@
             {
                 "key": "Note",
                 "outcome": [
-                    "success",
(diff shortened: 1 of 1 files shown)


packs/classfeatures/bravery.json
```diff
@@ -27,6 +27,10 @@
         "rules": [
             {
                 "key": "Note",
+                "outcome": [
+                    "success",
+                    "criticalSuccess"
```
TikaelSol, line 32:
```suggestion
                    "success"
```

PR #12641, opened 2023-12-31
Add automation effects to various talismans

pedrogrullada:


stwlam:
Would you mind sitting on these? We'll soon have a way to utilize them without creating new effects.

pedrogrullada:
> Would you mind sitting on these? We'll soon have a way to utilize them without creating new effects.

Not at all, I'll close this one for now and wait for the new technology

PR #12721, opened 2024-01-03
Relax equipped check for invested non-worn items

CarlosFdez:
From rudimentary testing, rule elements don't trigger on items that are invested but not in the right usage status, which I believe is what we want.


src/module/item/physical/document.ts
```diff
@@ -123,7 +123,11 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
     get isInvested(): boolean | null {
         const traits: Set<string> = this.traits;
         if (!traits.has("invested")) return null;
-        return this.isEquipped && this.isIdentified && this.system.equipped.invested === true;
+        return (
+            (this.isEquipped || this.system.usage.type !== "worn") &&
```
stwlam, line 127: Exclude stowed

PR #12733, opened 2024-01-04
Add support for inline /act syntax for using actions

nikolaj-a:
Add support for inline slash-style syntax, specifically `/act`, for actions.

Also give priority to the action objects instead of the old action functions when using both the old HTML syntax and the new slash syntax.
### Examples

`[[/act administer-first-aid variant=stop-bleeding dc=25]]`

`[[/act escape statistic=acrobatics dc=fortitude]]`

`[[/act grapple]]`

Examples

~~Depends on https://github.com/foundryvtt/pf2e/pull/13043 and https://github.com/foundryvtt/pf2e/pull/13047~~

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -18,6 +18,7 @@ import { USER_VISIBILITIES, UserVisibility, UserVisibilityPF2e } from "@scripts/
 import {
     createHTMLElement,
     fontAwesomeIcon,
+    getActionGlyph,
     htmlClosest,
     localizer,
     objectHasKey,
@@ -211,6 +212,8 @@ class TextEditorPF2e extends TextEditor {
         const [_match, inlineType, paramString, inlineLabel] = data;
 
         switch (inlineType) {
+            case "act":
+                return this.#createAction(data.groups?.slug ?? "", data.groups?.options ?? "", data.groups?.label);
             case "Check": {
                 const actor = options.rollData?.actor ?? item?.actor ?? null;
                 return this.#createCheck({ paramString, inlineLabel, item, actor });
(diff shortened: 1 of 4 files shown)


src/global.ts
```diff
@@ -80,7 +81,7 @@ interface GamePF2e
         UserPF2e
     > {
     pf2e: {
-        actions: Record<string, Function>;
+        actions: Record<string, Function> & Map<string, Action>;
```
stwlam, line 84: How is this both a `Map` and a `Record`?

src/module/apps/sidebar/chat-log.ts
```diff
@@ -8,9 +8,61 @@ import { TokenDocumentPF2e } from "@scene";
...
+    override _processMacroCommand(command: string, match: RegExpMatchArray): unknown {
+        if (match && match.length > 2 && match[1] === "act" && match.groups?.slug) {
+            const slug = match.groups.slug;
+            const action = game.pf2e.actions.get(slug ?? "");
+            if (action) {
+                const options: Partial<ActionUseOptions> = {};
+                for (const option of (match.groups.options ?? "").split(/\s+/)) {
+                    const [key, value] = option.split("=");
+                    switch (key) {
+                        case "dc": {
```
stwlam, line 40: The example shows a defense (fortitude) being used instead of a number

src/module/apps/sidebar/chat-log.ts
```diff
@@ -8,9 +8,61 @@ import { TokenDocumentPF2e } from "@scene";
...
+                            const dc = Number(value);
+                            options["difficultyClass"] = Number.isInteger(dc) ? { value: dc } : value;
+                            break;
+                        }
+                        case "stat": {
+                            options["statistic"] = value;
+                            break;
+                        }
+                        case "traits": {
```
stwlam, line 49: Let's use `options` here since it's really for feeding in roll options

src/module/apps/sidebar/chat-log.ts
```diff
@@ -8,9 +8,61 @@ import { TokenDocumentPF2e } from "@scene";
...
+        return super.parse(message);
+    }
+
+    /**
+     * Process messages which execute a macro.
+     * @param {string} command  The chat command typed.
+     * @param {RegExpMatchArray} match  The RegExp matches.
+     * @private
+     */
+    override _processMacroCommand(command: string, match: RegExpMatchArray): unknown {
```
stwlam, line 31:
```suggestion
    protected override _processMacroCommand(command: string, match: RegExpMatchArray): Promise<void> {
```

src/module/apps/sidebar/chat-log.ts
```diff
@@ -8,9 +8,61 @@ import { TokenDocumentPF2e } from "@scene";
...
+        }
+        return super.parse(message);
+    }
+
+    /**
+     * Process messages which execute a macro.
+     * @param {string} command  The chat command typed.
+     * @param {RegExpMatchArray} match  The RegExp matches.
+     * @private
+     */
```
stwlam, line 30: Instead of repeating the upstream jsdocs, write ones relevant to the override

types/foundry/client/application/sidebar/sidebar-tab/chat-log.d.ts
```diff
@@ -139,6 +139,14 @@ declare class ChatLog<TChatMessage extends ChatMessage = ChatMessage> extends Si
         createOptions: ChatMessageModificationContext,
     ): Promise<void>;
 
+    /**
+     * Process messages which execute a macro.
+     * @param {string} command  The chat command typed.
+     * @param {RegExpMatchArray} match  The RegExp matches.
+     * @private
+     */
+    protected _processMacroCommand(command: string, match: RegExpMatchArray): unknown;
```
stwlam, line 148:
```suggestion
    protected _processMacroCommand(command: string, match: RegExpMatchArray): Promise<void>;
```

src/module/apps/sidebar/chat-log.ts
```diff
@@ -8,9 +8,61 @@ import { TokenDocumentPF2e } from "@scene";
...
+    override _processMacroCommand(command: string, match: RegExpMatchArray): unknown {
+        if (match && match.length > 2 && match[1] === "act" && match.groups?.slug) {
```
stwlam, line 32:
```suggestion
        if (match[1] === "act" && match.groups?.slug) {
```

stwlam (changes requested):
It looks like we'd be better off overriding `ChatLog.parse` and `ChatLog#processMessage`

src/global.ts
nikolaj-a: It has the old action functions directly on the object (the `Record` part) _and_ it’s an instance of a Foundry `Collection`. I couldn’t get the types right for the Collection part, though.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: I used two different examples on purpose - one with a number and one with a string.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Hmmm, it really should support both rollOptions and traits 🤔 Maybe I need to take another look to ensure it works properly for both the old action functions and the new action objects.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Actually it returns whatever the macro returns. I think the Foundry API docs are inconsistent with the actual code.

nikolaj-a:
> It looks like we'd be better off overriding `ChatLog.parse` and `ChatLog#processMessage`

That was my initial attempt. I gave up on that once I basically copied the whole thing and added just another case for the switch in there.

It works, but we’d have to maintain it ourselves since it’s not currently set to be changed in core.

types/foundry/client/application/sidebar/sidebar-tab/chat-log.d.ts
nikolaj-a: Moved to https://github.com/foundryvtt/pf2e/pull/12772 instead.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Moved to https://github.com/foundryvtt/pf2e/pull/12772 instead.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Moved to https://github.com/foundryvtt/pf2e/pull/12772 instead.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Moved to https://github.com/foundryvtt/pf2e/pull/12772 instead.

src/module/apps/sidebar/chat-log.ts
nikolaj-a: Moved to https://github.com/foundryvtt/pf2e/pull/12772 instead.

src/global.ts
nikolaj-a: I managed to change the type to `Record<string, Function> & Collection<Action>`. Not sure why it didn't work when I tried the first time 🤷

The long-term goal is for it to be only a Collection and all the old action functions to go away. They are currently used by the action macros and even a few places in the system itself, so we'd need some sort of deprecation period and/or migration.

nikolaj-a:
Rough draft for the Style Guide:

# Inline Action Syntax

```
[[/act action-slug key1=value1 key2=value2]]
```

```
[[/act grapple]]
```

```
[[/act administer-first-aid variant=stabilize]]
```

```
[[/act perform variant=acting difficulty-class=19]]
```

```
[[/act escape statistic=acrobatics difficulty-class=fortitude]]
```

## Supported parameters

| Parameter        | Alias        | Values                   | Examples        |
| ---------------- | ------------ | ------------------------ | --------------- |
| difficulty-class | dc           | number or statistic slug | 25 or fortitude |
| statistic        | stat, skill  | statistic slug           | perception      |
| variant          |              | variant slug             | stop-bleeding   |

## What is a "statistic slug"?

A language-agnostic representation of a creature statistic. For most actions the supported statistics are perception, any saving throw, and any skill.

perception
fortitude
reflex
will
acrobatics
arcana
athletics
crafting
deception
diplomacy
intimidation
medicine
nature
occultism
performance
religion
society
stealth
survival
thievery

src/module/system/text-editor.ts
```diff
@@ -347,6 +350,133 @@ class TextEditorPF2e extends TextEditor {
...
+                    const STATISTIC_LABELS: Record<string, string> = {
+                        perception: "PF2E.PerceptionLabel",
+                        ...CONFIG.PF2E.saves,
+                        ...CONFIG.PF2E.skillList,
+                        unarmed: "PF2E.TraitUnarmed",
+                    };
+                    element.appendChild(document.createTextNode(" "));
+
+                    const details = document.createElement("span");
+                    if (dc && showDC && Number.isNumeric(dc)) {
```
stwlam, line 431: We should have our own helper for this: mind making something like `isIntegerString` that does a strict regex test?

nikolaj-a: Sure, but I think it's Foundry providing that `Number.isNumeric` function. Though I guess it's originally from jQuery.

Alternatively, there is something like `Number.isInteger(Number("10"))`.

stwlam (approved):

PR #12734, opened 2024-01-04
Use sheet actor for general tab swim/climb checks

CarlosFdez:


nikolaj-a:
It should probably also use the action objects instead of the old action functions.

```
game.pf2e.actions.get("swim").use({ event, actors: this.actor });
```

CarlosFdez:
Are you sure those are available?
image

nikolaj-a:
It's a [Collection](https://foundryvtt.com/api/classes/foundry.utils.Collection.html) of Action objects keyed by the sluggified action name. The contents of the collection is build from this list: https://github.com/foundryvtt/pf2e/blob/master/src/module/system/action-macros/index.ts#L161

CarlosFdez:
The type of the object in game.pf2e is Record.

PR #12738, opened 2024-01-04
Add metagame information setting to show or hide roll breakdowns

stwlam:


FeistyViking:
So looks like this change kinda sorta breaks Dice So Nice. Because the default now is to hide roll breakdowns, when that is the case players only see "?" dice from the Dice So Nice module. I think that has to do with the roll log entry now only having the result and hiding the d20 result as well as the modifiers. Is there a way to hide all the modifiers but still allow players to see the base dice result?

stwlam:
If they know the natural die result and the final total, they know the modifier.

FeistyViking:
Ah, right that does make sense. No way else to really hide that information.

I would imagine this will potentially come up as an issue that some folks submit, and since this change modified the new default metadata setting it will affect everyone that updates. It may be worthwhile to add a note about how this setting interacts with dicesonice (just since that is SUCH a popular module) or potentially change the default to having Show Roll Breakdowns checked.

PR #12762, opened 2024-01-05
Allow non-kingmaker worlds to enable kingdom and army sheets

CarlosFdez:
Supercedes https://github.com/foundryvtt/pf2e/pull/9922

### DIFF
--- a/src/module/system/settings/automation.ts
+++ b/src/module/system/settings/automation.ts
@@ -18,6 +18,7 @@ export class AutomationSettings extends SettingsMenuPF2e {
     protected static override get settings(): Record<ConfigPF2eListName, PartialSettingsData> {
         return {
             rulesBasedVision: {
+                prefix: "automation.",
                 name: CONFIG.PF2E.SETTINGS.automation.rulesBasedVision.name,
                 hint: CONFIG.PF2E.SETTINGS.automation.rulesBasedVision.hint,
                 default: true,
@@ -28,6 +29,7 @@ export class AutomationSettings extends SettingsMenuPF2e {
                 },
             },
             iwr: {
+                prefix: "automation.",
                 name: CONFIG.PF2E.SETTINGS.automation.iwr.name,
                 hint: CONFIG.PF2E.SETTINGS.automation.iwr.hint,
                 default: BUILD_MODE === "development",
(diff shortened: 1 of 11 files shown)


src/module/system/settings/automation.ts
```diff
@@ -18,6 +18,7 @@ export class AutomationSettings extends SettingsMenuPF2e {
     protected static override get settings(): Record<ConfigPF2eListName, PartialSettingsData> {
         return {
             rulesBasedVision: {
+                prefix: "automation.",
```
stwlam, line 21: Can there be a default prefix?

static/lang/en.json
```diff
@@ -3184,6 +3184,7 @@
                     "Name": "Spell Traits"
                 },
                 "Tabs": {
+                    "Campaign": "Campaign Settings",
```
stwlam, line 3187:
```suggestion
                    "Campaign": "Campaign",
```

static/lang/en.json
```diff
@@ -3089,8 +3089,8 @@
                 "Name": "Campaign Feats"
             },
             "CampaignType": {
-                "Hint": "Extended campaign rule type for parties",
-                "Name": "Campaign Type",
+                "Hint": "Extended campaign rule support for campaigns such as kingmaker",
+                "Name": "Campaign Specific Features",
```
stwlam, line 3093:
```suggestion
                "Hint": "Extended campaign rule support for campaigns such as Kingmaker",
                "Name": "Campaign-Specific Features",
```

src/module/system/settings/automation.ts
CarlosFdez: There can be, though given the number of places that have to prepend the prefix, as well as a helper, that seemed like the worse of two evils.

The other concern was how do we remove a prefix. Thinking now, `prefix: null` could work.

stwlam: If it's worse with a default, then that's fine.

static/templates/system/settings/homebrew.hbs
```diff
@@ -2,6 +2,7 @@
     <nav class="sheet-tabs tabs">
         <a class="item" data-tab="traits">{{localize "PF2E.SETTINGS.Homebrew.Tabs.Traits"}}</a>
         <a class="item" data-tab="languages">{{localize "PF2E.Actor.Creature.Language.Plural"}}</a>
+        <a class="item" data-tab="settings">{{localize "PF2E.SETTINGS.Homebrew.Tabs.Campaign"}}</a>
```
stwlam, line 5:
```suggestion
        <a class="item" data-tab="campaign">{{localize "PF2E.SETTINGS.Homebrew.Tabs.Campaign"}}</a>
```

PR #12787, opened 2024-01-06
Remove necessity for having multiple marshal's aura effects

stwlam:


pedrogrullada:
There's also a [Devrin's Cunning Stance](https://2e.aonprd.com/Feats.aspx?ID=4134) you might want to modify this for too

PR #12830, opened 2024-01-07
Fix overlapping images in compendium browser

jfn4th:
Also slightly reduces padding and removes unnecessary margin from rarity tags in compendium browser to keep within 2em height.

Before:
Screenshot 2024-01-07 at 3 59 06 PM

After
Screenshot 2024-01-07 at 4 01 15 PM

Could also set `aspect-ratio: 1/1;` and `object-fit: contain;` on the image itself if keeping a uniform height across all rows is desired, though some images will be hard to see.
Screenshot 2024-01-07 at 4 03 18 PM

### DIFF
--- a/src/styles/packs/_tab.scss
+++ b/src/styles/packs/_tab.scss
@@ -135,7 +135,7 @@
             > * {
                 align-items: center;
                 display: flex;
-                height: 2rem;
+                min-height: 2rem;
                 justify-content: center;
             }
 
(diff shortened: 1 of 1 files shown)


stwlam:
How does everything look in item results now?

jfn4th:
Screenshot 2024-01-07 at 4 59 44 PM

Looking through equipment, feats, spells, etc. everything seems to look fine. Items with two lines of text get 2px taller, looks like the text was previously overflowing into the padding. As an added benefit, text remains readable when wrapping to 3+ lines on a narrow window.

Before: 
Screenshot 2024-01-07 at 4 53 42 PM

After: 
Screenshot 2024-01-07 at 4 53 48 PM

PR #12834, opened 2024-01-08
Update REs in Burn It! feat to match text in Player Core

stwlam:


packs/feats/burn-it.json
```diff
@@ -48,14 +48,6 @@
...
                 "damageType": "fire",
                 "key": "FlatModifier",
-                "predicate": [
-                    {
-                        "or": [
-                            "item:trait:alchemical",
-                            "item:type:spell"
-                        ]
-                    }
-                ],
```
TikaelSol, line 58: `You also gain a +1 status bonus to any persistent fire damage you deal.` We'd need to keep a predicate for the damage type and category.

stwlam: It'll work without

stwlam: (note there wasn't one to begin with)

PR #12836, opened 2024-01-08
Fix setting Champion class DC on Champion Dedication

TikaelSol:


packs/feats/champion-dedication.json
```diff
@@ -71,7 +71,7 @@
             {
                 "key": "ActiveEffectLike",
                 "mode": "override",
-                "path": "system.proficiencies.classDCs.champion",
+                "path": "system.proficiencies.classDCs.champion.ability",
```
stwlam, line 74:
```suggestion
                "path": "system.proficiencies.classDCs.champion.attribute",
```

PR #12914, opened 2024-01-10
Convert hazard-sheet gap/margin/padding to use em/rem

stwlam:


src/styles/actor/hazard/_sidebar.scss
```diff
@@ -102,17 +103,16 @@
...
         }
     }
 
-    .rollable {
-        &:hover {
-            text-shadow: 0 0 8px var(--color-shadow-primary);
-            i {
-                @include die-spin;
-            }
+    .rollable :hover {
```
CarlosFdez, line 111: should this be `.rollable:hover` instead?

PR #12932, opened 2024-01-10
Update Mask of Mercy action cost

### LINKED ISSUE #12931: Mask of mercy cost 1 free action instead of 1 action to activate
Hi, the item "mask of mercy" should have a 1 action cost to activate the effect as by treasure vault new version.

jfn4th:
Closes #12931 

Screenshot 2024-01-10 at 10 21 11 AM

### DIFF
--- a/packs/equipment/mask-of-mercy.json
+++ b/packs/equipment/mask-of-mercy.json
@@ -9,7 +9,7 @@
         },
         "containerId": "null",
         "description": {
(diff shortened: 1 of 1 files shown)


TikaelSol:
I think this was made as a free action because single actions don't have triggers, it's an obvious typo that we took the liberty of fixing (there's a few of those in the system). Free action rather than reaction because free actions tend to be what happens on your turn.

PR #12934, opened 2024-01-10
Add missing _id properties to spells

### LINKED ISSUE #12926: Spell overlays loosing their _id property
I have been having a few problems since 5.12 with macros that were working previously (in particular the spellstrike macro):
When using this macro with some spells that have overlays, it fails as it cannot find the _id property of the overlay. 

I made a few tests, creating a new spell with overlays and had the same issue. But it doesn't affect every overlay spell. At the moment, I found essentialy that it affects Gouging Claw (Piercing), but there might be others.

From my tests, it would appear that the _id is correctly added to the overlay when it is first created but as soon as you update it (opening and closing the overlay sheet is enough to do this), the _id property gets lost in the process: This seems to be due to the fact that in the "updateOverride" method, the "difference2" object does not have that property.

a-hodges:
Fixes #12926
_id property was missing in 11 places on the spells _falling stars_, _gouging claw_, _spiritual armament_, and _sun blade_.
The missing property seemed to cause no issues in the system itself, but broke certain macros that relied on the id existing.

### DIFF
--- a/packs/spells/falling-stars.json
+++ b/packs/spells/falling-stars.json
@@ -56,6 +56,7 @@
         },
         "overlays": {
             "3Ifk4ND7o6QBAYMp": {
+                "_id": "3Ifk4ND7o6QBAYMp",
                 "name": "Falling Stars (Plasma)",
                 "overlayType": "override",
                 "sort": 4,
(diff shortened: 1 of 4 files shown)


stwlam:
I'm not sure we really want that since it introduces a risk of the key and the `_id` being different

a-hodges:
> I'm not sure we really want that since it introduces a risk of the key and the `_id` being different

Is there a place that generates the spell files? Because if there isn't then a manual fix is the only real solution.

CarlosFdez:
_id was originally added in the original implementation, but nowadays it serves no purpose. The overlay's id is only used for lookup in order to recreate the variant, and once the variant spell is created it maintains the id (so that rule elements work).

I'm leaning more towards removing it entirely.

PR #12973, opened 2024-01-12
Use (r)em instead of px where appropriate in item-sheet styling

stwlam:


src/styles/item/_sidebar.scss
```diff
@@ -36,13 +36,15 @@ section.sidebar {
     .feat-details {
         display: flex;
         flex-flow: column nowrap;
-        gap: 0.25rem;
+        gap: var(--space-4);
         overflow: hidden scroll;
-        padding: 0 0.25rem 0 0;
+        padding: 0 var(--space-4) 0 0;
```
CarlosFdez, line 41:
```suggestion
        padding: var(--space-8) var(--space-4) 0 0;
```

Makes the alignment for the sidebar a bit prettier. I was gonna do this anyways, but doing it here avoids the conflict :)

src/styles/item/_index.scss
```diff
@@ -334,8 +330,9 @@
                 color: var(--text-light);
                 display: flex;
                 justify-content: space-between;
-                margin: -1px -1px 0 -1px;
-                padding: 1px;
+                margin: var(--space-neg-1) var(--space-neg-1) 0 var(--space-neg-1);
+                padding: var(--space-1);
```
CarlosFdez, line 334:
```suggestion
                margin: var(--space-neg-1) var(--space-neg-1) 0 var(--space-neg-1);
                padding: var(--space-1);
```

These are in use in affliction to do the stage header. Test there to make sure nothing broke, but I think these properties exist in order to cause it to overlap with the border of the content.
image

PR #12983, opened 2024-01-12
Add bleed damage to grim tendrils

MSAbaddon:


TikaelSol:
We removed the bleed damage in the past because the damage application is wrong for success.

MSAbaddon:
Ah, nevermind then - I completely overlooked that passage. Sorry for that.

PR #13021, opened 2024-01-13
Add Forceful Trait Weapon Damage Modifiers

### LINKED ISSUE #5974: Extra damage from "forceful" is not being applied.  Bug? Feature Request?
I don't know if this is a bug or a feature that needs to be implemented, but weapons that are forceful do not apply their extra damage on subsequent hits. I imagine it's not as easy to implement as it sounds.  Maybe just having a 2 more damage buttons on the weapon {Damage (2nd attack)] and [Damage (3rd+ attack)] each button being keyed to add the appropriate extra damage as per the number of weapon damage dice.

MrPrimate:
Adds options for adding extra damage for forceful weapons. The addition of the `_source` dice and `strikingDice` allows for it to account for both striking rune and runic weapon changes. 

I believe this will resolve #5974.

The only thing I'm unsure about is if I have add the language template changes in the right place.

### DIFF
--- a/src/module/system/damage/weapon.ts
+++ b/src/module/system/damage/weapon.ts
@@ -365,6 +365,26 @@ class WeaponDamagePF2e {
             );
         }
 
+        // Forceful trait
+        if (weaponTraits.some((t) => t === "forceful") && weapon.isOfType("weapon")) {
+            modifiers.push(
+                new ModifierPF2e({
+                    slug: "forceful-second",
+                    label: "PF2E.TraitForcefulSecond",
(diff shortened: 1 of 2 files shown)


stwlam:
We prefer to not add localization keys to the top level without good reason. Those can go somewhere in `PF2E.Item.Weapon`.

MrPrimate:
> We prefer to not add localization keys to the top level without good reason. Those can go somewhere in `PF2E.Item.Weapon`.

All adjusted!

PR #13024, opened 2024-01-13
Restrict Basic Finisher to require Finishing Precision archetype feat

pedrogrullada:


TikaelSol:
We'll want to restrict the roll option entirely on either being a swashbuckler or having a finisher, otherwise it will show up for multiclass swashbucklers without the feat.

PR #13026, opened 2024-01-13
Brush up Actions

DocSchlock:
This is a sweep through Actions to add inline checks, clean up publications, and fix text to the Remaster versions when noticed.


packs/actions/clear-a-path.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p><strong>Requirements</strong> You're wielding a two-handed firearm or two-handed crossbow.</p>\n<hr />\n<p>You push outward with your weapon to clear some space before quickly reloading a fresh round. You make an @Check[type:athletics] check to @UUID[Compendium.pf2e.actionspf2e.Item.Shove] an opponent within your reach using your weapon, then Interact to reload. For this Shove, you don't need a free hand, and you add the weapon's item bonus on attack rolls (if any) to the Athletics check. If your last action was a ranged Strike with the weapon, use the same multiple attack penalty as that Strike for the Shove; the Shove still counts toward your multiple attack penalty on further attacks as normal.</p>"
```
TikaelSol, line 14: For this kind of thing we want to use the shove action macro instead of using unautomated inline rolls.

DocSchlock: Changed to an action macro.

nikolaj-a:
Not the best time for this given the new syntax hopefully being introduced soon. Also, it might make sense to at least include the actions’ roll options.

PR #13059, opened 2024-01-15
Add rules elements to derring-do.json

chazpls:
Add automation for rolling twice from Derring-do for all swashbuckler styles

### DIFF
--- a/packs/feats/derring-do.json
+++ b/packs/feats/derring-do.json
@@ -24,7 +24,64 @@
             "remaster": false,
             "title": "Pathfinder Advanced Player's Guide"
         },
-        "rules": [],
+        "rules": [
+          {
+            "keep": "higher",
+            "key": "RollTwice",
+            "predicate": ["self:effect:panache", "action:tumble-through"],
+            "selector": ["acrobatics-check"]
+          },
+          {
+            "keep": "higher",
+            "key": "RollTwice",
+            "predicate": [
+              "self:effect:panache",
+              "feature:battledancer",
+              "action:perform"
+            ],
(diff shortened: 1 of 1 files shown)


stwlam:
This looks like it was hand-entered, and it definitely wouldn't work. Test in foundry.

stwlam:
`selector` is a string, though the arrays are getting coerced to strings by the data model.

chazpls:
Shoot, you're right -- I'm not sure how those brackets ended up there, and I didn't notice since it was working. I'll close this pull request, fix it, and resubmit. Thank you!

chazpls:
I've amended this to use strings for the selector rather than arrays, and I followed the correct procedure for updating compendium items, by editing directly in Foundry and using npm extractPacks

packs/feats/derring-do.json
```diff
@@ -24,7 +24,78 @@
...
+        "rules": [
+            {
+                "keep": "higher",
+                "key": "RollTwice",
+                "predicate": [
+                    "self:effect:panache",
+                    "action:tumble-through"
+                ],
+                "selector": "acrobatics-check"
+            },
```
TikaelSol, line 36: Instead of doing this with a bunch of REs we should condense this to a single rule with a larger predicate.

```json
"predicate": [
    "self:effect:panache",
    {
       "or": [
            "action:tumble-through",
            { "and": ["feature:battledancer", "action:perform"] }
           ...
        ]
    }
]```

TikaelSol: Actually, we have a different idea that will be a much larger project but end up being much cleaner.

PR #13061, opened 2024-01-15
Store Swashbuckler's Precise Strike value in a flag

pedrogrullada:
I also use the very same flag to automate Bleeding Finisher, which supports multiclass swashbucklers as well.  ~~I tried to do the same with Confident Finisher but quickly ran into the roadblock of inline precision damage...~~

Were this to be accepted I suppose it'd require to migrate existing swashbucklers for them to have the relevant flag?

### DIFF
--- a/packs/classfeatures/precise-strike.json
+++ b/packs/classfeatures/precise-strike.json
@@ -28,7 +28,7 @@
             {
                 "key": "ActiveEffectLike",
                 "mode": "override",
-                "path": "flags.pf2e.swashbuckler.preciseStrike.value",
+                "path": "flags.pf2e.swashbuckler.preciseStrike",
                 "predicate": [
                     "class:swashbuckler"
                 ],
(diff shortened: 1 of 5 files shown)


packs/classfeatures/precise-strike.json
```diff
@@ -25,6 +25,41 @@
             "title": "Pathfinder Advanced Player's Guide"
         },
         "rules": [
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.swashbuckler.preciseStrike.value",
```
TikaelSol, line 31: The only issue with this is people who already have this feature on their actors but have not taken bleeding finisher yet. We'll need to migrate this change since it would otherwise break on those actors.

pedrogrullada: Yeah, I'm unsure how much of a hassle on the dev side that migration would be. It's certainly beyond me, so please feel free to close this if doing so is not a possibility.

TikaelSol: Shouldn't be a hard migration to write, maybe if I find some time this weekend I'll toss one up (unlikely that I will have time unfortunately)

pedrogrullada:
Thanks to the Discord, I found a way to reference the damage type of the weapon a Note is attached to, so we can have some Confident Finisher automation too

packs/classfeatures/precise-strike.json
```diff
@@ -25,6 +25,15 @@
...
```
stwlam, line 31: Why `preciseStrike.value`?

pedrogrullada: If I remember, at the time I thought it'd be a good idea to leave it open to PC2 adding stuff to Precise Strike, like dice scaling or whatnot. So no good reason, really. I'll fix it later today.

PR #13091, opened 2024-01-17
Allows sending items to any container from `carryType` menu 

reonZ:
Foundry_Virtual_Tabletop_PFQ5BNtPgY

Lists all the containers inside the `carryType` menu and allow any non-container item to be moved to them.

I arbitrary set the max width of the menu in case of extra long container name.


stwlam:
My vote here is "hell no", but if Supe is at least lukewarm I'll defer to him.

reonZ:
figured !

CarlosFdez:
Perhaps with a divider

CarlosFdez:
I still need to resurrect the context menu branch, but a potential alternative is to do Send to X buttons on the right click menu instead.

stwlam:
Conflicts!

reonZ:
I can't figure out why the tooltip centers itself based on the max possible content instead of the outer part of the tooltip, this makes it move to the left with long container names.

image

When i do the same in my module, i don't have this issue, but i couldn't make head or tail of why this would happen here.

CarlosFdez:
text overflow ellipses might be your friend here, though you'll need a bit for that to work out.

Also, this may be extra work, but I think this should probably pop open a dialog to allow splitting the stack?

reonZ:
> text overflow ellipses might be your friend here, though you'll need a bit for that to work out.

i use it (you can see the ellipses in the screenshot), i have pretty much the same setting for my module and the tooltip does center properly

if you want a popup, i think the approach should be different and use another button instead of using the `carry-type` one

CarlosFdez:
By a popup, I mean the same thing that occurs when you drag to or from a loot sheet, where it asks for a quantity. It'd only show if there is more than one being moved to a stowing container.

CarlosFdez:
I'll close this since it hasn't been updated for a while, but the past few months and some conversations in the development help channel has convinced me that stowing functionality that allows you to split the stack is very likely where many people would like to go with this, if anyone wants to pursue that route.

Thanks for making this PR though.

PR #13095, opened 2024-01-17
Set roll option for Swashbuckler's Panache-granting actions

pedrogrullada:
Makes things a bit cleaner, and hopefully easier for 3PP to add their own swashbuckler styles. I left feats out of this due to the way Panache and Derring-Do are worded:

Panache says
> gain a +1 circumstance bonus to checks to Tumble Through or to take any actions that give you panache due to your style

Derring-Do says
> you can roll twice and use the higher result on checks to which the circumstance bonus for having panache applies (Tumble Through and any skill actions listed in your swashbuckler's style).

Which implies that different ways to gain Panache (such as Leading Dance, One For All, and Disarming Flair) do not qualify for these benefits. For the sake of convenience, I think they're better off with their own Note for gaining Panache, if we add any. Thoughts?

### DIFF
--- a/packs/feats/derring-do.json
+++ b/packs/feats/derring-do.json
@@ -24,7 +24,17 @@
             "remaster": false,
             "title": "Pathfinder Advanced Player's Guide"
         },
-        "rules": [],
+        "rules": [
+            {
+                "keep": "higher",
+                "key": "RollTwice",
+                "predicate": [
+                    "panache",
+                    "grants-panache"
+                ],
+                "selector": "skill-check"
+            }
+        ],
         "traits": {
             "rarity": "common",
             "value": [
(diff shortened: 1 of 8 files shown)


TikaelSol:
I won't have time to review this until tomorrow, but it should supersede #13059

packs/feats/derring-do.json
```diff
@@ -24,7 +24,17 @@
             "remaster": false,
             "title": "Pathfinder Advanced Player's Guide"
         },
-        "rules": [],
+        "rules": [
+            {
+                "keep": "higher",
+                "key": "RollTwice",
+                "predicate": [
+                    "panache",
```
TikaelSol, line 32:
```suggestion
                    "self:effect:panache",
```

pedrogrullada: I thought it would've been fine since panache's effect gives that roll option, but it's best to be safe

PR #13114, opened 2024-01-19
Add rule elements for Tremorsense and Wavesense Familiar Abilities

j-bs:


TikaelSol:
Just want to confirm that sense REs actually work on familiars, I know at one point they didn't

PR #13115, opened 2024-01-19
Add automation for the Resistance group of Familiar Abilities

j-bs:
The rule elements for the base `Resistance` ability are a bit clunky but it gets the job done and allows for the upgrades from the other two abilities.

Happy for some tips if there's a more succinct way to do the same though.

### DIFF
--- a/packs/familiar-abilities/resistance.json
+++ b/packs/familiar-abilities/resistance.json
@@ -48,7 +48,6 @@
                 ],
                 "flag": "resistanceOne",
                 "key": "ChoiceSet",
-                "priority": 50,
                 "prompt": "PF2E.SpecificRule.Prompt.Resistance",
                 "rollOption": "resistance"
             },
@@ -111,7 +110,6 @@
                 ],
                 "flag": "resistanceTwo",
                 "key": "ChoiceSet",
-                "priority": 60,
                 "prompt": "PF2E.SpecificRule.Prompt.Resistance",
                 "rollOption": "resistance"
             },
@@ -129,14 +127,12 @@
                 "key": "ActiveEffectLike",
                 "mode": "override",
                 "path": "flags.pf2e.resistance.resistanceOne",
-                "priority": 100,
                 "value": "{item|flags.pf2e.rulesSelections.resistanceOne}"
             },
             {
                 "key": "ActiveEffectLike",
                 "mode": "override",
                 "path": "flags.pf2e.resistance.resistanceTwo",
-                "priority": 100,
                 "value": "{item|flags.pf2e.rulesSelections.resistanceTwo}"
             }
         ],
--- a/packs/familiar-abilities/resistance.json
+++ b/packs/familiar-abilities/resistance.json
@@ -18,7 +18,128 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
+                "choices": [
+                    {
+                        "label": "PF2E.TraitAcid",
+                        "value": "acid"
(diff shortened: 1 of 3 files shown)


packs/familiar-abilities/resistance.json
```diff
@@ -18,7 +18,128 @@
...
+                    {
+                        "label": "PF2E.TraitPoison",
+                        "value": "poison"
+                    },
+                    {
+                        "label": "PF2E.TraitSonic",
+                        "value": "sonic"
+                    }
+                ],
+                "flag": "resistance1",
```
TikaelSol, line 49:
```suggestion
                "flag": "resistanceOne",
```
I would prefer if we kept flags to words.

j-bs: Noted, I'll change it and keep it in mind for future.

packs/familiar-abilities/resistance.json
```diff
@@ -18,7 +18,128 @@
...
+                        "value": "poison"
+                    },
+                    {
+                        "label": "PF2E.TraitSonic",
+                        "value": "sonic"
+                    }
+                ],
+                "flag": "resistanceOne",
+                "key": "ChoiceSet",
+                "priority": 50,
```
TikaelSol, line 51:
```suggestion
```
We shouldn't need to worry about setting priorities for anything in this.

packs/familiar-abilities/resistance.json
```diff
@@ -18,7 +18,128 @@
...
+                            {
+                                "not": "resistance:sonic"
+                            }
+                        ],
+                        "value": "sonic"
+                    }
+                ],
+                "flag": "resistanceTwo",
+                "key": "ChoiceSet",
+                "priority": 60,
```
TikaelSol, line 114:
```suggestion
```

packs/familiar-abilities/resistance.json
```diff
@@ -18,7 +18,128 @@
...
+                "value": "{item|flags.pf2e.rulesSelections.resistanceOne}"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.resistance.resistanceTwo",
+                "priority": 100,
+                "value": "{item|flags.pf2e.rulesSelections.resistanceTwo}"
+            }
+        ],
```
TikaelSol, line 142:
```suggestion
                "key": "ActiveEffectLike",
                "mode": "override",
                "path": "flags.pf2e.resistance.resistanceOne",
                "value": "{item|flags.pf2e.rulesSelections.resistanceOne}"
            },
            {
                "key": "ActiveEffectLike",
                "mode": "override",
                "path": "flags.pf2e.resistance.resistanceTwo",
                "value": "{item|flags.pf2e.rulesSelections.resistanceTwo}"
            }
        ],
```

TikaelSol:
Technically this should get a migration to set some defaults on familiars who already have resistance but not either of the other two, but that's likely a small number of actors and it would be us injecting some default choice for players which I like less than making a bit of console noise.

packs/familiar-abilities/resistance.json
j-bs: I know you've already merged, but I'm curious if you don't mind me asking:

I thought priority in this case would help ensure that the `resistanceOne` ChoiceSet is always offered first before the `resistanceTwo`, since I only added the selection filtering to two to reduce the clutter.

I guess REs are just resolved in their declared order though... 😅

So priority is primarily used for the math operations in AE-Likes as you've mentioned in the wiki?

PR #13123, opened 2024-01-19
Add `target` property to SimpleRollActionCheck options that accepts Token or Actor

Cuingamehtar:
The proposed change allows calling `action.use({target:token})` or `action.use({target:token.action})` and use the passed value instead of picking the first user target.

When called with an actor, the assumption is that the caller doesn't care about a specific token.

The intended use case is to call this from a macro or a module and utilize the system action checks and modifiers.

### DIFF
--- a/src/module/actor/actions/single-check.ts
+++ b/src/module/actor/actions/single-check.ts
@@ -1,4 +1,4 @@
-import type { ActorPF2e } from "@actor";
+import { ActorPF2e } from "@actor";
 import { ModifierPF2e, RawModifier, StatisticModifier } from "@actor/modifiers.ts";
 import { DCSlug } from "@actor/types.ts";
 import { DC_SLUGS } from "@actor/values.ts";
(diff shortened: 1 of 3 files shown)


src/module/actor/actions/single-check.ts
```diff
@@ -155,6 +156,18 @@ class SingleCheckActionVariant extends BaseActionVariant {
                     note.selector ||= selector; // treat empty selectors as always applicable to this check
                     return note;
                 }),
+            target: options.target
+                ? () => {
```
nikolaj-a, line 160: The outer check for the presence of a `target` field in options is unnecessary. Just always have the function and let it return `null` if the field is missing.

nikolaj-a (dismissed):
Address the comment and it looks good! 👍

nikolaj-a:
Oh, and squash the commit 😉

Cuingamehtar:
Did I do it correctly?

src/module/actor/actions/single-check.ts
Cuingamehtar: Removed external check

nikolaj-a:
Yep - LGTM 👍

PR #13126, opened 2024-01-19
Add inline spell attack rolls to Shadow Signet

pedrogrullada:
I'm sure the system is doing something more elegant in the future, but in the meantime I think this is a decent stopgap.


pedrogrullada:
The future is now!

PR #13137, opened 2024-01-20
Remove colored border on disabled `Cast` buttons in spellcasting entries

j-bs:
The border remaining blue kinda looks deceptively like a lingering selection highlight IMO.

spell-cast-button

### DIFF
--- a/src/styles/actor/_spell-collection.scss
+++ b/src/styles/actor/_spell-collection.scss
@@ -28,6 +28,7 @@ ol.spell-list {
 
             button.cast-spell {
                 background: var(--color-disabled);
+                border-color: transparent;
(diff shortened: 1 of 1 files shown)


stwlam:
How does it look transparent?

j-bs:
It actually has a transparent border on hover already, and without examining the pixels with a big zoom-in it basically looks the same as my _after_ pic.

Do you have a preference to transparent vs the disabled colour?

stwlam:
Yeah let's do transparent, though it's weird that it would still show a visible border.

j-bs:
Done. And for reference this is with `transparent`: (almost identical but has a (more) distinct border radius?)

spell-cast-button-v2

PR #13155, opened 2024-01-20
Add effects and icon for magnetic shot

chazpls:
Added effects with rule elements for each version of magnetic shot, automating the damage dice, deadly die, and circumstance bonus for the attack. Also added an icon for the items and effects.


simonward:
Rather than effects, these could potentially be rule elements on the ammunition itself. Rule elements on ammunition are applied to their firing weapon. See Eroding Bullet and Ghost Ammunition for some examples. These will probably need labels, but the `"label": "{item|name}"` pattern won't work since that will pick up the weapon's name, so a localized string will need to be created for that.

packs/equipment-effects/effect-magnetic-shot-greater.json
```diff
@@ -0,0 +1,68 @@
+{
+    "_id": "oGsTuU8u0lnpSmnE",
+    "img": "icons/commodities/tech/bearing-steel-green.webp",
+    "name": "Effect: Magnetic Shot (Greater)",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.JqKNiEJxSZX1sD06]{Magnetic Shot (Greater)}</p>\n<p>The activated ammunition grants a +2 circumstance bonus to attack rolls against such targets. Due to magnetic acceleration, the ammunition deals four additional weapon damage dice and is deadly d12.</p>"
```
simonward, line 7: These links are malformed.

chazpls: Thanks for the ammunition examples. I'll take another pass to set them up that way, and I'll take a look at the links. I was just setting up the links via drag + drop. I'll check the contributing page again for the correct practice

chazpls: Ok, the links were removed altogether and I set these up as rule elements directly on the ammunition itself. Thanks for pointing me in the right direction!

simonward:
> These will probably need labels

Looks like they don't need labels after all - the automatic labels work fine.
Screen Shot 2024-01-20 at 6 46 02 PM

PR #13157, opened 2024-01-21
Fill automation gaps in Stance: Tenacious Stance

stwlam:
Depends on #13156

### DIFF
--- a/packs/feat-effects/stance-tenacious-stance.json
+++ b/packs/feat-effects/stance-tenacious-stance.json
@@ -4,7 +4,7 @@
     "name": "Stance: Tenacious Stance",
     "system": {
         "description": {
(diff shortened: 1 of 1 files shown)


packs/feat-effects/stance-tenacious-stance.json
```diff
@@ -33,15 +32,29 @@
...
+                "option": "armor:strength-requirement-met",
+                "phase": "beforeDerived",
+                "value": false
+            },
+            {
+                "domain": "str-skill-check",
+                "key": "RollOption",
+                "option": "armor:strength-requirement-met",
+                "phase": "beforeDerived",
+                "value": false
```
TikaelSol, line 57: We should add to the description that we are automating this part now.

PR #13181, opened 2024-01-22
Allow `SpecialStatistic` to create item-spellcasting abilities

stwlam:


src/module/item/consumable/document.ts
```diff
@@ -205,33 +207,30 @@ class ConsumablePF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
...
-                });
+            return (
+                actor.spellcasting
+                    .filter(
+                        (e): e is SpellcastingEntry<ActorPF2e> & { statistic: Statistic } =>
+                            !!e.statistic && e.canCast(spell, { origin: this }),
+                    )
+                    .reduce((previous, current) => {
+                        if (!previous && current?.statistic) return current;
+                        if (!current?.statistic && previous?.statistic) return previous;
```
CarlosFdez, line 225: Wouldn't the below defaults to negative infinity render these if statements unnecessary?

PR #13195, opened 2024-01-22
Add missing Compendium Browser settings localization strings

pedrogrullada:


stwlam:
Instead of adding these, use the localization keys already in use for the actual tabs.

PR #13198, opened 2024-01-23
Simplify action toMessage() and use() methods

CarlosFdez:
getDefaultVariant() is not async, and defaults to "toActionVariant()" anyways.


nikolaj-a (approved):
LGTM 👍

PR #13200, opened 2024-01-23
Include subitems when refreshing items from compendiums

stwlam:


CarlosFdez:
Is it safe to remove the expandObject calls? From a glance, it seems the point is to avoid conflicts, so that "system.location" and "system": { "location": ??? } aren't in the same call.

stwlam:
```js
export function mergeObject(original, other={}, {
    insertKeys=true, insertValues=true, overwrite=true, recursive=true, inplace=true, enforceTypes=false,
    performDeletions=false
  }={}, _d=0) {
  other = other || {};
  if (!(original instanceof Object) || !(other instanceof Object)) {
    throw new Error("One of original or other are not Objects!");
  }
  const options = {insertKeys, insertValues, overwrite, recursive, inplace, enforceTypes, performDeletions};

  // Special handling at depth 0
  if ( _d === 0 ) {
    if ( Object.keys(other).some(k => /\./.test(k)) ) other = expandObject(other);
```

CarlosFdez:
Neato, already handled.

PR #13215, opened 2024-01-23
Add DC to Depart Save (Tempest of Shades)

Intervencion:
Since Tempest of Shades has two different saves, either the Depart one should be made inline (👍) or both Arrive and Depart should be made inline (👎).

I mean ideally you'd want both saves on the Spell Details but eh


TikaelSol:
This is definitely wrong though, you wouldn't use your class DC for this and we have to account for spellcasting entries with different DCs due to using a different attribute modifier.

Intervencion:
Hm. True. I didn't account for spellcasting with different attributes.

I didn't find the RE for SpellDC though, could only find ClassDC and classOrSpellDC

PR #13244, opened 2024-01-25
Add support for specifying which weapon to use for a combat maneuver

nikolaj-a:
As part of the changes, support for adding a passthrough object to the roll check helper function has also been implemented.


src/module/system/action-macros/athletics/shove.ts
```diff
@@ -50,10 +69,16 @@ function shove(options: SkillActionOptions): void {
...
 
 class ShoveActionVariant extends SingleCheckActionVariant {
-    protected override checkContext<ItemType extends ItemPF2e<ActorPF2e>>(
-        opts: CheckContextOptions<ItemType>,
-        data: CheckContextData<ItemType>,
-    ): CheckContext<ItemType> | undefined {
+    override async use(options: ShoveActionUseOptions = {}): Promise<CheckResultCallback[]> {
+        return ActionMacroHelpers.resolveWieldedWeaponFromItemParameter(options, REQUIRED_WEAPON_TRAIT).then(
+            super.use.bind(this),
+        );
```
stwlam, line 75: There isn't really a good reason to use `then` in an async environment

nikolaj-a: I guess awaiting should work just as well. Updated.

src/module/system/action-macros/helpers.ts
```diff
@@ -63,10 +66,59 @@ export class ActionMacroHelpers {
...
+        options: Partial<CombatManeuverActionUseOptions>,
+        traits: WeaponTrait | WeaponTrait[],
+    ): Promise<Partial<CombatManeuverActionPassthroughOptions>> {
+        const original = options.item;
+        const passthrough: CombatManeuverActionPassthroughOptions = { ...options, item: undefined };
+        if (original && typeof original === "string") {
+            if (original.match(/^Actor\.[a-zA-Z0-9]+\.Item\.[a-zA-Z0-9]+$/)) {
+                const item = await fromUuid<ItemPF2e<ActorPF2e>>(original);
+                if (!item) {
+                    return Promise.reject(
```
CarlosFdez, line 79: From my limited digging, I think you can just throw. In an async method it appears its the same either way.

nikolaj-a: Seems to be working the same 👍

CarlosFdez:
To avoid needing a CombatManeuverActionUseOptions and CombatManeuverActionPassthroughOptions and maintaining duplicates of every property (source and transformed), could we not resolve the item in the actions themselves? I think a resolveItem() helper that does that (which accepts null/undefined and returns null) in the disarm/trip/shove might not be out of line?

src/module/actor/actions/single-check.ts
```diff
@@ -156,6 +155,7 @@ class SingleCheckActionVariant extends BaseActionVariant {
                     note.selector ||= selector; // treat empty selectors as always applicable to this check
                     return note;
                 }),
+            passthrough: options,
```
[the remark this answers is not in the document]
nikolaj-a: I'm not a fan of the name either. It's a crowded space for naming with "context" and "options" already taken 😂

nikolaj-a:
> To avoid needing a CombatManeuverActionUseOptions and CombatManeuverActionPassthroughOptions and maintaining duplicates of every property (source and transformed), could we not resolve the item in the actions themselves? I think a resolveItem() helper that does that (which accepts null/undefined and returns null) in the disarm/trip/shove might not be out of line?

My plan is actually to take this a notch further by having some methods like `getAdditionalModifiers(...)` and the like to keep it more general. But it's going to be a bigger refactoring to get rid of the old `ActionMacroHelpers.simpleRollActionCheck()`.

src/module/system/action-macros/athletics/grapple.ts
```diff
@@ -51,10 +70,16 @@ function grapple(options: SkillActionOptions): void {
 }
 
 class GrappleActionVariant extends SingleCheckActionVariant {
-    protected override checkContext<ItemType extends ItemPF2e<ActorPF2e>>(
-        opts: CheckContextOptions<ItemType>,
-        data: CheckContextData<ItemType>,
-    ): CheckContext<ItemType> | undefined {
+    override async use(options: GrappleActionUseOptions = {}): Promise<CheckResultCallback[]> {
+        return super.use(
+            await ActionMacroHelpers.resolveWieldedWeaponFromItemParameter(options, REQUIRED_WEAPON_TRAIT),
```
CarlosFdez, line 75: I don't think we need to do this. If we handle this resolving in the actual action implementatiion, we can avoid a separate interface type for "resolved" data, and reduce the number of layers of abstraction.

src/module/system/action-macros/athletics/combat-maneuver.ts
```diff
@@ -0,0 +1,67 @@
...
+            weapon = traits
+                .flatMap((trait) => ActionMacroHelpers.getApplicableEquippedWeapons(actor, trait) ?? [])
+                .shift();
+        }
+        return weapon;
+    }
+
+    override async use(
+        options: Partial<SingleCheckActionUseOptions & CombatManeuverActionUseOptions> = {},
+    ): Promise<CheckResultCallback[]> {
```
CarlosFdez, line 61: Why do you need to subclass use? If its because of the params, couldn't it be part of the function generics?

src/module/system/action-macros/athletics/combat-maneuver.ts
```diff
@@ -0,0 +1,67 @@
...
+     */
+    item: EmbeddedItemUUID | WeaponPF2e<ActorPF2e> | false;
+}
+
+function isOwnedItemUUID(uuid: EmbeddedItemUUID | string, actor: ActorPF2e) {
+    return uuid.match(/^Actor\.[a-zA-Z0-9]+\.Item\.[a-zA-Z0-9]+$/) && uuid.startsWith(`Actor.${actor.id}`);
+}
+
+abstract class CombatManeuverActionVariant extends SingleCheckActionVariant {
+    static resolveWieldedWeapon(
```
CarlosFdez, line 21: I think the current trend for these functions is to have them as functions (not methods), and put them in a helpers.ts file somewhere. It does frustrate people since these helpers are not accessible outside of the pf2e project though when done that way, but it is consistent.

src/module/system/action-macros/athletics/combat-maneuver.ts
nikolaj-a: For compile-time type safety basically. It's redefined one more time in each of the specific actions for the same reason.

src/module/system/action-macros/athletics/combat-maneuver.ts
nikolaj-a: Yep, and I agree with that approach. But as long as we need to maintain the old action functions we need to expose some stuff outside the action class implementations until we can remove them.

PR #13263, opened 2024-01-26
Automate Overdrive Ally

pedrogrullada:
Since it seems there's no way currently to get roll options from the item's origin, I set up a flag in Overdrive Ally that could tell the effect what Overdrive degree of success the inventor has. The alternative is a Choice Set, but this seemed cleaner.

It doesn't look like there's console noise if the flag is missing (which would be the case for inventors with an old copy of the feat), but even then if they want to have the effect handy, it's likely they will get a new overdrive ally copy. If this isn't ok I'm happy to go with a Choice Set instead though.

### DIFF
--- a/packs/feat-effects/effect-overdrive-ally.json
+++ b/packs/feat-effects/effect-overdrive-ally.json
@@ -1,0 +1,42 @@
+{
+    "_id": "a4C89wVlcwV6Uxx4",
+    "img": "icons/commodities/tech/cog-brass.webp",
+    "name": "Effect: Overdrive Ally",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Overdrive Ally]</p>\n<p>Until the end of their next turn, that ally's Strikes deal additional damage equal to half your Intelligence modifier, or your full Intelligence modifier if you were in critical overdrive.</p>"
+        },
+        "duration": {
+            "expiry": "turn-start",
+            "sustained": false,
+            "unit": "rounds",
+            "value": 1
+        },
+        "level": {
+            "value": 1
+        },
+        "publication": {
+            "license": "OGL",
+            "remaster": false,
+            "title": "Pathfinder Guns & Gears"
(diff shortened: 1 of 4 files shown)


packs/feats/overdrive-ally.json
```diff
@@ -28,7 +28,28 @@
             "remaster": false,
             "title": "Pathfinder Guns & Gears"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.overdriveAlly.damage",
```
TikaelSol, line 35:
```suggestion
                "path": "flags.pf2e.inventor.overdriveAlly.damage",
```
Best to keep things organized.

packs/feats/overdrive-ally.json
```diff
@@ -28,7 +28,28 @@
...
+                "phase": "afterDerived",
+                "predicate": [
+                    "self:effect:overdrive-success"
+                ],
+                "value": "floor(@actor.system.abilities.int.mod/2)"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.overdriveAlly.damage",
```
TikaelSol, line 45:
```suggestion
                "path": "flags.pf2e.inventor.overdriveAlly.damage",
```

packs/feat-effects/effect-overdrive-ally.json
```diff
@@ -0,0 +1,42 @@
...
+        "publication": {
+            "license": "OGL",
+            "remaster": false,
+            "title": "Pathfinder Guns & Gears"
+        },
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "selector": "strike-damage",
+                "value": "@item.origin.flags.pf2e.overdriveAlly.damage"
```
TikaelSol, line 27:
```suggestion
                "value": "@item.origin.flags.pf2e.inventor.overdriveAlly.damage"
```

PR #13265, opened 2024-01-26
Fix Blast attack-modifier labels when using items that affect melee or ranged attack rolls

CarlosFdez:


src/module/actor/character/elemental-blast.ts
```diff
@@ -1,3 +1,4 @@
+import { ActorPF2e } from "@actor";
```
stwlam, line 1:
```suggestion
import type { ActorPF2e } from "@actor";
```

PR #13266, opened 2024-01-26
Show activatable spell consumables in spellcasting tab

stwlam:


src/styles/actor/character/_spellcasting.scss
```diff
@@ -1,4 +1,22 @@
 &.spellcasting {
+    padding: 0;
+    overflow-y: hidden;
```
j-bs, line 3: I know this is merged but seemed like easiest place for context to add a comment.

This isn't quite functioning together with the `.tab { overflow-y: scroll; }` below. Scrolling is gone in the spellcasting tab.

The `overflow-y: scroll` needs to be one level higher on the `div.spell-collections` element instead of its child `.tab` 

Should I put in a tiny PR for it?

PR #13274, opened 2024-01-26
Improve Offensive Boost related automation and automate Variable Core

pedrogrullada:
A bit of an invasive attempt at automating most Offensive Boost related features and Variable Core through the use of flags. Because of this, this will probably need a migration to work on older actors.  

There's a little snag with the Explode action, in that I can't find a way to change its fire trait into the trait from the chosen Variable Core, so any help on that front would be appreciated.


packs/actions/explode.json
```diff
@@ -11,7 +11,7 @@
...
... @Damage[max(2, @actor.level)d6[@actor.flags.pf2e.inventor.explode]|traits:area-damage] damage with a @Check[type:reflex|dc:resolve(@actor.attributes.classDC.value)|basic:true] to all creatures in a @Template[type:emanation|distance:5] around you (if you're wearing or holding the innovation) or around your innovation (if your innovation is a minion).</p>\n<p>At 3rd level, and every level thereafter, increase your explosion's damage by 1d6.</p>\n<p>If you have the breakthrough innovation class feature, you can choose either a 5-foot or @Template[type:emanation|distance:10] for the area when you use Explode; if you have the revolutionary innovation class feature, you can choose a 5-foot, 10-foot, or @Template[type:emanation|distance:15].</p>\n<hr />\n<p>@Check[type:flat|dc:17|name:Unstable Flat Check|showDC:all|traits:unstable]{Unstable Check}</p>\n<p>On a failure, the innovation malfunctions in a spectacular (though harmless) fashion, such as a belch of smoke or shower of sparks, and it becomes incapable of being used for further unstable actions.</p>\n<p>On a critical failure, you also take @Damage[(@actor.level)[@actor.flags.pf2e.inventor.explode]] damage equal to your level.</p>"
```
TikaelSol, line 14:
```suggestion
            "value": "<p>You intentionally take your innovation beyond normal safety limits, making it explode and damage nearby creatures without damaging the innovation... hopefully. The explosion deals @Damage[max(2, @actor.level)d6[@actor.flags.pf2e.inventor.explode]|traits:area-damage] damage with a @Check[type:reflex|dc:resolve(@actor.proficiencies.classDCs.inventor.value)|basic:true] to all creatures in a @Template[type:emanation|distance:5] around you (if you're wearing or holding the innovation) or around your innovation (if your innovation is a minion).</p>\n<p>At 3rd level, and every level thereafter, increase your explosion's damage by 1d6.</p>\n<p>If you have the breakthrough innovation class feature, you can choose either a 5-foot or @Template[type:emanation|distance:10] for the area when you use Explode; if you have the revolutionary innovation class feature, you can choose a 5-foot, 10-foot, or @Template[type:emanation|distance:15].</p>\n<hr />\n<p>@Check[type:flat|dc:17|name:Unstable Flat Check|showDC:all|traits:unstable]{Unstable Check}</p>\n<p>On a failure, the innovation malfunctions in a spectacular (though harmless) fashion, such as a belch of smoke or shower of sparks, and it becomes incapable of being used for further unstable actions.</p>\n<p>On a critical failure, you also take @Damage[(@actor.level)[@actor.flags.pf2e.inventor.explode]] damage equal to your level.</p>"
```
While we're in here may as well fix these I think. Since github diffs are useless on large blocks this is a change for the inline saves to `actor.proficiencies.classDCs.inventor.value` since this shouldn't use your fighter DC if you're multiclass.

pedrogrullada: Makes sense, I'm on it

packs/feats/variable-core.json
```diff
@@ -24,7 +24,33 @@
...
+                        "value": "electricity"
+                    }
+                ],
+                "flag": "variableCore",
+                "key": "ChoiceSet"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.inventor.explode",
```
TikaelSol, line 49: I think the path here should be `variable-core` since it changing the explode damage is secondary, no?

TikaelSol: Or maybe no, it does an override of the default damage. so disregard that.

TikaelSol (approved):
Will need a migration, but otherwise approved.

PR #13280, opened 2024-01-26
Migrate precise strike damage to actor flags

TikaelSol:
For #13061


CarlosFdez (dismissed):

pedrogrullada:
Per shark's suggestion, I've changed the flag from `flags.pf2e.swashbuckler.preciseStrike.value` to `flags.pf2e.swashbuckler.preciseStrike`, so this will need to change too. Sorry about that!

PR #13294, opened 2024-01-27
feat: attach item to inline template as json under data-pf2-template-data

digitalpacman:
Story: As a module developer I want to play an animation for "Kobold Breath" when a template is placed from a chat inline template.

Problem:  Inline templates lose context of the item that created the message. This prevents modules from having item aware context decisions. 

Solution: Use  the existing json deserialization of the field data-pf2-template-data to contain the origin item context data necessary for foundry core attaching the 'item' field in a template created event.  The data will be serialized into an html data field as full json.  The code reading from this field already deserializes the field.  But the way it patches uses property whole replace. Change this code to instead start with the existing fields if they exist instead of empty objects.


src/module/system/text-editor.ts
```diff
@@ -314,11 +314,32 @@ class TextEditorPF2e extends TextEditor {
...
+                                actor: item.actor?.uuid,
+                                traits: itemTraits ? Array.from(itemTraits) : [],
+                                type: item.type,
+                                uuid: item.uuid,
+                            }
+                        }
+                    }
+                });
+            }
+
```
digitalpacman, line 334: This matches the same fields populated from a pf2e spells "place XX template"

src/module/system/text-editor.ts
```diff
@@ -314,11 +314,32 @@ class TextEditorPF2e extends TextEditor {
...
             // Add the html elements used for the inline buttons
             const html = document.createElement("span");
             html.innerHTML = label;
             html.setAttribute("data-pf2-effect-area", params.type);
             html.setAttribute("data-pf2-distance", params.distance);
+            if (params.templateData) {
+                html.setAttribute("data-pf2-template-data", params.templateData);
+            }
```
digitalpacman, line 342: Using an existing field that currently has no writes

src/scripts/ui/inline-roll-links.ts
```diff
@@ -295,7 +295,7 @@ export const InlineRollLinks = {
                 }
 
                 const flags: { pf2e: Record<string, unknown> } = {
-                    pf2e: {},
+                    pf2e: templateData.flags?.pf2e ?? {},
```
digitalpacman, line 298: Used this technique in order to alter the least number of lines

src/scripts/ui/inline-roll-links.ts
```diff
@@ -315,7 +315,7 @@ export const InlineRollLinks = {
 
                 const actor = resolveActor(foundryDoc, link);
                 if (actor || pf2Traits) {
-                    const origin: Record<string, unknown> = {};
+                    const origin: Record<string, unknown> = flags.pf2e.origin as Record<string, unknown> ?? {};
```
digitalpacman, line 318: Used this technique in order to alter the least number of lines

src/module/system/text-editor.ts
```diff
@@ -314,11 +314,32 @@ class TextEditorPF2e extends TextEditor {
                 });
             }
 
+            if (item) {
+                const itemTraits = (item as any).traits;
```
digitalpacman, line 318: For some reason traits does not exist on the interface but it does exist in practice.

src/scripts/ui/inline-roll-links.ts
```diff
@@ -315,7 +315,7 @@ export const InlineRollLinks = {
 
                 const actor = resolveActor(foundryDoc, link);
                 if (actor || pf2Traits) {
-                    const origin: Record<string, unknown> = {};
+                    const origin: Record<string, unknown> = (flags.pf2e.origin as Record<string, unknown>) ?? {};
```
digitalpacman, line 318: Used this technique in order to alter the least number of lines

src/module/system/text-editor.ts
```diff
@@ -223,7 +223,7 @@ class TextEditorPF2e extends TextEditor {
             case "Localize":
                 return this.#localize(paramString, options);
             case "Template":
-                return this.#createTemplate(paramString, inlineLabel, item?.system);
+                return this.#createTemplate(paramString, inlineLabel, item, item?.system);
```
stwlam, line 226: If you're passing the whole item, you needn't additionally pass the `system` property.

src/module/system/text-editor.ts
```diff
@@ -314,11 +319,32 @@ class TextEditorPF2e extends TextEditor {
...
+            }
+
             // Add the html elements used for the inline buttons
             const html = document.createElement("span");
             html.innerHTML = label;
             html.setAttribute("data-pf2-effect-area", params.type);
             html.setAttribute("data-pf2-distance", params.distance);
+            if (params.templateData) {
+                html.setAttribute("data-pf2-template-data", params.templateData);
+            }
```
stwlam, line 347:
```suggestion
            if (params.templateData) {
                html.dataset.templateData = params.pf2TemplateData;
            }
```

src/scripts/ui/inline-roll-links.ts
stwlam, line 318:
```suggestion
                    const origin: Record<string, unknown> = flags.pf2e.origin ?? {};
```

src/module/system/text-editor.ts
digitalpacman: I wanted to change the least amount of code, and technically leaving it as an arg allows people to disconnect those two things. I'll keep it to a single argument.

src/module/system/text-editor.ts
```diff
@@ -314,11 +319,32 @@ class TextEditorPF2e extends TextEditor {
                 });
             }
 
+            if (item) {
+                const itemTraits = item.system?.traits?.value;
```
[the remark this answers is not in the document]
digitalpacman: fixed

src/module/system/text-editor.ts
digitalpacman: fixed

src/module/system/text-editor.ts
digitalpacman: Why do this differently as opposed to doing it the way the other properties are done?

src/scripts/ui/inline-roll-links.ts
digitalpacman: This won't compile because flags.pf2e is record<string, unknown> so .origin is unknown type

src/scripts/ui/inline-roll-links.ts
```diff
@@ -315,7 +315,7 @@ export const InlineRollLinks = {
 
                 const actor = resolveActor(foundryDoc, link);
                 if (actor || pf2Traits) {
-                    const origin: Record<string, unknown> = {};
+                    const origin = (flags.pf2e.origin as Record<string, unknown>) ?? {};
```
stwlam, line 318: Don't assert

digitalpacman: https://en.wikipedia.org/wiki/Assertion_(software_development)

digitalpacman: image

digitalpacman: image

digitalpacman: image

digitalpacman: I'm trying to find a way to rewrite this whole thing so explicit casting isn't required

src/scripts/ui/inline-roll-links.ts
```diff
@@ -294,39 +293,39 @@ export const InlineRollLinks = {
...
-                if (actor || pf2Traits) {
-                    const origin: Record<string, unknown> = {};
-                    if (actor) {
-                        origin.actor = actor.uuid;
-                    }
-                    if (pf2Traits) {
-                        origin.traits = pf2Traits.split(",");
-                    }
-                    flags.pf2e.origin = origin;
+                const origin = isRecord(pf2e.origin) ? pf2e.origin : {};
```
digitalpacman, line 314: @stwlam  I changed the code to use a type guard instead.  Is this good enough?  I could not find another way other way other than changing the documents types so that pf2e is Record<string, Record<string, unknown>> over Record<string, unknown>. That or using the any value from a json serialization.

PR #13284, opened 2024-01-26
Automate Psychic Amps

TikaelSol:
Need to chase down the the DamageDice REs aren't working to handle the additional heightening effects, but these are otherwise ready. The feat based amps are localized but not handled until we have a way to dynamically build suboptions.

PR #13312, opened 2024-01-27
Add Spell Effect: Dragon Wings that grants Spell Effect: Dragon Claws

Cuingamehtar:
Combine Dragon Wings flying speed RE;
~Add Spell Effect: Dragon Wings (2 Focus Points) that grants Spell Effect: Dragon Claws.~

Add Spell Effect: Dragon Wings (10 minutes)
Add 2 Focus Point toggle for Dragon Wings spell
Make Spell Effect: Dragon Wings grant effects of Dragon Claw spell when the roll option is toggled.


stwlam:
This could probably be done with the single effect and a toggle on the draconic and wymblessed class features.

Cuingamehtar:
I'm not sure that a toggle is necessary since Dragon Wings are tied to these sorcerer bloodlines (and Dragon Disciple, that is supposed to be associated with a dragon too).
I also missed the heightening.
This would likely mean that Dragon Claws RE would have to be copied to Dragon Wings, unless I'm missing a RE that can affect the effect duration (both this and granted).

stwlam:
The toggle would be to switch between one and two focus points

TikaelSol:
If we delete effects we have to migrate the links. Those effects definitely need to go, but if we're replacing them then we should use the same ID, and if not we should keep the old effect and add them to the list of effects to migrate away: #4144

Cuingamehtar:
I kept the same ID, just renamed the files. Thankfully, the number of items stays the same in this case. Although I didn't verify the ordering of references in the description of imported spells.
It would be great if it was possible to compress it in one (ItemAlter the duration based on item.level), but I expect that would require modifications to the RE to change the property only on item creation.

pedrogrullada:
We could use an Item Alteration and a toggle on the spell itself to make it cost 2 focus points and append the Dragon Claws effect

Cuingamehtar:
I'm not sure how ItemAlteration would work for Focus Point cost (additional RollOption toggle?). Can he information from the spell (number of FP used) be passed into the Spell Effect somehow? If not, this would grow the number of Spell Effects to 4: 1FP-1minute/2FP-1minute/1FP-10minutes/2FP-10minutes.
I can see how the Effect reference can be added to the chat card depending on the spell level. And what about looking at the spell card in a compendium, at least one of the Effects should be visible there, right?

pedrogrullada:
What I mean is 3 REs on the spell itself: a toggleable Roll Option, and two Item Alterations: one to alter the focus point cost and one to add the dragon claws spell effect to the description, in which case there wouldn't be a need for a Dragon Wing effect that grants Dragon Claws. Unfortunately, since heightened Dragon Wings make the claws last 10 minutes, you don't end up with any less effects at the end of the day, so maybe just disregard this.

Cuingamehtar:
Also for consideration (not necessarily in this PR, or ever) - putting the actual applicable effect into the ItemAlteration added text like this
image
Here I also removed the effect references in the base description, which, I think, looks neater in the chat, but means they disappear in the spellcasting panel and in the compendium.

packs/spell-effects/spell-effect-dragon-wings-10-minutes.json
```diff
@@ -0,0 +1,100 @@
...
+                        "value": 1
+                    },
+                    {
+                        "label": "2 Focus Points",
+                        "value": 2
+                    }
+                ],
+                "prompt": "Focus Points spent",
+                "rollOption": "dragon-wings-focus-points"
+            },
```
TikaelSol, line 44: Drop this and instead put an ItemAlteration on the spell and a toggle placed on the spellcasting tab. Then a note that will add a link to the dragon claws effect at the same time. see #13284 for how to do those things.

TikaelSol:
Sorry for the conflicts, been trying to get that big PR in for some time now. Should be easy to resolve though.

Cuingamehtar:
Please confirm that I did it the intended way.

Also, I'm not sure that `focus-point-cost` ItemAlteration works as expected: The system checks the number of FP correctly, but it subtracts 1 even if the checkbox is set.

Cuingamehtar:
This is the only Note (well, ItemAlteration description) that I can think of adding.
The Dragon Claw effect in the Spell Effect is toggled by that toggle in the spellcasting option. Otherwise there would have to be 4 Spell Effect items to cover all cases.

packs/spell-effects/spell-effect-dragon-wings-10-minutes.json
```diff
@@ -0,0 +1,84 @@
...
+        "tokenIcon": {
+            "show": true
+        },
+        "traits": {
+            "rarity": "common",
+            "value": []
+        }
+    },
+    "type": "effect"
+}
```
TikaelSol, line 84:
```suggestion
}

```

packs/spells/dragon-wings.json
```diff
@@ -29,7 +29,49 @@
...
+                        ]
+                    }
+                ],
+                "property": "description",
+                "value": [
+                    {
+                        "text": "@UUID[Compendium.pf2e.spell-effects.Item.Spell Effect: Dragon Wings (10 minutes)]"
+                    }
+                ]
+            }
```
TikaelSol, line 73: On one hand this is a neat way to approach this, on the other it will make migrating away this effect when we get duration alteration a huge pain.

packs/spells/dragon-wings.json
```diff
@@ -29,7 +29,49 @@
             "value": ""
         },
         "requirements": "",
-        "rules": [],
+        "rules": [
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "Spend 2 Focus Points on Dragon Wings",
```
TikaelSol, line 36: Localize this

PR #13313, opened 2024-01-27
Add Effects for Dragon's Eye Charm

DocSchlock:


TikaelSol:
Make this one effect with a choice set

DocSchlock:
I merged the two Sustained effects into a single one with ChoiceSet. I'm reluctant to put the non-Sustained Resistance in there, even if Sustained doesn't do anything aside from display to the users.

TikaelSol:
I don't think sustained is enough to justify having two different effects.

DocSchlock:
Combined all into one. I left the Sustained timing in the effect description so it's communicated in some way to the players.

PR #13317, opened 2024-01-27
Automate Spell Trickster notes

pedrogrullada:
On draft until we have dynamic suboptions.


static/lang/re-en.json
```diff
@@ -3036,6 +3036,122 @@
                 "Tenth": "10th-Rank",
                 "Third": "3rd-Rank"
             },
+            "SpellTrickster": {
+                "AgileHand": {
+                    "Label": "Agile Hand",
```
TikaelSol, line 3041: We shouldn't need the labels for these since the rules are on the items

packs/feats/spell-trickster-dedication.json
```diff
@@ -31,7 +31,187 @@
...
+        "rules": [
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.SpellTrickster.Label",
+                "option": "spell-trickster",
+                "placement": "spellcasting",
+                "suboptions": [
+                    {
+                        "label": "PF2E.SpecificRule.SpellTrickster.AgileHand.Label",
+                        "predicate": [
```
TikaelSol, line 43: Do any spells have multiple tricks possible? If so then we should use this approach but wait for us to have dynamic suboptions so 3pp can be supported. If not just a single toggle to use your trick.

pedrogrullada: You can only use one modification at a time per the dedication's rules
> You can only apply one modification to a spell at a time, even if you know more than one modification for that spell.

static/lang/re-en.json
pedrogrullada: The labels are for the suboptions in Spell Trickster Dedication. Or do you mean I should just use `{item|description}` the body of each Item Alteration instead of making localisation keys?

packs/feats/spell-trickster-dedication.json
TikaelSol: I don't think we should be doing prereqs as RE stuff, no. But yeah maybe we stew on this until we have dynamic suboptions.

TikaelSol: We'll need that for amps too.

static/lang/re-en.json
TikaelSol: If we get dynamic suboptions then no need for the label at least. The notes can stay. But let's hold here for a bit and put this in draft.

pedrogrullada:
I goof'd the rebase, so I'm just going to make a new PR when dynamic suboptions get here. Sorry about that!

PR #13329, opened 2024-01-28
Append flat check note to spells under non-spellcaster Ancestral Curse

pedrogrullada:


pedrogrullada:
It looks like I missed the class predicate for moderate and major stage. Will fix tomorrow.

PR #13338, opened 2024-01-28
Allows familiars to use their master's hero points to reroll checks.

### LINKED ISSUE #10155: Feature Request: Allow Familiar to use it's owner's Hero Point to reroll a save
Trying this out with no modules activated, I can use a hero point to reroll a PC's save, but not the Familiar assigned to them. RAW states that you should be able to spend a Hero Point on behalf of your familiar or animal companion, so it'd be nice if it were possible to do so here with the standard right click function rather than manually.

xdy:
Fixes #10155

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -402,22 +402,30 @@ class CheckPF2e {
         const actor = game.actors.get(message.speaker.actor ?? "");
         let rerollFlavor = game.i18n.localize(`PF2E.RerollMenu.MessageKeep.${keep}`);
         if (heroPoint) {
+            const rerollingActor = actor?.isOfType("familiar") ? actor?.master : actor;
+
             // If the reroll costs a hero point, first check if the actor has one to spare and spend it
-            if (actor?.isOfType("character")) {
-                const heroPointCount = actor.heroPoints.value;
+            if (rerollingActor?.isOfType("character")) {
+                const heroPointCount = rerollingActor?.heroPoints.value;
                 if (heroPointCount) {
-                    await actor.update({
-                        "system.resources.heroPoints.value": Math.clamped(heroPointCount - 1, 0, actor.heroPoints.max),
+                    await rerollingActor.update({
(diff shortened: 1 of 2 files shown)


src/module/system/check/check.ts
```diff
@@ -402,22 +402,30 @@ class CheckPF2e {
         const actor = game.actors.get(message.speaker.actor ?? "");
```
stwlam, line 402:
```suggestion
        const actor = message.actor;
```

src/module/system/check/check.ts
```diff
@@ -399,25 +399,33 @@ class CheckPF2e {
             return;
         }
 
-        const actor = game.actors.get(message.speaker.actor ?? "");
+        const actor = message.actor;
         let rerollFlavor = game.i18n.localize(`PF2E.RerollMenu.MessageKeep.${keep}`);
         if (heroPoint) {
+            const rerollingActor = actor?.isOfType("familiar") ? actor?.master : actor;
```
stwlam, line 405:
```suggestion
            const rerollingActor = actor?.isOfType("familiar") ? actor.master : actor;
```

src/module/system/check/check.ts
```diff
@@ -399,25 +399,33 @@ class CheckPF2e {
...
+        const actor = message.actor;
         let rerollFlavor = game.i18n.localize(`PF2E.RerollMenu.MessageKeep.${keep}`);
         if (heroPoint) {
+            const rerollingActor = actor?.isOfType("familiar") ? actor.master : actor;
+
             // If the reroll costs a hero point, first check if the actor has one to spare and spend it
-            if (actor?.isOfType("character")) {
-                const heroPointCount = actor.heroPoints.value;
+            if (rerollingActor?.isOfType("character")) {
+                const heroPointCount = rerollingActor?.heroPoints.value;
```
stwlam, line 409:
```suggestion
                const heroPointCount = rerollingActor.heroPoints.value;
```

PR #13340, opened 2024-01-28
Add Failure Note to Pernicious Spore Bomb

### LINKED ISSUE #2128: create an effect for the pernicious spore bomb in the re.json in case of failure
In GitLab by @rectulo on Mar 5, 2022, 03:23

This one is more tricky to handle (Why Paizo, Why !?)

_The bomb deals [[/r {1}[poison]]]{1poison Damage}, [[/r {1d4}[persistent,piercing]]] @Compendium[pf2e.conditionitems.lDVqvLKA6eF3Df60]{Persistent Piercing Damage}, and [[/r {1}[splash,poison]]]{1 poison splash damage}. **Except on a critical failure,** the bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round._

There is no special effect for the bomb if the Strike is a critical **failure** (not a critical success) as other bombs. But there is one in case of success, failure and greater success.

There should be a failure entry with only the special effects **without the damage**

DocSchlock:
Removes the unnecessary current Notes and adds the special Terrain-specific note to the _attack roll_ instead of damage, so it can show on Failure as well.

Closes #2128

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -382,22 +382,8 @@
                 }
             },
             "PerniciousSporeBomb": {
-                "Greater": {
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*3d4)[persistent,piercing]]] and [[/r (3[splash])[poison]]]{3 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round.",
-                    "success": "<strong>Effect</strong> The bomb also deals [[/r 3d4[persistent,piercing]]] and [[/r (3[splash])[poison]]]{3 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round."
-                },
-                "Lesser": {
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*1d4)[persistent,piercing]]] and [[/r (1[splash])[poison]]]{1 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round.",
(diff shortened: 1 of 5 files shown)


static/lang/re-en.json
```diff
@@ -382,22 +382,8 @@
...
-                },
-                "Major": {
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*4d4)[persistent,piercing]]] and [[/r (4[splash])[poison]]]{4 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 minute.",
-                    "success": "<strong>Effect</strong> The bomb also deals [[/r 4d4[persistent,piercing]]] and [[/r (4[splash])[poison]]]{4 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 minute."
-                },
-                "Moderate": {
-                    "criticalSuccess": "<strong>Effect</strong> The bomb also deals [[/r (2*2d4)[persistent,piercing]]] and [[/r (2[splash])[poison]]]{2 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round.",
-                    "success": "<strong>Effect</strong> The bomb also deals [[/r 2d4[persistent,piercing]]] and [[/r (2[splash])[poison]]]{2 poison splash damage}. The bomb's splash area is coated in vegetation, becoming difficult terrain for 1 round."
-                }
```
TikaelSol, line 400: We'd need to migrate away notes that include these. Though, maybe it would be fine since bombs are consumables and this would not be on many actors.

DocSchlock: Bah, didn't think of that. Should I revert the removal of the old notes for the time being?

TikaelSol: For now, yeah.

PR #13342, opened 2024-01-28
Add REs to Nimble Crawl

### LINKED ISSUE #7491: [Feature Request] Don't apply flatfooted from prone when you have Nimble Crawl and are Legendary in Acrobatic
The Acrobatics skill feat Nimble Crawl states "...and if you’re legendary, you aren’t flat-footed while prone."

I would like to request this functionality be added to the system.

DocSchlock:
Closes #7491

### DIFF
--- a/packs/feats/nimble-crawl.json
+++ b/packs/feats/nimble-crawl.json
@@ -28,7 +28,29 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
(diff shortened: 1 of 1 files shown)


TikaelSol:
This only suppresses the penalty, but abilities that fire off on off-guard targets would still be active, like sneak attack.

DocSchlock:
I can't see a way to suppress an entire condition, so I'm closing this for now as not actionable.

PR #13351, opened 2024-01-29
Add Defense to spell description headers

j-bs:
Following on from the change in #13140 this adds the `Defense [...]` label to the prepended spell descriptions.

I need input on which display we prefer though: inline or separate line to `Duration`. I prefer inline but it does mean that some spell chat cards have to wrap the Duration value if it's wordy. The current PR is the _inline_ implementation.

Comparisons:
spell-defense-prepend

I also experimented with displaying a "fully complete" defense label for niche cases that have _both_ an AC and stat defense like Disintegrate but I'm not convinced it's worth the mess to implement 🤷 

Here's what one iteration looked like (note slightly fatter spell slot row):
spell-defense-prepend-complex

IMO it does look nice in the spell cards but not in the new Defense column, however I'm not convinced it's worth all the janky code changes vs the current PR which is a super simple template change. Happy to polish the more complete version if that's more preferable though.

### DIFF
--- a/src/styles/legacy/_chat.scss
+++ b/src/styles/legacy/_chat.scss
@@ -58,6 +58,12 @@
         p {
             margin: var(--space-4) 0;
             min-height: unset;
+
+            &.spell-defense-duration {
+                display: flex;
+                flex-wrap: wrap;
+                gap: var(--space-4);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


stwlam:
Inline is fine: make the wrapping happen per flex column.

j-bs:
> Inline is fine: make the wrapping happen per flex column.

Took a bit of wrangling flex CSS but got it looking nice and consistent:

image

stwlam:
Hmm, can the columns themselves wrap?

PR #13366, opened 2024-01-29
Restrict Megaton Strike to specified Strikes

pedrogrullada:
Also switched `value` for `diceNumber` on the rule elements, as the dice were not being added


stwlam:
You're going to need to update the migration as well.

pedrogrullada:
> You're going to need to update the migration as well.

I'll see if I can figure that out, but if I can't I suppose I can just draft this until next release

PR #13374, opened 2024-01-29
Add roll options for spellcasting traditions

xyzzy42:
When an actor can cast spells it currently gets a roll option named "self:caster" (or "target:caster", etc. depending on context).

Extend this with "self:caster:divine", etc. for whatever tradition(s) the actor has.  "self:caster" is still there if any traditions are known.

Like the existing isSpellcaster() check, the tradition doesn't count unless it actually has spells.

The motiviation is to be able to add automation for things like Divine Disharmony, which triggers on, amongst other things, "...  creature with divine spells." Or Robe of the Archmagi, which "benefit only characters who can cast arcane spells."

### DIFF
--- a/src/module/actor/creature/document.ts
+++ b/src/module/actor/creature/document.ts
@@ -393,10 +393,6 @@ abstract class CreaturePF2e<
         const { attributes, rollOptions } = this;
 
         // Add creature-specific self: roll options
-        if (this.isSpellcaster) {
-            rollOptions.all["self:caster"] = true;
-        }
-
         if (this.hitPoints.negativeHealing) {
             rollOptions.all["self:negative-healing"] = true;
         }
--- a/src/module/item/spellcasting-entry/document.ts
+++ b/src/module/item/spellcasting-entry/document.ts
@@ -135,6 +135,11 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
                 4,
             ) as OneToFour;
         }
+
+        if ((this.spells?.size ?? 0) > 0) {
+            actor.rollOptions.all["self:caster"] = true;
+            actor.rollOptions.all[`self:caster:tradition:${this.tradition}`] = true;
+        }
     }
 
     /** Prepares the statistic for this spellcasting entry */


src/module/actor/creature/document.ts
```diff
@@ -156,6 +156,12 @@ abstract class CreaturePF2e<
         return itemTypes.spellcastingEntry.length > 0 && itemTypes.spell.length > 0;
     }
 
+    get spellcastingTraditions(): Set<string | null> {
+        return new Set(
+            this.itemTypes.spellcastingEntry.filter((e) => e.spells && e.spells.size > 0).map((e) => e.tradition),
+        );
+    }
```
stwlam, line 163: A single use during data preparation doesn't really warrant a new getter.

xyzzy42: Ok.  `isSpellcasting` is one, but it's used in an html template, so I suppose that is the difference.

src/module/actor/creature/document.ts
```diff
@@ -393,8 +393,12 @@ abstract class CreaturePF2e<
         const { attributes, rollOptions } = this;
 
         // Add creature-specific self: roll options
-        if (this.isSpellcaster) {
+        const traditions = new Set(
```
stwlam, line 396: This actually looks a tad heavy. It'd pretty be lighter if don't from the spellcasting entries.

xyzzy42: You're last sentence doesn't parse for me.  Don't make a set?  It's possible to have the same spellcasting type from multiple entries.  I suppose it doesn't matter to set the same roll option multiple times.

stwlam: Sorry about that, swyping from couch!

xyzzy42: Do you mean to add this instead to `item.spellcasting-entry.document.preapreActorData()`?

Should the base `self:caster` option go there too? 

Like this:

```diff
--- a/src/module/item/spellcasting-entry/document.ts
+++ b/src/module/item/spellcasting-entry/document.ts
@@ -135,6 +135,11 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
                 4,
             ) as OneToFour;
         }
+
+        if (this.spells && this.spells.size > 0) {
+           actor.rollOptions.all["self:caster"] = true;
+           actor.rollOptions.all[`self:caster:${this.tradition}`] = true;
+        }
     }

     /** Prepares the statistic for this spellcasting entry */
```

stwlam: Yeah, like that. Use `self:caster:tradition:${this.tradition}` as well.

PR #13417, opened 2024-01-31
Hide merchant coins and empty loot sections from players

CarlosFdez:
GM view
image

Player View
image

Coins in containers are purposefully not hidden as it would cause strange behavior if the container itself as sold (and if it also sold what was inside the container, which it should.....)


stwlam:
I still see coins at the top

CarlosFdez:
I'm ok with hiding those too if you prefer, my goal was more that gold coins are not for sale.

stwlam:
yeah let's hide those too

CarlosFdez:
The coinage section is now hidden from non-owners unless its not a merchant

image

stwlam:
Total bulk, too :)

CarlosFdez:
Done. I will have you imagine how bare this sheet is now.

stwlam:
I cannot wait to see!

PR #13432, opened 2024-01-31
Remove old darkvision and low-light-vision ancestry Features

Drental:
the system builds, looks like they were linked nowhere and without rules they are just traps.


TikaelSol:
I just removed these links in the last update. There were quite a few things that linked to them. We should migrate those links away.

PR #13442, opened 2024-02-01
Fix styling regression in Kingdom sheet sub-nav header

j-bs:
This should be a more reliable way of maintaining the centered, evenly spread sub-nav links now that we have a Kingdom sheet with more sub-nav links than the player sheet and party sheet.

image


stwlam:
Spells?

PR #13456, opened 2024-02-01
Correctly set all Feat Max Taking values

DocSchlock:
In instances where a feat can only be taken multiple times based on character level, I set the max taking to the total number of times the feat can be taken levels 1-20.

### DIFF
--- a/packs/feats/additional-companion.json
+++ b/packs/feats/additional-companion.json
@@ -16,6 +16,7 @@
         "level": {
             "value": 4
         },
+        "maxTakable": 4,
         "prerequisites": {
             "value": [
                 {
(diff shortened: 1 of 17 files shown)


TikaelSol:
I don't get what you mean with that comment.

DocSchlock:
> I don't get what you mean with that comment.

Just a clarifying note.

Example: at level 1, you can take the feat once. At level 14, you may take it again. At level 18, you may take it for a third and final time. In this case, I set the maxTake to 3x.

There are a couple feats like that.

TikaelSol:
Ah, I would avoid that in case we get level 21+ play at some point or for people who homebrew.

DocSchlock:
I'll set those ones to uncapped then.

PR #13460, opened 2024-02-01
Update Deities to add support for two skills instead of one

### LINKED ISSUE #2258: [Feature Request] Able to pick one out of several divine skills for deity items.
In GitLab by @yoann.felix on May 1, 2022, 11:23

Currently, you can pick only one divine skill for a deity.

image

but some for some deity, you have the option to pick one out of two.

image

Would it be possible to have the option to pick more than one?

DocSchlock:
Closes #2258

Includes migration for existing Deities. 

Allows Deities to have two skills instead of one and updates all existing Deity files.


src/module/item/deity/data.ts
```diff
@@ -15,7 +15,7 @@ type DeitySystemSource = ItemSystemSource & {
     };
     font: DivineFonts;
     attribute: AttributeString[];
-    skill: SkillAbbreviation | null;
+    skill: SkillAbbreviation[] | null;
```
stwlam, line 18: Let's change this to long form while we're at it

DocSchlock: Changed in latest commit.

static/lang/en.json
```diff
@@ -1839,7 +1839,10 @@
                     "Heal": "Heal",
                     "Label": "Divine Font"
                 },
-                "DivineSkill": "Divine Skill",
+                "DivineSkill": {
+                    "Hint": "Some deities offer the choice between two skills instead of one.",
```
stwlam, line 1843:
```suggestion
                    "Hint": "Most deities have a single divine skill, though some offer a choice between two.",
```

PR #13474, opened 2024-02-02
Refactor item casting collection and fix erroneous item casting rendered collections

CarlosFdez:
Also allows creating item casting entries bound to a specific item.


src/module/actor/creature/document.ts
```diff
@@ -297,13 +295,10 @@ abstract class CreaturePF2e<
                     actor: this,
                     statistic: ability.statistic,
                     tradition: ability.tradition ?? spell.traditions.first() ?? null,
-                    castPredicate: new PredicatePF2e([`item:id:${consumable.id}`, `spell:id:${spell.id}`]),
+                    consumable,
```
stwlam, line 298: This seems unwise as a key since it won't be true the moment we introduce staves or fix wands.

CarlosFdez: Staves are likely to use an entirely different thing (since each individual spell within a stave can be cast using different spellcasting modifiers??) but good point on wand, it should be renamed to item.

PR #13487, opened 2024-02-02
Add context-menu option to reroll as a blind roll

### LINKED ISSUE #1097: [Suggestion] Reroll as Blind Roll
In GitLab by @ImmortalLobster on Feb 23, 2021, 11:43

Frequently players roll skill checks that are meant to be secret but mistakenly roll them openly, and GMs will likely make players roll the check again as a blind roll, ignoring the open roll. It would be helpful if an additional reroll option was added to the menu when you right click on a roll to reroll the check as a blind roll with the same modifiers.

DocSchlock:
Closes #1097


BernhardPosselt:
Can confirm that this fixes #13660 and that re-rolling as a blind roll works

DocSchlock:
Bugfix for rerolls was moved out of this PR and merged separately.

PR #13495, opened 2024-02-03
Add create bonus feat button to kingdom

CarlosFdez:


stwlam:
Display as "feature"? :)

PR #13510, opened 2024-02-03
Add saving throws to the proficiencies tab and remove the proficiency rank from saving throws in the sidebar

Truuun:
I added the saving throws to the proficiency tab below the defenses section. Like the section for class DCs, you have a button for modifiers and can see a breakdown of the value by hovering over it. I didn't add the ability to roll the saving throw from the section as I figured you'd probably want to keep doing that from the sidebar and that saves some clutter.
Speaking of the sidebar, I removed the proficiency rank element for the saving throws there to make that area a tiny bit less packed.
SavingThrowsSection
Sidebar

### DIFF
--- a/static/templates/actors/character/partials/sidebar.hbs
+++ b/static/templates/actors/character/partials/sidebar.hbs
@@ -264,7 +264,6 @@
     {{#each data.saves as |save slug|}}
         <li class="roll-data" data-save="{{slug}}">
             <h2 class="sidebar_label">{{localize save.label}}</h2>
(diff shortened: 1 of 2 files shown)


TikaelSol:
Renamed to be in imperative case

stwlam:
After some discussion, we decided this wouldn't be a desirable change. Sorry!

Truuun:
Thanks for your response! I can see rationale behind not doing it like it becoming inconsistent with perception and initiative and ending up with duplicate data. I'll take in more considerations for next time!

PR #13533, opened 2024-02-04
Always create RitualSpellcasting instances

stwlam:
(and defer retrieval of rituals for their collections)


CarlosFdez:
If I recall, this deferral doesn't really lead to anything, since ActorSpellcasting retrieves the collections immediately. But I suppose that may change if collection retrieval is deferred.

PR #13535, opened 2024-02-04
Add Round Timer to Encounter

### LINKED ISSUE #2296: Feature Request: Add Time Elapsed next to Round Counter
In GitLab by @bland-html on May 16, 2022, 24:16

Foundry staff suggested this might be better as a system feature.

### Feature Summary

Showcase the time elapsed next to the round counter. This would help improve tracking ability while minimizing module dependency.

### User Experience

Instead of just showing Round 2, it could show Round 2 - 00:06. If needed, users could have a setting to input how much time 1 Round is equal to or just disable it altogether. As rounds move forward, the timer simply elapses however much time the setting is set to.

Ex. Setting - 1 Round = 6 Seconds. Round is currently Round 2. Timer shows 00:06. When Round 3 happens. Timer shows 00:12.

### Priority/Importance

Right now there are a crap ton of things to track. I think adding a simple visual Time Elapsed counter will help many DM's and Players make quicker decisions when considering durations of many different effects. Sometimes messing around with many modules slows down gameplay significantly. Simplistic visual information is always helpful at reducing this.

Since you can have multiple Encounters, this function would also essentially allow an Encounter to act as a makeshift Time Tracker. This would help reduce module dependency as well.

DocSchlock:
Closes #2296

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -149,6 +149,7 @@ interface GamePF2e
             };
             /** Rules-based vision */
             rbv: boolean;
+            showRoundTimer: boolean;
             tokens: {
                 /** Automatic scaling of tokens belong to small actor */
                 autoscale: boolean;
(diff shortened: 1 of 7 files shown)


stwlam:
What's this?

stwlam:
I think that issue was created before automatic effect expiration was added.

PR #13538, opened 2024-02-04
Add Protection aura spell effect

ChristosLabrou:
- Adds `Aura: Protection` spell effect for Protection heightened to rank 3+
- Adjusts the Protection spell to link the aura as well


pedrogrullada:
This might be an old Bless/Bane situation, but are we sure 3rd-rank Protection is an aura? The way it's written right now seems to suggest it affects the creatures within the emanation at the time of casting, rather than it being a persistent aura.

stwlam:
This is an instantly-resolving emanation, much like bard composition cantrips.

PR #13573, opened 2024-02-06
Show additional icons for character sheet choices

j-bs:
Thoughts on expanding the stylised icon that we use for deity to the other choices too?

The default icons for some of the choices are a little jarring but IMO it's still quite charming.

show-player-ABCH-icons


stwlam:
It's something we've considered but decided against due to lack of icons and extra space consumption.

j-bs:
Yeah that's fair, I figured that might be the case.

Soon™

PR #13580, opened 2024-02-06
Accommodate subitems in certain rule elements

stwlam:


src/module/rules/rule-element/item-alteration/rule-element.ts
```diff
@@ -116,6 +109,30 @@ class ItemAlterationRuleElement extends RuleElementPF2e<ItemAlterationRuleSchema
...
+        }
+
+        if (this.itemType === "condition") {
+            return this.actor.conditions.contents;
+        }
+
+        if (this.itemType) {
+            const physicalItemTypes: Set<string> = PHYSICAL_ITEM_TYPES;
+            return physicalItemTypes.has(this.itemType)
+                ? [this.actor.itemTypes[this.itemType], this.actor.inventory.map((i) => i.subitems.contents)].flat(2)
```
CarlosFdez, line 130: You need to also filter the subitems by the item type I think.l

PR #13622, opened 2024-02-08
Add casting flag to spell damage card

reonZ:
- It makes it that an embedded spell retrieved using `ChatMessagePF2e#item` isn't parent-less
- It adds the `casting` flag to the `DamagePF2e#roll`


CarlosFdez:
I'm not comfortable with adding the casting flag to the DamageRollContext. We don't expose DamagePF2e, but even if we did (and we kinda want to expose something along those lines in the future) casting shouldn't be a parameter of the context. Maybe the spellcasting entry if needed.

Would it be possible to instead in DamagePF2e create the casting flag yourself instead?

reonZ:
> Would it be possible to create the casting flag in DamagePF2e instead?

it definitely can, i totally missed that, makes more sense.

CarlosFdez (approved):

CarlosFdez:
Looks fine to me, but after the next hotfix release.

src/module/chat-message/data.ts
```diff
@@ -22,12 +22,19 @@ export interface ItemOriginFlag {
     rollOptions?: string[];
 }
 
+export interface ChatCastingFlag {
```
stwlam, line 25: export at bottom of file

reonZ: the interface just above was using `export` directly so i just mimicked, i will do it for both then

reonZ:
After reviewing with shark, this solution cannot work when generating the `casting` flag directly inside `DamagePF2e`, at least not for consumables that are destroyed on use like scrolls.

So let's scrap this all together and think of something different.

CarlosFdez:
What allows this solution to work when in the context that makes it not work in DamagePF2e? Isn't both the spellcasting entry and the parent item part of the spell itself and can be fished out?

reonZ:
The issue is that i wanted to reproduce what was already done, by doing
```js
if (item.parentItem) {
    casting.embeddedSpell = item.toObject();
}
```
But in `DamagePF2e`, the spell is already the parent-less version, so `item.parentItem` is always null.

The solution would either to have, inside the spell source, something that says "hey, i was actually an embedded spell" or, always add the spell source to the flags regardless of it was embedded or not, so removing the `item.parentItem` condition and just have a `casting.spellSource` flag or something similar.

PR #13623, opened 2024-02-08
Remove redundant retrieval of actor in `getCheckDC` inline roll links helper

Drental:
This will allow the PFS modules to implement Influence Statblocks without excessive duplication, but I'm sure there could be other uses.

I've checked to make sure that the Journal and Chat are not affected by this in any way.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -879,8 +879,8 @@ function getCheckDC({
 
     if (base) {
         const getStatisticValue = (selectors: string[]): string => {
-            if (item?.actor && params.immutable !== "true") {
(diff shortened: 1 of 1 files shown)


src/module/system/text-editor.ts
```diff
@@ -879,8 +879,7 @@ function getCheckDC({
 
     if (base) {
         const getStatisticValue = (selectors: string[]): string => {
-            if (item?.actor && params.immutable !== "true") {
-                const { actor } = item;
+            if (actor && params.immutable !== "true") {
```
stwlam, line 882:
```suggestion
            if (actor && params.immutable === undefined) {
```

PR #13641, opened 2024-02-10
Normalize setting/reading from boolean dataset attributes

stwlam:


src/module/system/text-editor.ts
```diff
@@ -553,34 +553,26 @@ class TextEditorPF2e extends TextEditor {
...
-                if (rawParams.traits) traits.push(...rawParams.traits.split(",").map((trait) => trait.trim()));
-
-                // Deduplicate traits
-                return Array.from(new Set(traits));
-            })(),
+            traits,
             // Set action slug, damaging effect for basic saves, and any parameterized options
             extraRollOptions: R.compact([
-                ...(rawParams.basic === "true" ? ["damaging-effect"] : []),
+                ...("basic" in rawParams ? ["damaging-effect"] : []),
```
CarlosFdez, line 572:
```suggestion
                ...(params.basic ? ["damaging-effect"] : []),
```

I wonder why it didn't do this from the start

stwlam: data entry code

src/module/system/text-editor.ts
```diff
@@ -749,8 +741,8 @@ class TextEditorPF2e extends TextEditor {
             return null;
         }
 
-        const item = args.rollData?.item instanceof ItemPF2e ? args.rollData?.item : null;
-        const actor = (args.rollData?.actor instanceof ActorPF2e ? args.rollData?.actor : null) ?? item?.actor ?? null;
+        const item = args.rollData?.item ?? null;
+        const actor = args.rollData?.actor ?? item?.actor ?? null;
```
CarlosFdez, line 745: I remember this was for a reason, but I can not recall it. We can gamble if you're feeling lucky.

src/module/system/text-editor.ts
```diff
@@ -901,7 +892,7 @@ function getCheckDC({
 
         const slugName = sluggify(name);
         if (type === "flat") {
-            return params.immutable === "false"
+            return params.immutable === undefined
```
CarlosFdez, line 895: this one isn't quite right

src/module/system/text-editor.ts
```diff
@@ -762,7 +754,7 @@ class TextEditorPF2e extends TextEditor {
         const traits = ((): string[] => {
             const fromParams = params.traits?.split(",").flatMap((t) => t.trim() || []) ?? [];
             const fromItem = item?.system.traits?.value ?? [];
-            return params.overrideTraits === "true" ? fromParams : R.uniq([...fromParams, ...fromItem]);
+            return "overrideTraits" in params ? fromParams : R.uniq([...fromParams, ...fromItem]);
```
CarlosFdez, line 757: needs updating

src/module/system/text-editor.ts
stwlam: Not feeling lucky, but would be interested in tripping over it if it is needed so we can restore and also document why.

PR #13669, opened 2024-02-10
Fix non-hero point rerolls

### LINKED ISSUE #13660: Re-rolling checks via context menu broken except for hero points
**Affected version**: latest system release (5.13.5)

Steps to reproduce:

* Create a new world
* Create a new actor and assign it to a player
* Log in as the player and roll an unarmed attack
* Right click on the roll result and choose either Reroll and keep the new result/lower result/higher result
* Getting an error: **No actor is associated to the roll**
```
No actor is associated to the roll
fetch @ foundry.js:68326
notify @ foundry.js:68223
error @ foundry.js:68260
rerollFromMessage @ pf2e.mjs:1357
callback @ pf2e.mjs:1658
#onClickItem @ foundry.js:64986
dispatch @ jquery.min.js:2
y.handle @ jquery.min.js:2
```

Looks like if you don't use a heropoint, this one just jumps straight into the warning block? https://github.com/foundryvtt/pf2e/blob/master/src/module/system/check/check.ts#L405

DocSchlock:
Closes #13660

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -401,7 +401,13 @@ class CheckPF2e {
         }
 
         const actor = message.actor;
+        if (!actor) {
+            ui.notifications.error("PF2E.RerollMenu.ErrorNoActor", { localize: true });
+            return;
+        }
+
         let rerollFlavor = game.i18n.localize(`PF2E.RerollMenu.MessageKeep.${keep}`);
+
         if (heroPoint) {
             const rerollingActor = actor?.isOfType("familiar") ? actor.master : actor;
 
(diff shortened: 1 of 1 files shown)


src/module/system/check/check.ts
```diff
@@ -401,7 +401,13 @@ class CheckPF2e {
...
         const actor = message.actor;
+        if (!actor) {
+            ui.notifications.error("PF2E.RerollMenu.ErrorNoActor", { localize: true });
+            return;
+        }
+
         let rerollFlavor = game.i18n.localize(`PF2E.RerollMenu.MessageKeep.${keep}`);
+
         if (heroPoint) {
             const rerollingActor = actor?.isOfType("familiar") ? actor.master : actor;
```
stwlam, line 412:
```suggestion
            const rerollingActor = actor.isOfType("familiar") ? actor.master : actor;
```

PR #13706, opened 2024-02-12
Add item links to item transfer chat messages

### LINKED ISSUE #13680: Loot & deposit chat message item links
Party members, and even GMs when running APs, might miss details of looted and deposited items. This happens when a greedy player loots a corpse and others only see the name of the item. It would be nice if this name was a link to the item in question.

image

j-bs:
Closes #13680

Inserts the item link for all of the valid A -> B item transfer chat messages.

The default item link styling was way too chunky without word wrap so I experimented with a simple underline instead. It still maintains all the clicky, hover etc effects:

image

### DIFF
--- a/src/module/actor/item-transfer.ts
+++ b/src/module/actor/item-transfer.ts
@@ -298,7 +298,7 @@ export class ItemTransfer implements ItemTransferData {
         const formatProperties = formatArgs[1];
         if (!formatProperties) throw ErrorPF2e("Unexpected item-transfer failure");
         formatProperties.quantity = this.quantity;
-        formatProperties.item = item.name;
+        formatProperties.item = await TextEditor.enrichHTML(item.link, { async: true });
 
         // Don't bother showing quantity if it's only 1:
         const content = await renderTemplate(this.#templatePaths.content, {
--- a/src/styles/ui/sidebar/chat/_action.scss
+++ b/src/styles/ui/sidebar/chat/_action.scss
@@ -6,6 +6,19 @@
         align-items: center;
         font-weight: normal;
 
+        a.content-link {
+            background: none;
+            border: none;
+            padding: 0;
+            text-decoration: underline;
+            white-space: unset;
+            word-break: normal;
+
+            > i {
+                margin-inline: 0.125rem;
+            }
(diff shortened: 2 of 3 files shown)


stwlam:
What does the standard format look like?

j-bs:
The default styling? Would be something like this without solving the lack of word wrap:

image

stwlam:
I see, good call. Perhaps just tweak it a little to bring the icon closer to the linked text.

j-bs:
Halving the existing right margin looks much better:

(Although specific combos like the angled treasure icon and an italics _A_ play tricks on the eyes 👀 )

image

stwlam:
It's more that it looks a lot closer to the text to the left with the italics and without the spacing from the content link box.

j-bs:
Yeah that's fair, I was focused on the right margin.

Here's two slightly different left margin values, `0.125rem` both sides or `left 0.25rem / right 0.125rem`:
image

CarlosFdez:
Does this still work if the user has no permission for the destination actor?

I like this underline variant. Not for this PR but it might be worth it as a mixin in the future if we can't bolt on additional css classes.

j-bs:
> Does this still work if the user has no permission for the destination actor?

No it should fail in whichever way already exists. This PR doesn't touch any of the transfer permissions.

From memory, you get a UI error explaining that you don't have permission.

> I like this underline variant. Not for this PR but it might be worth it as a mixin in the future if we can't bolt on additional css classes.
👍

stwlam:
Unidentified items?

j-bs:
Ah good call, my lack of experience in the system is showing 🥲 

Unidentified items are currently clickable but hilariously on click it just puts the mystified description in a `warn` UI notification 🤨

 Do we wanna lock it down and just remove the link for mystified items?

image image

stwlam:
Maybe, but the warning kind of has its own charm. We can change our minds later.

j-bs:
Yeah I agree it's a cute message.

Updated the PR with the balanced margin on the icon.

Shall we migrate off Draft? 👀

stwlam:
ok

PR #13711, opened 2024-02-13
Implement custom proficiency ranks and maximum level

Kindpokemon:
I added the ability to increase the Level Maximum, as well as Custom Proficiency Ranks (mythic etc). THe former can be done through modules, the latter can be done through modules and/or a new tab in the Homebrew page

Here's the format for implementing both of them in a module
            "additionalLevels":5,
            "proficiencyRanks":{
               "mythic":{
                  "label":"Mythic",
				  "color": "#005C09",
                  "value":10
               }
            }


stwlam:
I think we'd be okay with making the codebase a little more agnostic as far as what proficiency ranks there are, but this is far too niche to have a setting.

TikaelSol:
Renamed to be in imperative case.

Kindpokemon:
If need be I can remove the setting window.

Kindpokemon:
Anything else need to be done to see this added to release? I've removed the tab from the settings window.

Kindpokemon:
Continuing to keep this updated. Anything that can be done to add it to the main branch? @stwlam

Kindpokemon:
Continuing to update. Any chance it can get merged @stwlam ?

CarlosFdez:
Unfortunately after looking at this even I think this is too niche to add such complexity to the codebase, so I'm going to close it. What shark meant by his suggestion was likely to put it in CONFIG so that a module can introduce these settings, rather than having them in our menus. 

Even then though, I think the most I'd be willing to allow to be customized is the max value and the labels/color. An array of possible indices is way too risky to manage (especially with our reliance of increment/decrement and clamp).

I'd like to see a usecase in the wild before deciding how we handle it. The best I can think of as a good use case isn't homebrewed pf1e mythic mythic (that might be better served with an extra untyped bonus, or even custom bonus types) but a custom setting using the PF2e rules. If someone mentions a setting that plays around with proficiency settings, we can see about exposing some knobs so that such a module can tweak them during startup.

CarlosFdez:
Thank you though for caring enough about the codebase and this game to go through so much effort. I wish you luck either with future PRs or with adjusting it to more closely align with project goals.

PR #13714, opened 2024-02-13
Add support for weapon damage die size alterations

stwlam:


src/module/rules/rule-element/item-alteration/schemas.ts
```diff
@@ -168,6 +167,11 @@ const ITEM_ALTERATION_VALIDATORS = {
             initial: undefined,
         } as const),
     }),
+    "damage-die-size": new ItemAlterationValidator({
+        itemType: new fields.StringField({ required: true, choices: ["weapon"] }),
+        mode: new fields.StringField({ required: true, choices: ["downgrade", "upgrade"] }),
+        value: new StrictNumberField({ required: true, choices: () => [1, 2, 3, 4] }),
```
CarlosFdez, line 173: I don't see value referred to in the implementation, what's this for?

PR #13724, opened 2024-02-13
Fix non-mental NPC skills being affected by stupefied condition

### LINKED ISSUE #13721: Stupefied Condition Bug - Applying to Incorrect Skills
v5.13.6
Applying the stupefied effect on a monster applies a penalty to all skills instead of the skills specified by the condition. Tested without modules.

stwlam:
Closes #13721

### DIFF
--- a/src/module/actor/npc/document.ts
+++ b/src/module/actor/npc/document.ts
@@ -1,12 +1,11 @@
 import { CreaturePF2e } from "@actor";
-import type { Abilities, CreatureSkills } from "@actor/creature/data.ts";
+import type { Abilities } from "@actor/creature/data.ts";
 import { setHitPointsRollOptions, strikeFromMeleeItem } from "@actor/helpers.ts";
 import { ActorInitiative } from "@actor/initiative.ts";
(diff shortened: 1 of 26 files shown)


CarlosFdez:
The idea of doing a side effect in R.mapToObj() actually concerns me a bit. Is it possible to move the skills assignment out of that somehow? Perhaps it can assign statistics, and then afterwards create the trace objects from the statistics in a separate loop/map?

src/module/actor/npc/document.ts
```diff
@@ -286,92 +297,44 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
         this.saves = saves as Record<SaveType, Statistic>;
     }
 
-    private prepareSkills(): CreatureSkills {
+    private prepareSkills(): Record<string, Statistic<this>> {
         const modifierAdjustments = this.synthetics.modifierAdjustments;
 
         // Internal function to create trace data, since NPCs still use the lore item type
```
CarlosFdez, line 303:
```suggestion
```

PR #13744, opened 2024-02-15
Reorganize Kinetic Gate/Gate's Threshold class features

stwlam:


Wizard-Level-1:
There doesn't seem to be any functionality for the impulse junction of the Metal Kin.
image
Putting something in actions with a damage button to pop it up in chat might be nice.

PR #13771, opened 2024-02-16
Fix not-targetOwner roll links

Drental:
check the value of targetOwner is "true" - it's a string when read from the dataset.

### DIFF
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -92,18 +92,10 @@ export const InlineRollLinks = {
         }
 
         for (const link of links.filter((l) => l.dataset.pf2Check && !l.dataset.invalid)) {
-            const {
-                pf2Check,
-                pf2Dc,
-                pf2Traits,
-                pf2Label,
-                pf2Defense,
-                pf2Adjustment,
-                pf2Roller,
-                targetOwner,
-                pf2RollOptions,
-                overrideTraits,
-            } = link.dataset;
+            const { pf2Check, pf2Dc, pf2Traits, pf2Label, pf2Defense, pf2Adjustment, pf2Roller, pf2RollOptions } =
+                link.dataset;
+            const overrideTraits = "overrideTraits" in link.dataset;
+            const targetOwner = "targetOwner" in link.dataset;
 
             if (!pf2Check) return;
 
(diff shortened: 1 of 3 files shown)


src/scripts/ui/inline-roll-links.ts
```diff
@@ -184,7 +184,7 @@ export const InlineRollLinks = {
                             }
 
                             const targetActor = pf2Defense
-                                ? targetOwner
+                                ? targetOwner === "true"
```
stwlam, line 187: This will never be the case

Drental: how sure are you about this?
Because I tested it while writing the code.
image
image
image

stwlam: whoops

the bug is instead in `createHTMLElement`. Pop this in:

```diff
diff --git a/src/util/dom.ts b/src/util/dom.ts
index c2287df520..6ec7268361 100644
--- a/src/util/dom.ts
+++ b/src/util/dom.ts
@@ -35,8 +35,8 @@ function createHTMLElement<K extends keyof HTMLElementTagNameMap>(
     const element = document.createElement(nodeName);
     if (classes.length > 0) element.classList.add(...classes);
 
-    for (const [key, value] of Object.entries(dataset).filter(([, v]) => !R.isNil(v))) {
-        element.dataset[key] = String(value);
+    for (const [key, value] of Object.entries(dataset).filter(([, v]) => !R.isNil(v) && v !== false)) {
+        element.dataset[key] = value === true ? "" : String(value);
     }
 
     if (innerHTML) {
```

Drental: better for sure - I get `""` evaluated as falsy, but I can check against !== undefined, probably

stwlam: That's what the `"string" in dataset` does

Drental: I'm not sure if I need to tweak something and where 😓
"string" in dataset only rings a bell for `"targetOwner" in rawParams`. This evaluates as true or false
Then the dom.ts makes it that the dataset either doesn't have data-target-owner or to have data-target-owner=""
when it's read later in the code from destructuring the dataset I get either `undefined` or `""` and both evaluate falsy in the `targetOwner ? a : b` syntax

I feel like I have four options here
1. set the value for true as String(value) in createHTMLElement, I get `"true"` and can do a simple `targetOwner ? a : b`
2. somehow make the dataset understand that a value of `""` should be represented as `true`, this seems like a bad idea since I'd have to change the type for link.dataset or add a new helper function with a more relaxed type
3. instead of using a destructuring assignment from the dataset, write a line `const targetOwner = "targetOwner" in link.dataset`
4. check if targetOwner is undefined later in the code.

I'll push up 3, writing this out it's probably what you meant

PR #13774, opened 2024-02-16
Show Common-backing language and prevent it from double-counting and displaying separately

### LINKED ISSUE #13675: Languages: knowing both Common and the actual language count as two languages to the character's max
Assume the Common language is set to Taldane for argument's sake

If a character picks Common and Taldane, it counts as if it knew two languages.
While not technically a bug, it would be nice if it didn't count so.

I'd also like to take the opportunity to suggest that:

- The Common language's name gets appended to Common, similar to how it is displayed on the language selector.
- The actual language gets hidden from the list if Common is known to avoid confusion when counting the number of known languages.

DocSchlock:
Closes #13675

GMs can still alter the common-linked language on PC sheets in case it becomes needed.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -315,6 +315,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
 
         sheetData.abpEnabled = AutomaticBonusProgression.isEnabled(actor);
 
+        // get the backing language for common
         const commonLanguage = game.settings.get("pf2e", "homebrew.languageRarities").commonLanguage;
 
         sheetData.languages = ((): LanguageSheetData[] => {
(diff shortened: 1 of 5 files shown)


stwlam:
I'm not sure that this is the most intuitive solution, and the actual issue can be addressed just by not counting both Common:tm: and the language that equates to Common:tm: when tallying number of languages known.

DocSchlock:
@stwlam 

Changed it to show the backing language as a tooltip instead and reworked the logic to not count the Common-backing language against the total language slots in the sheet.

src/module/actor/character/sheet.ts
```diff
@@ -315,25 +315,48 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
+        const commonLanguage = game.settings.get("pf2e", "homebrew.languageRarities").commonLanguage;
+
         sheetData.languages = ((): LanguageSheetData[] => {
             const languagesBuild = actor.system.build.languages;
             const sourceLanguages = actor._source.system.details.languages.value.filter(
                 (l) => l in CONFIG.PF2E.languages,
             );
-            const isOverMax = languagesBuild.value > languagesBuild.max;
+
+            const deductBackingLanguage = commonLanguage
```
stwlam, line 327: This is (or should be) guaranteed to be present

CarlosFdez: Even if its not, the following logic will still work out fine, the actor wouldn't have a common language of null.

DocSchlock: Bad variable name there. I refactored it - it's supposed to be the backing language.

src/module/actor/character/sheet.ts
```diff
@@ -315,23 +315,50 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
         sheetData.languages = ((): LanguageSheetData[] => {
             const languagesBuild = actor.system.build.languages;
             const sourceLanguages = actor._source.system.details.languages.value.filter(
                 (l) => l in CONFIG.PF2E.languages,
             );
-            const isOverMax = languagesBuild.value > languagesBuild.max;
+
+            // find if the backing language is present in character's languages
+            // this means characters explictly selected it
+            const deductBackingLanguage = commonBackingLanguage
```
stwlam, line 329: Same

src/module/actor/character/sheet.ts
```diff
@@ -315,23 +315,50 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
 
         sheetData.abpEnabled = AutomaticBonusProgression.isEnabled(actor);
 
+        // get the backing language for common
+        const commonBackingLanguage = game.settings.get("pf2e", "homebrew.languageRarities").commonLanguage;
```
stwlam, line 319: Retrieve from cache

src/module/actor/character/sheet.ts
DocSchlock: Cleaned up

src/module/actor/character/sheet.ts
DocSchlock: Added to cache and called from there.

stwlam:
Organizing this a little differently so that actor data isn't being calculated in a sheet class

PR #13781, opened 2024-02-17
Fix broken non-land Base Speed tooltip labels in creature documents

j-bs:
This label was missed when the actor base speed labels were refactored.

Noticed while doing some Familiar Sheet work 👀


stwlam:
It doesn't look like it even appears from PC sheets

j-bs:
Yeah it didnt seem to, but it does on Familiar sheets when they have the corresponding Abilities

Issue #13723, opened 2024-02-13
Assurance doesn't toggle off in check roll menu


Cuingamehtar:
Way to reproduce:
1) Roll a skill check with an actor that has Assurance for that skill,
2) Turn assurance on,
3) Turn assurance off.

Expected result:
d20 is rolled as normal.

Actual result:
Assurance is still applied.

Also, attribute modifier is not toggled on automatically when assurance toggled off, but this is likely related.

PR #13821, opened 2024-02-19
Fix rendering and disabling check substitution

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/13723

Also fixes Off looking janky in roll dialog substitutions
image


CarlosFdez:
Alternatively (or in addition to) I can replace the substitutions reassignment with an implementation that only replaces the contents of context.substitutions.

PR #13824, opened 2024-02-19
Add support for dynamic suboptions

stwlam:
image

image

image

### DIFF
--- a/src/module/rules/rule-element/roll-option/rule-element.ts
+++ b/src/module/rules/rule-element/roll-option/rule-element.ts
@@ -1,10 +1,18 @@
-import { DataUnionField, PredicateField, StrictBooleanField, StrictStringField } from "@system/schema-data-fields.ts";
+import {
+    DataUnionField,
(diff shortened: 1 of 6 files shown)


src/module/rules/rule-element/roll-option/rule-element.ts
```diff
@@ -190,11 +202,16 @@ class RollOptionRuleElement extends RuleElementPF2e<RollOptionSchema> {
         }
 
         if (this.toggleable) {
-            const suboptions = this.suboptions.filter((s) => s.predicate.test(optionSet));
-            if (suboptions.length > 0 && (!this.selection || !suboptions.some((s) => s.value === this.selection))) {
+            const suboptions = this.#resolveSuboptions();
+            if (this.ignored) return;
```
CarlosFdez, line 206: We might need a comment, either here or on #resolveSuboptions that resolveSuboptions() checks for validation, so that we don't get a bug later on when this order breaks.

PR #13834, opened 2024-02-19
Use actor/owning actor if rolling inline check directly from actor/item sheet

stwlam:


CarlosFdez:
I believe there may be creatures or items that present non-save checks (such as perception?) that may not necessarily be performed by the owning actor.

stwlam:
Yes, but it clicked from the actor sheet, it is a uselessly obscure "feature" and otherwise defies expectations to still use the controlled token's actor.

CarlosFdez:
Oh I see, I misidentified it as any owning actor. That makes sense.

PR #13838, opened 2024-02-19
Add sound emission to Vehicles

### LINKED ISSUE #11596: Fix vehicles so they give of sound like pcs and npcs dp
You should be able to the swirly sound indicator, could not find a way to add sound to vehicles.

hazard
something like this

DocSchlock:
Closes #11596

Adds selector to the sidebar of the Vehicle sheet.

### DIFF
--- a/packs/vehicles/adaptable-paddleboat.json
+++ b/packs/vehicles/adaptable-paddleboat.json
@@ -77,6 +77,7 @@
             "collisionDamage": {
                 "value": "2d6"
             },
+            "emitsSound": false,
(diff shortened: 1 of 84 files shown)


packs/vehicles/adaptable-paddleboat.json
```diff
@@ -77,6 +77,7 @@
             "collisionDamage": {
                 "value": "2d6"
             },
+            "emitsSound": false,
```
stwlam, line 80: This should have the same default as hazards (set in template.json).

PR #13844, opened 2024-02-19
Add Hero Point RollOption on rerolls

### LINKED ISSUE #4516: [Request] Add option to look for Hero Point usage in Rule Elements
I have a house-rule I use, where if someone uses a Hero Point to reroll a check, and the flat d20 roll is 10 or less, they add +10 to the result. I'm looking at automating this, and a chat with Tikael suggested the best way is to have a callout in the code so that Rule Elements can look for it.

He said that this would require someone to dig in `rolls.ts` to figure out where best to put an `options.push("hero-point")`, this would add "hero-point" to the list of options for Rule Elements.

DocSchlock:
Closes #4516

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -438,7 +438,7 @@ class CheckPF2e {
 
         context.skipDialog = true;
         context.isReroll = true;
-        if(heroPoint) context.options.push("hero-point");
+        if (heroPoint) context.options.push("hero-point");
 
         const oldRoll = message.rolls.at(0);
         if (!(oldRoll instanceof CheckRoll)) throw ErrorPF2e("Unexpected error retrieving prior roll");
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -438,6 +438,7 @@ class CheckPF2e {
 
         context.skipDialog = true;
         context.isReroll = true;
+        context.options.push("check:reroll")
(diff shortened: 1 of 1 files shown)


src/module/system/check/check.ts
```diff
@@ -438,6 +438,7 @@ class CheckPF2e {
 
         context.skipDialog = true;
         context.isReroll = true;
+        if (heroPoint) context.options.push("hero-point");
```
stwlam, line 441:
```suggestion
        if (heroPoint) context.options.push("check:hero-point");
```

src/module/system/check/check.ts
```diff
@@ -438,6 +438,7 @@ class CheckPF2e {
 
         context.skipDialog = true;
         context.isReroll = true;
+        if (heroPoint) context.options.push("check:hero-point");
```
stwlam, line 441:
```suggestion
        context.options.push("check:reroll")
        if (heroPoint) context.options.push("check:hero-point");
```

src/module/system/check/check.ts
```diff
@@ -438,6 +438,8 @@ class CheckPF2e {
 
         context.skipDialog = true;
         context.isReroll = true;
+        context.options.push("check:reroll")
```
stwlam, line 441:
```suggestion
        context.options.push("check:reroll");
```

PR #13845, opened 2024-02-19
Return early in RE constructors when `DataModel` validation fails

stwlam:


src/module/rules/rule-element/iwr/base.ts
```diff
@@ -85,8 +85,8 @@ abstract class IWRRuleElement<TSchema extends IWRRuleSchema> extends RuleElement
     /** A reference to the pertinent property in actor system data */
     abstract get property(): IWRSource[];
 
-    #isValid(value: unknown): boolean {
```
CarlosFdez, line 88: Out of following classical conventions, can this be "#isValueValid()" instead?

PR #13850, opened 2024-02-20
Use underline modifiers convention for character proficiencies tab

CarlosFdez:
Seems it went over well, so I'm confident enough to do it in the proficiencies tab as well

image


stwlam:
Yaaay.

PR #13853, opened 2024-02-20
Improve flexibility of spell area localization, prefer "shape" over "type"

stwlam:


CarlosFdez:
I like calling it area shape (seems even 5e calls it that now, wow) but remembering 5280 for 1 mile seems in the realm of odd arcane rituals, and I'm an american. Localization is more flexible, but in that respect I don't know if I can call it improved, we super duper need a solution for that.

stwlam:
image

CarlosFdez:
If there's only one, then we can can delay for a better answer. My thought was that it'd be a sheet side thing anyways, rather than in the data.

CarlosFdez:
To clarify, I mean delay solving it, not delay the change.

PR #13859, opened 2024-02-20
Update several PC strike features and effects to use newer rule elements

stwlam:


packs/feats/arcane-fists.json
```diff
@@ -43,12 +43,10 @@
...
-                "definition": [
-                    "item:trait:unarmed"
-                ],
-                "key": "AdjustStrike",
+                "itemId": "xxxxxxFISTxxxxxx",
+                "key": "ItemAlteration",
                 "mode": "add",
-                "property": "weapon-traits",
+                "property": "traits",
                 "value": "arcane"
```
CarlosFdez, line 50: from what I can tell, this particular trait is for all unarmed attacks, not just fist

PR #13860, opened 2024-02-20
Adjust some PC-sheet styling to accommodate non-English text

stwlam:
Also correct name/description localization of bundled conditions

### DIFF
--- a/src/styles/actor/character/_proficiency-ranks.scss
+++ b/src/styles/actor/character/_proficiency-ranks.scss
@@ -1,16 +1,20 @@
 .pf-rank {
     @include micro;
-    align-items: center;
     background: var(--color-proficiency-untrained);
     border: 1px solid rgba(black, 0.5);
     border-radius: 2px;
     box-shadow: inset 0 0 0 1px rgba(white, 0.15);
     color: var(--text-light);
-    display: flex;
+    display: block;
     font-weight: 500;
-    height: var(--font-size-16);
-    justify-content: center;
-    width: 6.5em;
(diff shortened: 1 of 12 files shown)


src/styles/actor/character/_sidebar.scss
```diff
@@ -54,7 +54,7 @@ aside {
     .sidebar {
         margin-bottom: 2.625rem;
         overflow: hidden auto;
-        padding: 0 var(--space-2) var(--space-2) var(--space-8);
```
CarlosFdez, line 57: You tested how it looks like if it has a scrollbar?

src/styles/actor/character/_proficiency-ranks.scss
```diff
@@ -44,9 +48,6 @@
 
 select.pf-rank {
     cursor: pointer;
-    line-height: 0.9em;
-    padding: 0;
-    text-align: center;
```
CarlosFdez, line 49: also test on firefox if you can. A lot of the reason the pf-rank is the way it is, is because select rendering is pretty screwy on different browsers (they all handle it differently). There's also tiny minor font alignment and rendering differences.

src/styles/actor/character/_sidebar.scss
stwlam: Stable scrollbar gutter, my dude!

CarlosFdez: More wondering if its hugging the scrollbar if it shows up

PR #13862, opened 2024-02-20
Handle dropping persistent damage links on character and NPC Sheets

### LINKED ISSUE #7223: Persistent damage links cannot be dropped on character sheets
They can be dropped on tokens.

Screen Shot 2023-03-31 at 1 30 38 PM

DocSchlock:
Closes #7223

Intercepts the drop at the creature level so both characters and NPCs benefit from the fix.

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -363,8 +363,6 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
             }
         }
 
-        console.log(data);
-
         return super._onDropItem(event, data);
     }
 
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -15,6 +15,7 @@ import { ActorSheetPF2e, SheetClickActionHandlers } from "../sheet/base.ts";
 import { CreatureConfig } from "./config.ts";
 import { Language } from "./index.ts";
 import { SpellPreparationSheet } from "./spell-preparation-sheet.ts";
+import { DamageRoll } from "@system/damage/roll.ts";
 
 /**
  * Base class for NPC and character sheets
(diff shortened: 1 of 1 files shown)


CarlosFdez:
We already have handling for this somewhere, but I don't see this deleting any code. What happened to it?

DocSchlock:
The only handling I saw was a hook for dropping PersistentDamage on tokens, where I cribbed most of the code from.

src/module/actor/creature/sheet.ts
```diff
@@ -296,6 +297,43 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
...
+    protected override async _onDrop(event: DragEvent): Promise<boolean | void> {
+        const dataString = event.dataTransfer?.getData("text/plain");
+        const dropData = ((): Record<string, unknown> | null => {
+            try {
+                return JSON.parse(dataString ?? "");
+            } catch {
+                return null;
+            }
+        })();
+        if (!dropData) return;
```
CarlosFdez, line 309: Isn't this going to break regular onDrop handling? Might be best to omit this if.

DocSchlock: This is copied code from Character Sheet's _onDrop - so if it breaks here, wouldn't it break there as well? Are there any valid data payloads being dropped on a sheet that wouldn't be able to be parsed as JSON?

DocSchlock: I'll give it a shot. I agree that's a better, cleaner way to handle the non-PF2E drops.

CarlosFdez:
Someone can correct me but this seems more like a enhancement than a bugfix. It seems like this never worked before. I thought I made this work at one point but I guess not.

stwlam:
Looking at the issue, they're expecting damage-roll links to be droppable. I have some doubts about this since a damage roll can have persistent damage along with anything else. These aren't content links but roll links.

CarlosFdez:
They are already droppable on tokens (not sheets), but if we don't already, mixed damage stuff should be ignored. Its more of a feature that is expected to exist, rather than a break of existing functionality.

This has kinda become more complicated than expected. I can take this off your hands.

stwlam:
I don't see the benefit of this at all.

PR #13863, opened 2024-02-20
Add support for Effect Badge Labels that loop to start

### LINKED ISSUE #7600: Feature Request: Looping Effect Counters
I was helping someone come up with an effect for [Touch of the Moon](https://2e.aonprd.com/Spells.aspx?ID=460) the other day, and while it was mostly doable by predicating on the level of a labeled counter, there was no way to loop back to 1 after getting to 4. Adding a 'can loop' toggle to counters would enable this to be more fully automated (except for having to manually advance every turn). It might make sense to limit it to labeled counters, as otherwise you would need a way to specify a cap on unlabeled ones, though that might also be useful.

DocSchlock:
Closes #7600

### DIFF
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -44,7 +44,6 @@ interface EffectSystemData
     expired: boolean;
     badge: EffectBadge | null;
     remaining: string;
-    loop: boolean;
 }
 
 export type { EffectFlags, EffectSource, EffectSystemData };
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -44,7 +44,7 @@ interface EffectSystemData
     expired: boolean;
     badge: EffectBadge | null;
     remaining: string;
-    looping: boolean | null;
+    loop: boolean;
 }
 
 export type { EffectFlags, EffectSource, EffectSystemData };
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -44,6 +44,7 @@ interface EffectSystemData
     expired: boolean;
     badge: EffectBadge | null;
     remaining: string;
+    looping: boolean | null;
 }
 
 export type { EffectFlags, EffectSource, EffectSystemData };
--- a/src/module/item/effect/document.ts
+++ b/src/module/item/effect/document.ts
@@ -74,10 +74,10 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
     /** Increases if this is a counter effect, otherwise ignored outright */
     async increase(): Promise<void> {
         const badge = this.system.badge;
-        const looping = this.system.looping;
+        const loop = this.system.loop;
         if (badge?.type === "counter" && !this.isExpired) {
             const value =
-                looping && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
+                loop && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
             await this.update({ system: { badge: { value } } });
         }
     }
--- a/src/module/item/effect/document.ts
+++ b/src/module/item/effect/document.ts
@@ -188,8 +188,6 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
 
                 const currentValue = Number(badgeChange.value ?? currentBadge.value ?? 1);
                 badgeChange.value = Math.clamped(currentValue, minValue, maxValue);
-
-                if (badgeChange["-=labels"] === null) badgeChange["loop"] = false;
             }
 
             // Delete min/max under certain conditions. Foundry is a bit shakey with -= behavior in _preUpdates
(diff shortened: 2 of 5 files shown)


src/module/item/effect/data.ts
```diff
@@ -44,6 +44,7 @@ interface EffectSystemData
     expired: boolean;
     badge: EffectBadge | null;
     remaining: string;
+    looping: boolean | null;
```
stwlam, line 47: What sets this to null?

stwlam: And let's save few bytes as call it `loop`

stwlam: And, and, and: this is badge data, so it shouldn't be at the top level.

DocSchlock: Pushed it down into BadgeCounter as an optional property so it didn't break all the inheritors. At worst it would be `undefined` so I put a boolean conversion in the increase() function, but I can move that to data preparation if that would be cleaner or more efficient.

src/module/item/effect/document.ts
```diff
@@ -74,8 +74,11 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
     /** Increases if this is a counter effect, otherwise ignored outright */
     async increase(): Promise<void> {
         const badge = this.system.badge;
+
         if (badge?.type === "counter" && !this.isExpired) {
-            const value = badge.value + 1;
+            const isLoop = !!badge.loop;
+            const value =
+                isLoop && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
```
stwlam, line 81:
```suggestion
            badge.loop ??= false;
            const value =
                badge.loop && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
```

src/module/item/effect/data.ts
stwlam: Please do set it during data preparation.

src/module/item/effect/document.ts
```diff
@@ -74,8 +74,11 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
     /** Increases if this is a counter effect, otherwise ignored outright */
     async increase(): Promise<void> {
         const badge = this.system.badge;
+
         if (badge?.type === "counter" && !this.isExpired) {
-            const value = badge.value + 1;
+            badge.loop ??= false;
+            const value =
+                badge.loop && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
```
stwlam, line 81:
```suggestion
            const value =
                badge.loop && badge.labels ? (badge.value + 1 > badge.max ? badge.min : badge.value + 1) : badge.value + 1;
```

src/module/item/effect/data.ts
DocSchlock: Done in latest commit.

CarlosFdez:
Screenshot?

DocSchlock:
image

The setting is removed if you drop the labels.

src/module/item/effect/document.ts
```diff
@@ -74,8 +75,14 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
...
         const badge = this.system.badge;
+
         if (badge?.type === "counter" && !this.isExpired) {
-            const value = badge.value + 1;
+            const value =
+                badge.loop && badge.labels
+                    ? badge.value + 1 > badge.max
+                        ? badge.min
+                        : badge.value + 1
+                    : badge.value + 1;
```
CarlosFdez, line 85:
```suggestion
            const shouldLoop = badge.loop && badge.labels && badge.value >= badge.max;
            const value = shouldLoop ? badge.min : badge.value + 1;
```

src/module/item/effect/document.ts
```diff
@@ -74,8 +75,10 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
     /** Increases if this is a counter effect, otherwise ignored outright */
     async increase(): Promise<void> {
         const badge = this.system.badge;
+
         if (badge?.type === "counter" && !this.isExpired) {
-            const value = badge.value + 1;
+            const shouldLoop = badge.loop && badge.labels && badge.value >= badge.max;
```
CarlosFdez, line 80: You caught me before I could edit it, but also remove the && badge.labels. badge.loop should be false if there are no labels anyways.

DocSchlock: Did quick test and yeah, that appears to be the case. Commit fixed.

CarlosFdez:
Ensure badge.loop gets set to false if the labels are removed from a counter badge and then I think this is fine.

src/module/item/effect/document.ts
```diff
@@ -186,6 +188,8 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
 
                 const currentValue = Number(badgeChange.value ?? currentBadge.value ?? 1);
                 badgeChange.value = Math.clamped(currentValue, minValue, maxValue);
+
+                if (badgeChange["-=labels"] === null) badgeChange["loop"] = false;
```
CarlosFdez, line 192: Move this to outside, like how min/max are. You also need to handle the case where a formerly counter badge with labels that looped is converted to another type. Also delete it instead of setting to false since its optional in the source data.

CarlosFdez (approved):

PR #13871, opened 2024-02-21
Add notes and automation for champion reaction upgrade feats.

pedrogrullada:
also a sneaky additional predicate to Everstand Stance's damage upgrade.

### DIFF
--- a/packs/feat-effects/stance-everstand-stance.json
+++ b/packs/feat-effects/stance-everstand-stance.json
@@ -32,17 +32,19 @@
                 "value": 2
             },
             {
-                "key": "DamageDice",
-                "override": {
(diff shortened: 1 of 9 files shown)


packs/feat-effects/stance-everstand-stance.json
```diff
@@ -36,6 +36,9 @@
                 "override": {
                     "upgrade": true
                 },
+                "predicate": [
+                    "item:hands-held:2"
+                ],
```
stwlam, line 41: This would benefit from getting changed into an item alteration

pedrogrullada: Might as well, on it.

PR #13873, opened 2024-02-21
Add parent and granter roll options to choiceset filters

CarlosFdez:
....and fixes a rarity option regression

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -41,6 +41,9 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
     /** Additional item roll options set by rule elements */
     declare rollOptions: Set<string>;
 
+    /** The item that granted this item, if any */
+    declare grantedBy: ItemPF2e<ActorPF2e> | null;
+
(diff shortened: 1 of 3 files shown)


src/module/item/base/document.ts
```diff
@@ -285,6 +286,8 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
                 grant.onDelete ??= "detach";
             }
         }
+
```
stwlam, line 289: I don't know that this can safely be set this early.

CarlosFdez: I did a quick test, and it seems that actor.items is filled out pretty early. While the items themselves are incomplete, its the same reference allegedly so it works out.

The only other alternative construction context stuff and storing a private granted by param, but having an additional fake-private property seemed odd.

CarlosFdez: That said, on second glance, it could be that the only reason its safe is because of your prepareData() override that avoids preparing child objects until the parent itself is also initialized. If you'd rather not trust that continues to exist in the future, I could find another way.

stwlam: Prepare sibling data?

PR #13887, opened 2024-02-21
Add familiar ability RollOptions to Familiars

### LINKED ISSUE #12704: Make familiar abilities populate roll options
I'd like to request having familiar abilities populate rolloptions so that something like `self:ability:slug` could be used to predicate on whether a familiar has another ability or not. For example, `predicate: [{"not":"self:ability:tough"}]` or `predicate: ["self:ability:climber"]`.

Thank you!

DocSchlock:
Closes #12704

### DIFF
--- a/src/module/item/ability/document.ts
+++ b/src/module/item/ability/document.ts
@@ -55,9 +55,8 @@ class AbilityItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> exten
         const actor = this.actor;
 
         if (actor?.isOfType("familiar") && this.system.category === "familiar") {
-            const prefix = "familiar:ability";
             const slug = this.slug ?? sluggify(this.name);
-            actor.rollOptions.all[`${prefix}:${slug}`] = true;
+            actor.rollOptions.all[`ability:${slug}`] = true;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/module/item/ability/document.ts
```diff
@@ -50,6 +51,16 @@ class AbilityItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> exten
         }
     }
 
+    override prepareActorData(): void {
+        const actor = this.actor;
+
+        if (actor?.isOfType("familiar") && this.system.category === "familiar") {
+            const prefix = "familiar:ability";
+            const slug = this.slug ?? sluggify(this.name);
+            actor.rollOptions.all[`${prefix}:${slug}`] = true;
```
CarlosFdez, line 60:
```suggestion
            const slug = this.slug ?? sluggify(this.name);
            actor.rollOptions.all[`ability:${slug}`] = true;
```

I personally don't know if we should include self: or not (perhaps @stwlam can answer that) but we definitely shouldn't include familiar: in actor.rollOptions.

CarlosFdez (dismissed):

src/module/item/ability/document.ts
stwlam: It probably should include `self:` since the master may need to query it.

PR #13911, opened 2024-02-23
Brush up Cleric feats

pedrogrullada:
- Add note to Panic the Dead.
- Grant Raise a Shield effect along with Raise Symbol when shield is a religious symbol.
- Remove manipulate trait from 1-action heal when using Restorative Strike and add Note rule element.
- Automate Channeling Block.
- Append Deity's Protection effect to cleric domain spells.
- Make Premonition of Clarity a self-applied effect.
- Add note rule element to Shared Clarity
- Adjust degree of success of Dance of Intercessions' performance check based on remaining uses.

And a couple of very minor things: Fix typo in Sap Life's description, Add description to Prayer-Touched Weapon's effect

### DIFF
--- a/packs/feats/raise-symbol.json
+++ b/packs/feats/raise-symbol.json
@@ -24,7 +24,21 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.Cleric.RaiseSymbol.RollOptionLabel",
+                "option": "shield-religious-symbol",
+                "predicate": [
+                    "self:shield:equipped",
+                    {
+                        "not": "self:effect:emblazon-armament-shield"
+                    }
+                ],
+                "toggleable": true,
+                "value": true
+            }
+        ],
         "selfEffect": {
             "name": "Effect: Raise Symbol",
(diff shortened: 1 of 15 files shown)


packs/feats/raise-symbol.json
```diff
@@ -24,7 +24,21 @@
...
+                "option": "shield-religious-symbol",
+                "predicate": [
+                    "self:shield:equipped",
+                    {
+                        "not": "self:effect:emblazon-armament-shield"
+                    }
+                ],
+                "toggleable": true,
+                "value": true
+            }
```
TikaelSol, line 40: You can just predicate on having a shield raised. Presumably if you have emblazon armament shield and a shield raised it's meant to be used as a symbol.

TikaelSol: Also we can lose the stored value in this

pedrogrullada: > You can just predicate on having a shield raised. Presumably if you have emblazon armament shield and a shield raised it's meant to be used as a symbol.

This is precisely meant to account for someone that is wielding a religious symbol shield that isn't their emblazon armament (they may not even have the feat). The feat itself is open to that as written, although may not be worth accounting for?

> Also we can lose the stored value in this

Oops.

PR #13914, opened 2024-02-23
Add effect for Ancestral Geometry tattoo

pedrogrullada:


packs/equipment-effects/effect-ancestral-geometry-save-type.json
```diff
@@ -0,0 +1,93 @@
...
+    "system": {
+        "badge": {
+            "evaluate": true,
+            "labels": [
+                "Fortitude",
+                "Fortitude",
+                "Reflex",
+                "Reflex",
+                "Will",
+                "Will"
```
TikaelSol, line 14:
```suggestion
                "Fortitude",
                "Reflex",
                "Will"
```

Just make the actual effect use a d3. Paizo uses d6 because they don't want to introduce weird dice to tabletop games but we don't have to bother following their madness here.

packs/equipment-effects/effect-ancestral-geometry-save-type.json
```diff
@@ -0,0 +1,93 @@
...
+            ],
+            "reevaluate": null,
+            "type": "formula",
+            "value": "1d6"
```
TikaelSol, line 18:
```suggestion
            "value": "1d3"
```

packs/equipment-effects/effect-ancestral-geometry-save-type.json
```diff
@@ -0,0 +1,93 @@
...
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.ancestralGeometry",
+                "predicate": [
+                    {
+                        "or": [
+                            "parent:badge:value:1",
+                            "parent:badge:value:2"
+                        ]
+                    }
```
TikaelSol, line 48:
```suggestion
                     "parent:badge:value:1"
```

TikaelSol: etc

packs/equipment-effects/effect-ancestral-geometry.json
```diff
@@ -0,0 +1,55 @@
...
+            "title": "Pathfinder Treasure Vault"
+        },
+        "rules": [
+            {
+                "key": "GrantItem",
+                "uuid": "Compendium.pf2e.equipment-effects.Item.Effect: Ancestral Geometry (Save Type)"
+            },
+            {
+                "key": "SubstituteRoll",
+                "label": "PF2E.SpecificRule.Equipment.AncestralGeometry.Label",
```
TikaelSol, line 36: I don't think we need the label here, having effect in there is fine

packs/equipment-effects/effect-ancestral-geometry.json
```diff
@@ -0,0 +1,55 @@
...
+    "name": "Effect: Ancestral Geometry",
+    "system": {
+        "badge": {
+            "evaluate": true,
+            "reevaluate": null,
+            "type": "formula",
+            "value": "2d20kh1"
+        },
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Ancestral Geometry]</p>\n<p>Roll 2d20 and record the highest result. Then roll 1d6 and note a type of saving throw: 1–2 Fortitude, 3–4 Reflex, and 5–6 Will.</p>"
```
TikaelSol, line 13: There's a really tortured way to do this in one effect based on the fact that `10 * 1d20 + 1d3` technically encodes all the information in a separable way. If I actually merged it I think stwlam might drive up to Ohio to hit me.

TikaelSol: Badge: `(10 * 2d20kh1) + 1d3`

For the roll sub: `floor(badge/10)`

For the save type: `ternary(eq(roll remainder,3), "will", ternary(eq(roll remainder, 2), "reflex", "fortitude"))`
where `roll remainder` is `badge - 10 * floor(badge/10)`

TikaelSol: Maybe you can't. Let's say that's the reason we aren't implementing that terrible idea :D

PR #13927, opened 2024-02-23
Restrict elemental blast to chosen elements (effect traits only) instead of any element trait.

CarlosFdez:
Also refactors to use a data model.

The properties are still referred to as element, but they're no longer restricted via specific element traits, leaving behind `"element": "void"` as a shame on you. It now restricts by the elements that the kineticist currently possesses (and also restricts those to be effect traits).

### DIFF
--- a/build/run-migration.ts
+++ b/build/run-migration.ts
@@ -28,6 +28,7 @@ import { Migration922SwashbucklerFinishers } from "@module/migration/migrations/
 import { Migration923KineticistRestructure } from "@module/migration/migrations/923-kineticist-restructure.ts";
 import { Migration924JiuHuanDoa } from "@module/migration/migrations/924-jiu-huan-dao.ts";
 import { Migration925TouchOfCorruption } from "@module/migration/migrations/925-touch-of-corruption.ts";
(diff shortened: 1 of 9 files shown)


stwlam:
The PR title and changes made appear to be entirely different things

CarlosFdez:
I had to do a bit more than the PR's intent in order to try to appease you, but other than the data model refactor it now also restricts the EB element to match those present in the kineticist's chosen elements. Its in the second sentence of the description but I guess I can think of a way to slip it into the title.

CarlosFdez:
Ah, and the migration is for the data model refactor to work. Our RecordField doesn't quite like `{ infusion: Blah } & { Exclude<string, "infusion>: otherBlah };` last I checked.

stwlam:
I'm not sure who you were trying to "appease." I didn't ask for any of this, nor does it strike me as necessary or useful.

CarlosFdez:
Discussed privately but to summarize, you had said that you didn't like opening it up to action traits because it was too loose. So I made it less loose, choosing instead to validate according to class feature items. There was discussion on doing this in the dev channels as well.

I can close this if you think validating elemental blasts based on gate items isn't a good idea. This was something I did on someone else's behalf.

PR #13933, opened 2024-02-24
Add effect-badge reevaluation event of "initiative-roll"

stwlam:
Also log the initial value for RE availability


src/module/item/effect/document.ts
```diff
@@ -106,12 +110,16 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
      * Evaluate a formula badge, sending its result to chat.
      * @returns The resulting value badge
      */
-    private async evaluateFormulaBadge(badge: EffectBadgeFormulaSource): Promise<EffectBadgeValueSource> {
+    async #evaluateFormulaBadge(
```
CarlosFdez, line 113: I don't think the proxy private problem has been solved yet right?

PR #13936, opened 2024-02-24
Create Compendium for Monster Core and include blog content (Mummy)

SpartanCPA:
Pinging stwlam too because I'm touching IWR here

### DIFF
--- a/packs/pathfinder-monster-core/mummy-guardian.json
+++ b/packs/pathfinder-monster-core/mummy-guardian.json
@@ -1,0 +1,377 @@
+{
+    "_id": "XZWUQklzWF6YFPmG",
+    "img": "systems/pf2e/icons/default-icons/npc.svg",
+    "items": [
+        {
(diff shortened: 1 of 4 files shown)


packs/pathfinder-monster-core/mummy-guardian.json
```diff
@@ -0,0 +1,381 @@
...
+            "privateNotes": "",
+            "publicNotes": "",
+            "publication": {
+                "license": "ORC",
+                "remaster": true,
+                "title": "Pathfinder Monster Core"
+            },
+            "source": {
+                "value": "Pathfinder Monster Core"
+            }
```
TikaelSol, line 339:
```suggestion
            }
```
That shouldn't be needed. Was that added in by my macro?

SpartanCPA: Yeah it was

PR #13945, opened 2024-02-25
Make dice faces change for two-hand trait directly from weapon

stwlam:
This will allow rule elements that hinge on dice faces count to properly function


CarlosFdez:
I'll assume that the weapon sheet uses source value instead of prepared, but if not that's easy to fix separately.

PR #13955, opened 2024-02-25
Ensure ABC items are processed first

CarlosFdez:
This prevents a break in Battlezoo dragons if you add the Empowered Size feat before the Ancestry itself. In theory Automatons could trigger the same issue with their own equivalent "get big feat" and the CreatureSize RE in the ancestry itself.


stwlam:
Background after heritage?

CarlosFdez:
Sure

PR #13966, opened 2024-02-26
Prevent counting Common and its corresponding language as two allocated languages

stwlam:
Also display as single language on sheet

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -317,26 +317,44 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
 
         sheetData.languages = ((): LanguageSheetData[] => {
             const languagesBuild = actor.system.build.languages;
-            const sourceLanguages = actor._source.system.details.languages.value.filter(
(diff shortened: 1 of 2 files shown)


src/module/actor/character/sheet.ts
```diff
@@ -321,22 +321,32 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
+                if (language === commonLanguage && languageSlugs.includes("common")) {
+                    return [];
+                }
+                const label =
+                    language === "common" && commonLanguage
+                        ? game.i18n.format("PF2E.Actor.Creature.Language.CommonLanguage", {
+                              language: game.i18n.localize(CONFIG.PF2E.languages[commonLanguage]),
+                          })
+                        : game.i18n.localize(CONFIG.PF2E.languages[language]);
+                const overLimit = isOverMax && sourceLanguages.indexOf(language) + 1 > languagesBuild.max;
```
CarlosFdez, line 336: Does this handle dynamic positioning of the common language? Like if Taldane was the last language in the list, would it fail to mark one of the overlimit languages?

PR #13981, opened 2024-02-27
Move roll context resolution to dedicated class

stwlam:


CarlosFdez:
I am super uncertain about doing this on the final week of dev, and while I think what you christened as DamageDamageContext needs to go away, DamageDamageContextFlag is not in that category (that should be ChatDamageContextFlag I think).

I am also uncertain about the message signature of the method, and whether statistic used should be part of the origin data. That seems to be the cause of the branching in the statistic itself.

However, most of these are changeable, and I really like the idea of moving the context to a separate class.

PR #14016, opened 2024-02-29
Effect: add predators claw effect

MrPrimate:
Adds an effect for Predators claw that grants critical specialization


pedrogrullada:
I believe we are holding off on consumable effects until we get some new tech from the devs

TikaelSol:
Yeah, consumable activations are coming soon.

PR #14050, opened 2024-03-03
Restore secondary spellcasters for rituals

Cuingamehtar:
After attempting to correct the number of secondary casters for Awaken Portal ritual #14043 (that PR was not enough for it to work, which I somehow forgot to check), I wanted to correct the values for other rituals.

What I figured out was that in one of the migrations this data was stripped completely, and the field expected a numerical value.
Since the system doesn't do anything with the number of secondary ritual casters at the moment (as far as I'm aware), and the values there are not as straightforward as simple numbers or ranges, I propose changing the field into string type.

The comment on the data type (about the *maximum* number of spellcasters) was also incorrect, as that was the number of required spellcasters.

### DIFF
--- a/build/run-migration.ts
+++ b/build/run-migration.ts
@@ -29,6 +29,7 @@ import { Migration923KineticistRestructure } from "@module/migration/migrations/
 import { Migration924JiuHuanDoa } from "@module/migration/migrations/924-jiu-huan-dao.ts";
 import { Migration925TouchOfCorruption } from "@module/migration/migrations/925-touch-of-corruption.ts";
 import { Migration926RemoveVisionFeatureLinks } from "@module/migration/migrations/926-remove-vision-feature-links.ts";
+import { Migration927RitualSecondaryCasters } from "@module/migration/migrations/927-ritual-secondary-casters.ts";
 // ^^^ don't let your IDE use the index in these imports. you need to specify the full path ^^^
 
 const { window } = new JSDOM();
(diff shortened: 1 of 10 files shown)


stwlam:
I don't think we're going to return it to text.

Cuingamehtar:
This is reasonable. 
How about {minimum:number, maximum:number | null, details:string}?
Where null represents an optional upper bound?

stwlam:
That sounds good. Use `min` and `max` as shorthands, which is in use elsewhere. If you'd like to continue using this PR, I'll reopen it.

Cuingamehtar:
Yeah, I want to continue with this PR.

Cuingamehtar:
Here is how the Details tab currently looks
image
image
image

I have not attempted the migration of the imported data yet.

Cuingamehtar:
I think, I'm ready for another look.

I'm pretty confident about this migration - I've checked all rituals post-migration and the numbers match those in books.

I'm, however, concerned about migration 882. I had to modify it, since the type itself is now different, and tried to adapt it, but I'm much less confident since I'm not sure how to test it properly. 
Ideally, I think, it should make the type that is was making before, but I can't figure out how to define it in the file.

Cuingamehtar:
I might have just been overcomplicating things with 882. Right now the results should be the way they were before, if I understand correctly.

stwlam:
To make reviewing easier, go ahead and remove the migration result from the PR until it's otherwise approved.

Cuingamehtar:
I've reverted the migration and rebased the changes on top of the 5.14.1 release version.

Cuingamehtar:
Rebased on 130a83cc
Added Diabolic (prev. Infernal) pact and Demonic (prev. Abyssal) pact 
Incremented migration counter

PR #14055, opened 2024-03-03
Add Aura Junction Metal effect

reyzor1991:


packs/feat-effects/effect-aura-junction-metal.json
```diff
@@ -0,0 +1,52 @@
+{
+    "_id": "MFCcEWCC9PR46qPy",
+    "img": "systems/pf2e/icons/spells/cascade-countermeasure.webp",
+    "name": "Effect: Aura Junction (Metal)",
+    "system": {
+        "description": {
+            "value": "<p>Enemies in the aura take a -1 status penalty to attacks with metal objects, and they take a -1 status penalty to AC if they're wearing metal armor, have the metal trait, or are made of metal.</p>"
```
TikaelSol, line 7:
```suggestion
            "value": "<p>Creatures with the metal trait take a -1 status penalty to AC.</p>"
```
Effects should state only what they actually do, and the other bits are kind of impossible to do properly.

PR #14110, opened 2024-03-08
Retrieve updated check/damage domains from context clones

### LINKED ISSUE #13931: Devise a Strategem and Quicksilver Mutagen Automation Issue
The Devise a Stratagem feature appears to mark your finesse and agile weapons as Int based strikes rather than Dex based strikes. As a result, the bonus to Dex based strikes from Quicksilver Mutagen will not apply, even if you're rolling with Dex, as it doesn't recognize the character's weapon strikes as Dex based. It doesn't even come up as an option to toggle on manually in the roll dialog box.

stwlam:
Closes #13931


CarlosFdez (approved):

CarlosFdez:
I'd like to know if the enabled/ignored change works for check dialogs, if it does, this looks good to me.

PR #14126, opened 2024-03-09
Add biconditional predication operator

stwlam:


CarlosFdez:
How does this differ from and?

PR #14129, opened 2024-03-09
Replace `@Damage` enricher with custom parsing

In3luki:


src/module/system/text-editor.ts
```diff
@@ -228,6 +229,77 @@ class TextEditorPF2e extends TextEditor {
...
+                if (c === "[") brackets.open += 1;
+                if (c === "]") {
+                    brackets.open -= 1;
+                    brackets.lastIndex = index;
+                    if (brackets.open === 0) break;
+                }
+            }
+            return brackets.open === 0 && brackets.lastIndex > 0 ? firstBracketIndex + brackets.lastIndex : null;
+        };
+        const data: { replaced: boolean; target: Text | null } = { replaced: false, target: null };
```
CarlosFdez, line 259: I don't think I'm a fan of this. The data.blah actually threw me off a bit until I realized you were trying to do let without using let. I think using let here is fine and even preferred.

PR #14147, opened 2024-03-10
Update Character Sidebar's Shield Max HP label's width

mclemente:
This increases the width of the Max HP label from `5em` to `6em`.
The reason for this change is due to (Brazilian Portuguese) localization not fitting into the box and requiring custom CSS styling to fix it.

image image


stwlam:
What does it look like in English?

mclemente:
It doesn't affect the label in the English localization at all.

PR #14152, opened 2024-03-10
Show impulse junction note on elemental blast attack roll instead of damage roll

In3luki:
```
In addition, you gain an impulse junction, a benefit that
occurs when you use an impulse of the chosen element
that takes 2 actions or more.
```
Currently, when an elemental blast attack misses, there is no roll note shown for the impulse junction effect.

After this PR:
image

### DIFF
--- a/packs/classfeatures/air-gate.json
+++ b/packs/classfeatures/air-gate.json
@@ -251,9 +251,29 @@
                             2
                         ]
                     },
-                    "impulse",
                     "junction:air:impulse",
-                    "air"
+                    "self:action:trait:impulse",
+                    "self:action:trait:air"
+                ],
+                "selector": "impulse-attack-roll",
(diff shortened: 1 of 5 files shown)


packs/classfeatures/air-gate.json
```diff
@@ -255,6 +255,26 @@
...
+                    {
+                        "not": "self:action:slug:elemental-blast"
+                    },
+                    {
+                        "gte": [
+                            "action:cost",
+                            2
+                        ]
+                    },
+                    "impulse",
```
stwlam, line 274: Avoid using top-level roll options

In3luki: That is copied from the existing note but I can change them to `self:action:trait:x` if you want.

stwlam: Yeah, that'd be better

PR #14178, opened 2024-03-12
Prevent gaining vision from controlled tokens without at least observer ownership

### LINKED ISSUE #14174: When NPCs are converted to lootables PCs are able to use the tokens vision
When playing with NPC vision enabled and lootable NPCs, after the NPC dies and becomes lootable players are able to select the token and see from their point of view with their senses, this can also allow players to see where their tokens can't. Perhaps it would be possible to disable vision when converting an NPC to a lootable?

stwlam:
Closes #14174


VictorElHajj:
@stwlam I think this made it so that when a PC tries to loot a dead NPC they suddenly lose all vision.

PR #14201, opened 2024-03-13
Rethrow migration exceptions with added information

In3luki:
image

### DIFF
--- a/src/module/migration/runner/base.ts
+++ b/src/module/migration/runner/base.ts
@@ -74,31 +74,49 @@ export class MigrationRunnerBase {
         const currentActor = fu.deepClone(actor);
 
         for (const migration of migrations) {
-            for (const currentItem of currentActor.items) {
-                await migration.preUpdateItem?.(currentItem, currentActor);
-                if (currentItem.type === "consumable" && currentItem.system.spell) {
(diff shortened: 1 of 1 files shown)


stwlam:
This is already caught at an outer scope

In3luki:
The current error is something like this:
```
Error thrown while migrating Actor.L0pXT0ba2Ovsjir9: Cannot read properties of undefined (reading 'flags')
```
Which isn't very helpful to locate where the error actually occured.

stwlam:
What additional details does the example screenshot have?

In3luki:
The `MigrationRunner` method the error occured in, the `Migration` version to identify the actual migration that failed and the `Migration` method the error occured in.

stwlam:
It's already obvious that a failed actor or item migration is from one of those two methods, and the screenshot doesn't show the migration number.

stwlam:
Oh, my bad: I see the number!

In3luki:
Lose everything but the version number, got it. 😄

In3luki:
```
Error thrown while migrating Actor.sJE82Ad2HKyKAB4k: Migration 0.915: Cannot read properties of
undefined (reading 'value')
```

PR #14211, opened 2024-03-14
Add save type alteration in mindshift toggle

### LINKED ISSUE #14166: Mindshift toggle changes damage type but not the save
Along with changing damage to mental, the mindshift trait can change the required save to Will

Geliogabalus:
Closes #14166

Adds simple logic for changing save types in text description to use with trait toggle. Also changed Violent Unleash feature description to be consistent with other mindshift feats.

### DIFF
--- a/packs/feats/violent-unleash.json
+++ b/packs/feats/violent-unleash.json
@@ -11,7 +11,7 @@
         },
         "category": "class",
         "description": {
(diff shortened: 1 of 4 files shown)


stwlam:
A solution for this shouldn't only work on inline checks. A rule element would be appropriate here, with non-synthetic changes (such as for this particular trait) handled somewhere in the codebase.

Issue #14212, opened 2024-03-14
FlatModifier RE ignores item's slug


Cuingamehtar:
The RE ignores the item's slug and uses its label instead.

This causes AdjustModifier to stop working with this modifier if the label is different. This causes issues for localization.

Self-contained example : Intimidating Prowess on a character with +5 strength and mastery in Intimidation

When the feat has the default name (Intimidating Prowess), it works as expected. But if the name is changed (a simple typo is enough), then AdjustModifier RE can't find the mod, and the +2 bonus to Intimidation is not applied.

xyzzy42:
If there are multiple FlatModifiers on the same item, they will all get the same slug if they fall back to the item's slug.  They can be given slugs to avoid fallback.  But a problem appears if this might not be the only instance of the rule on the actor, such as an effect with a choiceset applies multiple times with different choices chosen.  The value of the choice set can be placed into the label, and thence to the slug, to make them unique.  But the choiceset value can not be placed directly into the item slug.

PR #14213, opened 2024-03-14
Add slugs to FlatModifier REs targeted by AdjustModifier REs

Cuingamehtar:
Shouldn't be a need to check for empty string as empty slug auto-nulls on item card update. #14212

### DIFF
--- a/packs/classfeatures/catharsis-emotion-love.json
+++ b/packs/classfeatures/catharsis-emotion-love.json
@@ -31,6 +31,7 @@
                     "self:effect:emotional-fervor"
                 ],
                 "selector": "will",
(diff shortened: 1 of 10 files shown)


stwlam:
I don't think we can safely do this right now

Cuingamehtar:
Well, as another option, FlatModifier REs can be changed to have mandatory slugs themselves.

stwlam:
That is also not an option without a migration. You seem to have some items with rules in mind, so please feel free to fill slugs in those.

Cuingamehtar:
These are the Items that I found that exhibit the same pattern - FlatModifier RE without slug + AdjustModifier RE targeting that slug.

PR #14237, opened 2024-03-16
Add world time setting for imperial calendar

CarlosFdez:
It turns out this was half done, so I'm now concerned if there's some gotcha I was not aware of.


stwlam:
One gotcha would be the lack of any associated data like name or lengths of months, days, etc.

CarlosFdez:
From what I gather, they're the same as AR. Honestly, strange.

CarlosFdez:
It might be good to wait till the Tian Xia book is out. If they don't address it then, they never will.

PR #14255, opened 2024-03-19
Remake item transfer dialog, gms can now gift items and receive warnings when funds are insufficient rather than fail silently.

CarlosFdez:
image
image


CarlosFdez:
I updated it to keep the same flow as the original, but to add a new Give checkbox instead. The internal is still changed from running directly to being a resolvable though in preparation for item selling.

PR #14265, opened 2024-03-20
Update and Add Level 3 and 5 Iconics from *Player Core*

avagdumortesin:
Clarifying a couple issues with Alex:
Level 5 Ezren in the PDF doesn't have his Wayfinder, and Hand of the Apprentice is only doing 1d4+4 damage, even though his weapon is a +1 striking staff (I don't think this will matter, since the system uses the damage roll from the weapon). Level 5 Lini has a typo on the first page with the ancestry feats; animal elocutionist should be gnome obsession. Additionally, she's untrained in crafting, yet is hauling around a repair toolkit.


TikaelSol:
Did you send those to Alex already or do you want me to reach out?

PR #14309, opened 2024-03-27
Fix Mirage Dragon innate spells and tail attack

surged20:


packs/pathfinder-monster-core/mirage-dragon-adult-spellcaster.json
```diff
@@ -2600,7 +2601,7 @@
                         "value": 2
                     }
                 ],
-                "slug": "1-status-to-all-saves-vs-magic",
+                "slug": "2-status-to-all-saves-vs-arcane",
```
SpartanCPA, line 2604: This slug does not need to be updated. It will just cause noise if we refresh the feature later

SpartanCPA (approved):

PR #14377, opened 2024-04-04
Add draconic barrage spell effect

MSAbaddon:


TikaelSol:
I don't think we should half ass it, and I don't know that there's a non-janky way to implement this.

MSAbaddon:
Can you briefly tell me what you don't like about this implementation? I'll try to figure out a way to solve it, or close the pull request if I can't find a solution.

TikaelSol:
Sorry, been very busy.

This just throws a bunch of effects on the actor to track stuff and I'm not huge on tracking only effects with no actual automation attached. I'd rather we wait for a bit on this to see if we can't get a clever way to approach it instead.

MSAbaddon:
Ok, understood - I'll close the PR then :-)

PR #14398, opened 2024-04-07
Remove redundant console.warn from outdated module notification

esheyw:
`ui.notifications.warn` already posts to console unless it's fed `{console: false}` in it's second parameter. Currently this produces two warnings in the console per module, one from pf2e, one from foundry: .


esheyw:
Okay IDK what's up with it yelling about PredicatePF2e. I synced my fork right before making this branch. Probably a simple git-noob mistake again.

PR #14401, opened 2024-04-07
Allow camel options in sluggify Handlebars helper

esheyw:
If there's some reason this shouldn't go in I'm missing, let me know, but we're not letting anything that could cause an error though, so hopefully it's fine? The way a couple of the other helpers were written implies that I don't need to do more checking for if options gets passed or if hash exists, but that's also a guess.


esheyw:
Yea I did this in a very sleep-deprived manic completionist headspace. I can't think of an actual need. I'll close this one.

PR #14406, opened 2024-04-07
Predicate on effect's roll option instead of badge value for Bones Oracle REs

### LINKED ISSUE #14394: Predicates on bone oracle's curse not working
I was testing a bone oracle's curse and I found some strange behaviour during those tests: some of the rule elements were triggered while others with the same predicate and selector weren't.
- Put the effect Curse of Living Death on the actor and increase it to moderate (or more)
- Have the actor roll a `@Check[type:fortitude|dc:20|traits:poison]`
The +4 status bonus FlatModifier RE should trigger and in case of success, the AdjustDegreeOfSuccess RE should have it become a critical success. But it doesn't.
On the other hand, the Note RE on the effect is displayed in the chat.
Those 3 REs have the same predicate and selector. And looking at them, I don't see a reason why it would work in one case and not in the other.
After some more tests, removing the {"gte":["parent:badge:value",2]} part of the predicate on the problematic REs allowed the 2 REs to trigger, but that's not really a solution^^

pedrogrullada:
Closes #14394 
For some reason `parent:badge:value` is not being satisfied as a predicate for Flat Modifier and Adjust Modifier rule elements. I'm not sure if this is bug in and of itself, but regardless using the equivalent roll option setup by the effect itself works.

### DIFF
--- a/packs/feat-effects/effect-curse-of-living-death.json
+++ b/packs/feat-effects/effect-curse-of-living-death.json
@@ -10,6 +10,7 @@
                 "Major",
                 "Extreme"
             ],
+            "loop": false,
             "type": "counter",
             "value": 1
         },
@@ -36,7 +37,7 @@
                 "predicate": [
                     {
                         "gte": [
-                            "parent:badge:value",
+                            "self:effect:curse-of-living-death",
                             2
                         ]
                     },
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This is a bug, and I have a fix for it in mind.

pedrogrullada:
Yeah, that makes sense. I'll go ahead and close this.

PR #14408, opened 2024-04-08
Fix predicating parent roll options in modifier clones

CarlosFdez:
This is the true cause behind https://github.com/foundryvtt/pf2e/issues/14394. Rule isn't enumerable and must be copied separately.

I don't understand why there is a mergeObject without arguments though. It was introduced in https://github.com/CarlosFdez/pf2e/commit/1f9ae8ea0e2244936453f6d8f7297c1d7eee2c2a . I kept it in just in case but my guess is that it should probably be removed.


CarlosFdez:
I removed the mergeObject, I can't see it as anything other than a mistake.

PR #14411, opened 2024-04-08
Adds a new `Charges` spellcasting entry

reonZ:
I thought it could be useful for people who want to handle spells with charges.
It works by only ever using and referencing the `slot1` of the spellcasting entry. Casting a spell consumes an amount of charges equal to its heightened rank.

Foundry_Virtual_Tabletop_bvajvBi617

Foundry_Virtual_Tabletop_pHFFqhEvqN
I wasn't super inspired by the NPC sheet design, with the lack of real-estate in the header. I didn't want to touch it too much, but i guess reducing the size of all the number inputs and removing the attribute dropdown that isn't really used could make it fit there.


stwlam:
Than you, but we do not intend to expand spellcasting entries as items.

CarlosFdez:
This is the kind of thing that would have been very helpful to you if you communicated it beforehand to avoid the work :(. The plan for staves is to embed the spell in the stave itself (changing its format) and then generating the spellcasting collection at runtime. The primary hard thing to solve will be handling spells created after data preparation, as well as converting retrieval to asynchronous in the chat message.

Spellcasting entries in general, from my suspicion, will become actor data bolstered by feat subfeatures (since we will have to support NPCs as well).

reonZ:
I actually don't use that for the staves, i do something similar to the `ItemSpellcasting`: a "virtual" entry that implements `SpellcastingEntry` and is added during `CharacterPF2e#prepareData`.

This was something annex i made for convenience because people were asking for it and thought it could maybe be added to the system.

PR #14412, opened 2024-04-09
Allow setting the indentation value in json handlebars helper

esheyw:
This one I can actually justify though: I am displaying the value of object settings in an 'are you sure you want to reset this to default' box, currently handling the stringification in the data, would be nice to move it to the template but I need the indentation parameter. I don't think it needs more guards than this, it doesn't choke on NaN.

### DIFF
--- a/src/scripts/handlebars.ts
+++ b/src/scripts/handlebars.ts
@@ -54,8 +54,8 @@ export function registerHandlebarsHelpers(): void {
         return sluggify(String(text));
     });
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Is this a system object settings? I'm leaning towards modules not relying on system handlebars helpers for anything except maybe alternative sheets. This is something that can easily be handled in getData() more reliably without risk of breaking.

CarlosFdez:
Feel free to reopen if this was a requirement for a system level thing.

PR #14428, opened 2024-04-12
Include travel speed rates in party sheet

CarlosFdez:
Chart derived from https://2e.aonprd.com/Rules.aspx?ID=2441.
Also includes a few minor style touchups for the sidebar box elements

image

image


TMunsie:
Rubber Ducked :)

Issue #14444, opened 2024-04-15
Support RollOption rule elements in Hazards


MrVauxs:
As the title describes. Here is the following rule element I was using with the ability that has it unfolded in the screenshot. Moving the ability to an NPC sheet added the toggle.
`{"key":"RollOption","label":"1 Gem Broken","option":"one-gem","toggleable":true}`

![chrome_y1PnPm4chl](https://github.com/foundryvtt/pf2e/assets/32039708/7ce380d7-312f-4401-a492-0d24ae6042f2)

CarlosFdez:
I have a thing ready, but it won't go in until after after V12 is functional.

PR #14452, opened 2024-04-17
Add roll option and aelike re support to hazards

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/14444

image

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -778,6 +778,10 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         }
 
         this.prepareDataFromItems();
+
+        for (const rule of this.rules) {
+            rule.onApplyActiveEffects?.();
+        }
     }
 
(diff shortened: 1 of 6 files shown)


CarlosFdez:
This should be good to look at now.

src/module/actor/hazard/document.ts
```diff
@@ -98,6 +98,14 @@ class HazardPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e |
         details.alliance = null;
     }
 
+    /** Run rule elements */
+    override prepareEmbeddedDocuments(): void {
+        super.prepareEmbeddedDocuments();
+        for (const rule of this.rules) {
```
stwlam, line 104: Why not just move this to the base class?

CarlosFdez: As a refactor? Sure. I'm down. I assumed there was a reason it wasn't.

stwlam: I think it was a redundant safeguard on them not getting executed by loot actors.

PR #14454, opened 2024-04-17
Set default level in party check level-based DC's to average party level.

### LINKED ISSUE #11664: Feature Request:  Generate Check Prompt macro's Level-Based DC should allow quickly picking the party's level
This is what the prompt currently looks like, after clicking the Level-Based DC tab:

image

Since most people use level-based DCs that are nearly or exactly the party's current level, I think that flow should be easier.

Ideas:

1. prefill the input with the party's level, e.g. 7
2. add a placeholder text that says "Leave empty to use the party's level (7)", and make it work that way
3. add a small button next to the input, that just says `[7]` (maybe with a tooltip, "Party level"), and pressing it sets the input's value to that number.
4. as 3, with extra buttons for all levels of all creatures or hazards that appear in the scene  (there's more than enough space in the window)
5. If party level is undefined (due to the users not using the Party feature) - perhaps look through actors directory, filter for Characters, pick the mode (the most common occurrence) of the levels

Mockup of option 3:

image

racehd:
splitting PR from #14450 to include only default level feature request

Potentially Closes #11664 
  - I did not like the idea of extra clicks to prefill the default value, so I went with the most common occurrence to use the party level as the default value automatically rather than have to push an extra button to do so.  I believe this will be a better user experience than a separate level button. 
  - I think this can close the request, but that depends if you see value in the request for the option to have a button(s) for levels of creatures and hazards on the scene. Practically speaking I am not sure why you would be using a level based DC from Creature's/Hazards rather than the party level.

### DIFF
--- a/src/scripts/macros/check-prompt/prompt.ts
+++ b/src/scripts/macros/check-prompt/prompt.ts
@@ -14,7 +14,7 @@ interface CheckPromptDialogOptions extends ApplicationOptions {
 interface CheckPromptDialogData {
     proficiencyRanks: SelectData[];
     dcAdjustments: SelectData[];
-    averageLevel: number;
+    partyLevel: number | null;
 }
 
 interface SelectData {
@@ -51,13 +51,10 @@ class CheckPromptDialog extends Application<CheckPromptDialogOptions> {
         this.#actions = await getActions();
         this.#lores = loreSkillsFromActors(this.options.actors ?? game.actors.party?.members ?? []);
 
-        const actors = this.options.actors ?? game.actors.party?.members ?? [];
-        const averageLevel = actors.reduce((acc, actor) => acc + actor.level, 0) / actors.length;
-
         return {
             proficiencyRanks: this.#prepareProficiencyRanks(),
             dcAdjustments: this.#prepareDCAdjustments(),
-            averageLevel: Math.round(averageLevel),
+            partyLevel: game.actors.party?.level ?? null,
         };
     }
 
--- a/src/scripts/macros/check-prompt/prompt.ts
+++ b/src/scripts/macros/check-prompt/prompt.ts
@@ -14,6 +14,7 @@ interface CheckPromptDialogOptions extends ApplicationOptions {
 interface CheckPromptDialogData {
     proficiencyRanks: SelectData[];
     dcAdjustments: SelectData[];
+    averageLevel: number;
 }
 
 interface SelectData {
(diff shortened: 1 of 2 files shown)


CarlosFdez:
So I'm double checking, and sorry for not catching this the first time, but there's already a level property that already grabs the average party level. {{document.level}} in handlebars should grab it without needing to do anything in sheet.ts.

CarlosFdez:
Actually I mixed up the sheets, my advice was wrong, I'll notate better advice.

src/scripts/macros/check-prompt/prompt.ts
```diff
@@ -50,9 +51,13 @@ class CheckPromptDialog extends Application<CheckPromptDialogOptions> {
         this.#actions = await getActions();
         this.#lores = loreSkillsFromActors(this.options.actors ?? game.actors.party?.members ?? []);
 
+        const actors = this.options.actors ?? game.actors.party?.members ?? [];
+        const averageLevel = actors.reduce((acc, actor) => acc + actor.level, 0) / actors.length;
+
         return {
             proficiencyRanks: this.#prepareProficiencyRanks(),
             dcAdjustments: this.#prepareDCAdjustments(),
+            averageLevel: Math.round(averageLevel),
```
CarlosFdez, line 60:
```suggestion
        return {
            proficiencyRanks: this.#prepareProficiencyRanks(),
            dcAdjustments: this.#prepareDCAdjustments(),
            partyLevel: game.actors.party?.level,
```

static/templates/gm/check-prompt.hbs
```diff
@@ -33,7 +33,7 @@
             <section class="tab active" data-tab="level-dc">
                 <div class="form-group">
                     <label for="check-prompt-level-dc">{{localize "PF2E.LevelLabel"}}</label>
-                    <input type="number" id="check-prompt-level-dc" name="level-dc" />
+                    <input type="number" id="check-prompt-level-dc" name="level-dc" value="{{averageLevel}}" />
```
CarlosFdez, line 36:
```suggestion
                    <input type="number" id="check-prompt-level-dc" name="level-dc" value="{{partyLevel}}" />
```

CarlosFdez (commented):
Going over this code again I see things that need to be changed elsewhere but those should be left for a different PR.

src/scripts/macros/check-prompt/prompt.ts
```diff
@@ -14,6 +14,7 @@ interface CheckPromptDialogOptions extends ApplicationOptions {
 interface CheckPromptDialogData {
     proficiencyRanks: SelectData[];
     dcAdjustments: SelectData[];
+    averageLevel: number;
```
CarlosFdez, line 17:
```suggestion
    partyLevel: number | null;
```

racehd:
> So I'm double checking, and sorry for not catching this the first time, but there's already a level getter property in actor that grabs the average party level (which is assigned in PartyPF2e in prepareBaseData) . {{document.level}} in handlebars should grab it without needing to do anything in sheet.ts.
> 
> EDIT: disregard, document isn't in this sheet, I'll post a more helpful response.

No need to apologize, I had looked through the party sheet but missed the partyLevel somehow. Thanks for pointing that out. I did find it and was testing when you did your review comments, but it was good to see I got too the right conclusion on my own. Sorry for not getting that in sooner to save you the 2nd(3rd?) review.

PR #14464, opened 2024-04-19
Automate Keen Follower feat

xyzzy42:
This will add the larger bonus when following an Expert or Master.

To get the necessary information, the Follow The Expert effect makes the ChoiceSets copy to actor flags.  The proficiency is also a roll option.

The proficiency roll option is used so the Keen Follower modifier can be disabled when it shouldn't apply, i.e. when following a trained or legendary "expert".

I also tried as an AdjustModifier, but without a relabel there's no clear indication the feat is working.  And with a relabel, the localized skill name is removed from the modifier, e.g. "Follow The Expert (Arcana)" would become "Keenly Follow The Expert".

### DIFF
--- a/packs/feats/keen-follower.json
+++ b/packs/feats/keen-follower.json
@@ -24,7 +24,24 @@
             "remaster": false,
             "title": "Pathfinder Advanced Player's Guide"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "AdjustModifier",
+                "mode": "upgrade",
+                "predicate": [
+                    "self:effect:follow-the-expert",
+                    {
+                        "or": [
+                            "follow-the-expert:proficiency:rank:2",
+                            "follow-the-expert:proficiency:rank:3"
+                        ]
+                    }
+                ],
+                "selector": "skill-check",
+                "slug": "follow-the-expert-circumstance",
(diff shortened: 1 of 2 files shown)


xyzzy42:
Found a problem with this.  If there is no Follow the Expert effect on the actor, the actor flag is not defined, and then the selector of the feat's RE fails to validate.  The predicate will also be false, as it should be, and so the RE isn't going to be used.  But it's still a validation failure.

So I guess there is no way to use actor flags, from some other item, as a selector as they won't validate when the flag isn't there.

A workaround would be use a wider selector, e.g. `skill-check`, and then narrow to the specific skill via predicate.

Maybe selectors that resolve to undefined actor flags could be considered to be a disabled RE, rather than a validation failure?  Seems like this would come up with various feat/feature rules that interact with effects.

TikaelSol:
I'm not terribly keen on doing it this way. AdjustModifier, even without a relabel, would be preferable.

xyzzy42:
What would be preferable, to not relabel and keep the modified effect name that has the skill in it, or to relabel to the Keen Follower title, but lose the skill from the effect name?

xyzzy42:
Changed to AdjustModifier.  This also was able to fix the validation problem.

PR #14465, opened 2024-04-19
fix compendium folder for Tian Xia world guide Compendium

Drental:
the other changes are format on save, is that fine or did it mess up some settings?


stwlam:
It didn't mess anything up, but fix before merge

Drental:
okay, done

PR #14469, opened 2024-04-19
Allow Relic Trait to be added to Armor and Shields

### LINKED ISSUE #14254: [Feature Request] Allow "Relic" Trait to be added to more items
As it stands, the Relic trait can't be given to armor or shields, as far as I've read there is no actual limit that exists that would prevent relics to be armor or shields. I understand relics aren't a high priority at the moment, but it would be nice to have this functionality.

DocSchlock:
Closes #14254


CarlosFdez:
I'm uncertain about the relic trait itself since its only a thing in sky king's tomb and treasure vault with no definition, but if we're gonna have it all then by its intent it should indeed probably be on every trait collection.

PR #14474, opened 2024-04-20
Remove `TextEditor._enrichContentLinks` override 

In3luki:
The override breaks enrichment. This should be solved by core `TextEditor._primeCompendiums` now anyway.


CarlosFdez:
We should probably solve it as part of removing it to avoid forgetting it.

CarlosFdez:
Oh I misunderstood. You meant that this is no longer necessary. I started typing TextEditor and you are completely correct.

PR #14475, opened 2024-04-20
Destroyed armor not provide AC

reyzor1991:


CarlosFdez:
Bit of a late reply, but if you handle broken, then sure!

PR #14483, opened 2024-04-21
Restore surface cantrip for Psi Development psychics

### LINKED ISSUE #14476: Psi Development does not allow conscious mind surface amp choice
https://github.com/foundryvtt/pf2e/pull/14344 removed surface amp spells from the flag for psychic dedication choices. Psi Development allows taking the unique surface amp spell, and looks at that flag for the choices. Since the spell is no longer included in the flag, and is not added by any other rule, the spell is not available as a valid choice.

pedrogrullada:
Closes #14476

The Choice Sets in each conscious mind and Psi Development were not checking if the predicates set in `flags.pf2e.psychic.dedication.psiCantrips` were satisfied. I've opted for a different approach, where Psychic Dedication sets up the previously mentioned flag as an empty array, and then each conscious mind adds its standard cantrips to this array depending on whether or not they've been chosen before. Then, Psi Development will also add to this array depending on your choice of conscious mind.

After testing, this works as expected, with the user having the choice of two standard cantrips when they take the dedication, and then when they take Psi Development they can choose between the surface cantrip and the standard cantrip they didn't take initially.

### DIFF
--- a/packs/feats/psi-development.json
+++ b/packs/feats/psi-development.json
@@ -37,6 +37,84 @@
                 "value": 1
             },
             {
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "predicate": [
+                    "feature:the-distant-grasp"
+                ],
+                "priority": 10,
+                "value": {
+                    "label": "PF2E.SpecificRule.Psychic.PsiCantripLabels.TelekineticRend",
+                    "value": "telekinetic-rend"
+                }
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "predicate": [
+                    "feature:the-infinite-eye"
+                ],
+                "priority": 10,
+                "value": {
+                    "label": "PF2E.SpecificRule.Psychic.PsiCantripLabels.GlimpseWeakness",
+                    "value": "glimpse-weakness"
+                }
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "predicate": [
+                    "feature:the-oscillating-wave"
+                ],
+                "priority": 10,
+                "value": {
+                    "label": "PF2E.SpecificRule.Psychic.PsiCantripLabels.ThermalStasis",
+                    "value": "thermal-stasis"
+                }
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "predicate": [
+                    "feature:the-silent-whisper"
+                ],
+                "priority": 10,
+                "value": {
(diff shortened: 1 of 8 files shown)


packs/feats/psychic-dedication.json
```diff
@@ -41,6 +41,13 @@
                 "path": "system.skills.occ.rank",
                 "value": 1
             },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "priority": 10,
+                "value": []
+            },
```
TikaelSol, line 50: This would require a migration

pedrogrullada: Sorry, I didn't realize this. I can't write it myself, unfortunately. Considering how busy you are, should I instead restore this to how it was in the first place in the meantime, or leave this be for now?

TikaelSol: I'll see if I can find time to write one, it's an easy one since it's a near carbon copy of an existing migration. Though I do wonder if there's a more clever way to do this...

pedrogrullada: The original way is quite good except the Choice Sets are seemingly ignoring the predicates in their choices. Do you remember if that actually worked at the time? Maybe it's actually a bug and we don't need to touch this.

TikaelSol: I could swear they worked when I did them, but I really haven't had time to dig into it

pedrogrullada: For the sake of my sanity, I went back and checked, and it looks like this part never worked (that is, surface cantrip appearing as a choice in the dedication, and not discarding an already taken standard cantrip from Psi Development)

packs/feats/psi-development.json
```diff
@@ -36,6 +36,84 @@
...
+                "path": "flags.pf2e.psychic.dedication.psiCantrips",
+                "predicate": [
+                    "feature:the-distant-grasp"
+                ],
+                "priority": 10,
+                "value": {
+                    "label": "PF2E.SpecificRule.Psychic.PsiCantripLabels.TelekineticRend",
+                    "value": "telekinetic-rend"
+                }
+            },
```
TikaelSol, line 51: Each of these should go on their respective mind, predicated on having this feat.

TikaelSol: I realize this just reinvents what we had before. I think we should figure out the bug then fix that and revert to the old way.

pedrogrullada: > I realize this just reinvents what we had before. I think we should figure out the bug then fix that and revert to the old way.

Yeah, the idea was that since the predicates were not being read in the Choice Set, they had to be processed before the cantrips even went into the array, so that's why I did it this way. But if the reason the original way is not working is a bug, then I will try to get some reproduceable steps and make an issue about it, and once it's fixed revert, as you say.

pedrogrullada: Drafting this for now. I've reported the bug in #14574.

pedrogrullada:
#14579 makes this unnecessary, I will revert to the original implementation in a separate PR.

PR #14501, opened 2024-04-23
Fix status effect icons

In3luki:
Allows the creation of active effects that carry the "dead" status only. This could be relaxed to allow all statuses registered in `CONFIG.statusEffects` as long as the `overlay` flag is set.
A positive side effect of this change should be that modules that rely on `actor.statuses.has("dead")` should work out of the box now.

### DIFF
--- a/src/module/canvas/status-effects.ts
+++ b/src/module/canvas/status-effects.ts
@@ -95,15 +95,15 @@ export class StatusEffects {
     static #updateStatusIcons(): void {
         const iconTheme = game.settings.get("pf2e", "statusEffectType");
         const directory = iconTheme === "default" ? "conditions" : "conditions-2";
-        CONFIG.statusEffects = Object.entries(CONFIG.PF2E.statusEffects.conditions).map(([id, label]) => ({
+        CONFIG.statusEffects = Object.entries(CONFIG.PF2E.statusEffects.conditions).map(([id, name]) => ({
             id,
-            label,
-            icon: `systems/pf2e/icons/${directory}/${id}.webp` as const,
+            name,
+            img: `systems/pf2e/icons/${directory}/${id}.webp` as const,
         }));
         CONFIG.statusEffects.push({
             id: "dead",
-            label: "PF2E.Actor.Dead",
-            icon: CONFIG.controlIcons.defeated,
+            name: "PF2E.Actor.Dead",
+            img: CONFIG.controlIcons.defeated,
         });
     }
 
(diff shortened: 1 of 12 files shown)


src/module/scene/token-document/document.ts
```diff
@@ -29,12 +29,11 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
 
     /** Check actor for effects found in `CONFIG.specialStatusEffects` */
     override hasStatusEffect(statusId: string): boolean {
-        if (statusId === "dead") return this.overlayEffect === CONFIG.controlIcons.defeated;
-
         const actor = this.actor;
         if (!actor || !game.pf2e.settings.rbv) {
             return false;
         }
+        if (statusId === "dead") return actor.statuses.has("dead");
```
CarlosFdez, line 36: This being after the rbv check seems odd, is that right?

In3luki: An actor is required for AEs so I moved it after the actor check but I think it makes no difference so it could stay at the top too. 
`if (statusId === "dead") return !!actor?.statuses.has("dead");` would lead to the same result I think.

CarlosFdez: My concern is with || !game.pf2e.settings.rbv

In3luki: Oh right!🤦‍♂️
I'll move it to the top.

src/module/canvas/status-effects.ts
```diff
@@ -281,16 +280,14 @@ export class StatusEffects {
             if (objectHasKey(CONFIG.PF2E.conditionTypes, slug)) {
                 const newCondition = game.pf2e.ConditionManager.getCondition(slug).toObject();
                 await token.actor?.createEmbeddedDocuments("Item", [newCondition]);
-            } else if (iconSrc && (event.shiftKey || control.dataset.statusId === "dead")) {
```
CarlosFdez, line 284: I wonder what shift key did here, I'm testing in live and its not doing anything. I'll assume what you did is correct.

In3luki: I think it was supposed to set the icon as overlay effect. I'm not sure when that broke but from the lack of an issue I'm assuming no one used it.

CarlosFdez: I tested in 5e and I can't get the icon to overlay anyways via the condition hud.

PR #14503, opened 2024-04-23
Fix damage dialog rollmode display

### LINKED ISSUE #14439: Damage rolls are supposed to override the default roll mode to be public rolls, but they don't.
When rolling in Private GM Roll mode, pf2e is supposed to override this mode for damage rolls, always making damage rolls public. It does set the roll mode to Public Roll in the damage roll menu, but then the roll itself isn't actually made public, and my players can't see it or click the damage. Notably, if I instead manually select Public Roll in the damage roll menu, then it does properly override my default roll mode. But when it's automatically set to public in the menu, it fails to override unless I manually unselect and then reselect Public Roll in the dropdown.
image
image
What my players see:
image

DocSchlock:
Closes #14439

This sets the initial roll mode of Damage Rolls from undefined to the current user default. Undefined was falling back on the current user default anyway but displaying as Public Roll, leading to assumptions from users.

~~Also changes the secret rollMode handling a bit in damage.ts to compensate for undefined no longer being the default pass-through.~~

### DIFF
--- a/src/module/system/damage/damage.ts
+++ b/src/module/system/damage/damage.ts
@@ -20,7 +20,8 @@ export class DamagePF2e {
     ): Promise<Rolled<DamageRoll> | null> {
         const outcome = context.outcome ?? null;
 
-        context.rollMode ??= (context.secret ? "blindroll" : undefined) ?? game.settings.get("core", "rollMode");
+        context.rollMode = context.secret ? "blindroll" : context.rollMode ?? game.settings.get("core", "rollMode");
+
         context.createMessage ??= true;
 
         // Change default roll mode to blind GM roll if the "secret" option is specified
--- a/src/module/system/damage/damage.ts
+++ b/src/module/system/damage/damage.ts
@@ -20,8 +20,7 @@ export class DamagePF2e {
     ): Promise<Rolled<DamageRoll> | null> {
         const outcome = context.outcome ?? null;
 
-        context.rollMode = context.secret ? "blindroll" : context.rollMode ?? game.settings.get("core", "rollMode");
-
+        context.rollMode ??= (context.secret ? "blindroll" : undefined) ?? game.settings.get("core", "rollMode");
         context.createMessage ??= true;
 
         // Change default roll mode to blind GM roll if the "secret" option is specified
--- a/src/module/system/damage/dialog.ts
+++ b/src/module/system/damage/dialog.ts
@@ -141,7 +141,7 @@ class DamageModifierDialog extends Application {
             " + ",
         );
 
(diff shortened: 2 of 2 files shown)


src/module/system/damage/damage.ts
```diff
@@ -20,7 +20,8 @@ export class DamagePF2e {
     ): Promise<Rolled<DamageRoll> | null> {
         const outcome = context.outcome ?? null;
 
-        context.rollMode ??= (context.secret ? "blindroll" : undefined) ?? game.settings.get("core", "rollMode");
+        context.rollMode = context.secret ? "blindroll" : context.rollMode ?? game.settings.get("core", "rollMode");
```
CarlosFdez, line 23: A few things regarding this:
1) If rollMode is supplied, we need to always accept it. If someone does a public secret roll, then by golly, they want a public secret roll. This means the first ??= is correct. 
2) Please put parenthesis. The order of operations for ?? is unclear without them.
3) The space after this probably shouldn't exist. The reason context.rollMode and context.createMessage were together with no gaps between is that both pieces of code involve setting default users if the user didn't supply data.

If this isn't the bugfix, and the bugfix involves what's in damage/dialog.ts, it might be best to omit this change.

src/module/system/damage/dialog.ts
```diff
@@ -216,7 +218,7 @@ class DamageModifierDialog extends Application {
                 R.pick(CONFIG.PF2E.damageCategories, Array.from(DAMAGE_CATEGORIES_UNIQUE)),
             ),
             rollModes: CONFIG.Dice.rollModes,
-            rollMode: this.context?.rollMode,
+            rollMode: currentRollMode,
```
CarlosFdez, line 221:
```suggestion
            rollMode: this.context?.rollMode ?? game.settings.get("core", "rollMode"),
```

src/module/system/damage/dialog.ts
```diff
@@ -141,6 +141,8 @@ class DamageModifierDialog extends Application {
             " + ",
         );
 
+        const currentRollMode = this.context?.rollMode ?? game.settings.get("core", "rollMode");
+
         type DamageDicePF2eWithOverride = DamageDicePF2e & { override: NonNullable<DamageDicePF2e["override"]> };
```
CarlosFdez, line 146:
```suggestion
        );

        type DamageDicePF2eWithOverride = DamageDicePF2e & { override: NonNullable<DamageDicePF2e["override"]> };
```

CarlosFdez (commented):
For large functions like this, if a variable doesn't see reuse, I prefer them being part of the getData() object

PR #14506, opened 2024-04-23
Set Inline Damage Traits to "item:trait" to interact with IWR

### LINKED ISSUE #14403: Blessed Blood (spirit damage with holy trait) don't trigger weakness
Despite having the holy trait, Spirit damage coming from Blessed Blood isn't applying the weakness damage to creatures weak to it.
image

DocSchlock:
Closes #14403

EDIT:
This sets Damage inline traits to "item:trait" as mentioned by Tikael so they interact with the IWR system correctly.

Also fixed a spelling error in the Holy trait description.


TikaelSol:
I actually think the fix is the other way, to spit inline traits into `item:traits:x`, since that will fix more than just this issue.

src/module/system/text-editor.ts
```diff
@@ -944,10 +944,13 @@ async function augmentInlineDamageRoll(
                   ].flat(),
               );
 
+        // set inline damage traits to item:trait for the IWR system
+        const modifyTraits = traits ? traits.map((t) => `item:trait:${t}`) : [];
```
DocSchlock, line 948: I'm aware this needs altered to avoid messing with traits like materials and such.

src/module/actor/data/iwr.ts
```diff
@@ -70,7 +70,7 @@ abstract class IWR<TType extends IWRType> {
             case "primal":
                 return [{ or: [`item:trait:${iwrType}`, `origin:action:trait:${iwrType}`] }];
             case "area-damage":
-                return ["area-damage"];
+                return [{ or: ["area-damage", "item:trait:area-damage"] }];
```
CarlosFdez, line 73: I'm not sure I can reasonably accept this. Inline damage needs to get in line with IWR, not the other way around.

Inline roll links should not be dumping things into item:trait. In fact, this was the why the `options` param was added to text editor. But no one ended up using it. Though that param should probably be renamed since just options is ambiguous in that use case.

DocSchlock: What if I used `damage:trait` as a prefix instead and expanded IWR predication to accept it?

Area Damage doesn't make this clear, but IWR only accepts `item:trait` in most non-material, non-damage-type trait cases - which is the root of the issue. Area Damage is the sole exception to that rule. Inline damage rolls are currently just passing through the plain rolloption which IWR doesn't pick up.

CarlosFdez: Since this seems to be a bug, I'll take a look at it later.

CarlosFdez: I went ahead and looked into it. Tikael is right that we should dump traits into item:trait:x, but we should only dump real traits. For area-damage and such we need to fix it using another approach (such as maybe expanding what options IWR can see, or nominating area-damage as something else like a damage tag. I'd rather leave how to handle inline area damage to @stwlam since IWR is their child.

CarlosFdez: That said, thanks for narrowing the problem down, it was actually a big help.

CarlosFdez: I tested area damage (cassisian helmet) and the IWR seems to work as is, so I think we're fine on that front.

PR #14508, opened 2024-04-23
Fix Gigaton Strike and Megavolt Class DC

Cuingamehtar:
Implemented the fix proposed [here](https://old.reddit.com/r/FoundryVTT/comments/1c9f90f/bug_with_inventors_megavolt/l0lfjf7/) to correctly calculate the class DC


TikaelSol:
It needs to use inventor class DC RAW. Constructs using the system version of the item aren't going to work properly and we shouldn't break one thing to make another thing work better. CC can include its own version of the feats with this change instead

PR #14509, opened 2024-04-23
Convert more instances of the deprecated select handlebars helper to selectOptions

In3luki:
The `selectOptions` helper is pretty neat.

image

### DIFF
--- a/static/templates/actors/spellcasting-dialog.hbs
+++ b/static/templates/actors/spellcasting-dialog.hbs
@@ -3,11 +3,7 @@
         <div class="form-group">
             <label>{{localize "PF2E.SpellcastingTypeLabel"}}</label>
             <select name="system.prepared.value" {{disabled object.id}}>
-                {{#select system.prepared.value}}
-                    {{#each spellcastingTypes as |label mode|}}
-                        <option value="{{mode}}">{{localize label}}</option>
-                    {{/each}}
-                {{/select}}
+                {{selectOptions spellcastingTypes selected=system.prepared.value localize=true}}
             </select>
         </div>
         {{#if object.isPrepared}}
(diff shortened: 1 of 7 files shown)


static/templates/actors/spellcasting-dialog.hbs
```diff
@@ -34,48 +27,30 @@
...
-                            <option value="">{{localize "PF2E.MagicTraditionUseSpellLabel"}}</option>
-                        {{/if}}
-                        {{#each magicTraditions as |label mode|}}
-                            <option value="{{mode}}">{{localize label}}</option>
-                        {{/each}}
-                    {{/select}}
+                    {{#if (eq system.prepared.value "items")}}
+                        <option value="">{{localize "PF2E.MagicTraditionUseSpellLabel"}}</option>
+                    {{/if}}
+                    {{selectOptions magicTraditions selected=system.tradition.value localize=true}}
```
CarlosFdez, line 40: I'm a bit concerned about this one. If selected is blank, it won't select the blank one. Maybe it'll work incidentally because of how a re-render works, Blank can't be contextfual I imagine? What if false is passed to blank?

In3luki: Oh I missed that. Yeah, that probably needs handling in `getData` I'll get to that tomorrow.

CarlosFdez: Maybe its ok to have an if/else for prepared value "items" and the only difference be the "blank" option.

CarlosFdez: Maybe its ok to have the if/else have repeated selectOptions but one defines blank

CarlosFdez:
the helper is indeed pretty neat tho. It doesn't cover everything we're doing but for the stuff it does, its significantly better!

static/templates/actors/spellcasting-dialog.hbs
In3luki: That is a very good idea that I didn't think of. I've changed it to an else clause.

PR #14515, opened 2024-04-24
Start converting VisionSource/LightSource/etc to new V12 structure

CarlosFdez:
Accessing these properties from the global scope is now deprecated (and they're renamed to boot). Dark/Light sources are now split apart.


In3luki (approved):
LGTM

PR #14520, opened 2024-04-24
Convert all select handlebars helper in weapon sheet to selectOptions

In3luki:
I _think_ everything still works but I ran out of time to test it again.


CarlosFdez:
meleeUsageDamageDie vs damageDie is pretty vague if the only real difference is that the former has 1d6 and the latter has d6.

PR #14522, opened 2024-04-25
Add button to show item roll options in item sheet

CarlosFdez:
image

Requires V12 first due to locked tooltip bugfixes.


CarlosFdez:
Rebased and targets v12 now. Should be good to go whenever.

PR #14524, opened 2024-04-25
Add `omit` and `pick` Handlebars helpers and revert some changes in weapon sheet

In3luki:


CarlosFdez:
Yeah I think this solves a large number of selectOptions incompatibilities!

PR #14532, opened 2024-04-26
Convert more select helpers to selectOptions

In3luki:


src/module/item/weapon/document.ts
```diff
@@ -189,6 +189,10 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
         return new Set(this.system.traits.otherTags);
     }
 
+    get ammoLabel(): string | null {
```
CarlosFdez, line 192: We can't really add any sort of public api to documents, at least not without *heavy* discussion about it. Even then, ammoLabel is a sheet concern, so I'd rather not do this anyways. Might be best to keep the select options that rely on this as is until the end.

src/module/item/class/sheet.ts
```diff
@@ -11,7 +11,10 @@ export class ClassSheetPF2e extends ABCSheetPF2e<ClassPF2e> {
         return {
             ...sheetData,
             skills: CONFIG.PF2E.skills,
-            proficiencyChoices: CONFIG.PF2E.proficiencyLevels,
+            proficiencyChoices: CONFIG.PF2E.proficiencyLevels.map((p, index) => ({
```
CarlosFdez, line 14: As of the newest release, we no longer have to do this.

In3luki: yeah, that works now.

src/module/item/weapon/document.ts
In3luki: It turns out that the ammo item that was referenced in `strike.ammos` is not used anywhere so it was possible to change that to from `(ConsumablePF2e<ActorPF2e> | WeaponPF2e<ActorPF2e>)[];` to `{ id: string; label: string }[]` and everything still works.

PR #14535, opened 2024-04-27
Update user config overrides for v12

CarlosFdez:
The appearance of core side is different from pf2e side due to some unfortunate text wrapping of the label in standard-form mode. We might have to make some changes later on to form styles, or see if core bumps into the same issue as they continue to convert. We'll see!

image

image


src/module/user/sheet.ts
```diff
@@ -1,14 +1,62 @@
...
+
+    override tabGroups = {
+        primary: "core",
+    };
+
+    #getTabs() {
+        const DEFAULTS = { group: "primary" as const, active: false, cssClass: "" };
+        const tabs = [
+            { ...DEFAULTS, id: "core", icon: "fa-solid fa-cog", label: "Core" },
+            { ...DEFAULTS, id: "pf2e", icon: "fa-solid fa-cog", label: "PF2E.Pathfinder" },
```
In3luki, line 26: I don't know if `Pathfinder` is the best label for this. `Pathfinder Second Edition` like in the settings is probably too long. Maybe `System`?

types/foundry/build/generate-client-base-mixes.ts
```diff
@@ -470,6 +470,12 @@ const clientDocs: Record<string, { hasSheet?: boolean; isCanvasDoc?: boolean; pa
     Wall: { isCanvasDoc: true },
 };
 
+// imports
+console.log(
+    String.raw`import type { ApplicationV2, DocumentSheetV2 } from "../../../client-esm/applications/api/module.d.ts";
+
+`,
+);
```
In3luki, line 478: I have no idead how this works so it's probably not an error, just checking that this is intended.

In3luki: I've looked at the rest of the file and this is intended after all.

src/module/user/sheet.ts
CarlosFdez: Sure. If you got any idea for what would be good font awesome choices as well I'm all ears.

In3luki: `fa-user` for the first tab and maybe `fa-dice` (like the systems tab in the setup screen) for the other?

CarlosFdez: Done! Good choice.

PR #14537, opened 2024-04-27
Restore darkvision vision mode

CarlosFdez:


types/foundry/client/pixi/perception/vision-mode.d.ts
```diff
@@ -1,60 +1,24 @@
 import type { PointVisionSource } from "../../../client-esm/canvas/sources/module.d.ts";
+import type * as fields from "../../../common/data/fields.d.ts";
 
 declare global {
+    class ShaderField extends foundry.data.fields.DataField {
+        override _cast(value: unknown): unknown;
+    }
```
In3luki, line 7: It's weird that this isn't in `common/data/fields.mjs` but that is something to bring up with core.

CarlosFdez: It is a bit weird, but I'm guessing they didn't wanna expose it just yet. Its a pretty specific thing.

PR #14550, opened 2024-04-28
Fix combat tracker config and `actorsDeadAtZero` option

In3luki:


src/module/actor/base.ts
```diff
@@ -1153,7 +1153,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
                 (this.isOfType("character") && setting === "both" && !!instantDeath);
 
             if (
-                updated?.isDead &&
+                updated?.hitPoints?.value === 0 &&
```
CarlosFdez, line 1156: I think isDead is necessary because of things like regeneration? Was there a check for that in isDead before?

In3luki:
```ts
    get isDead(): boolean {
        const deathIcon = game.settings.get("pf2e", "deathIcon");
        if (this.token) return this.token.overlayEffect === deathIcon;
        const tokens = this.getActiveTokens(true, true);
        return tokens.length > 0 && tokens.every((t) => t.overlayEffect === deathIcon);
    }
```
It's just checking for the overlay icon. I'll recheck the old logic but I don't think it's possible to to that anymore. That said my brain is fried from too many `selectOptions` so maybe I'm wrong here.

CarlosFdez: I'll test on v11 later to be sure. But I think I remember that in some cases the death icon wasn't added automatically.

In3luki: I figured it out and reverted the change. The HP check is already happening in `CreaturePF2e#isDead`. It didn't work because I tested it with a character instead of an NPC. Characters only get marked as dead if they take fatal damage.

PR #14551, opened 2024-04-28
Fix `DamageDice` with critical set to `false` being included in critical damage

In3luki:
> True means the dice are added to critical without doubling; false means the dice are never added to critical damage; omitted means add to normal damage and double on critical damage.

The image shows: true, false, null

image

### DIFF
--- a/src/module/system/damage/formula.ts
+++ b/src/module/system/damage/formula.ts
@@ -93,7 +93,7 @@ function createDamageFormula(
     const BONUS_BASE_LABELS = ["PF2E.ConditionTypePersistent"].map((l) => game.i18n.localize(l));
 
     // Test that a damage modifier or dice partial is compatible with the prior check result
-    const outcomeMatches = (m: { critical: boolean | null }): boolean => critical || m.critical !== true;
+    const outcomeMatches = (m: { critical: boolean | null }): boolean => m.critical === null || m.critical === critical;
 
     // Add damage dice. Dice always stack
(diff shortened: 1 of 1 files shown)


src/module/system/damage/formula.ts
```diff
@@ -93,7 +93,10 @@ function createDamageFormula(
     const BONUS_BASE_LABELS = ["PF2E.ConditionTypePersistent"].map((l) => game.i18n.localize(l));
 
     // Test that a damage modifier or dice partial is compatible with the prior check result
-    const outcomeMatches = (m: { critical: boolean | null }): boolean => critical || m.critical !== true;
+    const outcomeMatches = (m: { critical: boolean | null }): boolean => {
+        if (critical && m.critical === false) return false;
+        return critical || m.critical !== true;
```
CarlosFdez, line 98: Hmm, so I ran a truth table on it mentally, and if this truth table is correct, we might need more comments on this section. It doesn't seem like outcomeMatches observes the overall rule of the critical prop (m.critical=true and critical = false returns false). I'm guessing its because doubling from null is achieved by getting true twice.

Would it be cleaner then if instead the condition was:
`m.critical === null || (m.critical === critical)`

| m.critical | critical |  PR   |
|----------|--------|------|
| null        | false    | true  |
| null        | true    | true   |
| true       | false   | false   |
| true       | true   | true   |
| false      | false   | true   |
| false      | true    | false  |

In3luki: Yeah, that is much cleaner. Updated.

PR #14560, opened 2024-04-29
Refactor `ChatMessagePF2e` flags

In3luki:
Deprecations:
`ChatMessagePF2e#flags#pf2e#appliedDamage` moved to `ChatMessagePF2e#system#appliedDamage`
`ChatMessagePF2e#flags#pf2e#origin` moved to `ChatMessagePF2e#system#origin` and refactored to:
```ts
interface ChatMessageOriginData {
    actor: ActorUUID | TokenDocumentUUID | null;
    token?: TokenDocumentUUID;
    item?: ItemUUID;
    rollOptions?: string[];
    /** Is the origin the roller? */
    self?: boolean;
}
```

`ChatMessagePF2e#flags#pf2e#context#target` refactored to:
```ts
interface ChatMessageTargetData {
    actor?: ActorUUID | TokenDocumentUUID | null;
    token?: TokenDocumentUUID;
    item?: ItemUUID;
    distance: number | null;
    rangeIncrement: number | null;
}
```
`ChatMessagePF2e#flags#pf2e#context` moved to `ChatMessagePF2e#system#context` and partially refactored.

~~I've opted to move origin to top level system data to avoid having to add bogus `origin`s to `context` data that has no origin.~~
```ts
interface ChatMessageSystemData {
    /** A record of applied damage that can be undone */
    appliedDamage?: AppliedDamageData | null;
    /** Message context data that describes the type of this chat message */
    context?: ChatMessageContext;
}
```

Old messages still work thanks to `DataModel#migrateData`.

### DIFF
--- a/src/module/actor/roll-context/damage.ts
+++ b/src/module/actor/roll-context/damage.ts
@@ -1,7 +1,7 @@
 import type { ActorPF2e } from "@actor";
 import type { StrikeData } from "@actor/data/base.ts";
 import type { ItemPF2e } from "@item";
-import type { CheckContextChatFlag } from "@module/chat-message/data.ts";
+import type { CheckContextChatData } from "@module/chat-message/data.ts";
 import { CheckRoll } from "@system/check/roll.ts";
 import type { Statistic } from "@system/statistic/statistic.ts";
 import { sluggify } from "@util";
@@ -26,7 +26,7 @@ class DamageContext<
         }
     }
 
-    #findMatchingCheckContext(): CheckContextChatFlag | null {
+    #findMatchingCheckContext(): CheckContextChatData | null {
         const { origin, target } = this.unresolved;
         const item = origin?.item;
         if (this.viewOnly || !origin?.actor || !item?.isOfType("melee", "weapon") || !target?.token) {
@@ -54,7 +54,7 @@ class DamageContext<
                 );
             });
 
-        return (checkMessage?.flags.pf2e.context ?? null) as CheckContextChatFlag | null;
+        return (checkMessage?.system.context ?? null) as CheckContextChatData | null;
     }
 }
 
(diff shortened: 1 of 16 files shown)


CarlosFdez:
Why is this v12? These are pf2e flags.

In3luki:
`ChatMessage`s don't have system data in v11.

CarlosFdez:
I double checked, and at least as far as context data is concerned, I suppose this is symmetrical with effects.

CarlosFdez:
Looking deeper, I think being able to move them is good, but I don't think it can be 1-1. Some of these properties should probably be nested (maybe origin should be in context, unsure) or still contained in flags (journalEntry?).

I'll do a quick runthrough of v12 to see if we're missing anything, but otherwise I'll try to think of a new structure.

CarlosFdez:
So off the top of my head, very quickly (do not act on this, I'm currently speculating, I have very little knowledge of the chat pipeline):

**modifiers**, **dice**, **origin** (which is really just the origin's item) and **casting** should probably merged into **context**.

**appliedDamage** can stay where it is, its fine.

I have no idea about **journalEntry**. Maybe its context.

preformatted and suppressDamageButtons should probably remain as flags. Maybe. I'm super uncertain about this, but maybe suppressDamageButtons can be moved somewhere.

CarlosFdez:
I went over the code, and I'm kinda torn between one of these two general structures. I'd like some more opinions on what sort of approach we should go for before doing we do anything

https://gist.github.com/CarlosFdez/6269e76ab8932d77558a9783dd9a5c44

I think because this is V12, we don't need to shim the old code. Migration only should suffice.

In3luki:
I would prefer Choice A. Having the source item information, like the variant and rank, under `item` makes more sense than dumping them directly into origin IMO.

CarlosFdez:
This got absolutely demolished by the rebase, but it already had some merge conflicts anyways so it might be worth redoing.

PR #14563, opened 2024-04-30
Fix typo in Artisan's Toolkit description

Valinora:
Caught this while playing last night. Not sure if this is large enough to really merit a PR on it's own, but here it is nonetheless.


TikaelSol:
Changed the title to be in imperative case

PR #14570, opened 2024-05-01
Fix token config `Infinity` warning

In3luki:
I think `maxR` should be equivalent to `Infinity`.

Also removes override for an issue that was fixed in build 320.

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -273,7 +273,6 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         this.sight.visionMode = visionMode;
 
         const visionModeDefaults = CONFIG.Canvas.visionModes[visionMode].vision.defaults;
-        const maxRadius = scene.dimensions.maxR;
         this.sight.brightness = visionModeDefaults.brightness ?? 0;
         this.sight.saturation = visionModeDefaults.saturation ?? 0;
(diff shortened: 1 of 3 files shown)


src/module/scene/token-document/document.ts
```diff
@@ -272,12 +273,13 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         this.sight.visionMode = visionMode;
 
         const visionModeDefaults = CONFIG.Canvas.visionModes[visionMode].vision.defaults;
+        const maxRadius = scene.dimensions.maxR;
```
CarlosFdez, line 276:
```suggestion
```

src/module/scene/token-document/document.ts
```diff
@@ -286,7 +288,6 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
             }
         }
 
-        const maxRadius = scene.dimensions.maxR;
         if (actor.perception.senses.has("see-invisibility")) {
```
CarlosFdez, line 291:
```suggestion
        const maxRadius = scene.dimensions.maxR;
        if (actor.perception.senses.has("see-invisibility")) {
```

CarlosFdez (commented):
Don't really need to move that property anymore if we don't need it earlier. That way it can stick close to where its reference.

PR #14577, opened 2024-05-02
Fix ignored choicesets not disabling other rule elements

CarlosFdez:
Fixes issues like not selecting a champion key score bricking an actor.

### DIFF
--- a/src/module/rules/rule-element/choice-set/rule-element.ts
+++ b/src/module/rules/rule-element/choice-set/rule-element.ts
@@ -64,20 +64,24 @@ class ChoiceSetRuleElement extends RuleElementPF2e<ChoiceSetSchema> {
             if (this.choices.unarmedAttacks) this.choices.predicate.push("item:category:unarmed");
         }
 
-        // Assign the selection to a flag on the parent item so that it may be referenced by other rules elements on
-        // the same item. If a roll option is specified, assign that as well.
-        this.item.flags.pf2e.rulesSelections[this.flag] = this.selection;
-        if (this.actorFlag) {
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/choice-set/rule-element.ts
```diff
@@ -64,21 +64,24 @@ class ChoiceSetRuleElement extends RuleElementPF2e<ChoiceSetSchema> {
...
-        if (this.actorFlag) {
-            this.actor.flags.pf2e[this.flag] = this.selection;
-        }
-
-        if (this.selection !== null) {
+        if (!this.ignored && this.selection !== null) {
             this.#setRollOption(this.selection);
         } else if (!this.allowNoSelection && this.test()) {
-            // Disable this and all other rule elements on the item until a selection is made
             this.ignored = true;
```
In3luki, line 70: You could test if `this.ignored` is already `true` in the else if clause but I think it's fine either way.

In3luki (approved):

src/module/rules/rule-element/choice-set/rule-element.ts
CarlosFdez: It could be written a bit better, but I plan to be a bit  careful. I could instead add !this.ignored and this.selection === null to the second branch, and then move setRollOption() to later after we check that its not ignored....perhaps its worth doing since the control flow will be easier to follow.

CarlosFdez: Done.

In3luki (approved):

PR #14591, opened 2024-05-03
Fix `TokenLayerPF2e#cycleStack`

In3luki:
Also fixes a small error in the status effects HUD. `token.actor.isDead` always returns `true` for creatures at 0 hp so we need to check for the status.


CarlosFdez:
I recall isDead and the status effect should be 1-1, but I'd have to double check. But even if it was 1-1, checking the (foundry) status effects in the status effects section is the correct way to go about it anyways.

Issue #14587, opened 2024-05-03
Duplicate of a token no longer off-guards the target when flanking with origin token


Intervencion:
See mirror thaum flanking with themselves. It used to provide flanking bonuses but now it doesn't

(Thanks to arkeoo on Discord for reporting this issue)

CarlosFdez:
I distinctly remember conversation about this needing to be supported for mirror thaum, and I do recall it working at some point, so I think I can reasonably assume its a bug and not working as intended.

CarlosFdez:
Seems this behavior changed a long time ago: https://github.com/CarlosFdez/pf2e/commit/58379d2f59ad216d29d11734047ad02b9ab3f951 the foundry system used to allow you to be your own ally.

I need to add a different check to flanking code to handle this scenario.

PR #14597, opened 2024-05-04
Allow duplicates of a linked actor to flank amongst themselves

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/14587

This stopped working a year ago when the ally check changed. The ally check change is correct (it was changed for the purposes of auras, but you can't be your own ally in pf2e) but flanking code wasn't changed in response.


In3luki (dismissed):

CarlosFdez:
In retrospect, I think being a bit more paranoid and comparing their ids instead of the instances themselves might be safer. So I changed it.

In3luki (approved):

PR #14600, opened 2024-05-04
Revert setting `Document#_stats#coreVersion` in the pack builder

In3luki:
Maybe this should only be added to the packs in development builds?

### DIFF
--- a/build/lib/compendium-pack.ts
+++ b/build/lib/compendium-pack.ts
@@ -8,8 +8,6 @@ import { RuleElementSource } from "@module/rules/index.ts";
 import { isObject, recursiveReplaceString, setHasElement, sluggify, tupleHasValue } from "@util/misc.ts";
 import fs from "fs";
 import path from "path";
-import type { JournalEntryPageSource } from "types/foundry/common/documents/journal-entry-page.ts";
-import type { JournalEntrySource } from "types/foundry/common/documents/journal-entry.ts";
 import coreIconsJSON from "../core-icons.json" assert { type: "json" };
 import "./foundry-utils.ts";
 import { getFilesRecursively, PackError } from "./helpers.ts";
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Probably so. It might be good to pull from system.json as well if we can (the verified version? The min version? Might need to be the min version)

In3luki:
The `json` arg is only used for releases so this should be safe.
The only caveat is that now every system release will trigger a pack migration for the user.

CarlosFdez:
Wouldn't that already happen because every system reason replaces all of the database files?

In3luki:
The trigger for a server-side migration is `_stats.coreVersion` being lower than the current foundry build version or missing.
As we wouldn't include `_stats` for production builds after this change, it would always run for new release builds.

In3luki:
It's a bit of a mess to manually keep track of supported server-side migrations as it involves unminifying a core file to look for changes in a `_migrationRegistry` property.

In3luki:
I think it would actually be better to not strip the value on extract instead of adding it back on build.
Then I'd be:
- Build packs
- Start server and let it run migrations
- Extract packs with updated `_stats.coreVersion`
- Done

CarlosFdez:
Wouldn't it also migrate for dev builds as well if set to the min version and the dev is currently not running the min version?

Could be that there's really no good way around it. It could be that the slow load time is mostly console log noise. Unless of course its possible for foundry to load a db for a core version that's greater than it?

Not stripping the value on extract would lead it to be based on who contributed, which is going to not really be that consistent.

In3luki:
I tried commenting out the logs but it didn't do much. It's really just the system having a ton of documents to traverse.

CarlosFdez:
I meant foundry server logs primarily. I don't recall it being nearly that slow in v11, but in v11 I run the windows application. If that is what you meant by commenting out, I suppose not much can be done for it.

CarlosFdez:
If foundry can load documents of later foundry versions, then setting it to the verified version is what I think it should do. Otherwise perhaps not including it at all is the correct play.

In3luki:
Reverted back to v11 state.

PR #14603, opened 2024-05-04
Normalize search queries with diacritics in compendium sidebar and browser

### LINKED ISSUE #14601: [Localization QOL request] Strip accent codepoints when searching the Compendium Browser
I cannot speak for other languages, but the strictness of Compendium Browser search causes issues when implementing grammatically correct Russian localization.
In Russian, letter 'ё' is rarely used online (likely because it's in a really inconvenient location on a keyboard) and is commonly replaced with 'e' (which is a different letter entirely). This causes an issue where if we are to localize, for example, "spear" as "копьё" (grammatically correct), people searching for "копье" won't find it.
Core Foundry works around it by lowering the text and stripping accent codepoints (both in the needle and in the haystack) by doing `query.trim().normalize("NFD").replace(/[\u0300-\u036f]/g, "")` in `SearchFilter.cleanQuery`, which works well enough on the right-bar tabs (except, coincidentally, the compendium tab, but I've already reported that issue on Foundry Discord).

In3luki:
I'm not sure if this is a bugfix or an enhancement.

image

image

Closes #14601

### DIFF
--- a/src/module/apps/compendium-browser/tabs/base.ts
+++ b/src/module/apps/compendium-browser/tabs/base.ts
@@ -70,7 +70,9 @@ export abstract class CompendiumBrowserTab {
                     return null;
                 }
                 return Array.from(wordSegmenter.segment(term))
-                    .map((t) => t.segment.toLocaleLowerCase(game.i18n.lang).replace(/['"]/g, ""))
+                    .map((t) =>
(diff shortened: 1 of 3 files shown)


CarlosFdez:
Me neither, but I'm leaning enhancement!

PR #14646, opened 2024-05-07
v20230506 update - added RE to Strix Flight feats

ammalagonc:
3rd attempt.

- Made approved changes to the Juvenile Flight feat.
- Made changes to the Juvenile Flight feat effects so they won't stack with Fully Flighted.
- Added RE to Fully Flighted feat


packs/feats/fledgling-flight.json
```diff
@@ -29,7 +29,33 @@
...
+                    },
+                    {
+                        "not": "feat:fully-flighted"
+                    }
+                ],
+>>>>>>> Stashed changes
+                "toggleable": true
+            },
+            {
+                "key": "BaseSpeed",
```
TikaelSol, line 51: This is a merge conflict.

This is also deleting several system files that shouldn't be touched, and the familiar focus familiar ability.

It will be best at this point to hard reset your branch and start again, pushing only the files you are actually changing. To hard reset do

```
git checkout master
git reset --hard upstream/master
```
If you can't do that hard reset then your upstream isn't set, in which case run `git remote add upstream https://github.com/foundryvtt/pf2e.git`

ammalagonc:
Redid the pull per Tikael's instructions. 

Hopefully it's right this time.

PR #14648, opened 2024-05-07
Move inline link handlers to separate methods

CarlosFdez:
This is preparation for eventually handling the listeners globally. This only moves them, so that future changes are easier to see.

Done in V12 since master is mostly no touchy for now.


In3luki:
I've seen the tag more as a "targets the v12 branch" indicator.

PR #14649, opened 2024-05-07
Fix highlighting for cone and custom circle templates

CarlosFdez:
It doesn't fix the templates center-point not snapping properly (which is apparently on purpose for smoother behavior), but it does fix the highlight when placed at a corner for a circle template created directly from canvas controls. It also restores mid edge placements for cones going in a lateral direction.

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -1,4 +1,3 @@
-import type { EffectAreaShape } from "@item/spell/types.ts";
 import type { MeasuredTemplatePF2e } from "../measured-template.ts";
 
 export class TemplateLayerPF2e<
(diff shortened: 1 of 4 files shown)


src/module/canvas/layer/template.ts
```diff
@@ -15,11 +15,11 @@ export class TemplateLayerPF2e<
...
     /** Set a grid-snapping mode appropriate for an effect area type */
     snapFor(areaShape: EffectAreaShape | null): void {
+        const M = CONST.GRID_SNAPPING_MODES;
         if (areaShape && canvas.grid.type === CONST.GRID_TYPES.SQUARE) {
-            this.#snappingMode =
-                areaShape === "burst" ? CONST.GRID_SNAPPING_MODES.CORNER : CONST.GRID_SNAPPING_MODES.CENTER;
+            this.#snappingMode = areaShape === "burst" ? M.CORNER : M.CENTER;
         } else {
-            this.#snappingMode = CONST.GRID_SNAPPING_MODES.CENTER;
+            this.#snappingMode = M.CENTER | M.CORNER;
```
In3luki, line 22: That doesn't look right.

CarlosFdez: They're bitwise flags,that's how you say "center or corner".

CarlosFdez: If you look at placeables-layers.js at getSnappedPoint() you'll see them being combined in that way and I was like "oh, that's how you're supposed to use it...."

In3luki: Oh, yeah. That's a thing.

In3luki (approved):

PR #14670, opened 2024-05-10
Update character _preUpdate logic for v12

CarlosFdez:
Fixes the unnecessary cloning

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -1841,10 +1841,14 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         if (!changed.system) return super._preUpdate(changed, options, user);
 
         // Clamp level, allowing for level-0 variant rule and enough room for homebrew "mythical" campaigns
-        if (changed.system.details?.level || changed.system.build?.attributes) {
+        const newLevel = changed.system.details?.level?.value ?? this.level;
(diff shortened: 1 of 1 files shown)


src/module/actor/character/document.ts
```diff
@@ -1841,10 +1841,14 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
-        if (changed.system.details?.level || changed.system.build?.attributes) {
+        const newLevel = changed.system.details?.level?.value ?? this.level;
+        const attributeChanged =
+            !!changed.system.build?.attributes &&
+            !R.isEmpty(fu.diffObject(this._source.system.build?.attributes ?? {}, changed.system.build.attributes));
+        if (newLevel !== this.level || attributeChanged) {
             const level = changed.system.details?.level;
-            if (typeof level?.value === "number") {
-                level.value = Math.clamp(Number(level.value) || 0, 0, 30) || 0;
+            if (level) {
```
In3luki, line 1850: This doesn't handle level 0

CarlosFdez: good catch

CarlosFdez: Wait no, it does, level is an object. level.value is what would be level 0.

In3luki: oh true

PR #14671, opened 2024-05-10
Add PF2e menu to prosemirror editor

In3luki:
This removes `protected` from two methods in `ProseMirrorMenu` that are needed to make the menu work. The alternative is subclassing it, which also would make the hook unnecessary.

Using `{ _preserve: { class: "action-glyph" } }` is currently the only way to add a class attribute to the prosemirror elements. It uses an internal property added by Foundry and circumvents prosemirror schema validation. This is not officially part of the API and might break without notice.

image

The menu could also be added as a top-level dropdown:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,6 +55,7 @@
                 "pixi.js": "7.2.4",
                 "prettier": "3.2.5",
                 "prompts": "^2.4.2",
+                "prosemirror-commands": "^1.5.2",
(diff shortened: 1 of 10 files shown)


CarlosFdez:
What's the likelyhood of it working well if we make the icons a font instead?

CarlosFdez:
We can handle it separately if we need to do that. I like it being in format, since it'll be less likely to wrap in certain smaller usages.

PR #14674, opened 2024-05-10
Add automation to Pitborn

TikaelSol:


SpartanCPA:
Do we want to do this manually like this or make a dynamic list to future-proof against additional "Trained in Athletics" level 1 skill feats? 

Do we even have the data structure of pre-reqs ablt to do that?

TikaelSol:
We don't have anything for that, so manual it is.

PR #14678, opened 2024-05-11
Update Thaumaturge Features

mysurvive:
This adds the `thaumaturge-implement-initiate` otherTag. 

While not inherently or necessarily important for the Thaumaturge class, it allows implement features that have grantItem features the ability to predicate on the tag - this means that the Thaumaturge Dedication doesn't grant these actions unless they also take the Implement Initiate feat.

The Thaumaturge Dedication is also given a choiceSet to choose an implement, and the `thaumaturge-implement-initiate` otherTag is granted only when the Implement Initiate feat is granted.

### DIFF
--- a/packs/classfeatures/amulet.json
+++ b/packs/classfeatures/amulet.json
@@ -27,6 +27,10 @@
         "rules": [
             {
                 "key": "GrantItem",
+                "predicate": [
+                    "parent:tag:thaumaturge-implement-initiate"
+                ],
+                "reevaluateOnUpdate": true,
                 "uuid": "Compendium.pf2e.actionspf2e.Item.Amulet's Abeyance"
             }
         ],
--- a/packs/classfeatures/bell.json
+++ b/packs/classfeatures/bell.json
@@ -27,6 +27,10 @@
         "rules": [
             {
                 "key": "GrantItem",
+                "predicate": [
+                    "parent:tag:thaumaturge-implement-initiate"
+                ],
+                "reevaluateOnUpdate": true,
                 "uuid": "Compendium.pf2e.actionspf2e.Item.Ring Bell"
             }
         ],
--- a/packs/classfeatures/chalice.json
+++ b/packs/classfeatures/chalice.json
@@ -27,6 +27,10 @@
         "rules": [
             {
                 "key": "GrantItem",
+                "predicate": [
+                    "parent:tag:thaumaturge-implement-initiate"
+                ],
+                "reevaluateOnUpdate": true,
                 "uuid": "Compendium.pf2e.actionspf2e.Item.Drink from the Chalice"
             }
         ],
--- a/packs/classfeatures/first-implement-and-esoterica.json
+++ b/packs/classfeatures/first-implement-and-esoterica.json
@@ -37,16 +37,17 @@
                 "prompt": "PF2E.SpecificRule.Prompt.FirstImplement"
             },
             {
+                "alterations": [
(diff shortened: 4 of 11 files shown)


TikaelSol:
This one is going to take a bit longer to review.

xyzzy42:
This seems reasonable to me.  The Thaumaturge class features already have an implement selection choiceset and grantitem.  The archetype also grants an implement, so it seems like it should have the choiceset and grantitem implement too.

But the archetype doesn't give the same things at first, so the flag and predicate are needed to avoid giving actions until the archetype Implement Initiate feat is taken.

The only thing I thought of was the tag could be reversed, so it's `thaumaturge-implement-archetype`. All the predicates would be inverted from what they are now.  This way the ItemAlteration on the first/second/third implement features could go away, since no tag is the default case and it only needs to be added for archetype + not feat:implement-initiate.

TikaelSol:
So this looks good but we don't have many v11 releases left and don't want to have to chase down anything this may cause issues with going forward, so we'll delay merging until after the last v11 release next week. Then we can get it in and ready for the non-beta v12 release where we could fix anything that needs fixing.

xyzzy42:
So there wouldn't be a v11 release with this change?  I guess if the Thaum Vuln module wants to become v12-only and depend on this change or remain v11 compatible and would around it would be something to consider.

mysurvive:
I think it's fine and makes sense to push this to v12, especially since existing thaums/diet thaums will need to level dance/refresh feats to apply the new rules. I'll also let Tik decide whether it makes more sense to merge this as-is or changing to Trent's suggestions. Not having an `initiate` roll option always slightly triggered me, but that's more of a personal issue than a mechanical one.

TikaelSol:
I still haven't had time to actually look through and figure out what this is doing or even how thaum works beyond what I half remember (not much at this point, brain is completely fried). So mostly it's down to me just being unable to dedicate time for a while. If one of the other devs wants to hit merge feel free to.

mysurvive:
There is actually a bug with this that I just found. It looks like the grantItem rule elements on the implements don't apply until the next update when there is a predicate in place. I'm not sure if this is something that can be worked around, but I'm looking into it.

xyzzy42:
Maybe the ItemAlteration REs are running after the GrantItem REs?

mysurvive:
I thought so too, so I changed the dedication GrantItem predicate to `["feat:implement-initiate"]` and it still didn't update until the next cycle _after_ the feat was applied.

mysurvive:
Actually, it does work if you close the sheet, then re-open it. I guess the sheet re-renders weird.

mysurvive:
This will require a change which allows ItemAlteration to run during preCreate, per Supe (which he has no bandwidth to work on until sometime after the v12 release). Currently, everything works hunky-dory except the Dedication, which will require closing/re-opening the sheet once the Implement Initiate feat is granted.

mysurvive:
Implement Initiate has been given the Item Alteration rule instead of directly on the feat per some recommendations in discord (https://discord.com/channels/880968862240239708/880969304365994034/1241244156635119696, https://discord.com/channels/880968862240239708/880969304365994034/1241244150230290513). This may or may not be acceptable since it adds another actor flag from the dedication feat.

This will work better (won't need to close/open the character sheet) if a trigger is added to run re-evaluates when feats are inserted (per Supe https://discord.com/channels/880968862240239708/880969304365994034/1241255907422109787)

PR #14690, opened 2024-05-12
Improve prosemirror dropdown menu

In3luki:
- Create `ProseMirrorMenuPF2e` subclass and remove hook
- Ensure the editor is refocused after a command is executed
- Override `ProseMirrorMenu#_isMarkActive` to work with `attrs: { _preserve }` values
- Give PF2e menu items a priority of 1 when they are active to ensure they are shown as active in the closed dropdown
- Override `FormApplication#_configureProseMirrorPlugins` to add the PF2e menu in all base sheet classes.


CarlosFdez:
It would be nice if this was the behavior for all the buttons in the editor. Might be worth looking at that, but I'm inclined to just merge this tomorrow if you can't figure it out or don't have the time to figure it out, this is more than good enough to use.

I took a very quick glance at the issue tracker and I'm surprised no one asked about this. I'll look again later once I'm more awake.

In3luki:
Looks like everything is working now. I've updated the description.

PR #14697, opened 2024-05-13
Add automation for the Mechanical Symbiosis background

LJSmith066:
Added a choice set for all skills other than Arcana, added a toggleable roll option for the +1 circumstance bonus to said skill, and added notes with linked Stunned x conditions for failure and crit failure of checks for selected skill.

### DIFF
--- a/packs/backgrounds/mechanical-symbiosis.json
+++ b/packs/backgrounds/mechanical-symbiosis.json
@@ -129,8 +129,7 @@
                     "entity-is-aiding"
                 ],
                 "selector": "{item|flags.pf2e.rulesSelections.entitySkill}",
-                "text": "You become @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 1}",
-                "title": "Mechanical Symbiosis (Failure):"
+                "text": "PF2E.SpecificRule.MechanicalSymbiosis.FailureNote"
             },
             {
                 "key": "Note",
@@ -141,8 +140,7 @@
                     "entity-is-aiding"
                 ],
                 "selector": "{item|flags.pf2e.rulesSelections.entitySkill}",
-                "text": "You become @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 2}",
-                "title": "Mechanical Symbiosis (Critical Failure)"
+                "text": "PF2E.SpecificRule.MechanicalSymbiosis.CriticalFailureNote"
             }
         ],
         "trainedLore": "",
--- a/packs/backgrounds/mechanical-symbiosis.json
+++ b/packs/backgrounds/mechanical-symbiosis.json
@@ -96,7 +96,7 @@
                 ],
                 "flag": "entitySkill",
                 "key": "ChoiceSet",
-                "prompt": "GM-Selected Skill"
+                "prompt": "PF2E.SpecificRule.Prompt.Skill"
             },
             {
                 "key": "ActiveEffectLike",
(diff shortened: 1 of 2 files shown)


packs/backgrounds/mechanical-symbiosis.json
```diff
@@ -30,7 +30,121 @@
...
+                        "value": "survival"
+                    },
+                    {
+                        "label": "PF2E.SkillThi",
+                        "value": "thievery"
+                    }
+                ],
+                "flag": "entitySkill",
+                "key": "ChoiceSet",
+                "prompt": "GM-Selected Skill"
```
TikaelSol, line 99:
```suggestion
                "prompt": "PF2E.SpecificRule.Prompt.Skill"
```
We want to localize strings in rule elements as much as possible, and this can just use the generic "Select a skill" localization.

packs/backgrounds/mechanical-symbiosis.json
```diff
@@ -30,7 +30,121 @@
...
+                "key": "Note",
+                "outcome": [
+                    "failure"
+                ],
+                "predicate": [
+                    "entity-is-aiding"
+                ],
+                "selector": "{item|flags.pf2e.rulesSelections.entitySkill}",
+                "text": "You become @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 1}",
+                "title": "Mechanical Symbiosis (Failure):"
```
TikaelSol, line 133:
```suggestion
                "text": "PF2E.SpecificRule.MechanicalSymbiosis.FailureNote"
```
We need to localize this. To do this open up `/static/lang/re-en.json` and go under SpecificRule. Make a new object for `MechanicalSymbiosis` (be sure to keep the file alphabetized) and put entries in for the text used in the rules here. We don't need specific titles for these notes though, that can just use the item name as normal.

packs/backgrounds/mechanical-symbiosis.json
```diff
@@ -30,7 +30,121 @@
...
+                "key": "Note",
+                "outcome": [
+                    "criticalFailure"
+                ],
+                "predicate": [
+                    "entity-is-aiding"
+                ],
+                "selector": "{item|flags.pf2e.rulesSelections.entitySkill}",
+                "text": "You become @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 2}",
+                "title": "Mechanical Symbiosis (Critical Failure)"
```
TikaelSol, line 145:
```suggestion
                "text": "PF2E.SpecificRule.MechanicalSymbiosis.CriticalFailureNote"
```
As above

packs/backgrounds/mechanical-symbiosis.json
LJSmith066: One thing I'd like to keep labels for is since the majority of the time you're not rolling skill checks against a given DC, so both the failure and the critical failure notes will show up regardless of the actual degree of success, which means you get handed "You become Stunned 1" and "You become Stunned 2" without any context - is there any other solution for this?

TikaelSol: It will have a label from the item name automatically, pointing back to the source.

LJSmith066: Yes, what I'm saying is that 99% of the time the results will look like this: 
image

since normally skill checks will be rolled without a targeted DC, so the check usually won't know whether it's a success/failure, so it prints all notes regardless of success state.

PR #14700, opened 2024-05-13
Restore blank select option to weapon sheet range dropdown

### LINKED ISSUE #14698: [BETA] Reach weapons being considered ranged weapons
version:
image

Bring the weapon from the compendium it works fine
image

Click in edit the "range 10ft appear"
image

After close it without change anything:
image

In3luki:
Closes #14698

### DIFF
--- a/static/templates/items/weapon-details.hbs
+++ b/static/templates/items/weapon-details.hbs
@@ -27,8 +27,8 @@
             {{/if}}
         </label>
         <select name="system.range" id="{{fieldIdPrefix}}range" data-dtype="Number" {{disabled mandatoryMelee}}>
-            {{#if mandatoryMelee}}
-                <option value="">{{localize "PF2E.Item.Weapon.NoRangeMelee"}}</option>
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I don't think this fixes the bug, it seems like this is a deviation from behavior still. I brought up the original select and this is what it was:
```hbs
{{#select item._source.system.range}}
    {{#unless mandatoryRanged}}<option value="">{{localize "PF2E.Item.Weapon.NoRangeMelee"}}</option>{{/unless}}
    {{#each weaponRanges as |label range|}}
        <option value="{{range}}">{{localize label range=range}}</option>
    {{/each}}
{{/select}}
```

It seems that mandatoryRanged weapons should not have a blank entry. I suspect what isn't actually working is the disabled on the entire select.

In3luki:
I'm in a game right now and probably shouldn't have done this right before. I think this is correct now?

CarlosFdez:
Looks correct!

PR #14706, opened 2024-05-14
Fix Hag Bloodline addendum parsing as markdown URL

### LINKED ISSUE #14660: Hag bloodline blood magic note has broken damage roll
grafik

as you can see there is an @ to much in the textbox for the hag bloodline.

jfn4th:
Resolves #14660 

This adds an escape character to the localization string so the Damage link followed by a parenthesis enclosed Check isn't interpreted as a markdown URL.

Then, I noticed sending the item to chat multiple times without closing the character sheet, or opening the item description and _then_ sending to chat would result in the links being malformed again.

The first time `#sanitizeMarkdown` runs for this item, `text` would be the key of the localization string and the escape would occur as expected. On subsequent runs, `text` contains `#sanitizeMarkdown`'s previous result for the item and would then be interpreted as a markdown url as it no longer contains the escape. Added a property to skip sanitization if already run for the item.


CarlosFdez:
Given the idea is to switch to markdown, the approach I think we are going to want to go with is to detect what areas are `@Check` and `@Damage` and omit them from markdown processing in one way or another, perhaps via the creation of special placeholder tags or some other approach. But regardless of our approach, V11 is on code freeze for everything except simple easy bugfixes, and this is a project that we can only reasonably support on the V12 branch.

It might be worth digging to see what foundry markdown sheet does, though it doesn't support our @ enriches, only inline rolls.

jfn4th:
Gotcha. In that case, I'll update this PR to specifically fix the bloodline issue by putting the word "damage" in between `@Damage[]` and `(@Check)`. The word should probably be there anyway, and I see some other items handle it that way.

PR #14709, opened 2024-05-14
Continue to improve prosemirror dropdown menu

In3luki:
- Make `Font` dropdown a child of `Format` dropdown
- Fix toggling of inline header
- Reduce menu item gap by 1px

This does not fix toggling of the action glyphs and inline gm text. The problem  there is that it's using a `ProseMirror.Mark` instead of a `ProseMirror.Node` because the foundry schema has no `span` `Node` only a `Mark`.
`toggleMark` only toggles a selection so to remove the span it is necessary to select the text that should be toggled instead of just toggling it off with the menu.

image


In3luki:
Now uses the core `compact` option where necessary instead of moving the fonts in the override.

Also the PF2e dropdown is styled now:
image

PR #14717, opened 2024-05-14
Convert all skills from short form to long form

CarlosFdez:


stwlam:
u crazy

CarlosFdez:
That is besides the point!

CarlosFdez:
(but its gonna stay in draft until some point later)

PR #14723, opened 2024-05-15
Assign default skill ranks in data prep instead of template

CarlosFdez:


stwlam:
For anything that i has a high chance of being written to disk, it makes sense to have those as defaults to save CPU overhead

CarlosFdez:
Talked separately, but the performance impact is nearly imperceptible (this time) and this might enable future work.

PR #14728, opened 2024-05-16
Start filling out scene region types

In3luki:
I've never added a new document type before so I've tried to model if after `Actor` and `Token` and it looks like nothing is blowing up so far. Adding the `Clipper` types might be unnecessary, I'm not sure if the placeable `Region` needs to be typed so accurately.


stwlam:
Mind pulling out the types in a second PR so we can get those merged in sooner?

In3luki:
I can do that but I'm not really sure which types you mean.

types/foundry/client-esm/canvas/regions/polygon-tree.d.ts
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * The Clipper path of this node.
+     * It is empty in case of the root node.
+     */
+    get clipperPath(): readonly ClipperLib.IntPoint[] | null;
+
+    /**
+     * The polygon of this node.
+     * It is `null` in case of the root node.
+     * @type {PIXI.Polygon|null}
```
CarlosFdez, line 50:
```suggestion
```

types/foundry/client-esm/canvas/regions/polygon-tree.d.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Test circle containment/intersection with this node.
+     * @param {Point} center     The center point of the circle.
+     * @param {number} radius    The radius of the circle.
+     * @returns {-1|0|1}          - -1: the circle is in the exterior and does not intersect the boundary.
+     *                            - 0: the circle is intersects the boundary.
+     *                            - 1: the circle is in the interior and does not intersect the boundary.
+     */
+    testCircle(center: Point, radius: number): -1 | 0 | 1;
```
CarlosFdez, line 83: The {Point} and {number} stuff should be removed here as well.

types/foundry/client-esm/canvas/regions/shape.d.ts
```diff
@@ -0,0 +1,78 @@
...
+    protected override _createClipperPolyTree(): (ClipperLib.PolyTree | ClipperLib.IntPoint)[];
+}
+
+/**
+ * An ellipse of a {@link Region}.
+ * @extends {RegionShape<foundry.data.EllipseShapeData>}
+ *
+ * @param {foundry.data.EllipseShapeData} data    The shape data.
+ */
+declare class RegionEllipse extends RegionShape<foundry.data.EllipseShapeData> {
```
CarlosFdez, line 66: param type and extends aren't necessary here I think

types/foundry/client/pixi/placeables/region.d.ts
```diff
@@ -0,0 +1,132 @@
...
+    /** The polygon tree of this Region. */
+    get polygonTree(): foundry.canvas.regions.RegionPolygonTree;
+
+    /** The Clipper paths of this Region. */
+    get clipperPaths(): readonly (readonly ClipperLib.IntPoint[])[];
+
+    /**
+     * The triangulation of this Region.
+     * @type {Readonly<{vertices: Float32Array, indices: Uint16Array|Uint32Array}>}
+     */
```
CarlosFdez, line 58:
```suggestion
    /** The triangulation of this Region. */
```

In3luki:
I think this should be good now. I can subclass `RegionBehavior` and `RegionBehaviorType` without issues.

```ts
import type { BooleanField } from "types/foundry/common/data/fields.d.ts";

class DifficultTerrainBehaviorTypePF2e extends foundry.data.regionBehaviors
    .RegionBehaviorType<DifficultTerrainTypeSchema> {
    static override defineSchema(): DifficultTerrainTypeSchema {
        const fields = foundry.data.fields;
        return {
            active: new fields.BooleanField({ label: "Active" }),
        };
    }
}

interface DifficultTerrainBehaviorTypePF2e
    extends foundry.data.regionBehaviors.RegionBehaviorType<DifficultTerrainTypeSchema>,
        ModelPropsFromSchema<DifficultTerrainTypeSchema> {}

type DifficultTerrainTypeSchema = {
    active: BooleanField;
};

```

image

types/foundry/client-esm/canvas/regions/mesh.d.ts
```diff
@@ -0,0 +1,55 @@
...
+     */
+    get tint(): number;
+    set tint(tint: number);
+
+    /** The tint applied to the mesh. This is a hex value. A value of 0xFFFFFF will remove any tint effect. */
+    protected _tintColor: PIXI.Color;
+
+    /**
+     * Cached tint value for the shader uniforms.
+     * @type {[red: number, green: number, blue: number, alpha: number]}
```
CarlosFdez, line 36: This looks odd being here given we have typescript, but is the type of _cachedTint then [number, number, number, number] instead of number[] ?

In3luki: Oh true, the type is wrong. It should be [number, number, number, number].

In3luki: I've changed the `@type` to `@returns`. I'm not sure what our stance on editorializing the docs is.

PR #14729, opened 2024-05-16
Add missing links to grapple action on several PFS NPCs

TikaelSol:


SpartanCPA:
Do we want to use the grapple action macros linked to "Athletics"?

TikaelSol:
We haven't been doing that thus far. If we start we should sweep all of them up at once.

PR #14737, opened 2024-05-17
Update types and system subclass of `ClientDatabaseBackend`

stwlam:


types/foundry/common/abstract/backend.d.ts
```diff
@@ -11,94 +17,114 @@ export default abstract class DatabaseBackend {
...
      */
     delete(
-        documentClass: ConstructorOf<Document>,
-        context: DatabaseBackendDeleteContext,
-        user: BaseUser,
-    ): Promise<Document>;
+        documentClass: typeof Document,
+        operation: DatabaseDeleteOperation,
+        user?: BaseUser,
+    ): Promise<Document | undefined>;
```
In3luki, line 115:
```suggestion
    ): Promise<Document[] | undefined>;
```
A bit weird that it returns undefined if the ids property is undefined.

src/module/system/client-backend.ts
```diff
@@ -1,39 +1,49 @@
...
+            operation.parent ||
+            operation.pack?.startsWith("pf2e.")
+        ) {
+            return super._getDocuments(documentClass, operation, user);
         }
 
         // Dispatch the request
-        const request = { action: "get", type, ...R.pick(context, ["query", "options", "pack"]) };
-        const response = await SocketInterface.dispatch("modifyDocument", request);
+        const request = { action: "get", type: documentClass.documentName, ...operation };
```
BernhardPosselt, line 25: you don't want a spread here, just 
```js
const request = { action: "get", type: documentClass.documentName, operation };
```
works for me after this change

PR #14788, opened 2024-05-21
Add support for canGangUp: eidolon

MrVauxs:
A small addition to allow for creation of Eidolon Side By Side mechanics.


CarlosFdez:
This should be aimed at the v12 branch currently. And @stwlam will have to be one to review this as he's the one that built it. Personally I'm starting to wonder if we should cache actor roll options and predicate instead.

Issue #14790, opened 2024-05-21
Request: Give 'Duo's Aim' Feat the Same RE as 'Sniper's Aim' in System


GorgoPrimus:
Sniper's Aim is a ranged strike where you "gain a +2 circumstance bonus to this Strike's attack roll and ignore the target's concealment. If you're using a Kickback firearm, you don't take the normal circumstance penalty on this Strike for not having the required Strength score or firing without using a stabilizer". 
Duo's Aim is "Make a ranged weapon Strike against a foe either within your spotter's melee reach or the first range increment of a ranged weapon your spotter is wielding. On this Strike, you gain a +2 circumstance bonus to the attack roll and ignore the target's concealment. If you're using a firearm with the Kickback trait, you don't take the normal circumstance penalty to this Strike for not having the required Strength score or firing without using a tripod".

So while the limitations as to who you can strike aren't the same, the effect when you do are. Since they share the same effect, and only Sniper's Aim has an RE implemented for that effect, I'm asking for that to be copied over to it's Sniper's Duo twin.

https://2e.aonprd.com/Feats.aspx?ID=3176
https://2e.aonprd.com/Feats.aspx?ID=3262

PR #14794, opened 2024-05-22
Adds REs/Effects for feats Duo's Aim and Eagle Eyes

GorgoPrimus:
Closes https://github.com/foundryvtt/pf2e/issues/14790

Adds REs/effects for the feats Duo's Aim and Eagle Eyes.


TikaelSol:
I won't have time to review this for a bit, and we're in a data freeze currently, so it's going to sit until we unfreeze for the 6.0.x releases.

GorgoPrimus:
I know you're not taking a look at this now, but when you have a moment I was wondering if it was permissible within System norms to just make one Effect that 'upgrades' the base Sniping Duo Dedication effect to handle both feats. Since Eagle Eyes is basically an add-on upgrade to the benefit/condition of being chosen as a Sniping Duo partner I mean. 

This would enable you to remove the basic Sniping Duo Effect and replace it with one granted by this feat - allowing you to use toggles instead of requiring you to have two Effects on you that would always go together or having to temporarily remove one and then readd it when you'd end up in a situation some uncommon blind/deaf combo hit one of the partners.

If allowed, it'd look like this in practice:
Effect
Toggles

If not, then what I've already submitted in this Pull Request can stay as is unless you also think something is also amiss with those. Thanks for your time.

GorgoPrimus:
Given the length of time that’s passed and that we’re heading into 6.1.0, would it be possible to get some direction soon as to if my last suggestion is something worth submitting instead along with whatever updates I’d need to make given PRs merged in the meantime like the change to Sniper’s Duo’s label in 6.1.0? Or if what I have now is fine and I just need to update it for the aforementioned, or if the whole thing doesn’t work?

GorgoPrimus:
Alright, well for now I've updated the original PR for version for 6.1 so it's good to go if it the 'combined with toggles' plan is rejected and the rest works.

GorgoPrimus:
Am closing and will resubmit this with the proper file formatting in future.

PR #14806, opened 2024-05-23
Remove cycle-token-stack keybinding and `TokenLayerPF2e`

stwlam:


CarlosFdez:
What happened to it?

stwlam:
core feature now using `[` and `]` to cycle in either direction

PR #14824, opened 2024-05-25
Fix grant item RE form error with resolvable UUIDs

In3luki:
`fromUuid` now throws an error instead of returning `null` when it encounters an invalid UUID.


stwlam:
jeez that's harsh

src/module/item/base/sheet/rule-element-form/grant-item.ts
```diff
@@ -7,7 +7,7 @@ class GrantItemForm extends RuleElementForm<GrantItemSource, GrantItemRuleElemen
     override async getData(): Promise<GrantItemFormSheetData> {
         const data = await super.getData();
         const uuid = this.rule.uuid ? String(this.rule.uuid) : null;
-        const granted = uuid ? await fromUuid(uuid) : null;
+        const granted = uuid && !uuid.startsWith("{") ? await fromUuid(uuid) : null;
```
stwlam, line 10: use `UUIDUtils.isItemUUID`

In3luki: Oh right, there was a helper for that. Updated.

PR #14849, opened 2024-05-26
Set level taken when adding a feat

CarlosFdez:
This makes the advanced archetype feats possible to implement. One possibility is https://github.com/foundryvtt/pf2e/pull/14636 but I had another idea floating around that also requires this change.

Technically a bugfix but not worth announcing until the feature is complete.

### DIFF
--- a/src/module/actor/character/feats.ts
+++ b/src/module/actor/character/feats.ts
@@ -360,7 +360,9 @@ class FeatGroup<TActor extends ActorPF2e = ActorPF2e, TItem extends FeatLike = F
 
         // If this is a new feat, create a new feat item on the actor first
         if (!alreadyHasFeat && (isFeatValidInSlot || !location)) {
-            const source = fu.mergeObject(feat.toObject(), { system: { location } });
+            const source = fu.mergeObject(feat.toObject(), {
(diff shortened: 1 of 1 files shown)


src/module/actor/character/feats.ts
```diff
@@ -360,7 +360,9 @@ class FeatGroup<TActor extends ActorPF2e = ActorPF2e, TItem extends FeatLike = F
 
         // If this is a new feat, create a new feat item on the actor first
         if (!alreadyHasFeat && (isFeatValidInSlot || !location)) {
-            const source = fu.mergeObject(feat.toObject(), { system: { location } });
+            const source = fu.mergeObject(feat.toObject(), {
+                system: { location, level: { taken: slot?.level ?? this.actor.level } },
+            });
```
stwlam, line 365: This looks fine, but I'm wondering if there's somewhere else that--maybe only in a subset of cases--handles this, like a drop handler.

CarlosFdez: I think currently the only place level.taken is set is lower down in the same function (in locationUpdates when its an existing feat being moved to a new slot). This attempts to handle the case for a newly created feat, and falls back to actor level in case its a bonus feat.

Or did you mean something else?

PR #14854, opened 2024-05-27
Automate Werecreature Dedication and Change Shape

### LINKED ISSUE #14813: Werecreature Dedication doesn't grant the Change Shape action or allow you to pick your type of werecreature.
The new werecreature archetype's dedication feat should have you select which type of werecreature you are upon taking it, since that determines the unarmed attacks you get. It should also grant you the change shape action that then gives you the unarmed attacks while transformed.

DocSchlock:
Closes #14813

### DIFF
--- a/packs/actions/change-shape-werecreature-dedication.json
+++ b/packs/actions/change-shape-werecreature-dedication.json
@@ -1,0 +1,102 @@
+{
+    "_id": "9zdI8S6ceGIgnAkT",
+    "img": "systems/pf2e/icons/actions/OneAction.webp",
+    "name": "Change Shape (Werecreature Dedication)",
+    "system": {
+        "actionType": {
+            "value": "action"
+        },
+        "actions": {
+            "value": 1
+        },
+        "category": "interaction",
+        "description": {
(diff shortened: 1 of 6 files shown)


packs/actions/change-shape-werecreature-dedication.json
```diff
@@ -0,0 +1,102 @@
+{
+    "_id": "9zdI8S6ceGIgnAkT",
+    "img": "systems/pf2e/icons/actions/OneAction.webp",
+    "name": "Change Shape (Werecreature Dedication)",
```
TikaelSol, line 4:
```suggestion
    "name": "Change Shape (Werecreature)",
```
File will need renamed, but I think this is a cleaner name for it

packs/feat-effects/effect-change-shape.json
```diff
@@ -0,0 +1,367 @@
+{
+    "_id": "Q8pD29NcGGuRxWwh",
+    "img": "systems/pf2e/icons/spells/beastmaster-trance.webp",
+    "name": "Effect: Change Shape",
```
TikaelSol, line 4:
```suggestion
    "name": "Effect: Change Shape (Werecreature)",
```
We shouldn't make it seem like there's one generic effect for all change shape

PR #14877, opened 2024-05-28
Add `RegionDocumentPF2e` and `RegionBehaviorPF2e` and implement `RegionBehaviorPF2e#isOfType`

In3luki:
There might be a better way to do this but it works and nothing blows up.

### DIFF
--- a/src/module/actor/helpers.ts
+++ b/src/module/actor/helpers.ts
@@ -13,7 +13,7 @@ import {
     extractRollSubstitutions,
     extractRollTwice,
 } from "@module/rules/helpers.ts";
-import { EnvironmentTypeData } from "@scene/region-behaviors/terrain.ts";
+import { RegionBehaviorTypeInstances } from "@scene/region/region-behavior/types.ts";
 import { eventToRollParams } from "@scripts/sheet-util.ts";
 import { CheckCheckContext, CheckPF2e, CheckRoll } from "@system/check/index.ts";
 import { DamageDamageContext, DamagePF2e } from "@system/damage/index.ts";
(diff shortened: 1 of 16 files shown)


src/module/actor/helpers.ts
```diff
@@ -271,9 +270,9 @@ function createEnvironmentRollOptions(actor: ActorPF2e): Record<string, boolean>
             const top = region.elevation.top ?? Infinity;
             if (token.elevation < bottom || token.elevation > top) continue;
 
-            for (const behavior of region.behaviors.filter((b) => b.type === "pf2eEnvironment")) {
```
CarlosFdez, line 274: tbh the reason I'd prefer to filter like this is because I think having separate loops for environment or terrain or whatever we do would be easier to read. The performance impact is just the creation of new arrays, since looping twice with one if per is the same performance impact as looping once with two if statements per.

In3luki: Filter is restored but I had to change the `RegionBehaviorTypeInstances` `TParent` to always have a `Scene` because apparently the token parent is narrowed to `NonNullable` somewhere. Tsc complains that `ScenePF2e` isn't `ScenePF2e | null` and I can't find a reason for that.

CarlosFdez: I think its because until the next ts version, you have to make all filters into type guards. so `filter((b): b is whatever =>`. The next TS version will make that unnecessary.

In3luki: image

This is what I get when I don't change the `RegionBehaviorTypeInstances` and I've already spent an hour before trying to figure out why that happens. 🥲

In3luki: But I agree that in 5.5 it would probably enough to filter on `b.isOfType("pf2eEnvionment")`. Can't wait for that release.

CarlosFdez: It seems to be because RegionBehaviorTypeInstances uses ScenePF2e and not ScenePF2e | null.

In3luki: The screenshot happens with `RegionBehaviorTypeInstances<TParent extends RegionDocumentPF2e = RegionDocumentPF2e>`. Explicitly settings `RegionDocumentPF2e<ScenePF2e>` fixes it.

In3luki:
I forgot to mark this as a Draft after waking up. 😨
Shark suggested going in another direction for this.

PR #14879, opened 2024-05-28
Stamp actors and items with partial `_stats` data on build

stwlam:
This will prevent unnecessary server migrations due to a lack `_stats` (specifically the `coreVersion` field)


CarlosFdez (approved):

In3luki:
Should this include journals and pages?

stwlam:
> Should this include journals and pages?

Not until we actually migrate them.

PR #14890, opened 2024-05-29
Rename `pf2Environment` region behavior type to `environment`

In3luki:


CarlosFdez (approved):
Implementation is fine, even if the entire idea is still seeped with madness.

stwlam:
Squee!

PR #14896, opened 2024-05-30
Reset contained actors when an environment region behavior is updated

In3luki:
Also adds `RegionBehaviorPF2e#isOfType` although I didn't end up really needing it for this PR


stwlam:
image
oi, singular, mr scofflaw

In3luki:
I actually wanted to change it but then thought it would dillute the diffs too much. Pushing it up.

PR #14901, opened 2024-05-30
Work around `TokenDocument##identifyRegions` using the token center to test region inclusion

In3luki:
This tries to solve the problem by overriding `Region#testPoint` to check all grid squares a token occupies if possible. Sadly this only works when a single token is moving.

[large-token.webm](https://github.com/foundryvtt/pf2e/assets/41452412/b1b447a7-9256-48b6-a8d7-53646041d4d9)

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -18,6 +18,9 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
 
     declare auras: Map<string, TokenAura>;
 
+    /** The new position of this token. Only set between `_preUpdate` and `update` if the token moves */
+    newPosition: Point | null = null;
+
     /** Returns if the token is in combat, though some actors have different conditions */
     override get inCombat(): boolean {
         if (this.actor?.isOfType("party")) {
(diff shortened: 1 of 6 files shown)


src/module/scene/token-document/document.ts
```diff
@@ -18,6 +18,9 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
 
     declare auras: Map<string, TokenAura>;
 
+    /** Whether this token is currently moving */
+    isMoving = false;
```
stwlam, line 22: Core handles this with `Token#_animation` (worth checking that it is still current in V12)

In3luki: The region test is called from `_preUpdateOperation` so the animation hasn't started yet. That also means any region triggers are fired instantly instead of at the end of the animation. Might be worth bringing up as a core issue.

In3luki: Renamed to `positionChanged` which might be clearer.

In3luki: This is now `newPosition`. The post-update top left Point of the token.

In3luki:
Updated to work with multiple tokens and exclude regions with `teleportToken` behaviors.

PR #14912, opened 2024-05-31
Convert actor skills and certain roll options to longform

CarlosFdez:
We also add some shims to map the existing short form names to the new long form stuff. Once we're fully done migrating everything we can make them show deprecation warnings with a very fast expiration (6.2 or 6.3) just in case we miss anything..

From a quick test, Raging Intimidation, Rogue Dedication, Acrobat Dedication, and Skill Mastery all continued to function, even with some of them in an incomplete migration state.


src/module/actor/creature/data.ts
```diff
@@ -91,7 +91,7 @@ interface CreatureSystemData extends Omit<CreatureSystemSource, "attributes">, A
     /** Saving throw data */
     saves: CreatureSaves;
 
-    skills: Record<string, SkillData>;
+    skills: Record<SkillLongForm, SkillData>;
```
stwlam, line 94: Where did lore go?

CarlosFdez: Character already doesnt account for lore in the type. That said, i could see about improving the type.

stwlam: I think the only other option is `string`, unfortunately.

CarlosFdez: Record<SkillLongForm, SkillData> & Record<string, SkillData | undefined> wouldn't work?

src/module/actor/character/document.ts
```diff
@@ -742,9 +742,10 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
             rollOptionsAll[`attribute:${key}:mod:${mod}`] = true;
         }
 
-        for (const key of SKILL_ABBREVIATIONS) {
+        for (const key of SKILL_LONG_FORMS) {
```
stwlam, line 745: Can we rename to `SKILL_SLUGS` now? :grin:

src/module/actor/character/document.ts
```diff
@@ -742,9 +742,10 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
             const rank = this.system.skills[key].rank;
             rollOptionsAll[`skill:${key}:rank:${rank}`] = true;
+            rollOptionsAll[`skill:${SKILL_DICTIONARY_REVERSE[key]}:rank:${rank}`] = true;
```
stwlam, line 748: Mind commenting that this is for backward compatibility?

CarlosFdez: Sure.

src/module/actor/character/document.ts
CarlosFdez: Not yet but that would be great to do after the next step. I want to keep the diffs as small as I can for now to not overwhelm the review process.

stwlam (dismissed):

stwlam (approved):

PR #14920, opened 2024-06-01
Add Witch trait to Whisper of Wings

rectulo:


TikaelSol:
Do we have the witch trait on other patrons? It's not necessary, and they don't have them in the books.

PR #14922, opened 2024-06-01
Add option to disable experience tracking.

Truuun:
I have added an option to remove experience bars from character sheets. Although this can already be achieved by modules such as Custom CSS I think this is more convienient and also less likely to break when the html of the character sheet changes. The setting can currently be found between the "Critical Damage Rule" setting and the "Player Rule Elements Access" setting.


Truuun:
Left a console log in there, one second..

PR #14923, opened 2024-06-01
Add option to disable experience tracking.

Truuun:
I have added an option to remove experience bars from character sheets. Although this can already be achieved by modules such as Custom CSS I think this is more convienient and also less likely to break when the html of the character sheet changes. The setting can currently be found between the "Critical Damage Rule" setting and the "Player Rule Elements Access" setting.

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -143,6 +143,7 @@ interface GamePF2e
                 buttons: boolean;
                 cards: boolean;
             };
+            disableExperienceTracking: boolean;
             /** Encumbrance automation */
             encumbrance: boolean;
             gmVision: boolean;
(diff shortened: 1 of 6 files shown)


stwlam:
The field is so inconspicuous that there doesn't really seem to be a need for a way to hide it, much less a setting.

Truuun:
While I agree that it doesn't particularly stand out or anything, I think it's a nice thing to have for those running "story based progression" rather than using xp. I think it would add just a little bit more clarity to players for those kind of campaigns. A similar setting exists in both the Starfinder 1e and D&D 5e system where those elements are also rather inconspiciuous.

CarlosFdez:
I dont disagree with idea of hiding, but more the approach. A single setting to set level up hp that could be set to 0 is my current guess on how to go about it.

CarlosFdez:
Yeah I'm leaning on what I said. I think if you instead convert it to an option to set level up XP globally, you could special case 0 to hide the XP bar. Right now XP needs to be set per player which is a bit annoying I imagine for people that actually use XP. 

That will likely fly better than a separate "yes/no" toggle.

CarlosFdez:
Given very little of this can transfer over to a new PR I'm going to close it to make space, but if you want to pursue that route feel free to open a new one! I would actually like the ability to hide the experience bar myself as I run a milestone game, but we have too many options so the more we can condense the better.

Truuun:
That's fair, thanks for the input!

PR #14925, opened 2024-06-01
Localize attach button

MSAbaddon:


static/lang/en.json
```diff
@@ -2316,6 +2316,7 @@
             },
             "Physical": {
                 "Attach": {
+                    "Label": "Attach",
```
stwlam, line 2319: Alphabetical order :blush:

MSAbaddon: Ah damn - changed :-)

PR #14929, opened 2024-06-01
Update `Token` type for v12

In3luki:


src/module/canvas/token/object.ts
```diff
@@ -38,16 +38,15 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
         return canvas.visibility.testVisibility(this.center, { tolerance, object: this });
     }
 
+    get _animation(): Promise<boolean> | void {
```
stwlam, line 41: I take it they removed this getter, but make it less javascripty if you want to keep it around (no underscore, etc.).

In3luki: Done

PR #14930, opened 2024-06-01
Make retrieving mergeable roll options to toggle more consistent

CarlosFdez:
The previous algorithm attempted to match rules to sources on both index and sourceindex, but the index of the rule was its position in the filtered results. This meant that a bug could occur if there were rule elements above the RollOption.

The new algorithm works by iterating on the rules we're trying to search for instead, and then pulling the source based on the source index. We then check if its a RollOption RE just in case.

### DIFF
--- a/src/module/rules/rule-element/roll-option/rule-element.ts
+++ b/src/module/rules/rule-element/roll-option/rule-element.ts
@@ -335,13 +335,13 @@ class RollOptionRuleElement extends RuleElementPF2e<RollOptionSchema> {
 
         if (this.mergeable && selection) {
             // Update the items containing rule elements in the merge family
-            const rules = R.groupBy(R.uniq(this.#resolveSuboptions().map((s) => s.rule)), (r) => r.item.id);
-            for (const itemId of Object.keys(rules)) {
+            const rulesByItem = R.groupBy(R.uniq(this.#resolveSuboptions().map((s) => s.rule)), (r) => r.item.id);
+            for (const [itemId, rules] of Object.entries(rulesByItem)) {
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/roll-option/rule-element.ts
```diff
@@ -335,13 +335,13 @@ class RollOptionRuleElement extends RuleElementPF2e<RollOptionSchema> {
...
+            const rulesByItem = R.groupBy(R.uniq(this.#resolveSuboptions().map((s) => s.rule)), (r) => r.item.id);
+            for (const [itemId, rules] of Object.entries(rulesByItem)) {
                 const item = actor.items.get(itemId, { strict: true });
                 const ruleSources = item.toObject().system.rules;
-                const rollOptionSources = ruleSources.filter(
-                    (_r, index): _r is RollOptionSource => rules[itemId][index]?.sourceIndex === index,
-                );
+                const rollOptionSources = rules
+                    .map((r) => (typeof r.sourceIndex === "number" ? ruleSources[r.sourceIndex] : null))
+                    .filter((r): r is RollOptionSource => r?.key === "RollOption");
```
stwlam, line 344: Filter then map?

CarlosFdez: the reason is because it could theoretically fail the array lookup , and we need to get rid of falsy. Might as well type assert while we're at.

CarlosFdez: I guess r as the prop in both might have obscured what's going on. rule vs ruleSource.

PR #14940, opened 2024-06-02
Update Rule Elements UI setting to manage access by user role

### LINKED ISSUE #14203: Option to give Rules access only to trusted players
Currently it's DM-only or all players. It would be nice to be able to just give those with the Trusted Player role access to rules.

jfn4th:
Closes #14203

Allows setting a minimum role for Rule Elements UI access. Choices are Game Master/Assistant GM, Trusted Player, and Player.

Screenshot 2024-06-02 at 3 05 18 PM

### DIFF
--- a/src/module/system/settings/index.ts
+++ b/src/module/system/settings/index.ts
@@ -88,13 +88,19 @@ export function registerSettings(): void {
         },
     });
 
-    game.settings.register("pf2e", "enabledRulesUI", {
-        name: "PF2E.SETTINGS.EnabledRulesUI.Name",
-        hint: "PF2E.SETTINGS.EnabledRulesUI.Hint",
+    game.settings.register("pf2e", "minimumRulesUI", {
+        name: "PF2E.SETTINGS.MinimumRulesUI.Name",
+        hint: "PF2E.SETTINGS.MinimumRulesUI.Hint",
         scope: "world",
         config: true,
-        default: false,
-        type: Boolean,
+        default: 4,
+        type: Number,
+        choices: {
+            1: "USER.RolePlayer",
+            2: "USER.RoleTrusted",
+            3: "USER.RoleAssistant",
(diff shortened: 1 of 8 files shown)


CarlosFdez:
It mostly looks fine to me, though I don't know if we should assume assistant gm as well. It might also be wise to make the value a number analogous to the values of CONST.USER_ROLES (and see if foundry has localization for those roles) and see if that makes the check any easier.

jfn4th:
I'd grouped the GM and Assistant GM together as it's more consistent with the current behavior, though I see what you mean. I suppose I could split it up and still maintain the state of affairs for existing worlds by setting Assistant GM as the minimum in the migration.

I'll take another look sometime this week.

jfn4th:
Thanks for the suggestions @CarlosFdez. All in.

CarlosFdez (approved):

src/module/system/settings/index.ts
```diff
@@ -88,13 +88,19 @@ export function registerSettings(): void {
...
-        name: "PF2E.SETTINGS.EnabledRulesUI.Name",
-        hint: "PF2E.SETTINGS.EnabledRulesUI.Hint",
+    game.settings.register("pf2e", "minimumRulesUI", {
+        name: "PF2E.SETTINGS.MinimumRulesUI.Name",
+        hint: "PF2E.SETTINGS.MinimumRulesUI.Hint",
         scope: "world",
         config: true,
-        default: false,
-        type: Boolean,
+        default: 4,
```
ThiefMaster, line 96: Wouldn't it make more sense to default to `3` here considering that you migrate "no players access" to "Assistant DM access"? This would also go in line with the Foundry defaults of Assistant DMs having the same level of permissions than full DMs?

stwlam: GM*, but it would.

PR #14941, opened 2024-06-02
Update spell effect for Fire Shield/Update damage to Glass Shield and re-add effect.

ammalagonc:
- Update Glass Shield to re-add the effect back to the description (somehow got yeeted during the last update).

- update Glass Shield damage to remove modifier and update the dice damage.

- update Fire Shield to include Shield HP and hardness that scales with level.


ammalagonc:
Redoing this again

PR #14956, opened 2024-06-03
Fix Crossbow Crack Shot automation

### LINKED ISSUE #14769: Crossbow crack shot doesn't interact with Backstabber trait
Crossbow crack shot doesn't adjust the backstabber damage
image
Backstabber should be +4 with the feat active (2 * 2 dices from having a striking rune)

Selector is set to an invalid "crossbow-group-damage" rather than "crossbow-weapon-group-damage"

Once adjusted, the value isn't resolved correctly for the 2*[number of weapon damage dice]
image

kanathan:
Fixes Crossbow Crack Shot to correctly adjust range and replace the standard damage from the Backstabber trait
- Change selector to crossbow-weapon-group-damage to remove need for item:group:crossbow predicate
- Removes circumstance type, as feat text implies damage is just additional damage
- Bypasses backstabber damage as workaround for @weapon not currently resolving in AdjustModifier rule

Fixes #14769

### DIFF
--- a/packs/feats/crossbow-crack-shot.json
+++ b/packs/feats/crossbow-crack-shot.json
@@ -26,7 +26,7 @@
         },
         "rules": [
             {
-                "domain": "damage",
+                "domain": "all",
                 "key": "RollOption",
                 "option": "crossbow-crack-shot",
                 "toggleable": true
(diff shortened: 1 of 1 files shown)


TikaelSol:
The selectors are fine but I'd rather we just make AdjustModifier work with `@weapon` references than bundle in a new modifier

Issue #14958, opened 2024-06-04
Allow for parsing injected properties in Strike RE's `ability` property


MrVauxs:
As the title says, it would be great if Strike REs `ability` resolved strings such as `{item|flags.pf2e.rulesSelections.mysticTraining}` in situations where you get to choose which ability you use. 

Currently it fails validation. This is in support of Summoners+

PR #14959, opened 2024-06-04
Make Strike attribute resolvable

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/14958


stwlam:
Use case?

CarlosFdez:
Linked in the issue, but they seem to want to implement this feature, and resolvables are usually in the case of "if its easy and if you want it for any reason".

image

PR #14964, opened 2024-06-04
Add basic diffcult terrain region support

In3luki:
This adds a few new roll options for difficult terrain: `terrain:difficult`, `terrain:difficult:greater`, and `terrain:difficult:magical`. These are not mutually exclusive.

The ruler has two layers of caching to make the impact as low as possible. First the cost per region is cached because the behavior is probably not chaning and we don't have to recalculate everything for the same region. The second caching happens by grid coordinates to avoid calling `Region#testPoint` over and over again for alle regions in the scene.
The cost can be changed by using an `ActiveEffectLike` rule element that adds a new predicate statement to `flags.pf2e.difficultTerrain.ignore` or `ignoreGreater` on the actor. 

```json
{
  "key": "ActiveEffectLike",
  "mode": "add",
  "path": "flags.pf2e.difficultTerrain.ignore",
  "value": {
    "and": [
      "terrain:forest",
      {
        "not": "terrain:difficult:magical"
      }
    ]
  }
}
```

Screenshot 2024-06-04 224819

Screenshot 2024-06-04 225440

### DIFF
--- a/src/module/actor/character/data.ts
+++ b/src/module/actor/character/data.ts
@@ -6,6 +6,7 @@ import {
     CreatureAttributes,
     CreatureDetails,
     CreatureDetailsSource,
+    CreatureFlags,
     CreatureLanguagesData,
     CreaturePerceptionData,
     CreatureResources,
@@ -18,7 +19,6 @@ import {
 import { CreatureInitiativeSource, CreatureSpeeds, Language } from "@actor/creature/index.ts";
 import {
     ActorAttributesSource,
-    ActorFlagsPF2e,
     AttributeBasedTraceData,
     HitPointsStatistic,
     InitiativeData,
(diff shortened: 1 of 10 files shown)


Drental:
about ground only - regions have a max/min elevation, don't they? wouldn't that be a better place to put the ground only?

In3luki:
> about ground only - regions have a max/min elevation, don't they? wouldn't that be a better place to put the ground only?

Regions are create with a min/max of infinity so you'd have to manually set it to 0,0 or 0,5 everytime you add a difficult terrain that should only affect the ground.
With the checkbox you could also stack two terrains on top of each other in the same region with one affecting the ground and one the air above it.

In3luki:
Mostly superseded by #15365

PR #14968, opened 2024-06-05
Improve automation of Fire Shield

ammalagonc:
- Update spell effect to account for variable hardness by using a toggle.

- Update spell heightening damage from +1 to +2 as per the description found in PC1.

### DIFF
--- a/packs/spell-effects/spell-effect-fire-shield.json
+++ b/packs/spell-effects/spell-effect-fire-shield.json
@@ -99,6 +99,33 @@
                 }
             },
             {
+                "alwaysActive": true,
+                "domain": "all",
+                "key": "RollOption",
+                "label": "Fire Shield Block",
+                "option": "water-trait",
+                "predicate": [
+                    {
+                        "or": [
+                            "self:type:character",
+                            "self:type:npc"
+                        ]
+                    }
+                ],
+                "suboptions": [
+                    {
+                        "label": "All traits - Except Water",
+                        "value": "10"
+                    },
+                    {
+                        "label": "Water trait",
+                        "value": "5"
+                    }
+                ],
+                "toggleable": true,
+                "value": true
+            },
+            {
                 "key": "ActiveEffectLike",
                 "mode": "override",
                 "path": "system.attributes.shield.hardness",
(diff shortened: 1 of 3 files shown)


packs/spell-effects/spell-effect-fire-shield.json
```diff
@@ -98,6 +98,33 @@
...
+                "suboptions": [
+                    {
+                        "label": "All traits - Except Water",
+                        "value": "10"
+                    },
+                    {
+                        "label": "Water trait",
+                        "value": "5"
+                    }
+                ],
```
TikaelSol, line 124: This shouldn't be a suboption, since it's just a binary thing. So just a "Defending against a water effect" with a toggle and then two AE likes. This would also need to be localized, but I'd like to see if we could instead find a way to actually automate this part.

packs/spell-effects/spell-effect-fire-shield.json
```diff
@@ -98,6 +98,33 @@
...
+                "label": "Fire Shield Block",
+                "option": "water-trait",
+                "predicate": [
+                    {
+                        "or": [
+                            "self:type:character",
+                            "self:type:npc"
+                        ]
+                    }
+                ],
```
TikaelSol, line 114: This shouldn't be necessary

TikaelSol:
Renamed PR to be in imperative form. PR titles should complete the sentence "If merged this PR will..."

ammalagonc:
Update v20240605 - Made changes to the spell effect as suggested.

packs/spell-effects/spell-effect-fire-shield.json
```diff
@@ -98,19 +98,21 @@
                     "field": "item|system.level.value"
                 }
             },
+            {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "Defending against a water effect",
```
TikaelSol, line 104:
```suggestion
                "label": "PF2E.SpecificRule.FireShield.DefendingAgainstWaterLabel",
```

Localize this. If you haven't localized something before go to `static\lang\re-en.json` and find the SpecificRule section. Insert `FireShield` as a new object where it should be alphabetically then make the subobject `"DefendingAgainstWaterLabel": "Defending against a water effect"`. Then reference it in this rule like the suggestion above.

ammalagonc (commented):
Made the changes - could not get the pull request to recognize the changes to the re-en.json file. Will try to see if I can add this.

PR #14987, opened 2024-06-06
Add damage button to Thermal Nimbus matching choice of damage type

pedrogrullada:


TikaelSol:
Wait, did I really enter this as two effects?

pedrogrullada:
> Wait, did I really enter this as two effects?

I didn't really question it. I could compress them into one while I'm at it, if you want

TikaelSol:
That may need a migration, I can add it to the list to look into

Issue #13609, opened 2024-02-07
Containers tagged "Stowing Container" which causes items to not appear in activations


maplealmond:
Issue detected with a thaumaturge who has scroll esoterica and sleeves of storage.  Scrolls within the sleeves did not appear in the spellcasting tab.  Sleeves are intended to be an item from which one can quick draw.

Ideally drawing items from all sources including backpacks would appear in activations tab, ideally with the action cost of drawing the item indicated somehow.

PR #15002, opened 2024-06-07
Allow Stowed Items to Appear on Activations Tab

maplealmond:
[Resolves ](https://github.com/foundryvtt/pf2e/issues/13609)

Currently, any items stored in a stowing container do not show up in the activations tab.

It is a common use case to keep scrolls or wands in a stowing container until needed (e.g. a wand of heal or a scroll of translate) and pull them out when needed.

This PR allows activations to appear even if the item is stowed.  To highlight a stowed item is being grabbed, the text is changed from "Draw" to "Retrieve".  (This necessitates a new localize field, since the current Retrieve buttons are weapon specific.)

A related incidental fix to the localization, both Retrieve1H and Retrieve2H have the same Title.  I can split this out if necessary.


src/module/item/physical/document.ts
```diff
@@ -188,6 +188,14 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         return !!this.container?.system.stowing;
     }
 
+    get actionCostToDraw(): ZeroToThree {
```
CarlosFdez, line 191: Creating a helper in physical item to figure out the number of actions needed to draw a thing makes sense to me, but I don't know about this name / signature. @stwlam any thoughts?

maplealmond: I had a hard time with the name so I'm more than happy to take a better suggestion.

Worth nothing that while I copied the logic from existing weapon draw, it assumes that the draw value is really only 1 or 2, whereas certain items (e.g. the legacy Gloves of Storing) may actually have a retrieval cost of 0.  Not sure if that's out of scope here.

static/templates/actors/partials/spell-collection.hbs
```diff
@@ -168,10 +168,18 @@
                             {{#if (eq spell.parentItem.isEquipped false)}}
                                 <button
                                     type="button"
-                                    class="cast-spell blue"
+                                    class="cast-spell"
                                     data-action="draw-item"
                                     data-item-id="{{spell.parentItem.id}}"
-                                >{{localize "PF2E.Actions.Interact.Draw.Title"}}</button>
+                                    >
+                                    {{#if (eq spell.parentItem.isStowed true)}}
```
CarlosFdez, line 175: You don't have to compare to true, {{#if spell.parentItem.isStowed}} is enough.

src/module/item/physical/document.ts
CarlosFdez: Super duper out of scope. There's also the option of moving it to a physical/helpers function and then calling that, in lieu of a better option.

Issue #14073, opened 2024-03-06
Sheets for non-stowing container show form elements for reducing bulk


Ghostdesu:
# To reproduce
1. Create a new container.
2. Make sure "Is Stowing Container" is not checked.
3. Set its Bulk Capacity and Bulk Ignored to anything above 0 .
4. Add any items to it.
5. Note that the items are counting towards the actor's bulk.

# Expected behavior
The specified of Bulk should be ignored regardless of whether or not a container is designated as stowing. 

# Example
There is a homebrew quiver in my game, which is meant to ignore the bulk of the arrows inside it. However, it is currently impossible to implement in the system because only stowing containers ignore bulk (which is not usable due to stowing containers removing your ability to actually use the ammo stored in them).

PR #15003, opened 2024-06-07
Sheets for non-stowing container show should not show form elements for reducing bulk

maplealmond:
Addresses: https://github.com/foundryvtt/pf2e/issues/14073

The page for containers shows reduced bulk even if the container is non-stowing.  The definition of a non-stowing container within the system is a "not real" container such as a belt, where the items are functionally "worn" as opposed to "stowed".

The FVTT PF2e system does not allow non-stowing containers to reduce bulk, so the form element is confusing.  This causes them to be hidden, so the user has greater clarity that these numbers will not apply.  The stowing container checkbox is moved upwards as the relevance of the maximum and ignored bulk elements is subordinate to the existence of the stowing field.

https://github.com/foundryvtt/pf2e/assets/22515/6083e8f7-a7c3-40a4-b388-33c183d160da

### DIFF
--- a/static/templates/items/backpack-details.hbs
+++ b/static/templates/items/backpack-details.hbs
@@ -13,15 +13,13 @@
 </div>
 
 {{#if item._source.system.stowing}}
+    <div class="form-group">
+        <label for="{{fieldIdPrefix}}capacity">{{localize "PF2E.Item.Container.Bulk.Capacity"}}</label>
+        <input type="number" name="system.bulk.capacity" id="{{fieldIdPrefix}}capacity" value="{{data.bulk.capacity}}" />
+    </div>
 
-<div class="form-group">
-    <label for="{{fieldIdPrefix}}capacity">{{localize "PF2E.Item.Container.Bulk.Capacity"}}</label>
-    <input type="number" name="system.bulk.capacity" id="{{fieldIdPrefix}}capacity" value="{{data.bulk.capacity}}" />
-</div>
-
-<div class="form-group">
-    <label for="{{fieldIdPrefix}}ignored">{{localize "PF2E.Item.Container.Bulk.Ignored"}}</label>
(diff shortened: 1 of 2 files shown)


stwlam:
Have it clean up after itself if not stowing (hidden fields may be sufficient)

maplealmond:
Apologies, I think I might need a bit more detail to understand the request.  It's not clear to me what needs to be cleaned up. 

Should the Bulk Capacity and Bulk Ignored fields be cleared if the contained is marked non-stowing?

CarlosFdez:
I think the normal approach we take is to override _preUpdate() in order to do it.

maplealmond:
@CarlosFdez That's a good pointer, but I'm still taking a bit of a guess at what the request is.  

The values in question being hidden are non nullable numeric, so I have adjusted this PR so that on a non-stowing container they are zeroed.  If this is not what you mean @stwlam let me know and I'll adjust accordingly.

static/templates/items/backpack-details.hbs
```diff
@@ -17,7 +24,4 @@
     <input type="number" name="system.bulk.ignored" id="{{fieldIdPrefix}}ignored" value="{{data.bulk.ignored}}" />
 </div>
```
CarlosFdez, line 25: You'll want to indent these form groups and remove the empty space after the if and before the end/if to maintain convention.

CarlosFdez:
Other than what I mentioned it looks fine. I think clearing out the values might run into some quirks if someone toggles the property on an existing bag trying to see what it does and then has to put the value back in again, but that is a minor thing.

static/templates/items/backpack-details.hbs
maplealmond: Done.

TikaelSol:
You can run `npm run lint --fix` to fix most formatting issues that prettier flags automatically. After that `npm run test` will run the tests locally to make sure the pipeline will pass here on github

PR #15007, opened 2024-06-07
Fix incorrect med to medicine conversions in some selections

CarlosFdez:


packs/paizo-pregens/jaithe.json
```diff
@@ -3009,7 +3009,7 @@
                         "flag": "fleshwarpSize",
                         "key": "ChoiceSet",
                         "prompt": "PF2E.SpecificRule.Prompt.CreatureSize",
-                        "selection": "medicine"
+                        "selection": "med"
```
stwlam, line 3012: Mind getting the AE-like as well?

stwlam: https://github.com/foundryvtt/pf2e/pull/15004/files

CarlosFdez: Do you mean converting the AELike to CreatureSize as part of an unrelated migration? I'm not against it since migrations are a bit hacky regardless, but just making sure.

stwlam: No just in that file

CarlosFdez: If its only to update the json manually with the updated value, done.

PR #15008, opened 2024-06-07
Remove `ProseMirrorMenuPF2e` subclass and patch some `ProseMirrorMenu` methdos instead

In3luki:
We still need the `_configureProseMirrorPlugins` overrides because you can't set `compact` any other way apparently. 🙃


In3luki:
I wasn't sure where the functions should live. Maybe they should be in 🐵 🩹 ?

CarlosFdez:
There's where I'm generally leaning.

Issue #14468, opened 2024-04-19
[Enhancement] Support inline link callbacks created from enrichHTML without having to invoke InlineRollLinks


typhonrt:
Greets,

An [enhancement issue](https://github.com/League-of-Foundry-Developers/foundryvtt-forien-quest-log/issues/39) was filed for Forien's Quest Log in relation to `pf2e` regarding inline links support from the game system. There are several editor sections of FQL that can benefit from being able to augment the editor content in a similar manner that the game system [handles journals](https://github.com/foundryvtt/pf2e/blob/master/src/scripts/hooks/render-journal-text-page-sheet.ts). At minimum I'd need global access to [InlineRollLinks](https://github.com/foundryvtt/pf2e/blob/master/src/scripts/ui/inline-roll-links.ts) and I can take it from there. 

No worries if this is undesirable to make globally accessible or located in a system related exposed area, however this may be useful for external integration to provide editor augmentation features that `pf2e` provides in any 3rd party modules. Another option is perhaps exposing a custom `pf2e` hook that can take HTML content and optional document and perform the augmentation; I actually like the latter approach best. 

Normally FQL stays game system neutral, but this is a compelling enhancement since `pf2e` is a well used / maintained system.

CarlosFdez:
Sorry for the late reply. What I'm considering doing is moving inline roll link handling to TextEditor.activateListeners(), and putting the handler in the body instead of each individual link. I wanted to do that because its cleaner, but a nice side effect is that integrations like Forien's Quest Log will work without you needing to do anything special.

I don't plan to do this until we have V12 support though, in order to make sure that our V11 final releases are as stable as can be.

Thanks for looking into this though. Hopefully you'll be able to keep that system neutral.

typhonrt:
Thanks for taking a look... This week I'm actually working on the v12 FQL update. Is there a way I can test this with a recent build of PF2E? I gather there may not be a system release with this code yet. Let me know if one will be available this week or general suggestion such as cloning PF2E and building locally to verify. I usually only do one release / update for FQL per Foundry core version, so I'd be glad to confirm that this change is working as intended later this week.

CarlosFdez:
No problem! I would recommend cloning and building though. The build process is the usual, make sure nodejs is installed, then `npm ci` followed by `npm run watch`. You can optionally run `npm run link` if you want to automatically create a symlink from the dist folder to the foundry systems folder.

typhonrt:
[...]

Issue #14466, opened 2024-04-19
[Improvement] Fullfill Roll requests from chat with User Main Actor when Party Token is selected


STBaf:
First things first: Foundry v11.315, PF2 5.15.1, verified in a empty PF2 test world without active Modules (Active Module count: 0!)

When you request rolls from players from chat (for example with the system included "Generate Check Prompt" Macro) the system seem to use the following options to determine which Actor is used to roll the Check Prompt for the player:
If you have selected a token, the stats from that token are used to roll. 
If you have not selected a token on the actual scene the stats from the "Main Character" selected in the User configuration seem to be used.  This, up to this point, seem to make total sense for me.

Following improvement: If you use the Party Token and give your users access to that token (for example: let the user move around the token on a map to indicate where the Party is located) and the Party Token is selected the Check Prompt Rolls does not work out since the system seem not to be able to do the roll with that special actor (which, absolutely, make sense as this is a special actor). On the UI side nothing seem to happen, on the Javascript Console you get the message "PF2e System | Skip rolling unknown statistic athletics".

My wish for improvement: Could you include a "special" rule for the selected party actor token (since Party Actor is anyway build-in special this should be possible to simple hard code) to act like "if you have the party token selected act like no token is selected and just roll with the main character of the user".

This seem to make totally sense for me to behave that way since, meanwhile, it will get a little bit annoying to remind your players continously with "you must deselect the party token first before you try to role that check, that's the reason why nothing happens".

(And yes, I know I could unpack the party token or directly place only user tokens or don't give users access to that token and only move it by myself. But I just want let my users control the party token and they should still be able to roll checks uncomplicated :) ).

Intervencion:
(And in general just default to User's Character when no token selected and user has Ownership of multiple actors)

STBaf:
> This occurs because of Kingmaker. I'll try to find a way to fulfill this without breaking kingmaker. EDIT: Fixing this one is incredibly risky due to the code's fragility. I will wait till after V12 for this, given its not a major bug.

Maybe this is at least possible to look at the Kingmaker Campaign Feature setting in a way like "ok if the Kingmaker feature is not activated then.." to decide to handle the Party Token special. That make it work for most games then (but yes, as a developer myself I understand the inner struggle that the system starts to behave differently depending on a setting that the user will never connect to that one Kingmaker checkbox in the settings is awful, too).

And v12 is totally fine, as said would be a nice improvement 🙂

[...]

PR #15009, opened 2024-06-08
Move inline check/action/template listeners to global handlers

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/14468

This creates a listener on the document in order to handle the inline checks and removes redundant visibility updates. The result should hopefully be better performance and module compatibility with things like Forien's Quest Log. 

**Nerd Details**
Its a bit verbose because we're not using jquery. In order to make it work we needed the resolve methods to be a bit smarter such as when in an item summary. TextEditor.enrichHTML() already calls `UserVisibilityPF2e#process()` (calling it extra times is redundant), so I moved the damage roll flavor tweaks as well as the repost button generation there since they also are temporary changes based on document ownership. There was one gotcha from this approach: it failed to detect properly for journal pages that were shared from a gm to a player via "Show Players". I managed to work around it by using relativeTo, which is the document foundry passes when calling enrichHTML to resolve relative links.


CarlosFdez:
When this can be resumed, I should look into https://github.com/foundryvtt/pf2e/issues/14466 at the same time.

PR #15012, opened 2024-06-08
Update Typescript to 5.4.5

In3luki:
All changes are to accommodate `Type instantiation is excessively deep and possibly infinite.` errors with the new version.
I've tracked it down to a problem with `Item<this>` types but can't find a good fix apart from asserting the type to avoid recursion.


In3luki:
I wanted to use `Promise.withResolvers` but I guess I can wait 2 weeks. 😢

PR #15018, opened 2024-06-09
Allow sheets to still open despite invalid data

CarlosFdez:
The skills themselves are still broken, and we need to do a bit more filtering, but the sheet should be openable now.

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -38,7 +38,7 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
 
         sheetData.data.perception.senses = R.sortBy(sheetData.data.perception.senses, (a) => a.label ?? "");
         const initiativeOptions = [{ value: "perception", label: "PF2E.PerceptionLabel" }];
-        initiativeOptions.push(...Object.values(sheetData.data.skills).map((s) => ({ value: s.slug, label: s.label })));
+        initiativeOptions.push(...Object.values(this.actor.skills).map((s) => ({ value: s.slug, label: s.label })));
 
         return {
             ...sheetData,
--- a/src/scripts/handlebars.ts
+++ b/src/scripts/handlebars.ts
@@ -150,6 +150,22 @@ export function registerHandlebarsHelpers(): void {
     Handlebars.registerHelper("raw", function (this: unknown, options: Handlebars.HelperOptions): string {
         return options.fn(this);
     });
+
+    /**
+     * Work around core issue with numberFormat where null/undefined values crash the sheet
+     * https://github.com/foundryvtt/foundryvtt/issues/11165
+     */
+    Handlebars.registerHelper(
+        "numberFormat",
+        (value: unknown, options: { hash: { decimals: number; sign: boolean } }) => {
+            const numValue = Number(value);
+            const dec = options.hash.decimals ?? 0;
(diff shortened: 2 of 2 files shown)


stwlam:
Why not just make sure they're numbers?

CarlosFdez:
I need a bit more detail than that.

The issue I'm trying to resolve is that the numberFormat handlebars helper crashes outright if its invalid. Invalid numbers shouldn't be an expectation, but allowing us to modify items to repair a sheet would be nice.

Do you mean to early return "NaN" as a string if its not a string or number? The core built in method does this
```js
  static numberFormat(value, options) {
    const dec = options.hash.decimals ?? 0;
    const sign = options.hash.sign || false;
    if ( typeof value === "string" ) value = parseFloat(value);
    let strVal = sign && (value >= 0) ? `+${value.toFixed(dec)}` : value.toFixed(dec);
    return new Handlebars.SafeString(strVal);
  }```
  
  which explodes if its not a string or number.

**Comment by stwlam** [comment]
Yeah and that's not great, but it's an upstream concern. Down here we can just make sure we're passing actual numbers.

**Comment by CarlosFdez** [comment]
In handlebars land that's not a reasonable guarantee, especially when there are AELikes running rampant. The sheets absolutely have to gracefully handle errors, including errors we do not forsee.

For context: the actual error was that someone had a failed skills migration, and I couldn't even investigate properly what was going on for a bit because the sheet wouldn't open. Their sheet could not be fixed by anything except messing with the console.

I'll see what Foundry does. If it gets added to a milestone, I'll treat it as a temporary patch. But otherwise its either making a safe mode sheet when rendering fails or a safeNumberFormat and replace all our references with that. Perhaps both.

**Comment by CarlosFdez** [comment]
I sent the issue yesterday here https://github.com/foundryvtt/foundryvtt/issues/11165 but its the weekend so we can't really tell right now.

File: src/module/actor/creature/sheet.ts [hunk 1]
```diff
@@ -38,7 +38,7 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
 
         sheetData.data.perception.senses = R.sortBy(sheetData.data.perception.senses, (a) => a.label ?? "");
         const initiativeOptions = [{ value: "perception", label: "PF2E.PerceptionLabel" }];
-        initiativeOptions.push(...Object.values(sheetData.data.skills).map((s) => ({ value: s.slug, label: s.label })));
+        initiativeOptions.push(...Object.values(this.actor.skills).map((s) => ({ value: s.slug, label: s.label })));
```
stwlam, line 41: Why this?

CarlosFdez: These two things should probably be separate PRs, but both together allowed the sheet to open despite bad data prep. If skills fail data prep for any reason, there are two points of failure: numberFormat on the skills tab, and the initiative selection.

The initiative selection is a new point of failure in V12, because of the deprecation of `#select` moving us towards `#selectOptions`. This is a point though where I saw we could reasonably guarantee that the data coming in was correct.

With the true bug fixed (updated the migration) both of these can wait till next release, and we can probably fix corrupted data period as well while we're at it (only for skills).

CarlosFdez:
Between numberFormat's crash being accepted to the V12 track and some measures I'm starting to take against corrupted skill data existing in the first place, we can probably do without. By the time the numberFormat becomes a problem again for something else, it should hopefully be resolved in core.

PR #15020, opened 2024-06-09
Begin fixing broken links in roll tables

TikaelSol:
This should be fine to do in parts. The remaster renaming some items broke the links because they are linked by name rather than ID 😱

This also catches a bunch of references to spell levels in wand names. There's still some more wands with names like (xth-level) rather than (xth-level spell). We should maybe consider standardizing these to (xth-rank spell), though now that rank and level are separate maybe just (xth-rank) is fine.


packs/rollable-tables/11th-level-permanent-items.json
```diff
@@ -63,7 +63,7 @@
                 24
             ],
             "text": "Bag of Holding (Type III)",
```
SpartanCPA, line 65: This is a spacious pouch now right?

packs/rollable-tables/10th-level-permanent-items.json
```diff
@@ -160,7 +160,7 @@
                 58,
                 63
             ],
-            "text": "Staff of Abjuration (Greater)",
+            "text": "Staff of Protection (Greater)",
             "type": 2,
```
SpartanCPA, line 164: this one isn't changing the types?

TikaelSol: That may have just got lost when I did some post extract editing

packs/rollable-tables/11th-level-permanent-items.json
TikaelSol: Yeah, I only fixed the 10th and 1st level table links

PR #15026, opened 2024-06-09
Update token config templates to latest core version

In3luki:
The whole section is copied directly from the core template.

Screenshot 2024-06-09 210318

Screenshot 2024-06-09 211934


CarlosFdez:
In a future release we definitely outta add in a subnav

PR #15032, opened 2024-06-10
Add class features for Revolutionary Innovation

ammalagonc:
Class Features granted by Revolutionary Innovation are not in the system. This is a first step to add these and then build towards adding automation.

### DIFF
--- a/packs/classfeatures/heavy-construction.json
+++ b/packs/classfeatures/heavy-construction.json
@@ -17,7 +17,11 @@
             "value": 7
         },
         "prerequisites": {
-            "value": []
+            "value": [
+                {
+                    "value": "Power Suit Only"
+                }
+            ]
         },
         "publication": {
             "license": "OGL",
--- a/packs/classfeatures/heavy-construction.json
+++ b/packs/classfeatures/heavy-construction.json
@@ -11,17 +11,13 @@
         },
         "category": "classfeature",
         "description": {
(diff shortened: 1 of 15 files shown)


packs/classfeatures/heavy-construction.json
```diff
@@ -17,7 +17,11 @@
             "value": 7
         },
         "prerequisites": {
-            "value": []
+            "value": [
+                {
+                    "value": "Power Suit Only"
```
pedrogrullada, line 22: We probably want to be consistent on whether these parenthesis are added as prerrequisites or as part of the description like it's done elsewhere in this PR. I'm leaning toward making it part of the description as these aren't prerrequisites in the same way feats have prerrequisites. 
image

ammalagonc: Fair enough - I'll change these to be part of the description then.

pedrogrullada: If we want to be consistent, omit them and use the predicates in the (future) Choice Set for Revolutionary Innovation to filter them out

ammalagonc: OK, those changes should be in now. Please check when you get a moment. 

Thanks :)

pedrogrullada: As I said, I think you're better off omitting these requirements for the sake of consistency with existing modifications. The ChoiceSet can do the filtering for prerequisites for you.

ammalagonc: Weird - I see them omitted on my end.

pedrogrullada: After checking, it looks like we don't do this consistently in the system. Sometimes the prerequisites are in the body of the text, sometimes they are omitted. This looks fine to me, but ultimately up to one of the leads which way we actually go with.

PR #15033, opened 2024-06-10
Convert NPC skills and skill variants to system data

CarlosFdez:
If approved, tikael has asked that I run the migration itself piecemeal so he can review it in chunks rather than dropping the atomic egg.

Includes sample migration.


CarlosFdez:
An alternative to having the variant skills in npc skills (if that's a problem?) is to have them top level in actor.system.variantSkills as a `Partial<Record<SkillSlug, NPCSkillVariant[]>>`

Issue #15037, opened 2024-06-10
CompendiumBrowserTab.createRollTable() not populating documentCollection or documentId 


matthewdbradley:
When I go to create a new RollTable from a compendium search it creates the RollTable with missing values for documentCollection and documentId for each TableResult. This is possibly due to the private method getRollTableResults returning with keys "collection" and "resultId" instead of "documentCollection" and "documentId" as expected by the TableResult class. Below is a screenshot of the data from a table created post- v6.0 update. I was able to confirm this was not the case with tables created pre- v6.0.

![bug](https://github.com/foundryvtt/pf2e/assets/87322227/b6f76743-7eef-4a9c-a48c-5301c0313ebe)

This bug appears as the "Result Details" column defaulting to the first value, but worse is when it is rolled and returns null. 

For those looking for a work-around, I've pasted my console command to circumvent this issue for now. The table name and compendium name have to be specified.
```
const rollTable = game.tables.getName("Compendium Browser");
const compendium = game.packs.get("pf2e.equipment-srd");
await compendium.getIndex();
rollTable.results.forEach(result => {
    const res = compendium.index.find(entry => entry.name === result.text);
    rollTable.updateEmbeddedDocuments("TableResult", [{
      _id: result._id,
      documentCollection: compendium.collection,
      documentId: res._id
    }]);
});
```

PR #15038, opened 2024-06-10
Fix creating roll tables from compendiums for v12

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/15037

### DIFF
--- a/src/module/apps/compendium-browser/tabs/base.ts
+++ b/src/module/apps/compendium-browser/tabs/base.ts
@@ -298,21 +298,23 @@ export abstract class CompendiumBrowserTab {
         initial?: number;
         weight?: number;
     }): Partial<TableResultSource>[] {
-        return this.currentIndex.flatMap((e, i) => {
-            const data = fromUuidSync(e.uuid);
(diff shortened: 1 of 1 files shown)


src/module/apps/compendium-browser/tabs/base.ts
```diff
@@ -298,21 +298,23 @@ export abstract class CompendiumBrowserTab {
...
+                    type: CONST.TABLE_RESULT_TYPES.COMPENDIUM,
+                    documentCollection: data.pack,
+                    documentId: data._id,
+                    img: e.img,
+                    weight,
+                    range: [rangeMinMax, rangeMinMax],
+                    drawn: false,
+                };
+            })
+            .filter((r): r is Partial<TableResultSource> => !!r);
```
stwlam, line 317: I was going to suggest a flatMap but then turned my head left :upside_down_face:

CarlosFdez: for whatever reason, typescript just let flatMap() get the types completely wrong, which is how this bug even slipped through. Best I could do is keep the flat map and assign to a variable that has a specific type and return it.

PR #15050, opened 2024-06-11
Add custom `tagify-tags` HTML element that handles `Tagify` form data

In3luki:
The `HTMLTagifyTraitsElement` has an `HTMLInputElement` child that must be used to attach a `Tagify` instance. The `value` of the `HTMLTagifyTraitsElement` will always be an array so `FormDataExtended` won't have any problems with JSON data anymore.
The input child inherits the `HTMLTagifyTraitsElement`'s name as `data-tagify-name` for selector purposes. It can also be targeted with `tagify-traits > input` of course.
I've tested every input after converting it and it looks like everything is still working as expected.

### DIFF
--- a/src/util/tags.ts
+++ b/src/util/tags.ts
@@ -1,6 +1,7 @@
 import { TraitViewData } from "@actor/data/base.ts";
-import Tagify from "@yaireo/tagify";
-import { ErrorPF2e, objectHasKey } from "./misc.ts";
+import type { HTMLTagifyTraitsElement } from "@system/html-elements/tagify-traits.ts";
+import Tagify, { TagifySettings } from "@yaireo/tagify";
+import { objectHasKey } from "./misc.ts";
 
 type WhitelistData = string[] | Record<string, string | { label: string }>;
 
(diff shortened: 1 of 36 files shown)


CarlosFdez:
The idea of creating a custom type but then having to bind to it after is a bit odd but way better than what we were doing before. I'm down to give it a shot soon.

In3luki:
It could be expanded with a function that creates the element and then binds tagify to it. I tried binding directly to the custom element first but that didn't work out because of the way that `Tagify` handles updates of the parent element.

CarlosFdez:
Its probably not the sort of thing that's really easy to fix in a clean way with our current tooling. What you have now is fine.

The issue with making a function that creates the element and binds it is that handlebars returns a string.

In3luki:
I've made `HTMLTagifyTraitsElement#input` public and modified the `tagify` helper to also accept `HTMLTagifyTraitsElement`s and automatically bind to the input element.
I've also removed the direct access to `___tagify` and ensured that the element value is always the parsed version of the input element value.

In3luki:
I have resolved the conflicts and renamed it to `HTMLTagifyTagsElement`.

src/util/tags.ts
```diff
@@ -31,15 +32,23 @@ function transformWhitelist(whitelist: WhitelistData) {
...
+    options?: TagifyOptions,
+): Tagify<TagRecord> | null;
+function tagify(
+    element: HTMLInputElement | HTMLTagifyTagsElement | null,
+    { whitelist, maxTags, enforceWhitelist = true, editTags }: TagifyOptions = {},
 ): Tagify<TagRecord> | null {
-    if (input?.hasAttribute("name") && input.dataset.dtype !== "JSON") {
-        throw ErrorPF2e("Usable only on input elements with JSON data-dtype");
-    } else if (!input) {
+    // Avoid importing the HTMLTagifyTraitsElement class that references the foundry API
```
CarlosFdez, line 45: What does this comment mean?

In3luki: Importing the class (that extends `foundry.applications.elements.AbstractFormInputElement`) directly for an `instanceof` check breaks the pack builder because tsx tries to resolve the import from *somewhere*.

CarlosFdez: Ah, so its not that the code leads to avoiding the import, its that its an alternative to importing. Got it.

Issue #14431, opened 2024-04-13
The Striking RE does not increase striking damage dice for features like Gravity Weapon


CrackJackFlood:
The title, gravity weapons effect does account the damage dice increase from runic weapon. 
![image](https://github.com/foundryvtt/pf2e/assets/49882452/48563115-8315-4ed1-8c2f-baa192bed393)

This is related to issue #14409

CarlosFdez:
This is not related to that issue at all. Runic Weapon is supposed to count as if it was a striking weapon, because its the striking re. Increasing it via other ways is not a planned feature.

Issue #13368, opened 2024-01-29
Striking Rule Element and @weapon.system.damage.dice


dcoffee:
`@weapon.system.damage.dice` does not reflect the additional dice from a Striking Rule Element ~~when it is applied to a synthetic strike from a Strike Rule Element~~.

I came upon this when applying *Boost Eidolon* to an Eidolon's strike which was built from a Strike Rule Element, but it can be reproduced using ~~Monk stances as well~~ many weapons.

Reproduction:
Create an actor with any weapon, such as a greatsword, and *Watch This!* feats.  Add *Spell Effect: Runic Body* or *Spell Effect: Runic Weapon* to the actor, as appropriate.

Preview a damage roll for the strike.  Note that even when 2 damage dice are shown in the preview, due to *Runic Weapon*, the bonus for *Watch This!* remains at +2, instead of the expected +4 (2 * `@weapon.system.damage.dice`).

Most features which scale with weapon damage dice seem to be using `@weapon.system.damage.dice` as their scalar, and thus will not scale correctly with Striking Rule Elements.

dcoffee:
This feels more like a bug than a feature request, especially given the description of the "Striking" rule element in the documentation.

dcoffee:
Updated this- apparently it isn't just limited to synthetic strikes.

PR #15056, opened 2024-06-12
Fix Striking RE

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/14431 and closes https://github.com/foundryvtt/pf2e/issues/13368,


src/module/system/damage/weapon.ts
```diff
@@ -221,38 +223,6 @@ class WeaponDamagePF2e {
...
-
-            // Remove extra dice from weapon's etched striking rune
-            if (weapon.isOfType("weapon")) {
-                weapon.system.damage.dice -= weapon.system.runes.striking;
-                weapon.system.runes.striking = 0;
-            }
-        }
-
         // Critical specialization effects
         const critSpecEffect = ((): CritSpecEffect => {
```
stwlam, line 227: What of crit spec effects that depend on number of damage dice?

CarlosFdez: It seems to work perfectly. The crit specs are collected, but they're not computed yet. We can move the ordering if that makes you feel comfortable with regards to potential future breakage though.

Issue #15060, opened 2024-06-12
Motion Sense is always precise


MrVauxs:
Despite there being plenty of [instances](https://github.com/foundryvtt/pf2e/blob/6708894870abeeb70201ace2e4d941a0935ecd80/packs/feats/uncanny-awareness.json#L14) where its not.

Version 6.0.1

stwlam:
Again you are correct! It is by definition precise.

CarlosFdez:
Doesn't seem to be by definition, only by example. Out of curiosity to the two of you:
1) @stwlam Where is the definition that its precise? I can't find this definition anywhere (I looked)
2) @MrVauxs what are the other instances? It might be that it has different behavior for PCs, so I'd need to know those instances.

MrVauxs:
> 2. @MrVauxs what are the other instances? You said there's multiple.

Ultimately ended up being two after examining every instance of the word in the books more thoroughly. The feat I linked above and [Oozeform Chair](https://2e.aonprd.com/AnimalCompanions.aspx?ID=55). That is of course excluding homebrew (and thus Summoners+).

In either case if its [_only precise for creatures lacking vision_](https://github.com/foundryvtt/pf2e/pull/15061#issuecomment-2162444437), then the PR fixing this still should be merged so it can be an imprecise for creatures that _do_ have it.

PR #15061, opened 2024-06-12
Allow motion sense to be imprecise or vague

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/15060


stwlam:
Motion sense is a precise sense for creatures lacking vision

PR #15067, opened 2024-06-12
Use snapping mode when placing bursts and emanations as well

CarlosFdez:
The highlight is snapped, but not the placement nor circle itself current.

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -1,4 +1,4 @@
-import type { MeasuredTemplatePF2e } from "../measured-template.ts";
+import { MeasuredTemplatePF2e } from "../measured-template.ts";
 
 export class TemplateLayerPF2e<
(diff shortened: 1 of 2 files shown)


src/module/canvas/layer/template.ts
```diff
@@ -6,13 +6,26 @@ export class TemplateLayerPF2e<
     /** Preview event listeners that can be referenced across methods */
     #previewListeners: TemplatePreviewEventListeners | null = null;
 
+    template: MeasuredTemplatePF2e | null = null;
```
stwlam, line 9: What is the reason for having one of the layer's contained templates as a single property?

CarlosFdez: To allow the measured template itself to say "hey snap according to me".

In retrospect I should either make that communication occur with a method (and make this property private) or see if layer.hover gets me what I want. Don't merge this yet!

PR #15079, opened 2024-06-13
Automate Hunt Prey

pedrogrullada:
This is a draft for for a Hunt Prey implementation. Since there is *a lot* here, I will write the idea behind it to hopefully make reviewing easier. I have no doubts this would require a migration, which I unfortunately don't have the ability to write.

## Hunt Prey Effect
We add a Hunt Prey effect to be dragged onto the critter, which serves two purposes
1. Add a `self:prey:{item|origin.signature}` roll option to the hunted prey, which helps automate many ranger abilities by using its `target` or `origin` counterparts as a predicate. The signature makes this roll option unique per Hunt Prey user, allowing multiple of them to mark the same creature without any conflicts in automation.
2. Add an Ephemeral Effect rule element to the critter, which applies an effect containing a MAP rule element. This allows Flurry rangers, and those who share its benefits, to automatically get an adjusted MAP when targeting the hunted prey.

### Why not Token Mark?
I decided against using the Token Mark rule element because while it's sufficient for the ranger's own abilities, it falls short when it comes to sharing your Hunt Prey and Hunter's Edge benefits. The on-critter effect allows other characters to enjoy the same automation the ranger does without needing to mark the same target multiple times.

### What the effect can't handle
Since a few Hunt Prey and Hunter's Edge benefits either can be done out of combat (such as Tracking or Gathering Information) or don't have target data (such as Recall Knowledge, or "loose" skill checks), these are handled via toggles. For instance, Hunt Prey has a toggle to Track or Seek your Hunted Prey, or Outwit has one for its various skill checks against Hunted Prey.

These replace the existing, broader "Target is your Hunted Prey" toggle, which I felt necessary because, in my opinion, it would not be clear otherwise which features require the toggle to be automated and which don't, which would lead to an awkward player experience.

### Hunt Prey Action
The action itself contains the effect to be dragged from chat once posted. It also contains rule elements for the previously mentioned "exploration" benefits of Hunt Prey, as well as the roll option to ignore range increment, which is now predicated on the new Hunt Prey effect.

Once the effect is applied, we can leverage it in various forms.

### Flurry
The Flurry class feature and its Masterful Hunter upgrade have had their rule elements removed, instead they all rely on a new Flurry effect containing MAP rule elements, which is only applied ephemerally when attacking the hunted prey. This approach gets around the current limitation on target roll options on that specific rule element.

### Outwit
While Outwit still needs a toggle for its skill benefits, as these can't reliably contain target data, a Flat Modifier to their AC predicated on `"origin:prey:{actor|signature}"` allows them to automatically gain this bonus against the Hunted Prey.

### Precision
Due to its effects requiring a first attack against the Hunted Prey, something not kept track of in the system, Precision does not benefit as much from this change, but it does allow us to further predicate its Damage Dice bonus. We also add mergeable suboptions to keep its "first, second and third" attack toggles contained in a single toggle.

## Sharing Hunt Prey and Hunter's Edge
For feats that allow sharing your Hunt Prey benefits, such as Warden's Boon and the ranger Animal Companion feats, we modify the existing effects in the system to benefit from the same automation the Hunt Prey user has by predicating these benefits on the origin's signature. Flurry has a slightly different approach, where the flurry benefits are applied ephemerally when attacking the hunted prey, whose Ephemeral Effect rule element checks that flurry has been shared with the attacker (via the `self:shared-flurry:{item|origin.signature}` roll option to ensure these benefits come from the origin of the Hunt Prey effect)

Additionally, we use the Item Alteration rule element on these feats to override their description to only contain the correct effect according to their Hunter's Edge.

## Other feat and features
Ranger Weapon Expertise, Hunter's Aim, Monster Hunter and Hamstringing Strike and a couple of pieces of equipment have all been implemented using the Hunt Prey effect. I thoroughly searched the system to make sure I didn't miss any instance of the old Hunt Prey implementation in feats and features.

## Non-Ranger Hunt Prey
Namely the Game Hunter and Bounty Hunter archetypes have had their features that interact with Hunt Prey reworked to play nice with this implementation. Not much about them has been automated yet, so the changes are overall small.

---

That's about everything this PR encompasses. I know it's a lot, but any feedback here would be welcome.


CarlosFdez:
I need to read fully, but this seems like a case for token mark upgrades if token mark doesn't handle this scenario, given this is exactly the kind of thing token mark was built for.

pedrogrullada:
> I need to read fully, but this seems like a case for token mark upgrades if token mark doesn't handle this scenario, given this is exactly the kind of thing token mark was built for.

That'd be great. I think the things Token Mark is missing to accommodate a Hunt Prey implementation are, as far as I can tell
1. The ability for other characters to reference that Token Mark, that way ranger benefits can be shared.
2. Each mark needs to be able to be unique, so that multiple rangers don't trip over each other when marking critters.

ChiefMilesEdgeworth:
Just wanted to chime in and say I'd be really excited to see this automated. Is there anything I could do to help with this PR?

pedrogrullada:
> Just wanted to chime in and say I'd be really excited to see this automated. Is there anything I could do to help with this PR?

The main thing holding it back is some improvements needed on the Token Mark rule element, as mentioned here in my [earlier comment](https://github.com/foundryvtt/pf2e/pull/15079#issuecomment-2166995354). The PR does work, but ideally Token Mark should be extended to handle this use case, as opposed to getting around it, like I am here.

If you are interested in that, it may be worth getting a hold of the main developers on the #development-help channel in the Discord, just to make sure there it's something they're interested in currently.

xyzzy42:
Note that the PF2e Ranged Combat module automates Hunt Prey too.  It's similar in concept to this.

The PF2e Thaumaturge module also has the same issues with TokenMark.  It uses it, but also needs extra stuff, such as an effect on the thaum's target that contains a reference back to the thaum, like this PR does.  Same reason basic reason: allies can't see token marks.

How about an extension:
TokenMark: ` "share": "allies"|"enemies"|"all"`
Allow the specified group to also see this mark in their rolls against the target.

Roll options of this form are added:
`target:mark:${mark.slug}:ally` and `target:mark:${mark.slug}:${mark.origin.signature}`

pedrogrullada:
> Note that the PF2e Ranged Combat module automates Hunt Prey too. It's similar in concept to this.
> 
> The PF2e Thaumaturge module also has the same issues with TokenMark. It uses it, but also needs extra stuff, such as an effect on the thaum's target that contains a reference back to the thaum, like this PR does. Same reason basic reason: allies can't see token marks.
> 
> How about an extension: TokenMark: ` "share": "allies"|"enemies"|"all"` Allow the specified group to also see this mark in their rolls against the target.
> 
> Roll options of this form are added: `target:mark:${mark.slug}:ally` and `target:mark:${mark.slug}:${mark.origin.signature}`

I was actually contacted by the Ranged Combat author, since they wanted to use this PR's approach for their module. I'm not sure if that went anywhere. Anyway, I agree with your assessment here and is more or less what I had in mind. I've created an issue here #16674 to actually request the feature, so it doesn't get lost. 

I will close this PR now, since several things would need to be rewritten here, and there are several improvements I would make (plus, I mucked up the rebase a bit!)

PR #15081, opened 2024-06-14
Ensure `EnvironmentCanvasGroup` is initialized with prepared environment data

stwlam:


CarlosFdez:
There an issue you can possibly link to?

stwlam:
nope :\

CarlosFdez:
I think creating an issue and linking to it might be wise, but it'll be a waste if they don't accept it for some reason, so hopefully we remember once they actually do fix it.

PR #15083, opened 2024-06-14
Add elementTraits to HOMEBREW_TRAIT_KEYS

MrVauxs:
This PR acts both as an Issue and remedy for it.

When one wants to create custom elements for Kineticist, you are forced to use JS to add the new traits to `CONFIG.PF2E.elementTraits` ([said by stwlam](https://discord.com/channels/880968862240239708/880969253765935174/1196555759270506506)). I thought about skipping the user unfriendly JS method and instead use the `pf2e-homebrew` index to add the traits. 

It looks like you can't due to `HOMEBREW_TRAIT_KEYS` not including `elementTraits`

Searching through the repository, `elementTraits` is purely for the purposes of Elemental Blast, which this PR aims to make extendable. I don't foresee any side effects that are going to come from allowing `elementTraits` to be extended in this more conventional way.

### DIFF
--- a/src/module/system/settings/homebrew/data.ts
+++ b/src/module/system/settings/homebrew/data.ts
@@ -16,6 +16,7 @@ const HOMEBREW_TRAIT_KEYS = [
     "weaponTraits",
     "equipmentTraits",
     "environmentTypes",
+    "elementTraits"
 ] as const;
 
 /** Homebrew elements from some of the above records are propagated to related records */
--- a/src/module/system/settings/homebrew/data.ts
+++ b/src/module/system/settings/homebrew/data.ts
@@ -16,7 +16,7 @@ const HOMEBREW_TRAIT_KEYS = [
     "weaponTraits",
     "equipmentTraits",
     "environmentTypes",
-    "elementTraits"
+    "elementTraits",
 ] as const;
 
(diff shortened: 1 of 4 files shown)


CarlosFdez:
There may be no repercussions at this very moment, but there may be in the future. Elements exist in pf2e for more than just kineticist, and people seem to also want to create custom kineticists for things that aren't elements such as "void".

This almost occurred but got mixed up in other stuff, but I think the correct solution is to limit elemental blast to effect traits instead. Really what we want to prevent is "attack" being an element.

MrVauxs:
I presume its more complicated than just merging the two Sets and replacing it in elemental blast code

CarlosFdez:
For converting to effectTraits? Simpler. You'd replace the choices in the schema with effectTraits instead of elementTraits and then update the types everywhere to match.

I did that once, but I also mixed it up with converting all of it to data models and joint validating that the EB's element matched the actual element added by kineticist feats, but that was too much.

MrVauxs:
Wait, effectTraits include the `"attack"` trait, is that really what I should be replacing `elementTraits` with?

CarlosFdez:
they shouldn't. Effect traits are created by omitting certain traits from action traits, including attack.

MrVauxs:
Ah, I missed that its an omit list

PR #15127, opened 2024-06-18
Replace level with rank in Unimpeded Stride description

Cuingamehtar:
I'm splitting https://github.com/foundryvtt/pf2e/pull/14976 PR into several smaller-sized and logically contained PRs for easier review.

There are bestiary creatures that use legacy version of this spell that still use spell level instead of rank. Namely, those are
- Fallen Champion from Book of the dead
- Ordwi from Rusthenge

I did not change them here but tell me if you want me to.

### DIFF
--- a/packs/spells/unimpeded-stride.json
+++ b/packs/spells/unimpeded-stride.json
@@ -11,7 +11,7 @@
         "damage": {},
         "defense": null,
         "description": {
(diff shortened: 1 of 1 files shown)


TikaelSol:
Feel free to change the text in the NPCs too, just saw that note

PR #15138, opened 2024-06-19
update Ostilli Host archetype to add actions (with effects)

ammalagonc:
This PR adds all the actions granted by the feats in the Ostilli Host Archetype. Managed to include some automation and effects save for the following  feats/actions:

- Spit Ambient Magic
- Administer Ambient Magic
- Drape Ambient Magic


packs/feat-effects/effect-turn-aside-ambient-magic.json
```diff
@@ -0,0 +1,69 @@
...
+                    },
+                    {
+                        "label": "Sonic",
+                        "value": "sonic"
+                    }
+                ],
+                "flag": "damage",
+                "key": "ChoiceSet",
+                "prompt": "Choose a damage type"
+            },
```
TikaelSol, line 50: Be sure to localize these. They all already have existing strings. The choices should all be `PF2E.Trait<name>` and the prompt will be under `PF2E.SpecificRule.Prompt.DamageType`

TikaelSol:
The pipeline is failing because it's unable to find a link to something. When you make multiple new items that link to each other you want to extract both compendiums twice to make sure all links get resolved (basically first extract finds the new link, but doesn't know what to do with it. The second it now knows where the new items live and can make the link proper)

ammalagonc:
Re-do to make it link properly

PR #15175, opened 2024-06-21
Manage a subitems collection in actor inventory

CarlosFdez:
Doesn't do anything yet, of course. Will be used to fix some bugs later.


CarlosFdez:
An alternative take involving deeper nested uuids was merged instead, and another approach was useed to fix the sub items in choice sets bug. This is no longer necessary.

PR #15200, opened 2024-06-22
Add item and origin to appliedDamage messages

MrVauxs:
Adds information as to what item has caused damage to a token. For use with modules.

If linking the message that the damage taken message came from is of interest, I can also add that. However, I don't know where to put such property.


src/module/chat-message/document.ts
```diff
@@ -103,6 +103,14 @@ class ChatMessagePF2e extends ChatMessage {
             return item ?? null;
         }
 
+        if ("appliedDamage" in this.flags.pf2e) {
+            const origin = this.flags.pf2e?.origin ?? null;
+            const matchActor = /Actor\.(\w+)/.exec(origin?.uuid ?? "") ?? [];
+            const actor = game.actors.get(matchActor?.[1] ?? "");
+            const matchItem = /Item\.(\w+)/.exec(origin?.uuid ?? "") ?? [];
+            return actor?.items.get(matchItem?.[1] ?? "") ?? null;
+        }
```
CarlosFdez, line 112: For now can you remove this? Depending on which existing PR hack of mine for subitems gets approved, this will be handled alongside that. The rest of this looks perfectly fine.

PR #15201, opened 2024-06-22
Add engines: node to package.json

MrVauxs:
Add a node 20 dependency to the package.json to avoid issues in regards to parsing dice ([18-19 said to be broken by Tikael](https://discord.com/channels/880968862240239708/880969304365994034/1253736595056496680), and node 21 also broke my local build process.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "packages": {
         "": {
             "name": "foundry-pf2e",
-            "version": "6.0.3",
+            "version": "6.0.4",
             "license": "Apache-2.0",
             "dependencies": {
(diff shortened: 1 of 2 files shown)


CarlosFdez:
If we're setting nodes in the package.json, it should be in the engines flag. That said I'm using 21.7.2 and its functioning fine. The issue with the dice was because peggy was updated to version 4 but the code wasn't updated to handle the breaking changes.

MrVauxs:
The problem with engines is that its strictly advisory, unless we also add an `npmrc` file. 
I can change the PR to fit that format instead, I simply went with the most straightforward `npm i node -D` solution.

stwlam:
engines, no npmrc

PR #15202, opened 2024-06-22
fix typos in roar mimicry

rectulo:


TikaelSol:
Typos are in the PDF, so we should probably point Paizo to them too

PR #15212, opened 2024-06-23
Remove spans with unnecessary styles

Cuingamehtar:
Keep Stone Chunk / Ingot - header was black text on brown background. Needlessly tall table cells.
Smoke bomb - removed span that visually did nothing
Archetypes - some dedication feats had span instead of h2, there was a span with custom background color, some spans that specify a font (visible with a non-standard font)
Actor-owned ancestries - useless span with underline in Sample names
Boons - removed underline in Recall Knowledge
Internal Insurrection / Ioseff Xarwin - affliction name and level now strong (as in the sourcebook) instead of underline.
Auldegrund Grimcarver - removes span that visually did nothing.


stwlam:
Looked over the results to make sure nothing regressed?

Cuingamehtar:
I checked most of them (except for archetype dedication `h2`s) 2 weeks ago in the other PR (that I've recently closed).
I've just rechecked them and everything looks alright to me,

PR #15222, opened 2024-06-24
Use the values of other objects at runtime for determining valid dc slugs

CarlosFdez:
This is only used by action macros and its a bit confusing to understand what its being used for there. Please review @nikolaj-a when you get the chance, since I don't know how to actually test this.


nikolaj-a:
LGTM at a quick glance. I’ll do some testing once it’s merged, but I don’t anticipate any fallout.

nikolaj-a (approved):

PR #15227, opened 2024-06-24
Fix content links creation/detection for V12

stwlam:


CarlosFdez:
Can you also catch the one in grant item while you're at it? grant-item.hbs might also need the same data-link

stwlam:
Updated

PR #15232, opened 2024-06-24
Remove Conjure Bullet grant from Spellshot Dedication

### LINKED ISSUE #15231: Spellshot Dedication still grants Conjure Bullet
With the errata to Spellshot Dedication Conjure Bullet has been removed, however, the dedication feat within the system still incorrectly grants this.

pedrogrullada:
Closes #15231

Conjure Bullet is not referenced anywhere else in the system. Let me know if you want me to delete the action while I'm at it, too.

### DIFF
--- a/packs/feats/spellshot-dedication.json
+++ b/packs/feats/spellshot-dedication.json
@@ -32,10 +32,6 @@
         },
         "rules": [
             {
-                "key": "GrantItem",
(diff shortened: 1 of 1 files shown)


TikaelSol:
There's still going to be actors who reference the action on them, so it would need a migration and a migration that forces errata onto a character's items seems a bit much for me. Fine to keep the action just sitting around I think

PR #15243, opened 2024-06-25
Update choice REs for labelled objects and support choices in predicates

CarlosFdez:


src/module/actor/actions/single-check.ts
```diff
@@ -33,7 +33,7 @@ function isValidDifficultyClass(dc: unknown): dc is CheckDC | DCSlug {
     return (
         ["ac", "armor", "perception"].includes(slug) ||
         tupleHasValue(SAVE_TYPES, slug) ||
-        setHasElement(SKILL_SLUGS, slug)
+        objectHasKey(CONFIG.PF2E.skills, slug)
```
stwlam, line 36:
```suggestion
        slug in CONFIG.PF2E.skills
```

CarlosFdez: true, I forget object inclusion is not under the same restrictions as sets nor arrays, and only exists if we need to start poking in there. I changed a few others while I was at it.

PR #15254, opened 2024-06-26
Support retrieving sub items from messages (nested uuid variant)

CarlosFdez:
This does Actor.X.Item.Y.Item.Z crimes. Alternative of https://github.com/foundryvtt/pf2e/pull/15255

### DIFF
--- a/src/module/item/physical/document.ts
+++ b/src/module/item/physical/document.ts
@@ -224,6 +224,10 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         });
     }
 
+    override get uuid(): ItemUUID {
+        return this.parentItem ? `${this.parentItem.uuid}.${this.documentName}.${this.id}` : super.uuid;
+    }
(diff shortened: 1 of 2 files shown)


src/module/item/physical/document.ts
```diff
@@ -423,6 +427,28 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
+        embeddedName: string,
+        id: string,
+        { strict }: { strict: true },
+    ): foundry.abstract.Document;
+    override getEmbeddedDocument(
+        embeddedName: string,
+        id: string,
+        { strict }: { strict: false },
+    ): foundry.abstract.Document | undefined;
+    override getEmbeddedDocument(
```
stwlam, line 440: This needs the third signature overload

CarlosFdez: I tried adding it just now but it error'd, so I'm going to set the argument to options instead

image
image

PR #15262, opened 2024-06-27
Convert all uses of `R.equals` to `R.isDeepEqual`

stwlam:
Infuriating on some of these, but so it goes


CarlosFdez:
That kingmaker one is super infuriating but so it goes

PR #15284, opened 2024-06-29
Replace all uses of `R.compact` with filtering

stwlam:


src/module/actor/party/sheet.ts
```diff
@@ -212,7 +212,7 @@ class PartySheetPF2e extends ActorSheetPF2e<PartyPF2e> {
         );
 
         function getBestSkill(slug: string): SkillData | null {
-            const bestMember = R.maxBy(members, (m) => m.skills[slug]?.mod ?? -Infinity);
+            const bestMember = R.firstBy(members, (m) => m.skills[slug]?.mod ?? -Infinity);
```
CarlosFdez, line 215: firstBy returns min, but we need max here. If there's no lastBy, it seems there's a desc parameter you can use.

CarlosFdez (changes requested):

CarlosFdez (approved):

PR #15289, opened 2024-06-29
Add icons to Surki effects

ammalagonc:


TikaelSol:
I haven't checked, do these same icons exist in the core Foundry icons? We're trying to rely more on those since they are higher resolution than ours.

ammalagonc:
I just checked and these icons are unique to the PF2e system folder.

PR #15301, opened 2024-06-30
localize Soulforger dedication text and add GrantItem to Soulforger feat

ammalagonc:
I localized the Soulforger dedication text and added a GrantItem RE to the dedication so that it adds `Manifest Soulforged Armament` as part of the actions.

Still in the early stages of figuring out the best approach for applying automation to this dedication.

### DIFF
--- a/packs/feats/soulforger-dedication.json
+++ b/packs/feats/soulforger-dedication.json
@@ -28,7 +28,72 @@
             "remaster": false,
             "title": "Pathfinder Secrets of Magic"
         },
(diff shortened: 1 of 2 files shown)


TikaelSol:
Hold off doing any more automation until we have afflictions. This archetype is heavily tied to a curse, though maybe that makes more sense as an effect? Something to discuss in discord before diving into it either way

ammalagonc:
I'll put this one in Draft in the interim.

PR #15311, opened 2024-07-01
update rule element for Sparkling Targe to account for other traits

### LINKED ISSUE #15304: Shield Block does not recalculate Rule Elements for the shield
When you use Shield Block it does not seem to recalculate Rule Elements for the shield that is blocking.

For example the Magus' Sparkling Targe should add 2 to Hardness to the Steel Shield in this case (should be 10, not 8), but it isn't added automatically:

image

image

### LINKED ISSUE #15305: Sparkling Targe rule elements only check for the Magical trait
The rule elements for the Magus class feature Sparkling Targe only check for the Magical trait, but not for Arcane, Divine, Occult, or Primal, so it isn't always automatically triggered, when it could. Reporting as per https://github.com/foundryvtt/pf2e/issues/15302#issuecomment-2198633172

ammalagonc:
Closes #15304
Closes #15305

### DIFF
--- a/packs/classfeatures/sparkling-targe.json
+++ b/packs/classfeatures/sparkling-targe.json
@@ -43,11 +43,7 @@
                     "item:equipped",
                     {
                         "or": [
-                            "magical",
-                            "item:trait:arcane",
-                            "item:trait:divine",
-                            "item:trait:occult",
-                            "item:trait:primal",
+                            "item:magical"
                             "item:type:spell",
                             "defend-against-magic"
                         ]
--- a/packs/classfeatures/sparkling-targe.json
+++ b/packs/classfeatures/sparkling-targe.json
@@ -43,7 +43,7 @@
                     "item:equipped",
                     {
                         "or": [
-                            "item:trait:magical"
+                            "item:trait:magical",
                             "item:type:spell",
                             "defend-against-magic"
                         ]
--- a/packs/classfeatures/sparkling-targe.json
+++ b/packs/classfeatures/sparkling-targe.json
@@ -44,10 +44,6 @@
                     {
                         "or": [
                             "magical",
-                            "arcane",
-                            "divine",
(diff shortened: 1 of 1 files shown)


packs/classfeatures/sparkling-targe.json
```diff
@@ -44,6 +44,14 @@
...
                         "or": [
                             "magical",
+                            "arcane",
+                            "divine",
+                            "occult",
+                            "primal",
+                            "item:trait:arcane",
+                            "item:trait:divine",
+                            "item:trait:occult",
+                            "item:trait:primal",
```
TikaelSol, line 54: I'm not big on including both of these. We should probably be moving to only the `item:trait:x` ones.

iBoMbY:
This doesn't solve the issue https://github.com/foundryvtt/pf2e/issues/15304 though

packs/classfeatures/sparkling-targe.json
```diff
@@ -44,6 +44,10 @@
                     {
                         "or": [
                             "magical",
+                            "item:trait:arcane",
+                            "item:trait:divine",
+                            "item:trait:occult",
+                            "item:trait:primal",
```
stwlam, line 50:
```suggestion
                            "item:magical"
```

iBoMbY:
Also Sparkling Targe does work against all magical effects, not just item effects. The first commit was correct imho.

stwlam:
> Also Sparkling Targe does work against all magical effects, not just item effects. The first commit was correct imho.

"item" is in a foundry sense rather than a pathfinder one.

iBoMbY:
I see a different problem here though, I did try it with "item:trait:magical" and it seems always be set/true if the shield itself is magical, even though the item/effect it defends against isn't magical. I guess in this case there would be no way to differentiate between both sides.

PR #15320, opened 2024-07-01
Ensure spell slot index is valid when casting

stwlam:


CarlosFdez:
Out of curiousity, why not check if resolvedIndex is not negative 1 rather than fishing for an expended slot secondly? I'd assume if we're fishing for a slot to cast, it should fail if its expended, but perhaps there's some use case I don't have in mind.

stwlam:
It does both. Without both, the slots-expended error message would never be reached.

CarlosFdez:
I see, the if statement after it. A bit odd looking but I suppose we do want a distinction between "cast fireball without specifying a spell slot but none remains" and "cast fireball when fireball wasn't prepared".

PR #15321, opened 2024-07-01
Include `@Check` traits as item traits

stwlam:
This probably shouldn't be a longterm solution: once afflictions are in, the inline check should reference it for item roll options (including traits) exclusively.

### DIFF
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -167,10 +167,20 @@ export class InlineRollLinks {
             return;
         }
 
-        const extraRollOptions = [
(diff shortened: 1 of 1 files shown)


CarlosFdez:
We do something similar in augment damage rolls, however there we first check if its in actionTraits. This is what we do

```ts
        const rollOptions = new Set([
            ...(actor?.getRollOptions(domains) ?? []),
            ...(item?.getRollOptions("item") ?? []),
            ...traits,
            ...traits.filter((t) => t in CONFIG.PF2E.actionTraits).map((t) => `item:trait:${t}`),
            ...(extraRollOptions ?? []),
            ...actionOptions,
        ]);
```
Could a similar filter be employed here?

PR #15322, opened 2024-07-01
More consistent origin flags on templates

MrVauxs:
Adds `itemUuid` data attribute to `@Template` tags and adds `item.getOriginData()` results to templates created from inline buttons.

Example: `@Template` tag from Blazing Wave kineticist feat. Includes Fireball spell for comparison.
| Before | After |
| :-----: | :----: |
| image | chrome_yvI9kHcZya |

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -4,7 +4,6 @@ import { ActorSheetPF2e } from "@actor/sheet/base.ts";
 import { SAVE_TYPES } from "@actor/values.ts";
 import { ItemPF2e, ItemSheetPF2e } from "@item";
 import { ActionTrait } from "@item/ability/types.ts";
-import { ItemSystemData } from "@item/base/data/system.ts";
 import { EFFECT_AREA_SHAPES } from "@item/spell/values.ts";
 import { ChatMessagePF2e } from "@module/chat-message/index.ts";
 import {
@@ -236,7 +235,7 @@ class TextEditorPF2e extends TextEditor {
             case "Localize":
                 return this.#localize(paramString, options);
             case "Template":
-                return this.#createTemplate(paramString, inlineLabel, item?.system);
+                return this.#createTemplate(paramString, inlineLabel, item);
             default:
                 return null;
         }
@@ -288,7 +287,7 @@ class TextEditorPF2e extends TextEditor {
     }
 
     /** Create inline template button from @template command */
-    static #createTemplate(paramString: string, label?: string, itemData?: ItemSystemData): HTMLSpanElement | null {
+    static #createTemplate(paramString: string, label?: string, item?: ItemPF2e | null): HTMLSpanElement | null {
         // Get parameters from data
         const params = this.#parseInlineParams(paramString, { first: "type" });
         if (!params) return null;
@@ -316,7 +315,8 @@ class TextEditorPF2e extends TextEditor {
             return null;
         } else {
             // If no traits are entered manually use the traits from rollOptions if available
-            params.traits ||= itemData?.traits?.value?.toString() ?? "";
+            params.traits ||= item?.system?.traits?.value?.toString() ?? "";
+            params.itemUuid ||= item?.uuid ?? "";
 
             // If no button label is entered directly create default label
             if (!label) {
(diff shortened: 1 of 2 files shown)


src/module/system/text-editor.ts
```diff
@@ -316,7 +315,8 @@ class TextEditorPF2e extends TextEditor {
             return null;
         } else {
             // If no traits are entered manually use the traits from rollOptions if available
-            params.traits ||= itemData?.traits?.value?.toString() ?? "";
+            params.traits ||= item?.system?.traits?.value?.toString() ?? "";
```
stwlam, line 318:
```suggestion
            params.traits ||= item?.system.traits.value.toString() ?? "";
```

src/scripts/ui/inline-roll-links.ts
```diff
@@ -349,8 +349,16 @@ export class InlineRollLinks {
...
 
         const actor = resolveActor(foundryDoc ?? resolveDocument(link));
-        if (actor || pf2Traits) {
+        const item = resolveItem(foundryDoc ?? resolveDocument(link));
+        if (actor || pf2Traits || item) {
             const origin: Record<string, unknown> = {};
+            if (item) {
+                for (const [key, value] of Object.entries(item.getOriginData())) {
+                    origin[key] = value;
+                }
```
stwlam, line 358: `uuid` should just be `item`

src/scripts/ui/inline-roll-links.ts
```diff
@@ -349,8 +349,16 @@ export class InlineRollLinks {
         }
 
         const actor = resolveActor(foundryDoc ?? resolveDocument(link));
-        if (actor || pf2Traits) {
+        const item = resolveItem(foundryDoc ?? resolveDocument(link));
```
CarlosFdez, line 352: Because you injected itemUuid into the actual anchor's data, foundryDoc should already be the item. So `const item = foundryDoc instanceof ItemPF2e ? foundryDoc : null;` should be enough for your purpose without needing to declare a new helper.

Also, while you're here, please remove `?? resolveDocument(link)` from const actor. I just noticed that foundryDoc is already resolveDocument(link).

src/scripts/ui/inline-roll-links.ts
```diff
@@ -348,9 +348,26 @@ export class InlineRollLinks {
             flags.pf2e.messageId = messageId;
         }
 
-        const actor = resolveActor(foundryDoc ?? resolveDocument(link));
-        if (actor || pf2Traits) {
+        const actor = resolveActor(foundryDoc);
+        const item =
+            foundryDoc instanceof ItemPF2e
+                ? foundryDoc
+                : foundryDoc instanceof ChatMessagePF2e
```
CarlosFdez, line 355: Is there a reason you need to check for ChatMessagePF2e? The suggestion I made was to completely omit checking it. The only reason we check for it elsewhere is partly in case there isn't an item, and partly due to legacy. If we do have to check in chat message, that will likely apply to other things such as inline checks, and we would need the resolveItem.

CarlosFdez: Yup, I see it. Line 246 in the current inline-roll-links, inside #onClickInlineCheck. So long as you refactor that one, then you can have resolveItem().

src/scripts/ui/inline-roll-links.ts
```diff
@@ -348,9 +348,26 @@ export class InlineRollLinks {
...
+                  ? foundryDoc.item
+                  : null;
+        if (actor || pf2Traits || item) {
             const origin: Record<string, unknown> = {};
+            if (item) {
```
CarlosFdez, line 360: I double checked how origin is used elsewhere, including in strikes. item.getOriginData() is used instead of everything else if it exists. I would instead rewrite the entire origin assignment into an if/else, if there's an item, set flags.pf2e.origin to item.getOriginData()

src/scripts/ui/inline-roll-links.ts
MrVauxs: As the commit mentioned it was untested, so I wasnt entirely sure what types `foundryDoc` could be at the time.

Having looked at it in Foundry right now, this is indeed not needed.

src/scripts/ui/inline-roll-links.ts
MrVauxs: So to summarize, rewrite this to be `if (actor || pf2Traits) { ... } else if (item) { ... }` and get rid of stwlam's initial `uuid -> item` override?

CarlosFdez: almost, but item first.

MrVauxs: right.

PR #15355, opened 2024-07-04
Automate Concussive trait via damage "redirection"

stwlam:


CarlosFdez:
Could ignored resistances be set such that omitting max is infinity, or would that look weird in the implementation?

stwlam:
Maybe were it not all raw objects, but as it stands it would just require a bunch of new nullish coalescing for numeric operations

src/module/system/damage/iwr.ts
```diff
@@ -225,6 +248,38 @@ function applyIWR(actor: ActorPF2e, roll: Rolled<DamageRoll>, rollOptions: Set<s
...
+    highest,
+    candidates,
+}: {
+    resistances: WorkingResistanceData[];
+    highest: WorkingResistanceData;
+    candidates: RedirectedResistance[];
+}): { to: WorkingResistanceData | null; redirect: RedirectedResistance } | null {
+    return candidates
+        .filter((rr) => [highest.type, null].includes(rr.from) && rr.to !== highest.type)
+        .reduce(
```
CarlosFdez, line 266: Would this function be clearer if the comparison was made relative to `const highestValue = Math.max(highest.value, mostReduction?.to?.value ?? 0);`, returning mostReduction (even if its null) if not?

PR #15356, opened 2024-07-04
Update Typescript to 5.5.3

In3luki:
The type assertions are still necessary to prevent the `Type instantiation is excessively deep and possibly infinite.ts(2589)` errors.
Feel free to close this if there is a better way to handle it.


stwlam:
I think we'll be able to not have these type assertions once all actor/item subtypes are data models.

PR #15367, opened 2024-07-05
Add CSS root variable for the parchment background

nikolaj-a:
Useful for modules to use in a way that is easily modifiable by UI modules.

Example:
```css
background: var(--background-primary);
```


CarlosFdez:
Perhaps this is fine if it gets put to use, but I wonder if backgrounds are a one size fits all sort of deal. Rather than background primary, I wonder if a different name might not work better.

What's the goal of this PR, more specifically?

Issue #15349, opened 2024-07-03
[Bug] Party actors in folders are not visible


HeliumAnt:
Party actors can be added to folders, though this is not apparent since they're always shown at root level.
This means you can **delete Parties by accident** when deleting a folder (delete all), since there is no indication that the party actor may be contained in the folder. 
This _almost_ never deletes the Actors in the Party (Though (ofc) it did when I first discovered, deleting all PCs from my main game (very scary, all backed up thankfully), haven't been able to reproduce)

(Likely applies to any version of pf2e, tested 6.0.4, and indirectly 5.4ish, happens even in a blank world with no modules enabled)

**To reproduce**:
1. Create a party actor (and some PCs)
2. Drag party onto a folder (or just create a party actor inside a folder)
3. Nothing happens?
4. Delete (Delete All) the folder
5. The party is gone (any contained actors are moved to the root (usually))

PR #15369, opened 2024-07-06
Prevent parties from being created or moved into folders

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/15349

### DIFF
--- a/src/module/actor/party/document.ts
+++ b/src/module/actor/party/document.ts
@@ -250,11 +250,23 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
     /*  Event Handlers                              */
     /* -------------------------------------------- */
 
+    protected override _preCreate(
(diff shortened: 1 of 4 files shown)


src/module/actor/party/document.ts
```diff
@@ -255,6 +255,11 @@ class PartyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         options: PartyUpdateOperation<TParent>,
         user: UserPF2e,
     ): Promise<boolean | void> {
+        // Prevent party actors from being dragged to folders
+        if (changed.folder) {
+            changed.folder = null;
+        }
```
stwlam, line 261:
```suggestion
        changed.folder &&= null;
```

CarlosFdez: Not quite the same thing, but it also works and is one line.

PR #15379, opened 2024-07-06
Add `splitListString` helper for consistent splitting of roll link parameters

stwlam:


CarlosFdez:
For release note purposes, what does this fix? It looks fine to me, and you even caught the adjustments bit.

PR #15381, opened 2024-07-07
Restore z-cycling of token stacks

stwlam:


CarlosFdez:
The persisting is a bit odd, but I trust you on it, that's probably what foundry does for [ ] I imagine.

PR #15400, opened 2024-07-10
Fix regression from chat messages referring to compendium items

CarlosFdez:
As far as I know, they only print red in the console, but better not print that at all.


stwlam:
What's the regression?

CarlosFdez:
Foundry complains that you can't use fromUUIDSync on compendium items. It was an issue introduced recently after uuid lookups became a feature (before it wouldn't return anything).

PR #15402, opened 2024-07-10
Force party actors to retain actor link

CarlosFdez:
Prevents infinite recursion during initialization

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -3770,6 +3770,7 @@
         "Token": {
             "Appearance": "Appearance",
             "Flanking": "Flanking",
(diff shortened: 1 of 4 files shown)


static/lang/en.json
```diff
@@ -3770,6 +3770,7 @@
         "Token": {
             "Appearance": "Appearance",
             "Flanking": "Flanking",
+            "ActorLinkForced": "{type} actors cannot become unlinked",
```
stwlam, line 3773:
```suggestion
            "ActorLinkForced": "{type} actors cannot be unlinked.",
```

PR #15428, opened 2024-07-13
Change styling of stowed strikes toggle

CarlosFdez:
Feel free to close if you think its not an improvement. I was worried that the previous button didn't properly communicate what it was.

image

image


stwlam:
On second thought, maybe I can live with "show stowed" :o)

stwlam:
Though could we have be somewhat smart and not show at all if there aren't any stowed weapons?

PR #15450, opened 2024-07-15
Create drawAuras hook

MrVauxs:
Creates 3 informational hooks, `deletedAuras`, `createdAuras`, and `updatedAuras`. Hopefully self-explanatory.
The arguments they give are the `AuraRenderers` object itself, and then an array of slugs of the auras that were somehow changed.

Primarily for use with the WIP PF2e Graphics module, such as animating size-changing auras. Ex. Issue https://github.com/MrVauxs/pf2e-jb2a-macros/issues/157

### DIFF
--- a/src/module/canvas/token/aura/map.ts
+++ b/src/module/canvas/token/aura/map.ts
@@ -19,6 +19,7 @@ export class AuraRenderers extends Map<string, AuraRenderer> {
      * @param [slugs] A specific list of slugs to limit which auras are cleared
      */
     async reset(slugs?: string[]): Promise<void> {
+        const preUpdateSlugs = Array.from(this.keys());
         if (!slugs) {
             this.clear();
         } else {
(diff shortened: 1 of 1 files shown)


stwlam:
augury for this says "woe"

CarlosFdez:
I might be amenable if there's no other recourse to a drawAuras hook that passes in the previous slugs as an argument, but doing the computation ourselves in what is already a performance sensitive code just in case a module uses it isn't something I could really agree with.

MrVauxs:
> I might be amenable if there's no other recourse to a drawAuras hook that passes in the previous slugs as an argument, but doing the computation ourselves in what is already performance sensitive code just in case a module uses it isn't something I could really agree with.

So using this PR as an example, it would be a hook that provides the `preUpdateSlugs` and the `token` that's called after `this.draw()`?

So something akin to
```js
 async reset(slugs?: string[]): Promise<void> {
        const preUpdateSlugs = Array.from(this.keys());
        
        ...
        
        return this.draw().then(() => Hooks.callAll("drawAuras", this.token, preUpdateSlugs));
```

CarlosFdez:
I would await instead of using then, but yes, though perhaps in an object, so `Hooks.callAll("drawAuras", this.token, { preUpdateSlugs })`.

I don't mind if its before or after, whichever follows the pattern of most draw/render hooks.

MrVauxs:
I would use await as well if it was possible to have code run after returning it lol
And I don't know if anything relies on `reset()` returning that yet.

CarlosFdez:
its a Promise<void>. The only reason we return is because returning a promise is the same as awaiting.

MrVauxs:
> The only reason we return is because returning a promise is the same as awaiting.

TIL

Making the changes.

CarlosFdez:
As awkward as it may be to work around, remove the preUpdateSlugs. It has more of a chance going through without it. I'll do research on existing hooks to see if we need to change the name as well.

MrVauxs:
As a note, I have removed the preUpdateSlugs. The changeset is now just adding a `Hooks.callAll("drawAuras", this.token);` at the end.

CarlosFdez:
I've reconnected after, and without the diffing of before/after or some way to even store what it was before, this PR has no use. It would be better served by a post prepared hook, which would be able to handle this situation. I'd recommend asking core foundry for such a hook though.

PR #15454, opened 2024-07-15
localize Elementalist Class Archtype spellshape feats

ammalagonc:
- Added RE to the remaining Elementalist spellshape feats.
- Localization added to re-en.json


ammalagonc:
Close to re-do - unwanted change snuck in

PR #15473, opened 2024-07-18
Add mergeable to Debilitating Strike

MrVauxs:
... enabling the addition of homebrew debilitations.


MrVauxs:
Apparently this only works visually. Subsequent RE's Subchoices appear on the sheet but are not actually used.
chrome_Lo6a2XuVXb

TikaelSol:
Fine for now but we should move these to the individual features as a follow up

PR #15475, opened 2024-07-18
Keep last-used animation for tokens for later use

stwlam:


CarlosFdez:
animation is an object, so wouldn't === not behave as expected?

stwlam:
o yah I initially just had it retain the transition (a string)

PR #15476, opened 2024-07-18
Start refactoring character sheet get data

CarlosFdez:
This should make it easier to follow along and also create room to start adding new things.

Some more refactoring needs to be done, but it'll require more intrusive updates to other creature sheets.

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -119,13 +119,6 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             if (tab) tab.initial = "biography";
         }
 
-        sheetData.numberToRank = R.mapToObj([0, 1, 2, 3, 4] as const, (n) => [
-            n,
-            game.i18n.localize(`PF2E.ProficiencyLevel${n}`),
-        ]);
-
-        sheetData.senses = condenseSenses(this.actor.perception.senses.contents);
-
         // Attacks and defenses
         // Prune untrained martial proficiencies
         for (const section of ["attacks", "defenses"] as const) {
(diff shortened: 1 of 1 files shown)


stwlam:
This looks like it's all thrown in the returned object, making it harder to read.

CarlosFdez:
Its generally easier to read what data we're returning and where its from when its part of a final return object, even more so after some of the more complicated sub-sections have been moved to private helpers. Significantly better than getting the parent data and casting it to the new return type and then bolting things on there and in helper methods.

Unfortunately, this isn't complete. There's still that weirdness with how saves and proficiencies are set up. Skills and saves can be lifted up to the creature level, and we should choose between whether to use short (character) or labelShort (NPC) for saves.

CarlosFdez:
....in short, please read character sheet getData(), and then party sheet or deity sheet getData(), and let me know which is actually easier to follow along and get a sense of the structure. I was making the former like the latter, but we can instead make everything like character instead if we prefer it that way.

stwlam:
While not asserting is nice, giving up separation of individual components of the sheet data isn't. There's a mix of both here, making for a net loss in readability.

CarlosFdez:
I find that claim pretty tough. I could not follow along at all when I tried to add has-stowing to it. Its scattered, and having no type safety was scary on top of that. While there's more work to be done on that front, it definitely went from a 5 to a 20 when the scale is 100.

I could further the separations with some more info on where its problematic, perhaps with a more thorough refactor than I was planning for first pass. But let me know if you'd prefer the character sheet's paradigm, because I can convert all newer sheets to that style instead. Currently, my thoughts on where its problematic is that I should separate out proficiencies and attribute boost preparation, and I definitely do need to lift saves and skills higher up. And reorganize the ordering. Most of the ordering on the final data is actually the original ordering in the getData function (it was super scattered).

CarlosFdez:
Ah, and side effects. I had planned to get rid of all side effects and mutations on base sheet data. That's always confusing to trace back.

CarlosFdez:
Anyways, I'm going to need something actionable. I can either take this further, or consider the merge workflow we're doing elsewhere a lost cause and revert to this style. https://github.com/foundryvtt/pf2e/pull/15482 is a trial run converting other things to be like character sheet instead (but it still has the prepareX functions. Character sheet does not split its processing).

CarlosFdez:
I will take another crack at this in the future, but I'll need to do so from scratch.

PR #15481, opened 2024-07-19
undo breaking changes in en.json

ammalagonc:
I caught this one while checking on an issue. This was a miss on my part.

As it turns out, changing the whole container breaks the labels on the drop-down menu in the Property Runes section. 

Will see about possibly updating this with proper guidance once we get over the PC2 rush

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -5532,12 +5532,6 @@
             "greaterDecaying": {
                 "Name": "Greater Decaying"
             },
-            "greaterDisrupting": {
-                "Name": "Greater Vitalizing",
-                "Note": {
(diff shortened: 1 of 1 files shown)


static/lang/en.json
```diff
@@ -5600,6 +5594,12 @@
                     "success": "Sonic damage dealt by this weapon ignores the target's sonic resistance."
                 }
             },
+            "greaterDisrupting": {
+                "Name": "Greater Vitalizing",
+                "Note": {
+                    "criticalSuccess": "On a critical hit, the undead creature is @UUID[Compendium.pf2e.conditionitems.Item.MIRkyAjyBeXivMa7]{Enfeebled 1} and @UUID[Compendium.pf2e.conditionitems.Item.e1XGnhKNSQIm5IXg]{Stupefied 1} as long as it has the persistent damage from this rune."
+                }
+            },
```
stwlam, line 5602: ?

ammalagonc: Oh

That section got moved. Hold on. Let me fix this.

ammalagonc: That part is reverted. Should no longer ping as a change.

PR #15483, opened 2024-07-19
Fix unicode word boundaries in `sluggify`

### LINKED ISSUE #15341: System `sluggify` handling differs from JS property -> html attribute handling in the browser
This may or may not be a thing the system cares about, but it's something Idle noticed here: https://discord.com/channels/880968862240239708/880969304365994034/1258060472750702694

```js
element.dataset.actionId = 'blah'
// produces <element data-action-id="blah">
const attrToFind = `data-${game.pf2e.system.sluggify('actionId')}`;
// produces "data-actionid"
```
Tested in Firefox 127 and Chrome 126

As I mention in discord this seems to be because the `nonWordBoundry` regex positive lookahead means it will never match the final character of a string by itself. Just adding `|$` to the lookahead seems to make it match the browsers' behaviour:

And I can't see how that would cause other incorrectness, but I'm not a regex master by any definition, so could easily be missing a nuance.

stwlam:
Closes #15341


CarlosFdez:
Regex not my forte, so I be blind mergy

PR #15488, opened 2024-07-19
Remove redundant construction of enriched content patterns

stwlam:


CarlosFdez:
I'm not sure why this is a bugfix instead of a refactor, but it does look better to me.

Issue #15507, opened 2024-07-21
`Show Basic Unarmed Attack` isn't taken into account


reonZ:
Not sure if this is from the `6.1.0` or before, but unchecking the setting doesn't hide the attack from the sheet anymore.

PR #15513, opened 2024-07-21
Fix hiding strikes in character sheet

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/15507

### DIFF
--- a/static/templates/actors/character/tabs/actions.hbs
+++ b/static/templates/actors/character/tabs/actions.hbs
@@ -36,9 +36,7 @@
 
                     <ol class="actions-list item-list directory-list strikes-list" data-strikes>
                         {{#each data.actions as |strike ai|}}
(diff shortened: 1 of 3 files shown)


static/templates/actors/character/tabs/actions.hbs
```diff
@@ -36,9 +36,9 @@
 
                     <ol class="actions-list item-list directory-list strikes-list" data-strikes>
                         {{#each data.actions as |strike ai|}}
-                            {{#unless (and strike.item.isStowed @root.actor.flags.pf2e.hideStowed)}}
+                            {{#if strike.visible}}
```
stwlam, line 39: This may break something that depends on index matching (though I've tried to eliminate that before)

CarlosFdez: I'll trust you on that. Easy enough to go back on this.

PR #15514, opened 2024-07-21
Adjust drag measurement feature to play nice(r) with modules

stwlam:


CarlosFdez:
Might need a longer term solution later. Somehow. Not sure how.

PR #15522, opened 2024-07-21
Add traits and roll-options parameters to /act inline links

nikolaj-a:
```
[[/act demoralise roll-options=some:other:option traits=bravado]]
```
Or using the `options` alias:
```
[[/act demoralise options=some:other:option traits=bravado]]
```


stwlam:
In `@Check` and `@Damage`, we've moved to separating traits and roll options, with the former being validation. That should be reproduced here.

nikolaj-a:
> In `@Check` and `@Damage`, we've moved to separating traits and roll options, with the former being validation. That should be reproduced here.

Yep, I'll make a separate PR for the traits.

PR #15533, opened 2024-07-22
Restore `allowInteractive` to rolls not evaluated by `ChatMessage` initializer

stwlam:


CarlosFdez:
This looks fine to me, though an idea I just had is that perhaps CheckRoll/DamageRoll could override evaluate to add that data maybe, since its a behavior specific to those classes.

CarlosFdez (approved):

stwlam:
Rolls don't known their own mode: it's chat message data

PR #15543, opened 2024-07-22
Replace Disrupting with Vitalizing rune

### LINKED ISSUE #15472: Vitalizing Rune
A minor thing, but all weapon property runes are sorted alphabetically in the dropdown on weapons, except for the Vitalizing Rune, which appears in the "D" part of the list.

image

Duncan-Henry:
I refactored a lot of existing references to "Disrupting" and "Greater Disrupting" to be more in line with the remaster. I left OGL names to vitalizing as "Disrupting" as that it what it was called pre-remaster, but this bit of tech debt should make things a little cleaner as more content gets added.

This should also help resolve #15472 as the rune listing has been adjusted.

Let me know if there's any big references I might have overlooked.


src/module/item/weapon/values.ts
```diff
@@ -59,7 +59,6 @@ const WEAPON_PROPERTY_RUNE_TYPES = new Set([
     "deathdrinking",
     "decaying",
     "demolishing",
-    "disrupting",
```
TikaelSol, line 62: We would need to migrate this change

PR #15552, opened 2024-07-23
Implement "rules-based snapping" for lines

stwlam:
comically bad UX

[video](https://github.com/user-attachments/assets/53f3685d-c4f1-4aaf-817a-6076ea01261a)


FolkvangrForgent:
Does this take into account lines that are more than 5ft wide? Looking at the code I am guessing it will cause issues for any lines with an even multiple of five.

PR #15616, opened 2024-07-28
Adjust Hex Grid Logic for MeasuredTemplatePF2e

### LINKED ISSUE #15615: [BUG] MeasuredTemplatePF2e get snappingMode returns incorrect snapmodes on Hex Grid
As a residence hex grid user I was tracking down odd snapping behavior of template on hex grids. It looks like the `get snappingMode` function in `MeasuredTemplatePF2e` is written with the expectation that it is called from a square grid but it can also be called from a hex grid via the chat message button. As far as I can tell it still can't be called from gridless.

I would suggest replacing the current code:
https://github.com/foundryvtt/pf2e/blob/6e634270a883320cbb30c11a9bb0fc08665bf80d/src/module/canvas/measured-template.ts#L34-L45
with something like the following code:
```
        if (canvas.grid.isSquare) {
            switch (this.areaShape) {
                case "burst":
                    return M.CORNER;
                case "emanation":
                    return M.CENTER;
                case "cone":
                    return M.CENTER | M.CORNER | M.SIDE_MIDPOINT;
                case "line":
                    return M.SIDE_MIDPOINT | M.CORNER;
                default:
                    return M.CENTER | M.CORNER;
            }
        }
        if (canvas.grid.isHexagonal) {
            switch (this.areaShape) {
                case "burst":
                    return M.VERTEX;
                case "emanation":
                    return M.CENTER;
                case "cone":
                    return M.CENTER | M.VERTEX;
                case "line":
                    return M.CENTER;
                default:
                    return M.CENTER | M.VERTEX;
            }
        }
```
(I'll admit that I have left the `"cone"` to the default of ` M.CENTER | M.VERTEX` as I do not know if a more restrictive set of modes would be cleaner for hex grids)

Additionally with the `snappingMode` working on hex grids the following code:
https://github.com/foundryvtt/pf2e/blob/6e634270a883320cbb30c11a9bb0fc08665bf80d/src/module/canvas/layer/template.ts#L19-L19
can be changed to:
```
        if (!template || canvas.grid.isGridless) { 
```

I will lastly note that there is a current MR #15552 that add more square grid dependent logic.

FolkvangrForgent:
- add grid type check to `MeasuredTemplatePF2e` `snappingMode` function
- use appropriate `GRID_SNAPPING_MODES` for templates on hex grids
- use `snappingMode` on hex grids instead of using default implementation
- [BREAKING] remove `hasHexGrid` from `ScenePF2e` as `isHexagonal` serves the same purpose

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -41,9 +41,11 @@ export class TemplateLayerPF2e<
         const dimensions = canvas.dimensions;
 
         // Snap the destination to the grid
-        const { x, y } = canvas.grid.getSnappedPoint(destination, {
-            mode: template.snappingMode,
-        });
+        const { x, y } = canvas.grid.isSquare
+            ? canvas.grid.getSnappedPoint(destination, {
+                  mode: template.snappingMode,
+              })
+            : super.getSnappedPoint(destination);
         destination.x = x;
         destination.y = y;
         const ray = new Ray(origin, destination);
@@ -130,9 +132,11 @@ export class TemplateLayerPF2e<
                 const { document, position } = preview;
                 this.#deactivatePreviewListeners(initialLayer, event);
                 document.updateSource(
-                    canvas.grid.getSnappedPoint(position, {
-                        mode: preview.snappingMode,
-                    }),
+                    canvas.grid.isSquare
+                        ? canvas.grid.getSnappedPoint(position, {
+                              mode: preview.snappingMode,
+                          })
+                        : super.getSnappedPoint(position),
                 );
                 canvas.scene?.createEmbeddedDocuments("MeasuredTemplate", [document.toObject()]);
             },


stwlam:
Given there are no rules for hex snapping, I'd be inclined to say we should defer to upstream.

FolkvangrForgent:
I would argue that `burst` and `emanation` are fairly well defined in the area text and work as a good default in lack of a more specific rule for hex grids. I can adjust them to use the upstream's default flags if you still want. I'll go ahead and update `line` to upstream's default flags.

stwlam:
Thanks, but I don't think we're willing to continue maintaining hex-grid house rules in the system. This is going to have to be globally kicked upstream.

FolkvangrForgent:
Okay np, I'll make the change to return `M.CENTER | M.VERTEX` which is the default in fvtt's `getSnappedPoint` if we are on a hex grid. Any idea how the `snappingMode` function can be hooked from a module to refine the behavior?

FolkvangrForgent:
Just notice your MR #15620 which fixes the big issue. This MR now only does the following two things:
- use `snappingMode` on hex grids
- [BREAKING] remove `hasHexGrid` from `ScenePF2e` as `isHexagonal` serves the same purpose

src/module/canvas/layer/template.ts
```diff
@@ -16,7 +16,7 @@ export class TemplateLayerPF2e<
     /** Overriden to snap according to the dragged template's type */
     override getSnappedPoint(point: Point): Point {
         const template = this.preview.children.at(0);
-        if (!template || !canvas.grid.isSquare) {
+        if (!template || canvas.grid.isGridless) {
```
stwlam, line 19: We'll want to keep this so that hex snapping is retrieved from upstream.

FolkvangrForgent: The reason for using snappingMode on hex grids is that other parts of the code already do so, so avoiding it's use in this one case can lead to the wrong assumption about it's use. I can alternately go to each use of snappingMode and add if checks so it is only called on square grids.

FolkvangrForgent:
Just notice your MR #15619  which fixes the big issue. This MR now only does the following things:
- use `snappingMode` on hex grids as other parts of the code already do so, so avoiding it's use in this one case can lead to the wrong assumption about it's use

src/module/canvas/layer/template.ts
FolkvangrForgent: `_getGridHighlightPositions` currently uses snappingMode on hex
`_onDragLeftMove` currently uses snappingMode on hex
`activatePreviewListeners` currently uses snappingMode on hex

stwlam: Those will all need to follow suit.

FolkvangrForgent: Okay I'll go ahead and make sure that hex does not use `snappingMode`

FolkvangrForgent:
This MR now only does the following things:
- Fix uses of snappingMode on hex grids as it is a square grid only function

src/module/canvas/layer/template.ts
FolkvangrForgent: Okay I changed the two locations that use snappingMode when not on square... Let me know if you want me to format it differently.

FolkvangrForgent:
Also this is now more a of bugfix PR not an enhancement PR if that matters.

PR #15627, opened 2024-07-29
Remove Redundant Grid Type Logic

FolkvangrForgent:
The change in #15624 made one of the check I added useless so this MR seeks to remove it.

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -41,11 +41,9 @@ export class TemplateLayerPF2e<
         const dimensions = canvas.dimensions;
 
         // Snap the destination to the grid
-        const { x, y } = canvas.grid.isSquare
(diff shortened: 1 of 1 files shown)


src/module/canvas/layer/template.ts
```diff
@@ -41,11 +41,9 @@ export class TemplateLayerPF2e<
...
 
         // Snap the destination to the grid
-        const { x, y } = canvas.grid.isSquare
-            ? canvas.grid.getSnappedPoint(destination, {
-                  mode: template.snappingMode,
-              })
-            : super.getSnappedPoint(destination);
+        const { x, y } = canvas.grid.getSnappedPoint(destination, {
+            mode: template.snappingMode,
+        });
```
stwlam, line 46:
```suggestion
        const { x, y } = canvas.grid.getSnappedPoint(destination, { mode: template.snappingMode });
```

PR #15638, opened 2024-07-29
Support suppressing action chat message creation

nikolaj-a:


src/module/actor/actions/simple.ts
```diff
@@ -73,10 +73,11 @@ class SimpleActionVariant extends BaseActionVariant {
         });
         const results: SimpleActionResult[] = [];
         for (const actor of actors) {
-            const message = await ChatMessage.create({
+            const data = {
                 flavor,
                 speaker: ChatMessage.getSpeaker({ actor }),
-            });
+            };
+            const message = options.message?.create ? await ChatMessage.create(data) : new ChatMessage(data);
```
stwlam, line 80: defaults to false?

nikolaj-a: That's what you get for making last second changes. Good catch!

PR #15639, opened 2024-07-29
Fixed icon paths in kingdom builder when not served from the root

HeliumAnt:


CarlosFdez:
To be more specific, broken on forge?

stwlam:
This would be broken on forge but also any server in which foundry isn't serving from the route of the hostname.

PR #15640, opened 2024-07-29
Add optional `testDomains` field to AE-likes

### LINKED ISSUE #15632: Kineticist Weapon Infusion roll options no longer work
Foundry_Virtual_Tabletop_zDilSYgNDq

Probably related to the recent change (the one that modified their domain) but both roll options are currently doing nothing.

stwlam:
Closes #15632


CarlosFdez:
I can't think of better, and including all domains is likely to introduce problems. domains instead of testDomains might also communicate the wrong message.

PR #15648, opened 2024-07-31
add max dice to Dice Number formula for Amped Redistribute Potential

### LINKED ISSUE #15524: REdistrubute Potential (psychics) Amped not functioning properly.
redistriubute potential when amped is selected isnt heigthening the spell damage properly. 

https://2e.aonprd.com/Spells.aspx?ID=1138

ammalagonc:
Closes #15524

### DIFF
--- a/packs/classfeatures/the-oscillating-wave.json
+++ b/packs/classfeatures/the-oscillating-wave.json
@@ -241,7 +241,7 @@
                 "hideIfDisabled": true,
                 "key": "DamageDice",
                 "override": {
-                    "diceNumber": "2*@spell.rank",
+                    "diceNumber": "max(6,2*@spell.rank - 4)",
                     "dieSize": "d6"
                 },
                 "predicate": [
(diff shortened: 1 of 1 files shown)


packs/classfeatures/the-oscillating-wave.json
```diff
@@ -241,7 +241,7 @@
                 "hideIfDisabled": true,
                 "key": "DamageDice",
                 "override": {
-                    "diceNumber": "2*@spell.rank - 4",
+                    "diceNumber": "2*@spell.rank",
```
pedrogrullada, line 244: The formula is correct. What you're doing here would be giving 10d6 at rank 5 instead of 6d6.

```suggestion
                    "diceNumber": "max(6,2*@spell.rank - 4)",
```
It does throw a warning if you try to amp the spell at a spell rank lower than 5 because it resolves as a negative number. This should never happen because you only gain this unique Psi Cantrip at level 10, but go ahead and add a `max` here while you're at it so people don't trip up over that. I suspect that's what the one who reported the issue did.

PR #15651, opened 2024-07-31
Disable `RulerPF2e` when `drag-ruler` module is enabled

stwlam:


kanongil (commented):
The main issue seems to be that subclassing of this custom ruler is broken since the subclassed class [is used to access the private `#dragMeasurement` field](https://github.com/foundryvtt/pf2e/blob/0d81e5c0b78eb0bd76f609ea3564f398334f1e34/src/module/canvas/ruler.ts#L9C22-L9C37).

This can be fixed by replacing `this.#dragMeasurement ` with `RulerPF2e.#dragMeasurement `.

PR #15652, opened 2024-07-31
Replace inline roll defense param for against and support all statistics

CarlosFdez:


src/module/system/text-editor.ts
```diff
@@ -663,10 +663,11 @@ class TextEditorPF2e extends TextEditor {
                 pf2Roller: params.roller || null,
                 targetOwner: params.targetOwner,
                 pf2Check: sluggify(params.type),
+                pf2Against: params.against,
```
stwlam, line 666: no pf2 :o)

PR #15529, opened 2024-07-22
add separate effect for aura to the Bane spell

ammalagonc:
I was in a session of Kingmaker and noticed the spell didn't have an aura :|

There is no fancy automation, this is more for tracking purposes to see how far the Aura has been extended via sustain.

PR #15654, opened 2024-07-31
Add rule element to Eternal Bane

ezubairov:
Adds a rule element to Eternal Bane that adds aura that applies Bane spell effects to enemies in 15 feet, as per Player Core.


stwlam:
Duplicate of #11374 and #15529

PR #15707, opened 2024-08-02
Add support for homebrew shield traits

CarlosFdez:
Also lockout existing armor/consumable traits from equipment traits

### DIFF
--- a/static/lang/en.json
+++ b/static/lang/en.json
@@ -3345,7 +3345,7 @@
                     "Label": "Label"
                 },
                 "EquipmentTraits": {
-                    "Hint": "Traits applicable to equipment other than weapons",
+                    "Hint": "Traits applicable to equipment, armor, and consumables",
                     "Name": "Equipment Traits"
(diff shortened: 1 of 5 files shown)


static/lang/en.json
```diff
@@ -3345,7 +3345,7 @@
                     "Label": "Label"
                 },
                 "EquipmentTraits": {
-                    "Hint": "Traits applicable to equipment other than weapons",
+                    "Hint": "Traits applicable to equipment, armor, and consumables",
```
stwlam, line 3348: shields, too, now? approaching "equipment other than weapons" :grin:

CarlosFdez: nah, I made a separate one for shield. I thought about it a bit more and propagation works best when propagating from a child to a parent that are in a is-a relationship. Shield is kinda divorced and has too many unique traits.

Armor should probably also have been divorced originally, but I wasn't about to change it.

PR #15709, opened 2024-08-02
Automate player core 2 improvised pummel

n1xx1:
So this is kind of a mess. Probably with some more features in the rule elements (not sure about which ones) it could be improved. The reason it's like this is because I didn't want to write out every single rune, otherwise it would be a nightmare to update every time a new property rune is added. 

The handwraps rule elements set some roll options to specify that the actor is using the handwraps (`handwraps-of-mighty-blows`) and which property runes are filled (`handwraps-rune-property0`, `handwraps-rune-property1`, ...). They also add some flags, i.e. potency and striking rune level and the property runes for each slot.

The Improvised Pummel rule elements functions as normal WeaponPotency that uses the max between what the feat gives and what the handwraps give. For the DamageDice it uses either what the feat does or the handwraps striking when `improvised-pummel-use-handwraps` RollOption is on. Also if the RollOption is on then it also applies the property runes. This should pretty much do exactly what the feat does.

It's a draft because maybe there is a better way to do it with less rule elements, or maybe new predicates that can check the items the actor have?

### DIFF
--- a/packs/equipment/handwraps-of-mighty-blows.json
+++ b/packs/equipment/handwraps-of-mighty-blows.json
@@ -51,7 +51,78 @@
         "reload": {
             "value": ""
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "ActiveEffectLike",
+                "mode": "override",
+                "path": "flags.pf2e.handwrapsOfMightyBlows.potency",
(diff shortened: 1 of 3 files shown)


TikaelSol:
I would much rather opt for a cleaner solution than this

PR #15739, opened 2024-08-03
Add Abysium Material damage

reyzor1991:
Screenshot 2024-08-03 at 12 44 13

Screenshot 2024-08-03 at 12 44 28


stwlam:
Not how we're doing it, sorry.

reyzor1991:
I use same logic as for runes

If you say me how to do it, I will rewrite

PR #15756, opened 2024-08-03
Rename Sulfur Bomb effect

### LINKED ISSUE #15752: Sulfur Bomb effect name
Sulfur bombs have an effect is applied on a hit, but the label for the effect says "sulfur bomb failure". There isnt really any saving throw involved tho.
image
image

pedrogrullada:
I believe it's clear enough that the effect is meant to be on the successful, but not critically successful hit, but I'm open to suggestions on renaming it something else. Like `Sulfur Bomb (Success)` or `Sulfur Bomb (Hit)` perhaps

Closes #15752


pedrogrullada:
woops, will fix the conflicts in a moment

PR #15773, opened 2024-08-04
Add effects for Shields of the Spirit and Security feats

pedrogrullada:
I'm hoping the UX is good enough, the idea is:
- The spell effect for Shields of the Spirit will grant the effect to allies within Champion's Aura if dragged onto the Champion, but its AC bonus will still function if dragged onto an ally directly.
- The effect for the Security version of the spell is meant to be dragged onto one specific ally, which I'm hoping the text of the feat will imply.


packs/spell-effects/spell-effect-shields-of-the-spirit-security.json
```diff
@@ -0,0 +1,58 @@
...
+                "key": "FlatModifier",
+                "selector": "ac",
+                "type": "status",
+                "value": 1
+            },
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "parent:origin:item:tag:greater-security",
+                    "greater-security-bonus:true"
```
stwlam, line 38: true?

pedrogrullada: It's using the Roll Option rule element set up right above. The effect doesn't have the `origin:shield:raised` roll option, so this was a way to tell wether the origin has their shield raised or not.

PR #14960, opened 2024-06-04
Show post roll options in the inspector and include in notes

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/15457

Also supercedes https://github.com/foundryvtt/pf2e/pull/15784 (which was created afterwards)

[
image](https://github.com/foundryvtt/pf2e/issues/15457)

PR #15784, opened 2024-08-04
Add `check:total:...` roll options for Note REs, and add them to the resulting ChatMessage context aswell

### LINKED ISSUE #15457: Expand `check:total:natural` to more Rule Elements
A suggestion to expand the pre-existing `check:total:natural` roll options that are only available on `AdjustDegreeOfSuccess` elements to also apply to other REs, **primarily `Note`s and `FlatModifier`s.**

This would be great automation for abilities that trigger off of rolling a "natural" number, such as a nat 1 (`check:total:natural:1`).

There is an example of Paizo abilities using natural dice results as well, [`Ageless Patience`](https://2e.aonprd.com/Feats.aspx?ID=4412&Redirected=1), currently handled manually and appearing on all rolls regardless of result.

n1xx1:
Fixes #15457.

This also makes available the natural dice result as roll options on the resulting chat message, seems like @MrVauxs needed this too according to this message <https://discord.com/channels/880968862240239708/1006806454700875816/1269431811176530002>.

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -184,40 +184,41 @@ class CheckPF2e {
         const allowInteractive = context.rollMode !== "blindroll";
         const roll = await new CheckRoll(`${dice}${totalModifierPart}`, {}, options).evaluate({ allowInteractive });
 
-        // Combine all degree of success adjustments into a single record. Some may be overridden, but that should be
-        // rare--and there are no rules for selecting among multiple adjustments.
-        const dosAdjustments = ((): DegreeAdjustmentsRecord => {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This is kinda mostly done here https://github.com/foundryvtt/pf2e/pull/14960/files except for passing the check options to notes. The thing to keep in mind is that we don't want check options to be in the main options, because they're not really available for most things. Its important to display it separately.

I should, given there's been some time, clean up the PR I linked. Its my bad, there are two open issues kinda asking for similar things and the PR only referred to and attempted to tackle one of them.

PR #15790, opened 2024-08-04
Add A(H)BCD tabs to Compendium Browser

mysurvive:
This PR adds Ancestries, Heritages, Backgrounds, Classes, and Deities to the compendium browser and additionally redirects the search buttons of the Character page on the Character sheet to the corresponding compendium browser tabs. This allows A(H)BCD content that is stored in compendiums to be loaded into the compendium browser to browse and filter during character building instead of searching through a myriad of compendia to find available content.

There is one issue that I ran into that I don't know how to solve. `src/module/apps/compendium-browser/index.ts` throws errors in the `Tagify` section starting at line 642. I tried a myriad of fixes but couldn't figure it out - this happened after removing the `traits` filter from the new A(H)BCD tabs. I don't know why this is happening, I thought the `objectHasKey` method in the if statement at line 631 would handle this, but it appears not.

I realize this is an extremely large PR and probably has a lot of places where the code can be optimized and/or doesn't follow the system's standards, so take as much time as necessary to review. There's no rush on this, just a project I wanted to work on.


intrand:
There are several failing CI checks found in the Details link here in the PR. Please take a look at those and resolve them

mysurvive:
Fixed the linting issue.

Also... maybe fixed the Tagify errors? Seems kind of janky to me but maybe it's the right way, I don't know.

intrand:
I'm seeing errors when I enable a module that contains compendium packs with valid items inside of it (class, feats, spells, etc) and enable those packs in the settings of the compendium browser. I will make a new world and test more slowly. If you want, you may install PF2E Lifeblade and/or PF2E Companion Compendia to try to reproduce the issue.

Here's a copy of the error from the console:
```
pf2e.mjs:1341 TypeError: Cannot read properties of undefined (reading 'toString')
    at CompendiumBrowserClassTab.loadData (pf2e.mjs:1676:75999)
    at async CompendiumBrowserClassTab.init (pf2e.mjs:1676:50897)
    at async CompendiumBrowser.loadTab (pf2e.mjs:1676:112592)
    at async HTMLFormElement.sheetHandler (pf2e.mjs:1341:441966)
```

stwlam:
Screenshot?

intrand:
Here are some of it working as expected (without any 3rd party content):

image
image
image
image
image
image

mysurvive:
image
image
image
image
image
It works if the compendium only has one type of item in it (i.e. only classes, only backgrounds, etc). I saw that Lifeblade has effects as well as classes, etc. in them (I'm having trouble downloading Companion Compendia right now... probably a github issue so I can't test, but I'm guessing it probably is the same way). I thought this was a design philosophy to have a unique compendium for each item type. As soon as I added an effect to my class compendium, I got the same error as you. I didn't test with mixed compendia!

mysurvive:
This commit should fix the error you're seeing, and I added a warning if any of the index fields are missing to be more in line with the other tabs.

intrand:
> This commit should fix the error you're seeing, and I added a warning if any of the index fields are missing to be more in line with the other tabs.

It seems as though that fixed everything except Classes from being added as packs in settings. I tested on a new world. If you...

1. open browser dev tools
2. enable either of the modules above
3. open the compendium browser
4. go to settings
5. enable the checkbox for either module under Classes
6. click save

this will throw an error to the console. It appears to be the same error but from a different line number. So, I assume we made it past the earlier calls for ancestry, background, etc and now we're stuck on a later call to deal with Classes. Everything else can be enabled in the settings and shows up fine.

Thank you for all of your work on this!

intrand:
After fixing my own dumb mistake, it looks like everything is working! I can successfully add packs from various modules and everything seems to come up normally and filter correctly! Huzzah!

Thank you for your work on this, @mysurvive 😄

intrand:
image

for posterity

intrand:
Upon further testing, I see some deities missing from the browser. Some examples: Aakriti, Abraxas, Adanye. Opening the pf2e.deities pack next to the browser is a good way to see what's missing and what was found. I double-checked that the browser wasn't excluding sources (in settings and in filters).

mysurvive:
I found the problem - I will get a commit in later tonight or tomorrow to fix it.

CarlosFdez:
Unfortunately, I'm not a fan of how many entries are there at the top. I don't know how to tackle it though. One idea I had was to create different contexts somehow, and have bestiary/hazards into a completely separate context, but I don't think that'd be enough.

mysurvive:
Would a tab for "Character Building" be a better option, with tabs for A(H)BCD (like the settings tab has)? I have a commit that I've been working on that does that, but it's not working yet.

I think this is an important enhancement for the compendium browser, so I'm willing to keep pecking at it until it's in a good place.

mysurvive:
Does this look better @CarlosFdez ? 
image

mysurvive:
Fixed the CSS too
image

CarlosFdez:
I don't think we're going this route. It appears the tide is more turning towards rewriting it in svelte, and using special sheets for PC side stuff: https://github.com/foundryvtt/pf2e/pull/16896

As part of whatever inevitable rewrite it'll get, I would personally like it if we split actors from items since the use case is different, but that's a road we'll cross when we get there.

I do like the sub-tab for ABCD items though, and I'll pitch that as part of the rewrite. Thank you for your work here, as it did at spark discussion on what we want and don't want.

PR #15818, opened 2024-08-05
Allow `@Embed` of Items

farling42:
Adds _buildEmbedHTML to ItemPF2e so that `@Embed` can be used to reference the description (and prerequisites if present) stored in the item directly.

This will simplify the maintenance of the Archetypes journal (as well as anywhere else that text is duplicated from Items into Journal entries.

For example, in the Acrobat page of the Archetypes journal, the first feat reference would be replaced by:

```
<H2> @UUID[Compendium.pf2e.feats-srd.Item.aFygWxgSv82WyCsl]{Acrobat Dedication} Feat 2
<p> @Embed[Compendium.pf2e.feats-srd.Item.aFygWxgSv82WyCsl inline]
```

whereas **Dodge Away** would require an additional new parameter to prevent the HR being added immediately after the prerequisites (since a HR is added in the body of the feat description):

```
<H2> @UUID[Compendium.pf2e.feats-srd.Item.0qGLCpggCcOVkbtT]{Dodge Away} Feat 6
<p> @Embed[Compendium.pf2e.feats-srd.Item.0qGLCpggCcOVkbtT inline hr=false]
```

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -862,6 +862,33 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
             this.actor.update(actorUpdates);
         }
     }
+
+    /** To be overridden by subclasses to extend the HTML string that will become part of the embed */
+    embedHTMLString(_config: DocumentHTMLEmbedConfig, _options: EnrichmentOptions) : string {
+        return this.description;
+    }
+
+    async _buildEmbedHTML(config: DocumentHTMLEmbedConfig, options: EnrichmentOptions) {
+        //const embed = await super._buildEmbedHTML(config, options);
+        //if (embed) return embed;
+
+        // As per foundry.js: JournalEntryPage#_embedTextPage
+        options = { ...options, relativeTo: this };
+        const {
+          secrets = options.secrets,
+          documents = options.documents,
+          links = options.links,
+          rolls = options.rolls,
+          embeds = options.embeds
+        } = config;
+        foundry.utils.mergeObject(options, { secrets, documents, links, rolls, embeds });
+    
+        // Get correct HTML
+        const container = document.createElement("div");
+        container.innerHTML = await TextEditor.enrichHTML(this.embedHTMLString(config, options), options);
+        return container.children;
+      }
+    
 }
 
 interface ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item<TParent> {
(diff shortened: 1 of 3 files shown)


stwlam:
Mind posting a screenshot?

farling42:
An example replacement for the Druid class page would be like:
image

farling42:
Where there are Prerequisites and Requirements before the displayed HR (using the hr=false flag which is specific to my PR code, so could be renamed):

image

src/module/item/base/document.ts
```diff
@@ -862,6 +862,40 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
             this.actor.update(actorUpdates);
         }
     }
+
+    /** To be overridden by subclasses to extend the HTML string that will become part of the embed */
+    _embedHTMLString(_config: DocumentHTMLEmbedConfig, _options: EnrichmentOptions): string {
+        // prerequisites (PrerequisiteTagData[])
+        const prereq: { value: string }[] | null = foundry.utils.getProperty(this, "system.prerequisites.value");
```
stwlam, line 869: Retrieve exactly what you want (and know what you're retrieving) instead of using `getProperty`.

src/module/item/base/document.ts
```diff
@@ -862,6 +862,40 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
             this.actor.update(actorUpdates);
         }
     }
+
+    /** To be overridden by subclasses to extend the HTML string that will become part of the embed */
+    _embedHTMLString(_config: DocumentHTMLEmbedConfig, _options: EnrichmentOptions): string {
```
stwlam, line 867: Don't use underscore-prefixed methods: make it `protected` if appropriate.

src/module/item/base/document.ts
```diff
@@ -862,6 +862,40 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+        // prerequisites (PrerequisiteTagData[])
+        const prereq: { value: string }[] | null = foundry.utils.getProperty(this, "system.prerequisites.value");
+        let result = "";
```
stwlam, line 870: no `let`

farling42:
Code refactored to access prerequisites only inside Feat and CampaignFeature classes

stwlam:
Is dnd5e making use of this? I want if they're including the item name if so: it seems a bit onerous to have that separately.

farling42:
> Is dnd5e making use of this? I want if they're including the item name if so: it seems a bit onerous to have that separately.

I don't think that dnd5e has implemented anything for embed of Items yet.

The only reason to have the name displayed separately is so that different presentation can be made in different places. By example above uses the "inline" qualifier which removes the caption and another html feature which would normally be included when a figure element is added by the core embed function.

What is missing from my example above, is that the first line is a header (H2?) element (as per the existing PF2E journal page), whereas the `@embed` is just in a normal paragraph.

PR #15819, opened 2024-08-05
Tap Into Blood is 1 action

### LINKED ISSUE #15815: Sorcerer Feat in PC2 "Tap into Blood" is set to Passive when it should be 1-Action.
It's 1-Action in Player Core 2. However, in the system, it's set to Actions: Passive. This is incorrect.

intrand:
Fixes #15815


intrand:
Thanks!

PR #15825, opened 2024-08-05
Make confused condition flat check immutable

pedrogrullada:


stwlam:
?

pedrogrullada:
> ?

It is being lowered by penalties to DCs like frightened on actors it's on

stwlam:
That's a bug given it's a flat DC. It shouldn't be worked around with a data update.

pedrogrullada:
I would expect flat check DCs to be able to be modified, just the checks themselves being unable to, which is why I thought this just needed a data update. I'll go ahead and close it though.

stwlam:
You may be right, though should being frightened reduce an inventor's unstable DC?

pedrogrullada:
> Only abilities that specifically apply to flat checks can change the checks’ DCs
They need to explicitely say they do, which frightened doesn't

PR #15829, opened 2024-08-05
Separate You're Next feat into Rogue and Swashbuckler feats

pedrogrullada:
As much as I dislike having duplicates, these are different enough to warrant it, imo
- They have different restrictions (within 60 ft. vs can see/be seen by)
- One calls out an Intimidation check explicitly, while the other doesn't, thus being elligible for Versatile Performance, for example.


pedrogrullada:
Scratch that, I will do a description Item Alteration instead.

PR #15832, opened 2024-08-05
Add description override to You're Next depending on the class

### LINKED ISSUE #15828: The feat You're Next is set for both Rogue and Swashbuckler, but the two feats are fairly different
You're Next (rogue) is from Player Core page 171 and has the text in the system, You're Next (swashbuckler) is from Player Core 2 page 164.
The two feats are fairly different (reaction/action, creature you can see/60 feet)
Should they be conflated or should it be two feats?

pedrogrullada:
Closes #15828

### DIFF
--- a/packs/feats/youre-next.json
+++ b/packs/feats/youre-next.json
@@ -11,7 +11,7 @@
         },
         "category": "class",
         "description": {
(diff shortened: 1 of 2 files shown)


packs/feats/youre-next.json
```diff
@@ -38,6 +38,31 @@
...
+            },
+            {
+                "itemId": "{item|id}",
+                "key": "ItemAlteration",
+                "mode": "override",
+                "property": "description",
+                "value": [
+                    {
+                        "text": "PF2E.SpecificRule.Feat.YoureNext.Trigger"
+                    },
```
TikaelSol, line 50: We can use a title of `Trigger` here, which will need localized, preferably in a generic place under specific rule. `SpecificRule.Formatting` or something like that?

pedrogrullada: I believe we already have that, I just forgot to put it, will check

pedrogrullada: We didn't. Put it under `PF2E.FeatDescription` which is already being used for things like these.

PR #15849, opened 2024-08-06
Add effect for Scars of Steel

pedrogrullada:


TikaelSol:
Since this is instantaneous we should just make it a toggle

pedrogrullada:
I'm doing a whole bunch of minor Barbarian stuff right now, so I'll bundle this on that next PR.

PR #15857, opened 2024-08-06
Lower the color intensity of the Light spell effect

TikaelSol:
Change to a more natural white and drop the alpha


ammalagonc:
My eyes thank you :)

PR #15860, opened 2024-08-06
Set several class feats to use the proper class DC

### LINKED ISSUE #10741: Dragon's Rage Breath has DC 0 (inconsistent classDC usage?)
Thank you for an amazing PF 2e game system! It has made running PF 2e online a delight.

After upgrading to v11 of Foundry and 5.6.2 of the plugin, one of my players noticed this:

image

This is using the following, which does not seem to work:

```
@Check[type:reflex|dc:resolve(@actor.attributes.classDC.value)|basic:true]
```

Looking through `packs/feats` also finds some uses of this, which doesn't work either:

```
@Check[type:reflex|dc:resolve(@actor.system.attributes.classDC.value)|basic:true]
```

However, changing this to the following works fine:

```
@Check[type:reflex|dc:resolve(@actor.classDCs.barbarian.dc.value)|basic:true]
```

Searching for `classDC` through `packs/feats` shows examples of all three usages, plus a few others. I'm not sure what each of the different variations is intended to do, and which are currently supported.

I'd be happy to submit a PR, except I'm not quite sure which is correct. Please let me know if I can help!

TikaelSol:
Closes #10741 though it will probably be a common enough issue after large data entry projects.


packs/actions/spinning-crush.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p><strong>Requirements</strong> You're wielding a loaded firearm or loaded crossbow.</p>\n<hr />\n<p>You go into a vicious spin, smashing your weapon into those nearby and increasing your momentum by firing. All creatures adjacent to you take @Damage[(4d6+@actor.abilities.str.mod)[bludgeoning]]{4d6 bludgeoning damage plus your Strength modifier}; this increases to @Damage[(6d6+@actor.abilities.str.mod)[bludgeoning]]{6d6} if your firearm has a striking rune, @Damage[(8d6+@actor.abilities.str.mod)[bludgeoning]]{8d6} if it has a greater striking rune, and @Damage[(10d6+@actor.abilities.str.mod)[bludgeoning]]{10d6} if it has a major striking rune. This ability does not apply other effects that increase damage with your firearm Strikes such as weapon specialization. Creatures affected by this attack must attempt a @Check[type:reflex|dc:resolve(@actor.system.proficiencies.classDCs.gunslinger.dc)|basic:true] save. A creature that fails its save is also pushed 10 feet directly away from you.</p>"
```
stwlam, line 14:
```suggestion
            "value": "<p><strong>Requirements</strong> You're wielding a loaded firearm or loaded crossbow.</p>\n<hr />\n<p>You go into a vicious spin, smashing your weapon into those nearby and increasing your momentum by firing. All creatures adjacent to you take @Damage[(4d6+@actor.abilities.str.mod)[bludgeoning]]{4d6 bludgeoning damage plus your Strength modifier}; this increases to @Damage[(6d6+@actor.abilities.str.mod)[bludgeoning]]{6d6} if your firearm has a striking rune, @Damage[(8d6+@actor.abilities.str.mod)[bludgeoning]]{8d6} if it has a greater striking rune, and @Damage[(10d6+@actor.abilities.str.mod)[bludgeoning]]{10d6} if it has a major striking rune. This ability does not apply other effects that increase damage with your firearm Strikes such as weapon specialization. Creatures affected by this attack must attempt a @Check[type:reflex|against:gunslinger|basic:true] save. A creature that fails its save is also pushed 10 feet directly away from you.</p>"
```

TikaelSol: If we're going to do that we should do them all, which I can do easier if this gets in first since it would be a simple pattern to match

PR #15867, opened 2024-08-06
Automate increase and decrease to item repair

### LINKED ISSUE #10958: Increases and decreases to item repair should be automated
When repairing an item with the Crafter's Eyepiece invested, the amount of HP restored via the "Repair: Crafting" action macro doesn't change. For example, when expert in crafting and successfully repairing an item, with and without the Crafter's Eyepiece 15 HP is restored, but I believe it should be the 15 HP is restored when the Crafter's Eyepiece isn't invested and 30 HP is restored when it is invested (a similar case is also true for critical success).

Also I believe the descriptions for the Crafter's Eyepiece and Crafter's Eyepiece (Greater) need to be updated as they don't match what is found on [Archives of Nethys](https://2e.aonprd.com/Equipment.aspx?ID=427), in particular on foundry it says:

> When you Repair an item, increase the Hit Points restored by 15 per proficiency rank instead of 10.

whereas on Archives of Nethys it says:
> When you Repair an item, increase the Hit Points restored to 10 + 10 per proficiency rank on a success or 15 + 15 per proficiency rank on a critical success.

jesuisfrog:
Added automation for the repair macro.

It will now check the new "repairValue" data point under actor.system.crafting to allow for customization of the amount of HP that can be repaired with the macro. The default is set to 5 as per the base repair action.

Crafter's Eyepiece has been updated with two new rules to add to repairValue and to add a flavor note about the updated repair numbers.

This should close #10958

### DIFF
--- a/packs/equipment/crafters-eyepiece.json
+++ b/packs/equipment/crafters-eyepiece.json
@@ -49,8 +49,8 @@
             },
             {
                 "key": "Note",
-                "text": "When you Repair an item, increase the Hit Points restored to 10 + 10 per proficiency rank on a success or 15 + 15 per proficiency rank on a critical success.",
-                "title": "<b>Crafter's Eyepiece:</b>",
+                "text": "PF2E.SpecificRule.CraftersEyepiece.Note",
+                "title": "{item|name}",
                 "selector": "crafting-check",
                 "outcome": [
                     "success",
--- a/packs/equipment/crafters-eyepiece.json
+++ b/packs/equipment/crafters-eyepiece.json
@@ -40,6 +40,25 @@
                 "selector": "crafting",
                 "type": "item",
                 "value": 1
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "system.crafting.repairValue",
+                "value": 5
+            },
+            {
+                "key": "Note",
+                "text": "When you Repair an item, increase the Hit Points restored to 10 + 10 per proficiency rank on a success or 15 + 15 per proficiency rank on a critical success.",
+                "title": "<b>Crafter's Eyepiece:</b>",
+                "selector": "crafting-check",
+                "outcome": [
+                    "success",
+                    "criticalSuccess"
+                ],
+                "predicate": [
+                    "action:repair"
+                ]
             }
         ],
         "size": "med",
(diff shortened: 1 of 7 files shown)


packs/equipment/crafters-eyepiece.json
```diff
@@ -40,6 +40,25 @@
...
+                "key": "ActiveEffectLike",
+                "mode": "add",
+                "path": "system.crafting.repairValue",
+                "value": 5
+            },
+            {
+                "key": "Note",
+                "text": "When you Repair an item, increase the Hit Points restored to 10 + 10 per proficiency rank on a success or 15 + 15 per proficiency rank on a critical success.",
+                "title": "<b>Crafter's Eyepiece:</b>",
+                "selector": "crafting-check",
```
TikaelSol, line 54:
```suggestion
                "key": "Note",
                "text": "PF2E.SpecificRule.CraftersEyepiece.Note",
                "title": "{item|name}",
                "selector": "crafting-check",
```
Add the text to re-en.json at that location. Also need to add this to the greater eyepiece too.

src/module/migration/migrations/722-crafting-system-data.ts
```diff
@@ -22,5 +22,9 @@ export class Migration722CraftingSystemData extends MigrationBase {
         if (!Array.isArray(crafting.formulas)) {
             crafting.formulas = [];
         }
+
+        if(!R.isNumber(crafting.repairValue)){
+            crafting.repairValue = 5;
+        }
```
TikaelSol, line 28: Instead of running this on an old migration we can just add the value during data prep. See `prepareBaseData()` in [module/actor/creature/document.ts](https://github.com/foundryvtt/pf2e/blob/8a404f752ea493e4313b0595860b2c3f6ffbce5e/src/module/actor/creature/document.ts#L284) though this one would go in character instead.

TikaelSol:
You can run `npm run lint:fix` to fix the linting issues, and `npm run test` to run the tests locally.

packs/equipment/crafters-eyepiece.json
jesuisfrog: Good catch, thanks for pointing it out!

src/module/migration/migrations/722-crafting-system-data.ts
jesuisfrog: I actually had that in there already, removing it from the migration one.

PR #15874, opened 2024-08-06
Remove AC penalty and add Temporary Hit Points to Share Rage

pedrogrullada:


packs/feat-effects/effect-share-rage.json
```diff
@@ -45,10 +45,24 @@
...
+                "predicate": [
+                    {
+                        "nor": [
+                            "self:type:familiar",
+                            "self:type:hazard",
+                            "self:type:vehicle"
+                        ]
+                    }
+                ],
+                "value": "@actor.level + @actor.abilities.con.mod"
```
TikaelSol, line 61: I think technically this should just be in the temp HP rule and skip the AE like here. If this went on an ally that overrides rage HP for their own thing they should get the benefits as described in the barbarian rage, not their particular flavor. Like a summoned giant badger shouldn't get 0 temp HP from this.

pedrogrullada: I was thinking that it should be affected by the Rage temp HP immunity effect, since allies would also not be able to benefit from them again, as well as supporting Contagious Rage in the future, but maybe we'll cross that bridge when we get there.

TikaelSol: It's an effect, so at least no migration will ever be needed.

PR #15915, opened 2024-08-07
Automate increase and decrease to item repair

### LINKED ISSUE #10958: Increases and decreases to item repair should be automated

jesuisfrog:
Added automation for the repair macro.

It will now check the new "repairValue" data point under actor.system.crafting to allow for customization of the amount of HP that can be repaired with the macro. The default is set to 5 as per the base repair action.

Crafter's Eyepiece has been updated with two new rules to add to repairValue and to add a flavor note about the updated repair numbers.

This should close #10958

(Apologies for the two closed PRs, I had some issues with them but it's all resolved now)

### DIFF
--- a/src/module/actor/character/data.ts
+++ b/src/module/actor/character/data.ts
@@ -404,6 +404,7 @@ interface VersatileWeaponOption {
 interface CharacterCraftingData {
     formulas: CraftingFormulaData[];
     entries: Record<string, Partial<CraftingEntryData>>;
+    repairValue: 5;
 }
 
 interface CharacterResources extends CreatureResources {
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -456,7 +456,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         }
 
         // Indicate that crafting formulas stored directly on the actor are deletable
-        system.crafting = fu.mergeObject({ formulas: [], entries: {} }, system.crafting ?? {});
+        system.crafting = fu.mergeObject({ formulas: [], entries: {}, repairValue: 5 }, system.crafting ?? {});
         for (const formula of this.system.crafting.formulas) {
             formula.deletable = true;
         }
--- a/packs/equipment/crafters-eyepiece-greater.json
+++ b/packs/equipment/crafters-eyepiece-greater.json
@@ -40,6 +40,25 @@
                 "selector": "crafting",
                 "type": "item",
                 "value": 2
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "add",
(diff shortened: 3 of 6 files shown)


stwlam:
This looks could, but would you mind making it `crafting.repair` an object with a lone `value` field? Then we won't have regret later if there is other repair-related data we want to track.

jesuisfrog:
@stwlam Made the requested change!

src/module/actor/character/data.ts
```diff
@@ -404,6 +404,7 @@ interface VersatileWeaponOption {
 interface CharacterCraftingData {
     formulas: CraftingFormulaData[];
     entries: Record<string, Partial<CraftingEntryData>>;
+    repair: { value: 5 };
```
CarlosFdez, line 407: This should be value: number, as this is a type

src/module/actor/character/document.ts
```diff
@@ -461,7 +461,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         }
 
         // Indicate that crafting formulas stored directly on the actor are deletable
-        system.crafting = fu.mergeObject({ formulas: [], entries: {} }, system.crafting ?? {});
+        system.crafting = fu.mergeObject({ formulas: [], entries: {}, repair: { value: 5 } }, system.crafting ?? {});
```
CarlosFdez, line 464: This allows persisted values to override the default of 5, when only rule elements should be able to modify the repair value. You might need to do `system.crafting.repair = { value: 5 };` after this line.

CarlosFdez:
Can you foresee other values being inside system.crafting.repair? I'm wondering if we can't replace `system.crafting.repair.value` with `system.crafting.repair.success` and `system.crafting.repair.criticalSuccess`. Because AELikes set the crafting proficiency though, perhaps this can't be done very cleanly. I'm wondering if this might be a thing better attempted once we proper skill tracking.

PR #15963, opened 2024-08-09
Fix updating homebrew languages

CarlosFdez:
The render in the languages update method was eating the update to the languages itself as well, and since updateObject always leads to a re-render it was unnecessary.

### DIFF
--- a/src/module/system/settings/homebrew/languages.ts
+++ b/src/module/system/settings/homebrew/languages.ts
@@ -98,6 +98,14 @@ export class LanguagesManager {
         }
     }
 
+    #isValidLanguage(language: string | undefined): language is LanguageNotCommon {
+        return (
+            !!language &&
+            language !== "common" &&
+            (objectHasKey(CONFIG.PF2E.languages, language) || this.menu.cache.languages.some((l) => l.id === language))
+        );
+    }
+
     async #onDropLanguage(event: Sortable.SortableEvent): Promise<void> {
         const droppedEl = event.item;
         const dropTarget = htmlClosest(droppedEl, "ul[data-languages]");
@@ -107,7 +115,7 @@ export class LanguagesManager {
         if (oldRarity === newRarity) return;
 
         const language = droppedEl.dataset.slug;
-        if (!objectHasKey(CONFIG.PF2E.languages, language) || language === "common") {
+        if (!this.#isValidLanguage(language)) {
             throw ErrorPF2e("Unexpected update to language rarities");
         }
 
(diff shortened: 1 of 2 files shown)


src/module/system/settings/homebrew/languages.ts
```diff
@@ -107,7 +107,10 @@ export class LanguagesManager {
         if (oldRarity === newRarity) return;
 
         const language = droppedEl.dataset.slug;
-        if (!objectHasKey(CONFIG.PF2E.languages, language) || language === "common") {
+        const isValidLanguage =
+            language &&
+            (objectHasKey(CONFIG.PF2E.languages, language) || this.menu.cache.languages.some((l) => l.id === language));
```
stwlam, line 112: This looks like an unrelated change

src/module/system/settings/homebrew/languages.ts
```diff
@@ -155,22 +158,18 @@ export class LanguagesManager {
...
 
-        let render = false;
-
         if (
             source.commonLanguage &&
             !languageSet.has(source.commonLanguage) &&
             !updatedLanguages.includes(source.commonLanguage)
         ) {
             source.commonLanguage = null;
-            render = true;
```
stwlam, line 166: Ditto

static/templates/system/settings/homebrew.hbs
```diff
@@ -11,7 +11,7 @@
         <div class="form-group setting">
             <label for="{{options.id}}-languages">{{localize settings.languages.name}}</label>
             <div class="form-fields">
-                <input type="text" name="languages" class="homebrew pf2e-tagify" data-dtype="JSON" value="{{json traitSettings.languages.value}}">
+                <input type="text" name="languages" class="homebrew pf2e-tagify" data-dtype="JSON" value="{{json settings.languages.value}}">
```
stwlam, line 14: This is the fix, right?

src/module/system/settings/homebrew/languages.ts
CarlosFdez: Only marginally unrelated. There are two issues with updating homebrew languages.

1) Adding new ones
2) setting the rarity of freshly added ones before saved

This would be one entry in the changelog.

static/templates/system/settings/homebrew.hbs
CarlosFdez: No. Its still broken even with just this change. I think something changed in core that caused the rerender of the rarity update to not reflect changes in the languages themselves.

src/module/system/settings/homebrew/languages.ts
CarlosFdez: I can try without this, if you'd rather I split this off, though I can't really see this being a second changelog entry.

PR #15964, opened 2024-08-10
add roll option to Inventor feats to add the `Unstable` trait

ammalagonc:
Once this PR is merged, the plan is to go back and localize the unstable check and text of the feats referencing these options.

### DIFF
--- a/packs/feats/megavolt.json
+++ b/packs/feats/megavolt.json
@@ -24,7 +24,37 @@
             "remaster": false,
             "title": "Pathfinder Guns & Gears"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "RollOption",
+                "label": "PF2E.TraitUnstable",
+                "mergeable": true,
+                "option": "unstable",
+                "predicate": [
+                    {
+                        "not": "feature:construct-innovation"
+                    }
+                ],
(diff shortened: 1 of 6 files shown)


stwlam:
Why mutually exclusive options to make Deep Freeze and Electrify Armor unstable?

ammalagonc:
Now that you mention it, you are right - it should all be under one toggle. Let me update this.

ammalagonc:
Changes applied - please review when you get a moment.

packs/feats/megavolt.json
```diff
@@ -24,7 +24,31 @@
             "remaster": false,
             "title": "Pathfinder Guns & Gears"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "RollOption",
+                "label": "PF2E.TraitUnstable",
+                "mergeable": true,
```
stwlam, line 31: ?

ammalagonc: Same as the other ones, this ones has an Unstable function that should add the Unstable trait when used.

ammalagonc: Removed all instances of mergeable: true from the RE.

stwlam (approved):

PR #15991, opened 2024-08-11
brush up Sanctify Armament feat to account for the Lasting Armament feat

### LINKED ISSUE #11498: [Remaster] Warpriest feat Lasting Armament does not increase effect of Sanctify Armament
As the title says; having the Lasting Armament feat doest not cause the effect-sanctify-armament to last an hour rather than 1 round.

ammalagonc:
How this works
- Added a second effect for Sanctify Armament that has a 1-hour duration.
- Added an item alteration description that posts the corresponding effect - normal duration if you don't have the Lasting Armament feat, or extended duration if you do.
- Removed the self-applied effect.

If approved, 
closes #11498


ammalagonc:
After mulling it over - this one is too much of a hack. Closing this one.

PR #16032, opened 2024-08-14
Add item alteration RE to Unstable Function Inventor feats.

ammalagonc:
This will add the note with the check when using the Unstable Function toggle.


packs/feats/deep-freeze.json
```diff
@@ -46,6 +46,21 @@
...
                 "value": "unstable"
+            },
+            {
+                "itemType": "feat",
+                "key": "ItemAlteration",
+                "mode": "add",
+                "predicate": [
+                    "item:trait:unstable",
+                    "item:deep-freeze"
+                ],
```
stwlam, line 57:
```suggestion
                "itemId": "{item|id}",
                "key": "ItemAlteration",
                "mode": "add",
                "predicate": [
                    "item:trait:unstable"
                ],
```

PR #16035, opened 2024-08-14
Fix some GM Core items activations

Cuingamehtar:


Cuingamehtar:
Regarding Accolade Robe (Greater) - it doesn't make sense that the better version has a more restrictive item activation. My reading of the text is that Review still has 1-action activation, but it's free once a minute.

Cuingamehtar:
I believe I won't be introducing any more changes here.

PR #16040, opened 2024-08-14
Have `sourceId` getter fallback to `_stats.compendiumSource`

### LINKED ISSUE #16039: Make the `sourceId` getter fallback to `_stats.compendiumSource` if no flag exist
Foundry V12 seem to always add the source id of a document in `_stats.compendiumSource`, but the system `sourceId` getter only look for the flag and it doesn't always exist. A fallback would be helpful.

reonZ:
Resolves #16039

### DIFF
--- a/build/lib/compendium-pack.ts
+++ b/build/lib/compendium-pack.ts
@@ -263,7 +263,7 @@ class CompendiumPack {
             this.#assertSizeValid(docSource);
             docSource.system._migration = { version: MigrationRunnerBase.LATEST_SCHEMA_VERSION, previous: null };
             for (const item of docSource.items) {
-                item._stats = partialStats;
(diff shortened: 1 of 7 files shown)


reonZ:
Or not !

xyzzy42:
Given that the system has deprecated `flags.core.sourceId`, would it make sense to reverse the priority and make `compendiumSource` checked first?

stwlam:
@reonZ go ahead and implement xyzzy42's suggestion

xyzzy42:
> Actually, since core migrates source ID entirely, it should use `_stats.compendiumSource` exclusively.

The way the migration works, if an item has *both* `_stats.compendiumSource` and `flags.core.sourceId`, then it leaves them both alone.
So all the pf2e items have both.  But its pack generation system is good and they always match.  A less careful module could ship with non-matching values and I didn't see anything in core that would catch this.

PR #16051, opened 2024-08-15
Set default max crafting level to actor level and display in header

CarlosFdez:


stwlam:
How will this affect Munitions Crafter?

CarlosFdez:
Munitions crafter has maxItemLevel: 1 already set in its data.

stwlam (approved):

PR #16056, opened 2024-08-15
Fix barbarian armor mastery

### LINKED ISSUE #16053: Barbarian Armor Mastery feat at 19 gives mastery in heavy armor
Armor Mastery description reads "Your proficiency ranks for light, medium, and heavy armor..." and gives mastery in heavy armor, rather than only unarmored, light and medium

jesuisfrog:
Typo in PC2 for barbarian's 19th level feat: Armor Mastery should be Medium Armor Mastery.

Closes #16053


jesuisfrog:
> image
> 
> The wrong armor mastery was used, but the name is right.

It's almost certainly to be a typo from Paizo, apart from the flavor text, it's 100% Medium Armor Mastery. The feature is consistent across all classes with those proficiencies, doesn't make sense to use a different name that is used for proficiency of all armor types in this instance.

TikaelSol:
We go by the printed name, since we're readying for an update I just wrapped the fix up in #16071, and you can see how I went about it with an RE instead.

jesuisfrog:
That works, I'll close this then.

PR #16085, opened 2024-08-17
Remove duplicate icons already provided by foundry

Everspace:
My goal is to slowly remove all duplicate, redundant, and duplicate images from `static/` and point them to existing system provided icons to reduce file size of the package, and to help better curate what differences exist instead.


TikaelSol:
We can't remove the files unless we write a migration.

Everspace:
Ah, how would you do that?

Everspace:
After talking in the discord about how to do this and the current state, I'm going to reconstruct this PR in the form of a migration.

PR #16091, opened 2024-08-17
Fix Bestial Mutagen Strikes

### LINKED ISSUE #16090: All levels of Bestial Mutagen effects do not apply their strikes if you don't have Mutant Physique
From what I can gather, this possibly has to do with #15968.

The mutagen effects have 2 versions of each claw and jaws strike. The first one is predicated on not having a feat Mutant Physique. The second one is predicated on having the feat. They share the same slug. The second of each Strike is overriding the first, but since the predicate is not matched when you don't have the feat, the Strike is then removed from the sheet. Giving them the feat or deleting the second Strike rule puts the Strike on the sheet.

So looks like more checks need to be put into place that the later-added Strike that's going to override the earlier one is actually going to apply before you override.

pedrogrullada:
By moving the damage dice upgrades to the Mutant Physique feat, as well as its intimidation bonus and the resistance to physical damage from Juggernaut Mutagen.

Closes #16090

### DIFF
--- a/packs/equipment-effects/effect-bestial-mutagen-greater.json
+++ b/packs/equipment-effects/effect-bestial-mutagen-greater.json
@@ -40,6 +40,29 @@
                 "value": -2
             },
             {
+                "baseType": "jaws",
+                "category": "unarmed",
+                "damage": {
+                    "base": {
+                        "damageType": "piercing",
+                        "dice": 3,
+                        "die": "d10"
+                    }
+                },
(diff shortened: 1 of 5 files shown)


stwlam:
did ya test the effects with and without the feats?

pedrogrullada:
yup

pedrogrullada:
forgot to make the intimidation bonus an item bonus though, there we go

PR #16140, opened 2024-08-20
consolidate Current Spell and Water Step feats and update Elementalist Journal

ammalagonc:
- Feat has the same wording on the body of the text.
- Now adding Water Step to the list.
- Updated Elementalist Journal Entry to match update in Rage of Elements


packs/feats/current-spell.json
```diff
@@ -63,6 +70,7 @@
         "traits": {
             "rarity": "common",
             "value": [
+                "archetype",
```
TikaelSol, line 73: Not sure about adding the trait actually. Need to figure out how we want to handle these I think.

ammalagonc: Want me to remove `archetype` trait?

ammalagonc: PR updated - you should now see the two feats removed and the updated Journal Entry.

ammalagonc:
Close to re-do

PR #16141, opened 2024-08-20
Move top level traits in predicates to proper format

TikaelSol:
Not all of them, but this should be a good chunk of a larger project.

### DIFF
--- a/packs/abomination-vaults-bestiary/azvalvigander.json
+++ b/packs/abomination-vaults-bestiary/azvalvigander.json
@@ -717,11 +717,11 @@
                         "predicate": [
                             {
                                 "or": [
-                                    "magical",
-                                    "arcane",
-                                    "divine",
-                                    "primal",
-                                    "occult"
+                                    "item:magical",
+                                    "item:trait:arcane",
+                                    "item:trait:divine",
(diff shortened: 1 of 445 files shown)


packs/abomination-vaults-bestiary/azvalvigander.json
```diff
@@ -717,11 +717,11 @@
...
-                                    "arcane",
-                                    "divine",
-                                    "primal",
-                                    "occult"
+                                    "item:magical",
+                                    "item:trait:arcane",
+                                    "item:trait:divine",
+                                    "item:trait:primal",
+                                    "item:trait:occult"
                                 ]
```
stwlam, line 725: This should just be "item:magical"

packs/abomination-vaults-bestiary/barcumbuk.json
```diff
@@ -1346,11 +1346,7 @@
                         "predicate": [
                             {
                                 "or": [
-                                    "magical",
-                                    "arcane",
-                                    "divine",
-                                    "primal",
-                                    "occult"
+                                    "item:magical"
```
stwlam, line 1349: Didn't need an `or` now :)

PR #16144, opened 2024-08-20
Remove tag_alt from character sheet trait tags

CarlosFdez:
Certain elements like the character sheet and chat messages override padding to zero. Chat messages lost out due to specificity. 

A better solution is to remove the padding from the global styling but its not safe to do that during a patch release.

### DIFF
--- a/static/templates/actors/character/tabs/character.hbs
+++ b/static/templates/actors/character/tabs/character.hbs
@@ -16,7 +16,7 @@
                     <li class="tag size">{{localize (lookup actorSizes data.traits.size.value)}}</li>
                 {{/if}}
                 {{#each traits as |trait slug|}}
-                    <li class="tag tag_alt" data-slug="{{slug}}">{{trait.label}}</li>
+                    <li class="tag" data-slug="{{slug}}">{{trait.label}}</li>
                 {{/each}}
             </ul>
 


src/styles/_tags.scss
```diff
@@ -84,113 +85,113 @@
...
+        width: auto;
+    }
+}
 
-            x {
-                align-items: start;
-                margin: 0;
-            }
+// Tagify and non-tagify paizo style main traits row
+.tags.paizo-style {
```
stwlam, line 110: Unnesting these is introducing unnecessary risk of them getting forgotten when a nested block is updated

CarlosFdez: I'd say its pretty necessary. Global styles need to use as few identifying tags as possible so that they can be overriden by specific uses of them. The only alternative is a mixin, or fully divorcing paizo style from non-paizo style stylings so that they're repeats and one does not affect the other.

This kinda changes it back to the original style of tags + paizo style specific overrides separately, but changes the original tags to use a less specific variant if its paizo-style.

CarlosFdez: Now that I think more on it, I'm starting to lean towards divorcing the styles, though I'm hesistant to do that in a patch release. What I did here was effectively a soft revert, but I changed the main css class, fixed the actual source of the bug, and kept nesting the tag light into the main style.

CarlosFdez: I noticed .tags.paizo-style in the main stuff was unnecessary though so I removed it. I don't know a good selector to divorce the styles while maintaining low specificity though. We'd need a divergent name from paizo style.

PR #16193, opened 2024-08-23
Switch to use of `_stats.compendiumSource` instead of `flags.core.sourceId`

stwlam:
Supersedes #16040


src/module/encounter/combatant.ts
```diff
@@ -51,6 +52,19 @@ class CombatantPF2e<
         }
     }
 
+    /** Set combatant to be base64-encoded snapshot of dynamic token if applicable  */
+    static async #setImageFromDynamicTokens(data: (CombatantPF2e | PreCreate<CombatantSource>)[]): Promise<void> {
```
CarlosFdez, line 56: seems this snuck in, though I have no objections to the contents.

PR #16213, opened 2024-08-24
consolidate Catharsis Emotion class feature

ammalagonc:
Let me know if you want to go the uuid route with this one.

- Notice - I added an AE-like effect to store the selection. I will use that alter item description of the reaction so that it matches the emotion. At least that's what the plan is.

Some of the emotions had no effects coded to them. Future work.


ammalagonc:
Close to re-do

PR #16227, opened 2024-08-24
fix the advanced-alchemy to 2r

LeoChep:
now can creat  alchemist 1:1，and use the system.crafting.entries.alchemist.batchSize to control it.
fix the number of  Advanced Alchemy's item.


CarlosFdez:
We're currently in the process of rewriting the entire system to center around a combination of prepared crafting and dynamic special resources crafting, pending some private clarification on how we should handle Munitions Crafter. After that's done we should have an accurate implementation of remaster alchemist.

Thank you for your time.

PR #16238, opened 2024-08-25
Brush up Hand of the Apprentice wizard focus spell

ammalagonc:
- Added a choice set and roll option to get the spellcasting attribute used in the spellcasting entry.
- Added additional Flat Modifiers for the charisma and wisdom that only proc on the corresponding roll option

### DIFF
--- a/packs/spell-effects/spell-effect-hand-of-the-apprentice.json
+++ b/packs/spell-effects/spell-effect-hand-of-the-apprentice.json
@@ -61,8 +61,7 @@
                     "focus-spell:wisdom"
                 ],
                 "selector": "{item|flags.pf2e.rulesSelections.weapon}-damage",
-                "type": "ability",
-                "value": 0
+                "type": "ability"
             },
             {
                 "key": "CriticalSpecialization",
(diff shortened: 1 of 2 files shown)


packs/spell-effects/spell-effect-hand-of-the-apprentice.json
```diff
@@ -32,12 +32,38 @@
...
+            {
+                "ability": "wis",
+                "hideIfDisabled": true,
+                "key": "FlatModifier",
+                "predicate": [
+                    "focus-spell:wisdom"
+                ],
+                "selector": "{item|flags.pf2e.rulesSelections.weapon}-damage",
+                "type": "ability",
+                "value": 0
```
TikaelSol, line 65:
```suggestion
                "type": "ability"
```

packs/spells/hand-of-the-apprentice.json
```diff
@@ -29,7 +29,37 @@
...
+                        "value": "wisdom"
+                    }
+                ],
+                "flag": "handOfApprentice",
+                "key": "ChoiceSet"
+            },
+            {
+                "key": "RollOption",
+                "option": "focus-spell:{item|flags.pf2e.rulesSelections.handOfApprentice}"
+            }
```
TikaelSol, line 61: ?

ammalagonc: This sets the Roll Option as a variable from the previous choice set for predicate purposes. 

The way it works - the choice set provides a value based on `spellcasting:x`. This is then set to the RollOption, which in turn allows the effect to use the correct spellcasting attribute. 

I've been toiling at this for weeks trying to get this to work and this method that worked.

PR #16243, opened 2024-08-26
Update Alchemist traits to match PC2 text

ammalagonc:
- Updated trait descriptions for Additive and Infused traits.
- Added `TraitAdditive` without a number (changed per PC2) - should these be removed?
- added Coagulant trait and the corresponding description

If this PR gets approved, I think the next step would be to update the feats with these traits.


static/lang/en.json
```diff
@@ -3805,6 +3805,7 @@
         "TrainedSkillsLabel": "Trained Skills",
         "TraitAberration": "Aberration",
         "TraitAcid": "Acid",
+        "TraitAdditive": "Additive",
```
pedrogrullada, line 3808: If you're adding this then you're gonna want to actually add the trait too

ammalagonc: Where do I add it?

pedrogrullada: https://github.com/foundryvtt/pf2e/blob/6f190d941f17d03739a2e0e37e235b8241437795/src/scripts/config/traits.ts

PR #16265, opened 2024-08-28
Add implement selection to Thaumaturge Dedication

xyzzy42:
A ChoiceSet will present an implement selection upon taking the Dedication.  It does _not_ GrantItem the chosen implement class feature (e.g. Weapon), as none of the implement's benefits are provided yet.

The archetype feat Implement Initiate will GrantItem the chosen implement class feature, using the choice made in the Dedication above.  This is when the implement's benefits should show up, e.g. Weapon will grant the action Implement's Interruption.

### DIFF
--- a/packs/feats/implement-initiate.json
+++ b/packs/feats/implement-initiate.json
@@ -28,7 +28,12 @@
             "remaster": false,
             "title": "Pathfinder Dark Archive"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "GrantItem",
+                "uuid": "{actor|flags.pf2e.implementDedication}"
+            }
+        ],
         "traits": {
             "rarity": "common",
             "value": [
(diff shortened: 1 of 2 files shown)


xyzzy42:
@websterguy came up with the idea.  It's an alternative to #14678

The reason 14678 didn't work, was that the GrantItem of the implement feature was in the Dedication feat.  That GrantItem controlled if the implement's benefits were granted or not.  It needed to be re-evaluated when the Implement Initiate feat is added to the actor.  But adding a new feat to the actor doesn't trigger re-evaluation of existing feats.

This avoids that problem by not granting the implement feature until Implement Initiate is added.  The GrantItem is in Implement Initiate now.  Since none of the implement's benefits are provided until Implement Initiate gives them, it makes sense to not put the feature on the character sheet until it does something.

The implement type is still chosen in the Dedication.  The feat says to make the choice at this time.  It's also nice for the thaum module, as it would like to know what kind of implement to create even if it doesn't provide any benefits yet.

This is also quite a bit simpler than the 14678 version.

Here's what the feats look like on the character sheet.
image

PR #16269, opened 2024-08-28
Audit TXCG Goblin and Oni ancestry feats

pedrogrullada:
Some Oni horns automation still left to do once an underlying bug is fixed

### DIFF
--- a/packs/feats/storming-gaze.json
+++ b/packs/feats/storming-gaze.json
@@ -11,7 +11,7 @@
         },
         "category": "ancestry",
         "description": {
-            "value": "<p>You can open or energize a notable third eye on your forehead to strike with storming power. You deal 3d4 electricity or 3d4 sonic damage damage in a 15-foot cone; Storming Gaze gains this trait. Each creature in the area must attempt a basic Reflex saving throw against the higher of your class DC or spell DC. You can't use this ability again for 1d4 rounds.</p>\n<p>At 7th level and every 2 levels thereafter, the damage increases by 1d4.</p>"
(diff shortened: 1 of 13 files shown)


packs/feats/tempest-gaze.json
```diff
@@ -28,7 +28,36 @@
...
+                "key": "ItemAlteration",
+                "mode": "override",
+                "predicate": [
+                    "item:slug:storming-gaze",
+                    "storming-gaze:tempest-gaze"
+                ],
+                "property": "description",
+                "value": [
+                    {
+                        "text": "PF2E.SpecificRule.Oni.StormingGaze.TempestGaze"
```
TikaelSol, line 56:
```suggestion
                        "text": "PF2E.SpecificRule.Oni.StormingGaze.TempestGaze"
```
We can probably just use the description rather than localizing it. `{item|system.description.value}`

packs/feats/storming-gaze.json
```diff
@@ -24,7 +24,53 @@
...
+                "mode": "override",
+                "predicate": [
+                    {
+                        "not": "storming-gaze:tempest-gaze"
+                    }
+                ],
+                "property": "description",
+                "value": [
+                    {
+                        "text": "PF2E.SpecificRule.Oni.StormingGaze.Base"
```
TikaelSol, line 69: Why override the description here?

pedrogrullada: The idea is to use the suboptions to switch between electricity and sonic, and adding the traits as appropriate and overriding the description with the resolved damage type. Alternatively an actor flag could be used, or just not doing it and keeping the two damage buttons.

packs/feats/tempest-gaze.json
pedrogrullada: That's fair, in that case it could add to Storming Gaze instead of overriding it

packs/feats/storming-gaze.json
TikaelSol: Since it's on the same item you can reference flags just fine without item alterations. `@Damage` is fancy like that

TikaelSol: See Blasting Beams, for example

TikaelSol: Anyways if it's necessary to use actor flags I prefer that to shoving more into localization files

PR #16289, opened 2024-08-29
Fix broken localization paths in certain TXCG items

pedrogrullada:


static/lang/re-en.json
```diff
@@ -3454,6 +3454,9 @@
                 "TouchTelepathy": "Touch Telepathy"
             },
             "Magus": {
+                "AloofFirmament": {
+                    "RollOptionLabel": "You flew or jumped"
+                },
```
TikaelSol, line 3459: Move these under a more generic section

PR #16328, opened 2024-08-30
Automate Starlit Sentinel archetype

pedrogrullada:
- Add rule elements for Starlit Transformation. Unfortunately, the implementation of the bolts of starlight has its issues
  - The Strike is still considered melee in the system, so automation tied to melee attacks and damage would be wrongly applied.
  - It follows that ranged attack automation doesn't apply either.
  - A Strike RE could be used, but there is no sane way to replicate the runes from the chosen weapon onto the synthetic Strike.
  - This automation should still be correct in most cases, but there's enough wrong with it to give me pause. I'm happy to take suggestions or outright remove it if it's not good enough. It can always be tackled again in the future.
- Integrate Blade of the Heart with the weapon chosen in Starlit Transformation.
- Add Sentinel's Orbit fly speed.
- Add a choice of constelation to Starlit Sentinel Dedication and alter damage of Shining Starlight Attack based on said choice. Templates and save types are handled via inline templates and spell variants, respectively.
- Add feat descriptions to Starlit Sentinel journal page.
- Tidy up some formatting.


packs/feats/starlit-sentinel-dedication.json
```diff
@@ -25,6 +25,61 @@
             "title": "Pathfinder Lost Omens: Tian Xia Character Guide"
         },
         "rules": [
+            {
+                "choices": [
+                    {
+                        "label": "PF2E.SpecificRule.StarlitSentinel.Constellation.Archer",
+                        "value": "archer"
+                    },
```
TikaelSol, line 33: Instead of a simple choice do an object
```json
"value": {
    "constellation": "archer",
    "damageType": "piercing"
}
```
Then you can store those in actor flags and reference that instead of having a million DamageAlterations. If we need a roll option for the constellation we can set one with the RollOption RE.

PR #16345, opened 2024-08-31
fix truncated desc for effect: fey blood magic

mose:


packs/feat-effects/effect-fey-blood-magic.json
```diff
@@ -4,7 +4,7 @@
     "name": "Effect: Fey Blood Magic",
     "system": {
         "description": {
-            "value": "<p>Granted by @UUID[Compendium.pf2e.classfeatures.Item.Bloodline: Fey]</p>\n<p>Either you gain a +2 status bonus to Performance checks.</p>"
+            "value": "<p>Granted by @UUID[Compendium.pf2e.classfeatures.Item.Bloodline: Fey]</p>\n<p>Either you gain a +2 status bonus to Performance checks for 1 round, or you can become @UUID[Compendium.pf2e.conditionitems.Item.DmAIPqOBomZ7H95W]{Concealed} for 1 round. Such obvious concealment can't be used to @UUID[Compendium.pf2e.actionspf2e.Item.XMcnh4cSI32tljXa]{Hide}.</p>"
```
pedrogrullada, line 7: Effects should only say what they do. You could add a Choice Set and allow the user to choose the performance bonus or concealment, and then a GrantItem granting the condition.

Additionally, remember to work in Foundry and extract. This will give you human readable links, which the pipeline likes.

mose: Oh I was just trying to fix the description (cause I'm working on translations and I catch stuff sometimes), sentence was beginning with a `Either` and felt truncated. But yeah there is some ruling to fix in that file too, it seems. I have no idea how to propose a choice.

pedrogrullada: Oh, yeah, the `either` is definitely wrong there, sorry about that. Do you mind closing this and letting me add the additional automation and fixing the description then?

mose: Excellent ! That way I may begin to learn a bit more about something else than just fiddling with descriptions :)

pedrogrullada: Turns out the automation I wanted wasn't possible, but either way I smuggled the description fix in my most recent PR. Thank you!

PR #16360, opened 2024-09-01
Automate Qi-Infused Fabric and Unsheathing the Sword Light

pedrogrullada:
More specifically
- Tidy up some formatting
- Add missing Studious Spells and alphabetise list of Hybrid Studies.
- Automate Unfurling Brocade's Qi-Infused Fabric
  - The bladed scarf will still be considered two-handed by the system, which is something to consider.
- Automate Unsheathing the Sword Light through Note rule element.
- Add missing trait for Vermillion Thread


packs/classfeatures/unfurling-brocade.json
```diff
@@ -24,7 +24,49 @@
...
+                "suboptions": [
+                    {
+                        "label": "PF2E.TraitHeldTwoHands",
+                        "value": "2"
+                    },
+                    {
+                        "label": "PF2E.TraitHeldOneHand",
+                        "value": "1"
+                    }
+                ],
```
TikaelSol, line 41: Can we not just use `item:hands-held:x`?

pedrogrullada: You mean using that as the roll option in this rule element? Or using the automatically generated `item:hands-held:x` from wielding a weapon in one hand? If it's the latter, the problem is that since the Bladed Scarf can't normally be used in one hand, the character sheet won't let you use it.

TikaelSol: Rather than hacking this together with a weird toggle we should request usage alterations then.

pedrogrullada: I did add it to the wish list, albeit bounded to weapons rather than usage in general. I can broaden the request, and I guess just get rid of this part of the automation.

PR #16379, opened 2024-09-02
Update Catharsis Emotion class feature RE to account for case sensitive localization

ammalagonc:


packs/classfeatures/catharsis-emotion.json
```diff
@@ -39,23 +39,23 @@
                 "choices": [
                     {
                         "label": "PF2E.SpecificRule.CatharticMage.Anger.Label",
-                        "value": "anger"
+                        "value": "Anger"
```
TikaelSol, line 42: Ok, I see the problem. If you're going to set the roll option and the localization from this choice then we need an object with both.

```json
"value": {
    "label": "Anger",
    "option": "anger"
}
```
Then reference those with `{item|flags.pf2e.rulesSelections.catharticMageDedication.label}` or `.option`

ammalagonc: OK, those changes should be in now

Issue #16364, opened 2024-09-01
Wildcard tokens not shown in chat messages


Goliathcraft:
The new feature from update 6.30 that displays tokens on some actions does not work with tokens using the wildcard tokens feature, instead just displaying the broken image icon.  Problem is that this feature cannot be turned of in the settings, and that long character names can obscure part of the roll box.

![test](https://github.com/user-attachments/assets/02987374-aa3e-4ff8-9090-48f52a2262f7)

PR #16391, opened 2024-09-02
Show active token image rather than the prototype token

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/16364

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -199,7 +199,7 @@ class ChatMessagePF2e extends ChatMessage {
         if (actor && header && this.isContentVisible) {
             const token = this.token ?? actor.prototypeToken;
             const useToken =
-                token?.texture?.src && ImageHelper.hasImageExtension(token.texture.src) && !isDefaultTokenImage(token);
+                token.texture.src && ImageHelper.hasImageExtension(token.texture.src) && !isDefaultTokenImage(token);
             const [imageUrl, scale] = useToken
                 ? [token.texture.src, Math.max(1, token.texture.scaleX)]
                 : [actor.img, 1];
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -204,7 +204,7 @@ class ChatMessagePF2e extends ChatMessage {
                 ? [token.texture.src, Math.max(1, token.texture.scaleX)]
                 : [actor.img, 1];
 
-            const image = createHTMLElement("img");
+            const image = document.createElement("img");
             image.alt = actor.name;
             image.src = imageUrl;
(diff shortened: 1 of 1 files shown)


src/module/chat-message/document.ts
```diff
@@ -199,11 +199,9 @@ class ChatMessagePF2e extends ChatMessage {
         if (actor && header && this.isContentVisible) {
             const token = this.token ?? actor.prototypeToken;
             const useToken =
-                actor.prototypeToken?.texture?.src &&
-                ImageHelper.hasImageExtension(token.texture.src) &&
-                !isDefaultTokenImage(token);
+                token?.texture?.src && ImageHelper.hasImageExtension(token.texture.src) && !isDefaultTokenImage(token);
```
stwlam, line 202: Is this nullable? The type may just need to be tighter if it's a `string`.

CarlosFdez: Its not, but it seems if you try feeding it a blank string foundry turns it into mystery man.

stwlam: _Can_ it be blank?

src/module/chat-message/document.ts
```diff
@@ -199,11 +199,9 @@ class ChatMessagePF2e extends ChatMessage {
...
             const [imageUrl, scale] = useToken
-                ? [actor.prototypeToken.texture.src, Math.max(1, actor.prototypeToken.texture.scaleX)]
+                ? [token.texture.src, Math.max(1, token.texture.scaleX)]
                 : [actor.img, 1];
 
             const image = createHTMLElement("img");
```
stwlam, line 207:
```suggestion
            const image = document.createElement("img");
```

src/module/chat-message/document.ts
CarlosFdez, line 202:
```suggestion
                token.texture.src && ImageHelper.hasImageExtension(token.texture.src) && !isDefaultTokenImage(token);
```

src/module/chat-message/document.ts
```diff
@@ -199,14 +199,12 @@ class ChatMessagePF2e extends ChatMessage {
         if (actor && header && this.isContentVisible) {
             const token = this.token ?? actor.prototypeToken;
             const useToken =
-                actor.prototypeToken?.texture?.src &&
-                ImageHelper.hasImageExtension(token.texture.src) &&
-                !isDefaultTokenImage(token);
+                token.texture.src && ImageHelper.hasImageExtension(token.texture.src) && !isDefaultTokenImage(token);
```
stwlam, line 202:
```suggestion
                token.texture.src && !isDefaultTokenImage(token);
```

PR #16458, opened 2024-09-09
Automate Fortissimo Composition

MiddleTwin:
Previously players would have to manually roll performance, manually compare to the standard dcs by level, then manually adjust every attack and damage roll for inspire courage to match the correct bonus (assuming they remembered if they crit-succeeded or succeeded this round). This has been a major pain point in my game, so it seems worth fixing.


MiddleTwin:
force pushed update: use will instead of standard dc because I just noticed remaster wants that.

packs/spells/fortissimo-composition.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p>You call upon your muse to greatly increase the benefits you provide with your <em>courageous anthem</em>, <em>rallying anthem</em>, or <em>song of strength</em> composition. If your next action is to cast one of these compositions, attempt a <span data-pf2-check=\"performance\"  data-against=\"will\">Performance</span> check. The DC is usually the highest Will DC of the composition's targets, but the GM can assign a different DC based on the circumstances. The effect of your composition depends on the result of your check.</p>\n<hr />\n<p><strong>@UUID[Compendium.pf2e.spell-effects.Item.Spell Effect: Fortissimo Composition Critical]{Critical Success}</strong> The status bonus from your composition increases to +3.</p>\n<p><strong>@UUID[Compendium.pf2e.spell-effects.Item.Spell Effect: Fortissimo Composition]{Success}</strong> The status bonus from your composition increases to +2. </p>\n<p><strong>Failure</strong> Your composition provides only its normal bonus of +1, but you don't spend the Focus Point for casting this spell.</p>"
```
TikaelSol, line 14:
```suggestion
            "value": "<p>You call upon your muse to greatly increase the benefits you provide with your <em>courageous anthem</em>, <em>rallying anthem</em>, or <em>song of strength</em> composition. If your next action is to cast one of these compositions, attempt a @Check[performance|against:will] check. The DC is usually the highest Will DC of the composition's targets, but the GM can assign a different DC based on the circumstances. The effect of your composition depends on the result of your check.</p>\n<hr />\n<p><strong>Critical Success</strong> The status bonus from your composition increases to +3.</p>\n<p><strong>Success</strong> The status bonus from your composition increases to +2. </p>\n<p><strong>Failure</strong> Your composition provides only its normal bonus of +1, but you don't spend the Focus Point for casting this spell.</p>"
```
We don't use the old style inlines anymore, and this should not be two effects. It should be a single effect with a ChoiceSet for the outcome.

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,82 @@
...
+            {
+                "key": "AdjustModifier",
+                "selectors": [
+                    "attack-roll",
+                    "damage"
+                ],
+                "mode": "upgrade",
+                "value": 2,
+                "slug": "courageous-anthem-attack-bonuses"
+            },
```
TikaelSol, line 33: These can all be condensed to a single AdjustModifier. Rather than going by slug use a predicate that checks for the slugs `modifier:slug:courageous-anthem`, etc. That also saves us from needing to manually slug all the other effects which may interfere with other existing automation either in the system or in modules.

They also need to do a match against the signature of the origin of this effect and the origin of the composition. That way parties with two bards don't get the benefit on both compositions (This is not common in most home games, but definitely happens in PFS).

MiddleTwin: I'm afraid I don't know what you mean by `modifier:slug:courageous-anthem`, etc. My only guess is basically making a list of slugs in the predicate like: `["courageous-anthem-attack-bonuses", "rallying-anthem-defensive-bonuses",....]` but this would still require adding slugs to the composition cantrips, which I think you refer to as "manual slugging" and wish to avoid. I will see if I can find some other examples of slugs in selectors tomorrow when I'm more awake, but I do appreciate the advice.

SIgnature match requirement makes sense, I'll see if I can find where the signature is and what it's called. Also a search best done tomorrow.

TikaelSol: Effects store their signature in roll options like this, `item:origin:signature:<signature>`, but we may need to expand it a bit, since we want to compare the origin of the fortissimo effect and the composition effect. I'll try to think of a more compact way to do it but we may need to add a RollOption RE to each composition effect with something like `courageous-anthem:signature:{item|origin.signature}`.

Then the predicate will look like
```json
[
    "or": [
        {
            "and": [
                "bonus:slug:courageous-anthem",
                "courageous-anthem:signature:{item|origin.signature}"
            ]
        },
        ...
    ]
]
```
We need roll options like this for warrior muse as well, so we'll want to make sure whatever we do fits both use cases without doubling effort.

MiddleTwin: Worklog, writing cuz I've been silent for a bit. Sorry about that, work and illness waylaid me. I got the ChoiceSet working pretty easily, but struggled some with the predicate. Ultimately I resorted to printing every domain and scrolling through until I found what looked like a correct one (at bottom of message)

so I think the ideal predicate is something like
```js
[{and:[
  "bonus:type:status", 
  "parent:origin:signature:<however I indicate accessing the origin:signature buff on the rallying anthem effect>", 
  {or: [
    "bonus:slug:rallying-anthem",  "bonus:slug:courageous-anthem", "bonus:slug:song-of-strength"
  ]}
]}]
```
(my best guess for the signature is "parent:origin:signature:{item|origin.signature}")

Example predicate domain for rallying anthem bonus follows. Note that there is nothing like rallying-anthem:signature, but I've given my best gueses on the different prefixes here if you want to double check.

I think I need to cast the spell on someone else and then search console again to separate between self:signature and 
parent:origin:signature, but it should be one or the other. 

```js
new Set([
    "self:type:character", # Note: self is character buff is on (i think)
    "self:signature:867485f0-ef35-5b9c-965d-8f95054dc946", #Note: id of caster (caster put buff on self)
     .... 
    "bonus:slug:rallying-anthem", # Note: bonus is the actual bonus we're checking against the predicate
    "bonus:type:status",
    "bonus:value:1",
    "parent:type:effect", # Note: parent is the spell effect that has the action:adjustmodifier we're looking for targets for
    "parent:id:edrvMxrErBrbqU4v",
    "parent:fortissimo-composition",
    "parent:slug:spell-effect-fortissimo-composition",
    "parent:trait:spellshape", #Note: maybe parent is the spell itself? i don't think the effect has trait:spellshape
    "parent:rarity:common",
    "parent:level:5", 
    "parent:origin:type:character", # I think parent:origin is the origin of the spell or spell effect
    "parent:origin:signature:867485f0-ef35-5b9c-965d-8f95054dc946", # Note: Caster Actor's signature?
    ...
    "parent:origin:caster",
    "parent:origin:caster:tradition:occult",
    ...
    "parent:origin:item:type:spell", # Note: parent:origin:item is the actual spell. not sure why both parent: trait and parent:origin:item:trait have spellshape.
    "parent:origin:item:id:fNszHaiZ0iAddILT",
    "parent:origin:item:fortissimo-composition",
    "parent:origin:item:slug:fortissimo-composition",
    "parent:origin:item:trait:bard",
    "parent:origin:item:trait:concentrate",
    "parent:origin:item:trait:focus",
    "parent:origin:item:trait:spellshape",
    "parent:origin:item:rarity:uncommon",
    "parent:origin:item:magical",
    "parent:origin:item:level:5",
    "parent:magical",
    "parent:origin:item:rank:5",
    "parent:bard", #Note: caster cast these buffs on themself. I think parent is the character the effect is attached to?
    "parent:concentrate",
    "parent:focus",
    "parent:spellshape",
    "parent:origin:item:trait:occult",
    "parent:origin:item:duration:0",
    "parent:origin:item:frequency:limited",
    "parent:spellcasting:cha",
    "parent:spellcasting:attribute:cha",
    "parent:spellcasting:focus",
    "parent:spellcasting:category:focus",
    "parent:spellcasting:occult",
    "parent:spellcasting:tradition:occult"
])
``` 

I'll fiddle around some more and put out a new revision assuming it works.

MiddleTwin: I was right about parent:origin:signature being the signature of the actor who cast fortissimo composition, I was wrong about "parent:origin:signature:{item|origin.signature}" being a good predicate. It always matches. 

My best guess is that `item` resolves to the parent, so it's pretty much a tautology.

TikaelSol:
```json
{
    "or": [
        {
            "and": [
                "bonus:slug:courageous-anthem",
                "courageous-anthem:origin:signature:{item|origin.signature}"
            ]
        },
        {
            "and": [
                "bonus:slug:song-of-strength",
                "song-of-strength:origin:signature:{item|origin.signature}"
            ]
        }
    ]
}
```
Continue to add `and` statements for each composition. That should work to narrow it properly.

MiddleTwin: There does not seem to be any key that has the signature of the character who cast rallying anthem in the domain. All I have are multiple ways to access the character the rallying anthem buff was applied to ("self:" and "{actor|}") and multiple ways to access the character who cast fortissimo composition ("parent:origin:" and "{item|origin|}").

My initial plan was to add an injectable when calling `this.resolveInjectedProperties` in `AdjustModifierRuleElement::test`. This was wrong, we make a predicate once and then check it a bunch of times. Therefore I should add the needed signature to the domain instead.

`AdjustModifier::test` is called in `module/actor/modifiers.ts::adjustModifiers`. `adjustModifier`s gets modifiers passed in and calls `getRollOptions` (no arguments)

`ModifierPf2e::getRollOptions` does seem to be the place to add stuff to that domain. 

Unfortunately, `ModifierPF2e.source`, `ModifierPF2e.rule?.parent.getOriginData().actor`, `ModifierPF2e.rule?.actor.signature` and `ModifierPF2e.rule?.item?.parent` are all the actor that the buff is attached to.

I don't think that who made the effect is stored anywhere at the moment?

MiddleTwin: Ah heck this is what the roll options in #16468 are for...

MiddleTwin:
Force-push because I had to rebase for the new roll options

packs/spell-effects/spell-effect-fortissimo-composition.json
MiddleTwin: > ```json
> {
>     "or": [
>         {
>             "and": [
>                 "bonus:slug:courageous-anthem",
>                 "courageous-anthem:origin:signature:{item|origin.signature}"
>             ]
>         },
>         {
>             "and": [
>                 "bonus:slug:song-of-strength",
>                 "song-of-strength:origin:signature:{item|origin.signature}"
>             ]
>         }
>     ]
> }
> ```
> 
> Continue to add `and` statements for each composition. That should work to narrow it properly.

Ah, I completely missed this last night and spent several hours figuring it out on my own. That was good advice, sorry for wasting it lol.

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,75 @@
...
+            {
+                "key": "AdjustModifier",
+                "predicate": [{ "and" : [
+                    "bonus:type:status", 
+                    { "or": [
+                        {"and": ["bonus:slug:courageous-anthem", "courageous-anthem:origin:signature:{item|origin.signature}"]},
+                        {"and": ["bonus:slug:rallying-anthem", "rallying-anthem:origin:signature:{item|origin.signature}"]},
+                        {"and": ["bonus:slug:song-of-strength", "song-of-strength:origin:signature:{item|origin.signature}"]}
+                    ]}
+                ]}],
```
TikaelSol, line 48: Please build this then run extract packs so it gets formatted properly.

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,75 @@
...
+                        "value": "3"
+                    },
+                    {
+                        "label": "Success",
+                        "value": "2"
+                    }
+                ],
+                "flag": "fortissimoBonus",
+                "key": "ChoiceSet",
+                "prompt": "Success type"
```
TikaelSol, line 37:
```suggestion
                "choices": [
                    {
                        "label": "PF2E.Check.Result.Degree.Check.criticalSuccess",
                        "value": "3"
                    },
                    {
                        "label": "PF2E.Check.Result.Degree.Check.success",
                        "value": "2"
                    }
                ],
                "flag": "fortissimoBonus",
                "key": "ChoiceSet",
                "prompt": "PF2E.SpecificRule.Prompt.DegreeOfSuccess"
```
We need to use localizations in rule elements to prevent translators from having to dig through the array to translate.

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,75 @@
+{
+    "_id": "rAW5o8MjjuIhYhLg",
+    "img": "systems/pf2e/icons/spells/inspire-heroics.webp",
+    "name": "Spell Effect: Fortissimo Composition",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Fortissimo Composition]</p>\n<p>Improve composition bonuses from +1 to +2 or +3 on crit/p>"
```
TikaelSol, line 7:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Fortissimo Composition]</p>\n<p>Increase the status bonus from the composition to +2 (+3 on a critical success).</p>"
```

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,96 @@
...
+                "selectors": [
+                    "ac",
+                    "saving-throw",
+                    "attack-roll",
+                    "damage",
+                    "athletics",
+                    "fortitude",
+                    "reflex"
+                ],
+                "value": "{item|flags.pf2e.rulesSelections.fortissimoBonus}"
```
TikaelSol, line 80:
```suggestion
                "value": "@item.flags.pf2e.rulesSelections.fortissimoBonus"
```
Using the `@` format because we are referencing a number.

packs/spell-effects/spell-effect-fortissimo-composition.json
```diff
@@ -0,0 +1,96 @@
...
+            {
+                "choices": [
+                    {
+                        "label": "PF2E.Check.Result.Degree.Check.criticalSuccess",
+                        "value": "3"
+                    },
+                    {
+                        "label": "PF2E.Check.Result.Degree.Check.success",
+                        "value": "2"
+                    }
```
TikaelSol, line 33:
```suggestion
                    {
                        "label": "PF2E.Check.Result.Degree.Check.criticalSuccess",
                        "value": 3
                    },
                    {
                        "label": "PF2E.Check.Result.Degree.Check.success",
                        "value": 2
                    }
```
Making these explicitly numbers, and not strings that look like numbers.

MiddleTwin:
Thanks for your patience, this is by far the friendliest project I've ever submitted a PR to.

PR #16470, opened 2024-09-10
Reduce layout thrashing in activateListeners phase of encounter tracker update

Codas:
This PR aims to improve performance when updating the encounter tracker. The current problem is that `refreshTargetDisplay` forces a style recalculation each time it is executed, which is at least once per combatant per turn or whenever the status effects for any combatants change.

The cause for this repeated recalculation is that the combatants title is updated in a loop in `activateListeners` (modify) to add the user targets indicator. In the same loop, `refreshTargetDisplay` is then called to modify the newly created target indicator element, which first has to be found in the dom for this combatant (query). This combination leads to forced style recalculations, which can be quite expensive.
Bildschirmfoto 2024-09-10 um 17 22 47
In the above case, just the `activateListeners` phase took over 100ms to complete, in a very reasonably sized encounter with 9 participants in total.

The good news is that these modifications can easily be done without forcing style recalculations and layout thrashing by first removing the tracker node from the document tree, doing all the modifications and finally re-inserting the tracker at the original position. This reduces the time needed for the `activateListeners` phase to just under 2ms.
Bildschirmfoto 2024-09-10 um 17 32 08

### DIFF
--- a/src/module/apps/sidebar/encounter-tracker.ts
+++ b/src/module/apps/sidebar/encounter-tracker.ts
@@ -68,6 +68,8 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
 
         const encounter = this.viewed;
         if (!encounter) return super.activateListeners($html);
+        const trackerPlaceholder = document.createElement("div");
+        tracker.replaceWith(trackerPlaceholder);
 
         const tokenSetsNameVisibility = game.pf2e.settings.tokens.nameVisibility;
         const allyColor = (c: CombatantPF2e<EncounterPF2e>) =>
@@ -129,7 +131,7 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
                 }
             }
 
-            this.refreshTargetDisplay(combatant);
+            this.refreshTargetDisplay(combatant, [tracker]);
 
             // Hide names in the tracker of combatants with tokens that have unviewable nameplates
             if (tokenSetsNameVisibility) {
@@ -180,11 +182,12 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
             }
         }
 
+        trackerPlaceholder.replaceWith(tracker);
         super.activateListeners($html);
     }
 
     /** Refresh the list of users targeting a combatant's token as well as the active state of the target toggle */
-    refreshTargetDisplay(combatantOrToken: CombatantPF2e | TokenDocumentPF2e): void {
+    refreshTargetDisplay(combatantOrToken: CombatantPF2e | TokenDocumentPF2e, trackers?: HTMLElement[]): void {
         if (!this.viewed || !canvas.ready) return;
 
         const { combatant, tokenDoc } = combatantAndTokenDoc(combatantOrToken);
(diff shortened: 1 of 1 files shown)


src/module/apps/sidebar/encounter-tracker.ts
```diff
@@ -68,12 +68,16 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
 
         const encounter = this.viewed;
         if (!encounter) return super.activateListeners($html);
+        const trackerParent = tracker.parentElement!;
```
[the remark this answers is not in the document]
Codas: You are of cource sompletely right and this should not have made it into the PR to begin with.

src/module/apps/sidebar/encounter-tracker.ts
```diff
@@ -63,11 +63,18 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
...
         if (!tracker) throw ErrorPF2e("No tracker found");
 
         const encounter = this.viewed;
         if (!encounter) return super.activateListeners($html);
+        const trackerParent = tracker.parentElement;
+        const trackerPlaceholder = document.createElement("div");
+
+        if (trackerParent) {
+            trackerParent.insertBefore(trackerPlaceholder, tracker);
+            tracker = trackerParent.removeChild(tracker);
```
stwlam, line 76: Is this not referentially identical to the variable's existing value?

Codas: dang, yeah it is.

src/module/apps/sidebar/encounter-tracker.ts
```diff
@@ -180,19 +187,24 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
             }
         }
 
+        if (trackerParent) {
+            trackerParent.insertBefore(tracker, trackerPlaceholder);
+            trackerPlaceholder.remove();
```
stwlam, line 192: Last thing, I think: try a `replaceWith` here.

Codas: Didn't know that was a thing. It's been too long for me working with plain DOM APIs I guess.
Done, and thank you for your help here :)

src/module/apps/sidebar/encounter-tracker.ts
```diff
@@ -68,6 +68,13 @@ export class EncounterTrackerPF2e<TEncounter extends EncounterPF2e | null> exten
 
         const encounter = this.viewed;
         if (!encounter) return super.activateListeners($html);
+        const trackerParent = tracker.parentElement;
+        const trackerPlaceholder = document.createElement("div");
+
+        if (trackerParent) {
+            trackerParent.insertBefore(trackerPlaceholder, tracker);
+            trackerParent.removeChild(tracker);
+        }
```
stwlam, line 77: o here too!

Codas: In that case we don't even need the trackerParent reference anymore, do we?
Just two replaces, which are still valid and do not throw exceptions even if the element has no parent.
```js
const div = document.createElement('div');
const span = document.createElement('span');
div.replaceWith(span); // all good!
```

stwlam:
Thank you!

PR #16482, opened 2024-09-11
Split certain crafting functions and defer use of async

CarlosFdez:
This is a split of multiple refactors from an inprogress branch, in a more finalized state. To go into more detail, this is what changes:

* Scope several crafting functions in Character to actor.crafting.
* Lazily cache the list of known actor formulas instead of receiving each time 
* Add CraftingAbility#getSheetData() and pass that instead of CraftingAbility itself. Also remove the sheet data only formula getter. 
* Construct crafting abilities during data preparation instead of during sheet rendering. The sheet based functions that required the formulas to be preloaded have been converted to async.
* When updating the list of prepared formulas, updates all rule elements with the same selector instead of only the parent item. This makes it so we no longer need to pass the parent item, and allows merging to work in the future (required for advanced alchemy). 

Crafting 3 of a field discovery bomb doesn't work after nor before this change. You might get reports of that later, but its unimportant since we're throwing out the infused reagents handling anyways.

More refactoring will occur, but these parts can be done independently of all current dev.

### DIFF
--- a/src/module/actor/character/crafting/ability.ts
+++ b/src/module/actor/character/crafting/ability.ts
@@ -13,7 +13,7 @@ class CraftingAbility implements CraftingAbilityData {
 
     selector: string;
 
-    /** This crafting entry's parent actor */
+    /** This crafting ability's parent actor */
     actor: CharacterPF2e;
 
     preparedFormulaData: PreparedFormulaData[];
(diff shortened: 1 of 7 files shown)


src/module/actor/character/crafting/ability.ts
```diff
@@ -12,13 +13,9 @@ class CraftingAbility implements CraftingAbilityData {
 
     selector: string;
 
-    /** This crafting entry's parent item */
-    parent: ItemPF2e<CharacterPF2e>;
+    /** This crafting entry's parent actor */
```
stwlam, line 16:
```suggestion
    /** This crafting ability's parent actor */
```

PR #16541, opened 2024-09-15
Consolidate Archetype journal entries and update article links in dedication feats

ammalagonc:
This puts all of the Archetype base information on one journal entry page.


TikaelSol:
Would this create broken links on existing copies of the feats?

ammalagonc:
The original articles should still be in place, which would not break the links to existing actors. I was going to experiment with using the UUID redirect for documents but decided to just pose a question to Shark - and then I forgot about it.

Issue #11303, opened 2023-11-14
Auras (active effects) only work with a square grid


elespike:
Hello,

I can't figure out why, but auras only apply effects/conditions when using a square grid. If you have a **cinder rat** on the canvas, for example, and place another token within its aura, the aura effects will automatically apply to the token. But if you change the grid from square to hex (or any other, including gridless), the aura seems to stop working.

I've looked into the defined creatures and item effects, and there doesn't seem to be anything that would prevent auras from working in other grids. Could this be a Foundry issue?

TMunsie:
Paizo has only defined things in terms of squares in their ruleset. The system is reflective of that. The ruleset isn’t designed for gridless (a module could override this) and doesn’t have set rules for hex for combat (how does elevation and how does flanking work? Again - a module could implement something)

Until Paizo makes firm rules on how this works we likely won’t change the system implementation. If you want automated auras you’ll be, unfortunately, relegated to square maps.

elespike:
@TMunsie I see, thanks!

PR #16543, opened 2024-09-15
Expose aura logic for other grid type support

FolkvangrForgent:
## Description
This MR serves to expose the aura related classes and make a minor change. This will allow modules such as pf2e-hex to add support for additional grid types to the systems aura automation.
## Changes
- Move grid type check from `checkAuras` function to `containsToken`. This was done to make it so that modules extending the behavior only need to override `containsToken`.
- Expose `AuraRenderers`, `AuraRenderer`, `TokenAura`, and `EffectAreaSquare` classes via `CONFIG.PF2E` to allow modules to access them to wrap functions.

### DIFF
--- a/src/module/scene/helpers.ts
+++ b/src/module/scene/helpers.ts
@@ -9,7 +9,7 @@ let auraCheckLock = Promise.resolve();
 
 /** Check for auras containing newly-placed or moved tokens */
 const checkAuras = foundry.utils.debounce(async function (this: ScenePF2e): Promise<void> {
-    if (!(canvas.ready && this.isInFocus && this.grid.type === CONST.GRID_TYPES.SQUARE)) {
+    if (!(canvas.ready && this.isInFocus)) {
         return;
     }
 
--- a/src/module/scene/token-document/aura/index.ts
+++ b/src/module/scene/token-document/aura/index.ts
@@ -69,7 +69,13 @@ class TokenAura implements TokenAuraData {
     /** Does this aura overlap with (at least part of) a token? */
     containsToken(token: TokenDocumentPF2e): boolean {
         // If either token is hidden or not rendered, return false early
-        if (this.token.hidden || token.hidden || !this.token.object || !token.object) {
+        if (
+            this.token.hidden ||
+            token.hidden ||
+            !this.token.object ||
+            !token.object ||
+            this.scene.grid.type !== CONST.GRID_TYPES.SQUARE
+        ) {
             return false;
         }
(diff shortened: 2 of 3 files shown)


CarlosFdez:
This almost looks fine to me, though I wonder if those classes are exposed if we should not construct instances from the config instead of importing directly in our own uses of it, to discourage monkey patching and encourage subclassing. But this is something that @stwlam is more familiar with.

FolkvangrForgent:
Are other things in CONFIG setup to use subclassing? I can certainly look into changing it to a subclassing friendly implementation if that is the design the project is aiming for. I also just want to note the `AuraRenderers`, `AuraRenderer`, and `TokenAura` are the only things I am monkey patching. `EffectAreaSquare` is just exposed so I can create instances to use within my monkey patching.

FolkvangrForgent:
Hi Just wondering if you've had a chance to take a look at this MR @stwlam . I've been using a build with this in my home games and haven't run into any issues with the current implementation.

CarlosFdez:
So far I don''t think stwlam is amenable to exposing any aura logic at all, so I don't expect this will ever be merged.

You specifically just need some way to determine "is in aura"? Its kinda bad code design wise, but the best thing I can think of as a compromise is moving isInAura() to the token, as tokens are foundry api and thus overridable.

FolkvangrForgent:
Hi CarlosFdez,
If the system does not want to expose the aura objects then the token object having a function would be better than nothing. It would prevent me from piggybacking off the existing rendering calls but I could always try to implement my own render. I am curious what the hesitation for exposing the aura logic would be; I would certainly not be expecting a stable api or anything of that sort from it's exposure. I'd also love to add hex (and maybe even gridless) support directly to the aura logic in the pf2e system if this is something the system would be amenable to. Although currently I would want fvtt to expose some functions so I don't have duplicate versions of fvtt functions: https://github.com/foundryvtt/foundryvtt/issues/11795 .

shemetz:
@stwlam Could this be reconsidered?  The [latest position](https://github.com/foundryvtt/pf2e/issues/11303) on auras in gridless/hexes was "paizo didn't make firm rules, so the pf2e system won't try to support them, but a module could override this" -- and in cases like this, modules require the system to add some API support without increasing the maintenance load for the system devs.

SilverLous:
There is a module now that let's you start macros with hex-grid supported auras: https://github.com/Wibble199/FoundryVTT-Grid-Aware-Auras

PR #16547, opened 2024-09-16
Update NPC Spells template to support Spellcasting RollOptions

### LINKED ISSUE #16236: Support `"placement": "spellcasting"` for NPC RollOption toggles

DocSchlock:
Closes #16236


CarlosFdez:
How does this look like if there aren't any toggles? It seems like the toggles section still exists even if its empty.

DocSchlock:
> How does this look like if there aren't any toggles? It seems like the toggles section still exists even if its empty. Does it introduce any margins or padding?

image

Here's what it contributes empty. It is exactly the same as the toggles section on the main tab.

CarlosFdez:
I'll merge it, since I'm not sure of what the repercussions would be of changing this, but think we need to move the if statement one level higher.

PR #16571, opened 2024-09-17
Improve actor getRollOptions performance

Codas:
This PR aims to improve the performance of the actors `getRollOptions` function.

Performance is quite difficult to measure in the best of times, but from my profiling, this change alone means the function only takes about 60% the time to execute.
Absolute terms are even more complicated as they depend on even more variables, but in my test world this shaves off about 17ms when changing turns in the encounter tracker. This function is also a hot path when adding effects or manipulating actors in other ways. 

previous function:
Bildschirmfoto 2024-09-17 um 22 10 23

with `Object.keys`
Bildschirmfoto 2024-09-17 um 22 09 57

The reason `Object.keys` is faster in this case is due to the fact that no additional arrays have to be created for the [key, value] pair.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1558,8 +1558,9 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         const toReturn: Set<string> = new Set();
 
         for (const domain of withAll) {
-            for (const [option, value] of Object.entries(rollOptions[domain] ?? {})) {
-                if (value) toReturn.add(option);
+            const optionsRecord = rollOptions[domain] ?? {};
+            for (const option of Object.keys(optionsRecord)) {
+                if (optionsRecord[option]) toReturn.add(option);
             }
         }
 


CarlosFdez:
I'm kinda annoyed that they don't optimize out the array destructuring.

Codas:
Huh, interestingly, both Firefox and Safari optimize this much better than Chrome.

On my system (MacBook, M1 family), chrome is about twice as fast when using `Object.keys` instead of `Object.entries`, with 3k total runs in the attached jsperf benchmark. Safari and Firefox both have negligable performance differences between the two (about 4-5%). Both also run much faster overall, with 6k iterations for safari and 9k in firefox.
That means in this micro benchmark, Firefox for Object.entries is more than 6 times faster than chrome.

I still think that it makes sense to consider this change in this place, as even in Firefox `Object.keys` is at least as fast as `Object.entries` and I don't think that the code quality suffers too much from this one change. Maybe some day V8 will optimize this better overall.

JSPerf for reference: https://jsperf.app/yabije

CarlosFdez:
The interesting bit is that its not even the destructuring, its Object.entries() itself. We'll have to see where we can begin incorporating maps, but I can imagine that the initialization of a map is where most of its expense is.

PR #16575, opened 2024-09-18
Fix issue with permanently hidden Campaign Features in Comp Browser

### LINKED ISSUE #15555: Campaign tab of the compendium browser is empty after changing browser source settings
1. Turn on campaign specific features in the system homebrew elements settings
2. Open the compendium browser. Shows campaign feats and the source filter as normal
3. Turn off a source in the settings tab
4. Campaign tab no longer shows any feats and the source filters are blank

image

Re-enabling whatever source got turned off or resetting the source settings doesn't fix the issue.

v12 Build 329
pf2e v6.1.1

DocSchlock:
Closes #15555

Issue was the campaignFeature sources were getting permanently set to load = false.

I noticed there was no pack settings to toggle it, so I added it and it fixed the issue.

### DIFF
--- a/static/templates/compendium-browser/settings/pack-settings.hbs
+++ b/static/templates/compendium-browser/settings/pack-settings.hbs
@@ -53,6 +53,7 @@
     </dl>
 </div>
 
+{{#if showCampaign}}
 <div class="setting-section">
     <h3>{{localize "PF2E.Item.CampaignFeature.Plural"}}</h3>
     <dl>
@@ -61,3 +62,4 @@
         {{/each}}
     </dl>
 </div>
+{{/if}}
--- a/static/templates/compendium-browser/settings/pack-settings.hbs
+++ b/static/templates/compendium-browser/settings/pack-settings.hbs
@@ -53,7 +53,7 @@
     </dl>
 </div>
 
-{{#if showCampaign}}
+{{#if @root.showCampaign}}
 <div class="setting-section">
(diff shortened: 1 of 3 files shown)


CarlosFdez:
I remember hiding campaign features when kingmaker is disabled. Can you also hide the setting along with it?

DocSchlock:
> I remember hiding campaign features when kingmaker is disabled. Can you also hide the setting along with it?

So I took a look: the template's context for settings is narrowed using a with block that excludes showCampaign, the visibility flag.

I would need to unravel the context narrowing - if that's okay, I can do it, but otherwise I'm not sure how to get the flag to the partial's context to correctly hide the pack setting.

EDIT: I guess the other option would be cloning the flag into the settings structure so it can be referenced even with the narrowing.

DocSchlock:
Decided to go the clone route.

CarlosFdez:
You can always use @root. to get the root one, as well as ../ to go one step up, if that works better with less changes.

static/templates/compendium-browser/settings/pack-settings.hbs
```diff
@@ -52,3 +52,14 @@
         {{/each}}
     </dl>
 </div>
+
+{{#if @root.showCampaign}}
+<div class="setting-section">
```
CarlosFdez, line 57: Indent everything in the if

PR #16580, opened 2024-09-18
Automate Shield Augmentation

TikaelSol:


pedrogrullada:
For this it might be worth choosing a specific shield to alter too? Two-shield wielding maniacs are very likely to make use of Shield Augmentations.

TikaelSol:
The problem with that is that you have to make choices on item creation, but you may want to move the augmentation to another shield later.

TikaelSol:
We don't generally have good automation for dual shield weirdos so I kind of just ignored them

pedrogrullada:
I guess that part is a hurdle that could be crossed if attachments are expanded to support augmentations and have roll options to leverage.

SpartanCPA (approved):

PR #16581, opened 2024-09-18
Fix divine health modifier

mose:
I'm not 100% sure about this one, because people in the champion's aura get that +1, but the champion himself gets +2.


pedrogrullada:
You are correct that allies get only a +1, which is what this effect is for. The +2 is already implemented in the feat itself.

PR #16588, opened 2024-09-18
Revert 'Fix divine health modifier'

pedrogrullada:
I commented on #16581, but I guess I should've closed it. This effect is meant to have only a +1, since it's the bonus that allies receive through the champion's aura. The champion's +2 is on the feat itself.


TikaelSol:
I flipped the changes in my head, for some reason I saw the 1 in red and was like "yup, that's the right value"

PR #16593, opened 2024-09-19
Make Crossbow Crackshot's additional precision damage untyped and add increased backstabber damage

### LINKED ISSUE #14769: Crossbow crack shot doesn't interact with Backstabber trait

pedrogrullada:
also fix weapon group selectors. The AdjustModifier on the backstabber damage is not working due to some bug in retrieving the weapon damage dice, it seems. I'll try to confirm that, but that's unrelated to this particular fix.

Closes #14769 though the general limitation of AdjustModifier remains.

### DIFF
--- a/packs/feats/crossbow-crack-shot.json
+++ b/packs/feats/crossbow-crack-shot.json
@@ -38,17 +38,17 @@
                     "crossbow-crack-shot"
                 ],
                 "selector": "crossbow-weapon-group-damage",
+                "slug": "crossbow-crack-shot-bonus",
                 "value": "@weapon.system.damage.dice"
             },
             {
                 "key": "AdjustModifier",
-                "mode": "upgrade",
                 "predicate": [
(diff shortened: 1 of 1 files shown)


TikaelSol:
We can maybe work around the bug by suppressing the backstabber damage and adding a new flat modifier in its place?

pedrogrullada:
> We can maybe work around the bug by suppressing the backstabber damage and adding a new flat modifier in its place?

If we're ok with that workaround, then yeah, I'll get to it tonight

PR #16603, opened 2024-09-20
Add spell effect for Charged Javelin

pedrogrullada:
`item:material:metal` is not an actual roll option, but it will show the bonus in the roll dialog for users to toggle on when targeting the creature with the effect

### DIFF
--- a/packs/spell-effects/spell-effect-charged-javelin-attacker.json
+++ b/packs/spell-effects/spell-effect-charged-javelin-attacker.json
@@ -29,7 +29,23 @@
                             "item:trait:electricity",
                             {
                                 "and": [
-                                    "item:material:metal",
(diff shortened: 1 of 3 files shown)


packs/spell-effects/spell-effect-charged-javelin-attacker.json
```diff
@@ -0,0 +1,56 @@
...
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    {
+                        "or": [
+                            "item:trait:electricity",
+                            {
+                                "and": [
+                                    "item:material:metal",
```
TikaelSol, line 32: We can at least do an or with all the known metal types. adamantine, cold iron, etc

PR #16614, opened 2024-09-20
Condense specific Marshal Stance effects into single Marshal Aura effect

pedrogrullada:


TikaelSol:
Not sold on this one

pedrogrullada:
The idea was to reduce the number of effects being transmitted by the aura, and figured there was some rules support, in so far as the stances extend what Marshal Aura does, rather than being a brand new thing in and of themselves.

That's about it, feel free to close this if it's not what we want though, no worries.

PR #16624, opened 2024-09-22
Make compasses suppress Sense Direction penalty

pedrogrullada:


pedrogrullada:
Now that #16640 is merged, this one is ready to go

PR #16635, opened 2024-09-22
Remove attack bonus from Starlit Transformation effect

websterguy:
The action specifies a bonus to damage rolls, not attack rolls

### DIFF
--- a/packs/feat-effects/effect-starlit-transformation.json
+++ b/packs/feat-effects/effect-starlit-transformation.json
@@ -72,9 +72,9 @@
                     "bolts-of-starlight"
                 ],
                 "property": "damage-type",
-                "selectors": {
+                "selectors": [
                     "{item|flags.pf2e.rulesSelections.weapon}-damage"
-                },
+                ],
                 "value": "force"
             },
             {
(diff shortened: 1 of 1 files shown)


packs/feat-effects/effect-starlit-transformation.json
```diff
@@ -41,7 +41,6 @@
             {
                 "key": "FlatModifier",
                 "selector": [
-                    "{item|flags.pf2e.rulesSelections.weapon}-attack",
                     "{item|flags.pf2e.rulesSelections.weapon}-damage"
                 ],
```
pedrogrullada, line 45: You can lose the array too

packs/feat-effects/effect-starlit-transformation.json
```diff
@@ -75,9 +74,7 @@
                     "bolts-of-starlight"
                 ],
                 "property": "damage-type",
-                "selectors": [
-                    "{item|flags.pf2e.rulesSelections.weapon}-damage"
-                ],
+                "selectors": "{item|flags.pf2e.rulesSelections.weapon}-damage",
```
pedrogrullada, line 77: In the case of Damage Alteration, you do need an array even if it's a single one

websterguy: made the wrong change...

PR #16644, opened 2024-09-23
Remove charisma and wisdom options from Hand of the Apprentice

pedrogrullada:
No other way to access this spell besides Wizard

### DIFF
--- a/packs/spell-effects/spell-effect-hand-of-the-apprentice.json
+++ b/packs/spell-effects/spell-effect-hand-of-the-apprentice.json
@@ -4,7 +4,7 @@
     "name": "Spell Effect: Hand of the Apprentice",
     "system": {
         "description": {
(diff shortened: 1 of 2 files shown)


ammalagonc:
Wizard Dedication?

pedrogrullada:
> Wizard Dedication?
Your attribute would still be Intelligence in that case

ammalagonc:
OH
I remember why this was added - There was that one Livestream where they were showcasing this spell and there was an argument that because Sorcerer was using their already existing Focus Pool (which was keyed off CHA) it should have applied the CHA and not the INT.

pedrogrullada:
> OH I remember why this was added - There was that one Livestream where they were showcasing this spell and there was an argument that because Sorcerer was using their already existing Focus Pool (which was keyed off CHA) it should have applied the CHA and not the INT.

Is "they" Paizo? In any case, that contradicts the rules

>**Focus Points from Multiple Sources**
>It’s possible, especially through archetypes, to gain focus spells from more than one source. If this happens, you have just one focus pool, counting all your focus spells to determine the points in your pool. You can spend any of your Focus Points on any of your focus spells. Likewise, when you Refocus, you get back a point as long as you follow the guidelines of any abilities that granted you focus spells. Having Focus Points from multiple sources doesn’t change the tradition of your spells; if you had both cleric domain spells and druid order spells, your domain spells would remain divine and the order spells primal. **Similarly, you need to use the attribute modifier determined by the source of the focus spell.**

from Player Core pg. 298

ammalagonc:
Perfect - Paperwork is good someone asks why

PR #16661, opened 2024-09-24
Always include origin and target distance roll option

CarlosFdez:
I am not certain why we are typechecking distance, but I'm not about to change that.


stwlam:
There's no point in adding a roll option with a distance of `null`.

PR #16671, opened 2024-09-24
Propagate homebrew damage types to config

CarlosFdez:


src/module/system/settings/homebrew/damage.ts
```diff
@@ -19,6 +19,8 @@ import { HomebrewElements } from "./menu.ts";
 export class DamageTypeManager {
     // All collections that homebrew damage must be updated in
     collections = {
+        physicalConfig: CONFIG.PF2E.physicalDamageTypes as Record<string, string>,
+        energyConfig: CONFIG.PF2E.energyDamageTypes as Record<string, string>,
         physical: PHYSICAL_DAMAGE_TYPES as unknown as string[],
         energy: ENERGY_DAMAGE_TYPES as unknown as string[],
```
stwlam, line 25: ??

CarlosFdez: those fields are readonly and thus don't allow us to make them mutable without doing this. The alternative is changing the core type elsewhere to not be readonly.

PR #16740, opened 2024-10-03
Conserve compendium source for embedded items when extracting

CarlosFdez:
This doesn't fix the items that are missing a compendium source, not convert the ones that are using flags to use compendium source.

### DIFF
--- a/build/lib/compendium-pack.ts
+++ b/build/lib/compendium-pack.ts
@@ -258,7 +258,6 @@ class CompendiumPack {
 
         if (isActorSource(docSource)) {
             docSource.effects = [];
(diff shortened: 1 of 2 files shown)


stwlam:
This should only be done for embedded items.

CarlosFdez:
It should be only for embedded items yeah. Unfortunately, the tree prune works recursively, and we call the tree prune *again* for embedded items, so we can't get too aggressive there. I added an additional _stats strip to #sanitizeDocument() to fully strip from non-embedded documents. It might be possible to handle all _stats pruning there, but I'm a bit concerned if there might be repercussions so I haven't done that yet.

CarlosFdez:
I made the title more descriptive. This is for embedded and double embedded items.

PR #16743, opened 2024-10-03
Also set defaults for sub items when running migration script

CarlosFdez:


stwlam:
Any changes with a migration run?

CarlosFdez:
Well, ,currently it just crashes, so that's a bit hard to tell what changes are good/bad.

I'm not sure which changes are valid. A creature's wand loses their embedded spell, but we can avoid that by manually adding a source id. There's some changes when rerunning the migration on megaton strike, but I want data entry's opinion before I handle that.

PR #16769, opened 2024-10-04
Add mutagenist drawback suppression to Bestial Mutagen

pedrogrullada:
Proof of concept for a way to handle the mutagenist's drawback suppression. I don't have time to do them all right now, but the idea is to do this with all mutagens.
image

### DIFF
--- a/packs/equipment-effects/effect-bestial-mutagen-greater.json
+++ b/packs/equipment-effects/effect-bestial-mutagen-greater.json
@@ -32,6 +32,11 @@
             },
             {
                 "key": "FlatModifier",
+                "predicate": [
+                    {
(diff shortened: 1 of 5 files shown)


CrackJackFlood:
I have done similiar thing, was told system doesn't want to hardcode things like this

TikaelSol:
I don't want to bake suppression in to the effects, I'd rather find a way to suppress penalties more generally so that we can just do it once and be done.

pedrogrullada:
> I don't want to bake suppression in to the effects, I'd rather find a way to suppress penalties more generally so that we can just do it once and be done.
The drawbacks are a bit arbitrary which makes it tricky, but I'll give it some more thought. We might need new tools for something like that.

PR #16788, opened 2024-10-06
Clean up Campaign Effects compendium and brush up effects.

ammalagonc:
1. Folders were added to the Campaign Effects folder, making sorting items easier.
2. Moved Resonant Reflection and Season of Ghost specific actions to Adventure Specific Actions compendium 
3. Brush up some effects text to match standard conventions.


packs/campaign-effects/pathfinder-society/effect-heart-break.json
```diff
@@ -18,7 +19,7 @@
         "publication": {
             "license": "OGL",
             "remaster": false,
-            "title": ""
+            "title": "Pathfinder Society Scenario #4-11: Prisoners of the Electric Castle"
```
TikaelSol, line 22: This should be a bestiary effect. Maybe there's more like that in here too?

TikaelSol: Fine to fix that as a follow-up though.

PR #16819, opened 2024-10-08
Added Sovereign heritage to dragonblood

MrFlou:
Adding Sovereign Heritage to Dragonblood, since it seemed to be missing. Not sure if it was left out on purpose?

### DIFF
--- a/packs/heritages/versatile-heritages/dragonblood.json
+++ b/packs/heritages/versatile-heritages/dragonblood.json
@@ -100,7 +100,17 @@
                             "shape": "cone",
                             "tradition": "occult"
                         }
-                    }
+                    },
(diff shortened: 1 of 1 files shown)


pedrogrullada:
We only have the options presented in Player Core 2 at the moment. There is PFS's remaster dragons table but they didn't specify a shape for each dragon's breath weapon, so we need clarification on that first.

MrFlou:
Ah that makes sense! Was unsure about shape, so went with what the archive said the dragon had.

> We only have the options presented in Player Core 2 at the moment. There is PFS's remaster dragons table but they didn't specify a shape for each dragon's breath weapon, so we need clarification on that first.

PR #16842, opened 2024-10-10
Show label for saving throws roll if exists in @check

reyzor1991:
Text from Message
`The target is knocked @UUID[Compendium.pf2e.conditionitems.Item.j91X7x0XSomq8d60]{Prone} unless they succeed at a @Check[fortitude|against:class|name:Hammer Critical Specialization] save against your class DC.
`

Rigth now name of Check is ignored for saving throws rolls

Message 
image

Old roll & New roll
imageimage

### DIFF
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -275,6 +275,7 @@ export class InlineRollLinks {
                     if (defenseStat) {
                         return {
                             label:
+                                pf2Label ??
                                 defenseStat.dc.label ??
                                 game.i18n.format("PF2E.InlineCheck.DCWithName", { name: defenseStat.label }),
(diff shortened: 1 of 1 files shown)


stwlam:
Do you have a better example case? In the above, it's more informative that this is the fighter class DC.

reyzor1991:
Example would be very similar because if ```@Check``` contains info about against:class it would be post name of class 
like Fighter DC, but it also ignore name in check if it exists

```@Check[fortitude|against:class|name:Hammer Critical Specialization]```

reyzor1991:
Sometime it's more informative like here. One roll vs rune, one vs crit. spec., but if need make 3+ roll and all would contains agains:class, then saving throw message would be similar and need to spend more time to understand which effect apply based on result

image

PR #16865, opened 2024-10-12
Initial implementation of SpecialResource rule element

CarlosFdez:
While it is functional, it is still a work in progress, and will need to be run through a bit of rigor to be as good as it can be. The physical item binding, while *generally* functional, is untested and will be completed during the crafting rework. Token bar updates work though, as does updating the max via AELike rule elements.

These bits of functionality are missing or spotty:
1) Slugs are hyphen cased, and included in system.resources as is. This becomes awkward when you have `system.resources.heroPoints` and `system.resources.vitality-network`. We could make them bactrian in system.resources and kebab in synthetics.resources, which is a bit weird to me, but perhaps its fine to others?
2) They aren't shown for npcs yet. I'd like to get them shiny in character first.
3) It assumes full regeneration during daily prep. I'm not sure of a good syntax for configuring it.
4) itemUUID? itemUuid? item?
5) They always display in the sidebar as input fields. We may want pip displays for some. I can't think of a good syntax.
6) Localization

image

### DIFF
--- a/src/module/actor/creature/document.ts
+++ b/src/module/actor/creature/document.ts
@@ -30,7 +30,14 @@ import { PerceptionStatistic } from "@system/statistic/perception.ts";
 import { ErrorPF2e, localizer, setHasElement } from "@util";
 import { eventToRollParams } from "@util/sheet.ts";
 import * as R from "remeda";
-import { CreatureSpeeds, CreatureSystemData, LabeledSpeed, VisionLevel, VisionLevels } from "./data.ts";
+import {
+    CreatureResources,
+    CreatureSpeeds,
+    CreatureSystemData,
+    LabeledSpeed,
+    VisionLevel,
+    VisionLevels,
+} from "./data.ts";
 import { imposeEncumberedCondition, setImmunitiesFromTraits } from "./helpers.ts";
 import { CreatureTrait, CreatureType, CreatureUpdateOperation, GetReachParameters } from "./types.ts";
 
@@ -616,6 +623,19 @@ abstract class CreaturePF2e<
         });
     }
 
+    /** Updates a resource. Redirects to special resources if needed */
+    async updateResource(resource: string, value: number): Promise<void> {
+        const resources = this.system.resources;
+
+        const special = this.synthetics.resources[resource];
+        if (special) {
+            await special.update(Math.clamp(value, 0, special.max));
+        } else if (!!resources?.[resource] && ["heroPoints", "focus", "resolve"].includes(resource)) {
+            value = Math.clamp(value, 0, resources[resource]?.max ?? 0);
+            await this.update({ [`system.resources.${resource}.value`]: value });
+        }
+    }
+
     prepareSpeed(movementType: "land"): this["system"]["attributes"]["speed"];
     prepareSpeed(movementType: Exclude<MovementType, "land">): (LabeledSpeed & StatisticModifier) | null;
     prepareSpeed(movementType: MovementType): CreatureSpeeds | (LabeledSpeed & StatisticModifier) | null;
@@ -778,6 +798,15 @@ abstract class CreaturePF2e<
             const updatedPoints = Number(focusUpdate.value ?? this.system.resources.focus?.value) || 0;
             const enforcedMax = (Number(focusUpdate.max) || this.system.resources.focus?.max) ?? 0;
             focusUpdate.value = Math.clamp(updatedPoints, 0, enforcedMax);
+        }
+
+        // Remove special resources from update data
+        if (changed.system.resources) {
+            for (const special of Object.keys(this.synthetics.resources)) {
+                if (special in changed.system.resources) {
+                    delete (changed.system.resources as CreatureResources)[special];
+                }
+            }
         }
 
         // Preserve alignment traits if not exposed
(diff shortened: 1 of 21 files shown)


CrackJackFlood:
Possibility to have something like these show else where like in the spellcasting tab as seperate resourse similiar to focus points, could open the door to altnerative ways of spellcasting or just different forms of focus points.

CarlosFdez:
If Paizo implements it we'll consider something like that, but for the sake of avoiding assumptions and having to update it all later once we're wrong, we'll keep it to things that exist. Resources will be consumable for some things like abilities, but a spellcasting collection type won't be made for it until we need it.

CrackJackFlood:
> If Paizo implements it we'll consider something like that, but for the sake of avoiding assumptions and having to update it all later once we're wrong, we'll keep it to things that exist in published content. Resources will be consumable for some things like abilities, but a spellcasting collection type won't be made for it until we need it.

Not even some AElike or RE override redirecting pointing focus point path to a new resource path? I understand not wanting to implement something like that overall for spellcasting in case paizo do something but a cheeky liitle way to fake it would be nice. Maybe that is something to consider more when spellcasting tab gets some more love though.

src/module/actor/creature/sheet.ts
```diff
@@ -53,6 +53,14 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
                 remainingDying: Math.max(actor.attributes.dying.max - actor.attributes.dying.value),
                 remainingWounded: Math.max(actor.attributes.wounded.max - actor.attributes.wounded.value),
             },
+            specialResources: Object.values(this.actor.synthetics.resources)
+                .filter((r) => !r.itemUUID)
+                .map((r) => ({
```
stwlam, line 58: R.pick!

CarlosFdez: I suspect eventually we'll have something there that can't simply be picked, and its why I leaned on that at the start....but I suppose I definitely should convert that now.

src/module/actor/sheet/base.ts
```diff
@@ -347,7 +347,7 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
 
                 const value = (() => {
                     const value =
-                        element instanceof HTMLInputElement && element.type === "checbox"
```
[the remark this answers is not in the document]
CarlosFdez: 🙃

src/module/actor/creature/document.ts
```diff
@@ -778,6 +798,15 @@ abstract class CreaturePF2e<
             focusUpdate.value = Math.clamp(updatedPoints, 0, enforcedMax);
         }
 
+        // Remove special resources from update data
```
stwlam, line 801: Are we not able to prevent these from getting included from the template?

CarlosFdez: Its mostly a fail safe in case I missed a spot. It was originally added because of token hud updates, but that's been redirected now. Now it mostly just backhands module devs or temporary breaking changes due to updates.

Ideally I'd prefer to intercept it and do an embedded item update, but embedded item updates triggered as part of an actor update do not go through the _onX pipeline.

Would you prefer if I remove the failsafe? I'm good either way.

src/module/actor/sheet/base.ts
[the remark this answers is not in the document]
CarlosFdez: This is actually a piece of code that isn't in use, so I kinda did a naughty and slipped it in. There are currently no data-item-property input fields at the current moment, but it'd be a nasty surprise if someone adds one eventually and we kill the handling.

src/module/rules/rule-element/special-resource.ts
```diff
@@ -0,0 +1,198 @@
...
+    "crafting",
+    "focus",
+    "heroPoints",
+    "investiture",
+    "infusedReagents",
+    "resolve",
+];
+
+class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema> {
+    protected static override validActorTypes: ActorType[] = ["character", "npc"];
```
stwlam, line 21: OK for now, but we should look at expanding this as people request it.

src/module/rules/rule-element/special-resource.ts
```diff
@@ -0,0 +1,198 @@
...
+    /** Updates the rule element's max value to match alterations such as by ActiveEffectLike, and snapshots the resolved value */
+    #updateMax() {
+        const thisMax = Number(this.resolveValue(this.max));
+        const existingMax = this.actor.system.resources[this.slug]?.max ?? 0;
+        this.max = Math.max(thisMax, existingMax);
+    }
+
+    /** Updates the remaining number of this resource. Where it updates depends on the type */
+    async update(value: number, options: { commit: false }): Promise<SpecialResourcesCommitData>;
+    async update(value: number, options?: { commit?: boolean }): Promise<void>;
```
stwlam, line 52:
```suggestion
    async update(value: number, options: { commit: true }): Promise<void>;
    async update(value: number, options?: { commit?: boolean }): Promise<SpecialResourcesCommitData | void>;
```

src/module/rules/rule-element/special-resource.ts
CarlosFdez: This was originally character and I extended it to npc, but to extend further we would need to ensure that the resource object exists at least at the creature level.

CarlosFdez: I ended up removing NPC, as we're not using it for mythic points. There was a 50/50 chance when I wrote this PR but now its vanishingly small. This PR doesn't include any UI for the NPC resource either. We can add npc/familiars as people request them.

Issue #16510, opened 2024-09-12
Weapon Proficiencies: If Martial is treated as Simple, and Advanced as Martial, then Advanced is treated as Simple


ToonVanEysendeyk:
When implementing two Weapon Proficiency Rule Elements, one that sets an Advanced Weapon category to be treated as Martial, and a second that sets a similar Martial Weapon category to be treated as Simple, then the Advanced Weapon of that category is also treated as Simple.

As example, Dwarven Weapon Familiarity treats all Martial Dwarf weapons as Simple, and all Advanced Dwarf weapons as Martial. However, the Advanced Dwarven Dorn-Dergar is treated as a Simple Weapon as a result.

This also happens when choosing a specific weapon, such as with Samsaran Weapon Memory, which takes the chosen weapon and has two Rule Elements; one to treat it as Martial if it were Advanced, and one to treat it as Simple if it were Martial.

PR #16879, opened 2024-10-12
Fix advanced weapons being treated as simple instead of martial with certain feats

sbodegraven:
Fix for https://github.com/foundryvtt/pf2e/issues/16510


stwlam:
Hi there, this turned out to simply be a duplicate processing of "same-as" proficiencies.

PR #16884, opened 2024-10-13
Convert CompendiumBrowser to ApplicationV2

In3luki:
- Rewrite browser as a Svelte app
- Move browser settings to a separate Application
- Add "Settings" button to browser window controls
- Move roll table buttons to browser window controls 
- Use gsap for the drag fade animations
- Fix (or implement?) setting the level filter in the crafting tab of the character sheet
- Fix up dark mode styling
- Enable opening the browser with a selection of nagivation tabs (or no navigation at all)
- Cleaned up translations in `en.json`

Some filter changes:
- Removed `multiselects` and `sliders` from the filter as they were unused except for the `traits` and `level` filters
- `traits` and `level` are now top level filter data.
- The `level` filter now renders as two select elements for min and max

Controls:
c-light-settings

Level filter:
browser-levels

Light:
browser-light

Dark:
browser-dark


In3luki:
yeah I wanted to wait for svelte support initially but the version 5 release has taken so long that I thought it made sense to convert to the handlebars renderer first. Switching out the renderer later should be relatively straight forward.

In3luki:
But if that isn't something you want to do I can keep this in a stash until svelte is ready. It definitely helped me to better understand ApplicationV2 so it wasn't for nothing.

src/module/apps/compendium-browser/settings.ts
```diff
@@ -0,0 +1,199 @@
...
+        hazard: {
+            label: "PF2E.Actor.Hazard.Plural",
+        },
+        spell: {
+            label: "PF2E.Item.Spell.Plural",
+        },
+    };
+
+    static override DEFAULT_OPTIONS: DeepPartial<ApplicationConfiguration> = {
+        ...super.DEFAULT_OPTIONS,
```
stwlam, line 36: Upstream takes care of merging for you

src/module/apps/sidebar/item-directory.ts
```diff
@@ -68,7 +68,7 @@ export class ItemDirectoryPF2e<TItem extends ItemPF2e<null>> extends ItemDirecto
             game.i18n.localize("PF2E.CompendiumBrowser.Title"),
         );
         browseButton.addEventListener("click", () => {
-            game.pf2e.compendiumBrowser.render(true, { focus: true });
+            game.pf2e.compendiumBrowser.render(true);
```
stwlam, line 71:
```suggestion
            game.pf2e.compendiumBrowser.render({ force: true });
```

src/module/apps/compendium-browser/settings.ts
```diff
@@ -0,0 +1,199 @@
+import { htmlQuery, htmlQueryAll, localizer, objectHasKey } from "@util";
+import type {
+    HandlebarsRenderOptions,
+    HandlebarsTemplatePart,
+} from "types/foundry/client-esm/applications/api/handlebars-application.ts";
+import type { PackInfo, TabName } from "./data.ts";
+
+const foundryApp = foundry.applications.api;
+
+class CompendiumBrowserSettingsApp extends foundryApp.HandlebarsApplicationMixin(foundryApp.ApplicationV2) {
```
stwlam, line 10: A separate app?

src/module/apps/compendium-browser/index.ts
```diff
@@ -68,48 +73,164 @@ class CompendiumBrowser extends Application {
...
     }
 
-    static override get defaultOptions(): ApplicationOptions {
-        return {
-            ...super.defaultOptions,
-            id: "compendium-browser",
-            classes: [],
-            template: "systems/pf2e/templates/compendium-browser/compendium-browser.hbs",
+    static override DEFAULT_OPTIONS: DeepPartial<ApplicationConfiguration> = {
+        ...super.DEFAULT_OPTIONS,
```
stwlam, line 77:
```suggestion
```

src/module/apps/compendium-browser/index.ts
```diff
@@ -68,48 +73,164 @@ class CompendiumBrowser extends Application {
...
+                    label: "PF2E.CompendiumBrowser.RollTable.AddLabel",
                 },
                 {
-                    navSelector: "nav[data-group=settings]",
-                    contentSelector: ".settings-container",
-                    initial: "packs",
+                    action: "createRollTable",
+                    icon: "fa-solid fa-list",
+                    label: "PF2E.CompendiumBrowser.RollTable.CreateLabel",
                 },
```
stwlam, line 100: Header controls now?

src/module/apps/compendium-browser/settings.ts
In3luki: The settings tab is taking up real estate and doesn't really fit in the same list as the content tabs. Also I don't think it's used nearly as much as the other tabs so it makes sense to move it out of the way.

src/module/apps/compendium-browser/index.ts
In3luki: I didn't like the floating buttons from the start but there was no room to put them anywhere else. This moves them out of the way but keeps them still accessable.

In3luki:
I'll yoink the basic svelte stuff from the other PR and try it out for this. Converted to a draft in the meantime.

In3luki:
I've pushed the svelte browser and updated the description with new infos and current screenshots. The changed file count just kept rising, sorry. 🙃

src/module/apps/compendium-browser/browser.svelte.ts
```diff
@@ -0,0 +1,428 @@
...
+        const gmControls = ["addToRollTable", "createRollTable", "openSettings"];
+        for (const control of controls) {
+            if (!game.user.isGM && gmControls.includes(control.action)) {
+                control.visible = false;
+            }
+        }
+        return controls;
+    }
+
+    #getNavTabs(requested?: ContentTabName[]): CompendiumBrowserNavTab[] {
```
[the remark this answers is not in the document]
In3luki: Good call with the label. I'm not sure if a `visible` property would be better as that breaks easy reactivitiy. The property would have to be a `$state` itself to make it reactive as the `$state` proxying only works with simple objects and arrays.
That would make it necessary to make the tabs `.svelte.ts` files to allow the use of `$state` inside the classes.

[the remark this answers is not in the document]
In3luki: It was actually easier than I thought so now navigaton visibility is determined by `BrowserTab#visible`.

src/module/apps/svelte-mixin.svelte.ts
```diff
@@ -0,0 +1,43 @@
+import * as svelte from "svelte";
+import type {
+    ApplicationRenderContext,
+    ApplicationRenderOptions,
+} from "types/foundry/client-esm/applications/_types.d.ts";
+import type ApplicationV2 from "types/foundry/client-esm/applications/api/application.d.ts";
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+function SvelteApplicationMixin<TBase extends AbstractConstructorOf<ApplicationV2>>(Base: TBase) {
+    abstract class SvelteApplication extends Base {
```
CarlosFdez, line 10: Don't we have one of these? I ended up doing a bugfix to the replaceHTML() implementation that is somewhat similar to yours, if you wish to poach it, that maintains $state being a protected property.

In3luki: That is a vestige from me starting with the browser before the abc-picker was merged. I think the mixin location was moved multiple times and I forgot to update that-

static/lang/en.json
```diff
@@ -1609,7 +1608,11 @@
         "FeatActionTypeLabel": "Action Type",
         "FeatActionsLabel": "Number of Actions",
         "FeatBackgroundShort": "BG",
-        "FeatBrowserSearchHint": "You can search for name or custom attributes. Possible searchable attributes are:<br> source, spellType, level, materials, target, range, time, duration, damage, damageType, save, concentration, ritual, ability and classes. <br>Example: 'fire, damage:d6' to show all spells that have fire in their name and a d6 in the damage",
+        "FeatBonusHeader": "Bonus Feats",
+        "FeatCampaignHeader": "Campaign Feats",
+        "FeatClassHeader": "Class Feats",
+        "FeatDeityBoonCursesHeader": "Divine Intercessions",
+        "FeatGeneralHeader": "General Feats",
```
CarlosFdez, line 1615: These were moved elsewhere and should no longer exist at the top level

src/module/apps/compendium-browser/components/filters.svelte
```diff
@@ -0,0 +1,185 @@
...
+    </FilterContainer>
+    {#each Object.entries(filter.checkboxes) as [key, checkbox]}
+        <FilterContainer
+            isExpanded={checkbox.isExpanded}
+            clearButton={{ data: checkbox, options: { visible: checkbox.selected.length > 0 } }}
+            label={checkbox.label}
+        >
+            <Checkboxes bind:checkbox={filter.checkboxes[key as keyof BrowserFilter["checkboxes"]]} />
+        </FilterContainer>
+    {/each}
```
In3luki, line 103: `R.entries` adds a`never[]` here for some reason so I resorted to type assertion. A union type of all checkbox filters does not include `never` so I'm not really sure what's going on there.

CarlosFdez: I tested with converting everything to Record<"string" | "unions", CheckboxData> but that doesn't seem to work at all. Seems to be an issue with R.entries's implementation on a union of records with explicit keys.

Issue #14123, opened 2024-03-08
@Damage Missing Bracket Crash


Crowguard:
Using an `@damage` without a closing bracket and saving the item crashes Foundry; any subsequent attempts to edit the item also result in crashes (found on accident, didn't notice I was missing the bracket when I saved).

`@Damage[(floor((@actor.level)/2)+2){2 + half your level}`

https://github.com/foundryvtt/pf2e/assets/99844555/8eada9e0-c501-46b3-9724-4793b28cbcc1

Foundry 11.315
PF2e 5.13.6
Electron app & Forge

mclemente:
You can soft-brick a world if you post it to chat too. The only fix I found was deleting the messages DB file on the world's folder.
I had reported this privately to a dev the other day because I wasn't sure making it public wasn't dangerous.

PR #16895, opened 2024-10-14
Fix Crash/Freeze when missing @Damage closing bracket

sbodegraven:
Closes https://github.com/foundryvtt/pf2e/issues/14123


CarlosFdez:
Testing in regex101, it doesn't seem like your pattern matches `@Damage[{2d6[fire],4d6[acid]}]`

sbodegraven:
@CarlosFdez thanks for catching that, I didn't realize that curly braces were also allowed within the square brackets. 

I've updated the regex so it now matches your formula, and I believe it now completely resolves the reported issue.

CarlosFdez:
From my limited testing, new one seems to not be subject to catastrophic backtracking anywhere near as easily, and it seems to satisfy the same expressions. I'm down to give it a try.

PR #16898, opened 2024-10-15
Add interval area heightening to some spells via Item Alteration

pedrogrullada:


TikaelSol:
@stwlam Think this is something we could wire up automatically from the details instead of writing a bunch of REs?

CarlosFdez:
We absolutely should not be doing this with REs, this is a feature request. Area is a number so this is doable now.

PR #16899, opened 2024-10-15
Recharge turn and round abilities as the encounter changes turns

CarlosFdez:
There is a very strange bug where the self effect image disappears until next re-render if the ability got updated. It seems to be an issue with the fallback uuid sync fallback, so I'll leave that to shark to figure out as he's more familiar with that side of things.

Merging this or the special resource branch will necessitate updating the other. That's fine by me though.


src/module/actor/types.ts
```diff
@@ -10,6 +10,7 @@ import type { immunityTypes, resistanceTypes, weaknessTypes } from "@scripts/con
 import type { DamageRoll } from "@system/damage/roll.ts";
 import type { DegreeOfSuccessString } from "@system/degree-of-success.ts";
 import type { Predicate } from "@system/predication.ts";
+import { ActorSourcePF2e } from "./data/index.ts";
```
stwlam, line 13:
```suggestion
import type { ActorSourcePF2e } from "./data/index.ts";
```

PR #16911, opened 2024-10-17
Organize character feat collection classes and interfaces

CarlosFdez:
There is only one actual change here. FeatGroupOptions split into FeatGroupData and options. Level is used by kingmaker to pass in the level of the kingdom instead of the actor's level, but it was not part of the actual structural metadata. That is not the sort of data we'd want to store in the campaign feat settings when we finally make a UI for it, so I split it off.

### DIFF
--- a/src/module/actor/character/feats/collection.ts
+++ b/src/module/actor/character/feats/collection.ts
@@ -1,9 +1,9 @@
-import type { ActorPF2e, CharacterPF2e } from "@actor";
-import type { FeatPF2e, HeritagePF2e, ItemPF2e } from "@item";
-import { ItemSystemData } from "@item/base/data/system.ts";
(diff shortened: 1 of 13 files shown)


src/module/actor/character/feats/collection.ts
```diff
@@ -116,7 +116,7 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<FeatGroup<
         }
     }
 
-    createGroup(options: FeatGroupOptions): this {
+    createGroup(options: FeatGroupData): this {
         return this.set(options.id, new FeatGroup(this.actor, options));
```
stwlam, line 120:
```suggestion
    createGroup(data: FeatGroupData): this {
        return this.set(data.id, new FeatGroup(this.actor, data));
```

PR #16944, opened 2024-10-19
Add linting/formatting for svelte files

In3luki:
Note: `svelte-checker` emits a warning if no svelte files are found in the workspace.
Also moves the `vite-checker` plugin to development mode only as `svelete-checker` provides type checking in production.


In3luki:
formatting output sample:
```svelte
    {#each R.values(filter.checkboxes) as checkbox}
        <fieldset class="sortcontainer">
            <legend>
                {game.i18n.localize(checkbox.label)}
                <button
                    type="button"
                    class="expand-section"
                    onclick={() => (checkbox.isExpanded = !checkbox.isExpanded)}
                    aria-label="expand"
                    aria-expanded={checkbox.isExpanded}
                >
                    <i class="fa-solid {checkbox.isExpanded ? 'fa-chevron-down' : 'fa-chevron-right'}"></i>
                </button>
            </legend>
            {#if checkbox.isExpanded}
                <div class="checkbox-container" transition:fade>
                    {#if checkbox.selected.length}
                        <button type="button" class="clear-filter" onclick={() => onClearFilter(checkbox)}>
                            {game.i18n.localize("PF2E.CompendiumBrowser.Filter.ClearFilter")}
                        </button>
                    {/if}
                    {#each R.entries(checkbox.options) as [name, option]}
                        <label>
                            <input
                                type="checkbox"
                                {name}
                                checked={option.selected}
                                onchange={(event) => onChangeCheckbox(event, checkbox, { name, data: option })}
                            />
                            {game.i18n.localize(option.label)}
                        </label>
                    {/each}
                </div>
            {/if}
        </fieldset>
    {/each}
```

.prettierrc.json
```diff
@@ -3,19 +3,14 @@
     "tabWidth": 4,
     "overrides": [
         {
-            "files": [
-                "*.scss",
-                "*.css"
-            ],
+            "files": ["*.scss", "*.css"],
```
stwlam, line 6: Looks like you have some autoformatting going rogue here

eslint.config.js
```diff
@@ -3,33 +3,27 @@
...
-            parserOptions: { project: "./tsconfig.json" },
-        },
         rules: {
             eqeqeq: "error",
             "prettier/prettier": "error",
             "no-console": "off",
             "no-plusplus": ["error", { allowForLoopAfterthoughts: true }],
             "no-unused-expressions": ["error", { allowShortCircuit: true }],
             "spaced-comment": ["error", "always", { markers: ["/"] }],
-            "@typescript-eslint/await-thenable": "error",
```
stwlam, line 32: ?

package.json
```diff
@@ -17,11 +17,12 @@
         "pretest": "npm run lint",
         "test": "jest",
         "migrate": "tsx ./build/run-migration.ts",
-        "lint": "npm run lint:ts && npm run lint:json && npm run prettier:scss",
+        "lint": "npm run svelte-check && npm run lint:ts && npm run lint:json && npm run prettier:scss",
```
stwlam, line 20: Why do we need to run `svelte-check` if we're also adding a new svelte-eslint package?

eslint.config.js
In3luki: That rule was throwing an eslint error in the global rules scope at one point but it doesn't anymore so I'll move it back.

package.json
In3luki: I thought `eslint-typescript` didn't work because `no-undef` errors for globals like `game` and `fu` were generated in svelte files. I just noticed that `no-undef` is also off in the recommended `typescript-eslint` options so I'll recheck if it actually works.

In3luki: I've tested with the following svelte file with the current config:
```svelte
<script lang="ts">
    type Test = Array<string>;

    //Test type checking
    game.i19n.localize("test");
</script>

```
eslint-svelte-errors

It's not picking up on the wrong `i19n` type. I actually think `typescript-eslint` doesn't pick up on type errors at all because the `game.i19n.localize("test");` isn't picked up in `npm run lint` in master either.

Running `svelte-check` will emit errors in both `.ts` and `.svelte` files:

svelte-check

stwlam: The linter doesn't do type checking; building does

In3luki: yeah, but `vite-plugin-checker` (which runs `tsc` after a build unless im totally mistaken) doesn't support svelte hence the inclusion of `svelte-checker`.

PR #16945, opened 2024-10-19
Correct action cost and duration for Soaring Flight feet and text on effect

### LINKED ISSUE #16942: Please revisit: Soaring Flight - Tengu Talent 5
Heyho. 
There are a couple of issues with Soaring Flight (Tengu 5)
Its a 1 Action - Once a Round Talent. 
Text should be:
"You Fly. If you don't normally have a fly Speed, you gain a fly Speed of 20 feet for this movement. If you aren't on solid ground at the end of this movement, you fall."

Right at the moment, we have a 2 Action, a Talent with a wrong description for Level 5 and a Level 9 Effekt, also with a wrong description. Could you please have a look into it? Thanks :)
For Reference - I tested on Pathfinder 6.4.1 w/o Modules

ammalagonc:
Closes #16942

### DIFF
--- a/packs/feat-effects/effect-soaring-flight.json
+++ b/packs/feat-effects/effect-soaring-flight.json
@@ -7,10 +7,10 @@
             "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Soaring Flight]</p>\n<p>You gain a fly Speed of 20 feet for this movement.</p>"
         },
         "duration": {
-            "expiry": "turn-start",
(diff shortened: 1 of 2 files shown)


packs/feat-effects/effect-soaring-flight.json
pedrogrullada, line 1: This should've probably been a toggle, but since the effect already exists, might as well make it 0 rounds and expire at the end of the turn.

PR #16949, opened 2024-10-20
Remove unneeded rule elements from Fortress Shield

pedrogrullada:
Also clean up Cursebreak Bulwark

### DIFF
--- a/packs/equipment/cursebreak-bulwark.json
+++ b/packs/equipment/cursebreak-bulwark.json
@@ -52,8 +52,7 @@
                 "key": "AdjustModifier",
                 "mode": "upgrade",
                 "predicate": [
-                    "item:trait:curse",
-                    "item:type:spell"
+                    "item:trait:curse"
                 ],
                 "selector": "saving-throw",
                 "slug": "cursebreak-bulwark-bonus",
--- a/packs/equipment/cursebreak-bulwark.json
+++ b/packs/equipment/cursebreak-bulwark.json
@@ -10,7 +10,7 @@
         },
         "containerId": null,
         "description": {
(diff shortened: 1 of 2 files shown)


packs/equipment/cursebreak-bulwark.json
```diff
@@ -38,23 +38,24 @@
...
+                "key": "AdjustModifier",
+                "mode": "upgrade",
                 "predicate": [
-                    "raisedShield",
-                    "curse"
+                    "item:trait:curse"
                 ],
                 "selector": "saving-throw",
-                "type": "circumstance",
+                "slug": "cursebreak-bulwark-bonus",
```
pedrogrullada, line 58: I made this an Adjust Modifier since the wording of the item says that the bonus increases against curses, but it's still a bonus against spells nevertheless. Let me know if you think it should stay as a separate Flat Modifier.

packs/equipment/cursebreak-bulwark.json
```diff
@@ -10,7 +10,7 @@
...
+            "value": "<p>This tower shield is composed of interlocking hexagonal wooden tiles lined with metal. The tiles are painted a verdant green, and each is marked with a faded rune that protects against curses. While you have this shield raised, you gain a +1 circumstance bonus to saving throws against spells that target you, increasing to a +2 circumstance bonus against curses. You can unleash the energy stored in the shield to destroy a curse before it can affect you.</p><hr /><p><strong>Activate</strong> <span class=\"action-glyph\">r</span> command</p>\n<p><strong>Frequency</strong> once per day</p>\n<p><strong>Trigger</strong> You are targeted with a curse spell</p>\n<p><strong>Requirements</strong> The <em>cursebreak bulwark</em> is raised</p><hr /><p><strong>Effect</strong> You activate the shield's countermeasures. It casts 8th-rank @UUID[Compendium.pf2e.spells-srd.Item.Cleanse Affliction] with a counteract modifier of [[/r 1d20+27 #Counteract]]{+27} against the triggering curse.</p><table class=\"pf2e\"><thead><tr><th>Hardness</th><th>HP</th><th>BT</th></tr></thead><tbody><tr><td>14</td><td>56</td><td>28</td></tr></tbody></table>"
```
pedrogrullada, line 13: Removed the "AC increases to +4 if you are using the Take Cover action.", as that's just part of the base item, and the current UI seems to convey that clearly enough.

PR #16956, opened 2024-10-21
Floor special resource and make it more receptive to aelikes

CarlosFdez:


src/module/rules/rule-element/special-resource.ts
```diff
@@ -104,6 +117,34 @@ class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema>
         }
     }
 
+    /** Treat special resources as upgrades during the AELike phase */
+    override onApplyActiveEffects(): void {
+        if (this.ignored) return;
```
stwlam, line 122: I guess this should be responsive to a predicate, though I can't envision a use case at the moment.

CarlosFdez:
This is an unreleased thing with no final impact but the bugfix will be needed for starfinder, so I'll merge it as is.

PR #16957, opened 2024-10-21
Add effect roll options of "parent:origin" and "parent:origin:item" if applicable

stwlam:


src/module/item/abstract-effect/document.ts
```diff
@@ -89,6 +89,12 @@ abstract class AbstractEffectPF2e<TParent extends ActorPF2e | null = ActorPF2e |
                     ? (origin.getSelfRollOptions("origin").map((o) => `${prefix}:${o}`) ?? [])
                     : [];
             })();
+        if (originRollOptions.length > 0) {
+            originRollOptions.push(`${prefix}:origin`);
+            if (originRollOptions.some((o) => o.startsWith(`$${prefix}:origin:item`))) {
```
CarlosFdez, line 94: $$?

stwlam: rb

PR #16973, opened 2024-10-22
Ignore disabled region behaviors when determining effect on actors

farling42:
Region behaviors (environment and environmentFeature) were always being processed, even when the behavior was not active.

This PR updates the filter searches to only consider region behaviors which are active (`behavior.active`).

### DIFF
--- a/src/module/canvas/ruler.ts
+++ b/src/module/canvas/ruler.ts
@@ -148,13 +148,13 @@ class RulerPF2e<TToken extends TokenPF2e | null = TokenPF2e | null> extends Rule
                 .filter(
                     (r) =>
                         r.document.behaviors.some(
-                            (b) => !b.disabled && b.type === "environmentFeature" && b.system.terrain.difficult > 0,
+                            (b) => !b.disabled && b.type === "environmentFeature" && b.active && b.system.terrain.difficult > 0,
                         ) && token.testInsideRegion(r, toPoint),
                 )
                 .flatMap((r) =>
                     r.document.behaviors.filter(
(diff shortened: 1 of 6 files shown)


stwlam:
An active region is also one that is not disabled, so we don't need both checks. What purpose does this serve that checking that they're not disabled already achieve?

farling42:
The current code for calculating movement cost always treats difficult terrain behaviors as active, even when they have been disabled using the tick box in the Region's settings.

farling42:
I don't see any code that might reference "disabled" region behaviors

src/module/canvas/ruler.ts
```diff
@@ -148,13 +148,20 @@ class RulerPF2e<TToken extends TokenPF2e | null = TokenPF2e | null> extends Rule
                 .filter(
                     (r) =>
                         r.document.behaviors.some(
-                            (b) => !b.disabled && b.type === "environmentFeature" && b.system.terrain.difficult > 0,
+                            (b) =>
+                                !b.disabled &&
+                                b.type === "environmentFeature" &&
+                                b.active &&
+                                b.system.terrain.difficult > 0,
```
stwlam, line 155: ?

farling42: Ah, I hadn't noticed that when going through all the cases of checks that matched "environment

farling42: I will update my PR to only add !b.disabled in places where it is missing

farling42:
Whilst https://github.com/foundryvtt/pf2e/commit/b287239f1ae7435c659832c5d776559ed01eb43b fixed the situation for the Ruler 7 hours ago, the problem still existed for environment changes.

PR #16974, opened 2024-10-22
Exclude `node_modules` directory in tsconfig watch options

In3luki:
I did some tests with a script from [here](https://unix.stackexchange.com/questions/15509/whos-consuming-my-inotify-resources/502812#502812) to monitor the amount of watched files.

`tsc --watch` before exclusion of  `node_modules`:
tsc-watch-count-1

`tsc --watch` after exclusion of  `node_modules`:
tsc-watch-count-no-node_modules

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -78,6 +78,7 @@
         "fallbackPolling": "dynamicpriority",
         "excludeDirectories": [
             "dist",
+            "node_modules",
             "packs",
             "static"
         ]
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -28,6 +28,7 @@
         "resolveJsonModule": true,
         "useDefineForClassFields": true,
         "noErrorTruncation": true,
+        "disableSizeLimit": true,
         "paths": {
             "@actor": [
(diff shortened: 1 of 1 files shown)


In3luki:
ok, this doesn't stop the svelte language server from crashing either. What does seem to work (so far) however is enabling the `disableSizeLimit` option: https://www.typescriptlang.org/tsconfig/#disableSizeLimit .
I'm not sure if that is safe to enable tho.

In3luki:
Since I enabled `disableSizeLimit` in the compendium browser branch the Svelte language server didn't crash or stopped working once. I've not seen any negative effects so far.
I can remove the `node_modules` exclusion but I think it would still be beneficial to not include those files.

In3luki:
Creating a new .svelte file often triggers the "Svelte language server detected a large number of JS/Svelte files" warning, even in the master branch. While `disableSizeLimit` stops the warning, it unfortunately breaks IntelliSense for newly created .svelte files in VSCode.
Closing as neither of these settings completely resolves the problem.

PR #16981, opened 2024-10-23
Update Typescript to version 5.6.3 with minimal code changes

In3luki:
It looks like `ActorSpellcasting` is fully responsible for the previous "type instantiation is excessively deep" errors. Changing that to not reference `this` and the necessary follow up changes clears up all other errors.
The only caveat is that the `TActor` generic has to be manually passed to  `ActorSpellcasting#collections#get` to correctly resolve the return type.
I haven't managed to solve that another way yet.


stwlam:
Kind of looks like https://github.com/foundryvtt/pf2e/pull/15356

PR #16984, opened 2024-10-23
 Check for wall collision when determining flanking buddies in `TokenPF2e#isFlanking`

In3luki:
Screenshot 2024-11-10 184135

### DIFF
--- a/src/module/canvas/token/object.ts
+++ b/src/module/canvas/token/object.ts
@@ -145,6 +145,27 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
         return this._canControl(user, event);
     }
 
+    /** Determine wether this token can see another token on the canvas. */
+    canSee(token: TokenPF2e): boolean {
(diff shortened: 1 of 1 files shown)


xyzzy42:
The wall collision check here is different than the one used for the aura to square wall collision check, in this function https://github.com/foundryvtt/pf2e/blob/498dd946ccf924f9bc4c4830289bdcf4bf312f76/src/module/canvas/token/aura/util.ts#L7

The way that code works, is it creates four source points, each offset in one of the cardinal directions from the center of token by ⅛ the token size.  Then it checks for any of those points having an unblocked line to the center of the destination square.  So it's just four checks, rather than sixteen.

In3luki:
I totally forgot about this PR. The reason for the 16 tests was to catch edge cases like this:
Screenshot 2024-11-08 211714

Ideally it would just be a check if the flanked token is actually rendered on the canvas from the perspective of the flanker but sadly that's not currently possible without switching the canvas perception layer (or maybe it was some other layer that I  forgot).

PR #16989, opened 2024-10-25
Apply item alterations to weapon alt usages

CarlosFdez:
Also did a minor performance optimization, since I wasn't sure about constantly filtering on the entire rule element list in a hot loop.

### DIFF
--- a/src/module/item/weapon/document.ts
+++ b/src/module/item/weapon/document.ts
@@ -476,12 +476,21 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
     getAltUsages({ recurse = true } = {}): WeaponPF2e<TParent>[] {
         const meleeUsage = this.toMeleeUsage();
 
-        return [
+        const altUsages: WeaponPF2e<TParent>[] = [
             this.toThrownUsage() ?? [],
             meleeUsage ?? [],
             // Some combination weapons have a melee usage that is throwable
             recurse ? (meleeUsage?.toThrownUsage() ?? []) : [],
         ].flat();
+
+        // Apply item alterations to all alt usages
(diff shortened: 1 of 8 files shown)


src/module/item/weapon/document.ts
```diff
@@ -476,12 +476,23 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
...
-        return [
+        const altUsages: WeaponPF2e<TParent>[] = [
             this.toThrownUsage() ?? [],
             meleeUsage ?? [],
             // Some combination weapons have a melee usage that is throwable
             recurse ? (meleeUsage?.toThrownUsage() ?? []) : [],
         ].flat();
+
+        // Apply item alterations to all alt usages
+        if (this.actor) {
```
stwlam, line 487: You can skip checking here for an actor in the same way you did above

src/module/rules/rule-element/item-alteration/rule-element.ts
```diff
@@ -1,15 +1,20 @@
...
 import type { StringField } from "types/foundry/common/data/fields.d.ts";
 import { AELikeRuleElement } from "../ae-like.ts";
-import { RuleElementPF2e } from "../base.ts";
-import { ModelPropsFromRESchema, RuleElementSchema } from "../data.ts";
+import { RuleElementOptions, RuleElementPF2e } from "../base.ts";
+import type { ModelPropsFromRESchema, RuleElementSchema, RuleElementSource } from "../data.ts";
 import { ItemAlteration, ItemAlterationSchema } from "./alteration.ts";
 
 class ItemAlterationRuleElement extends RuleElementPF2e<ItemAlterationRuleSchema> {
+    constructor(source: RuleElementSource, options: RuleElementOptions) {
```
stwlam, line 13: This isn't appropriate in a constructor: it bypasses priority

PR #16997, opened 2024-10-26
Exclude PF2e Companion Compendia items from `ABCPicker` results

stwlam:


CrackJackFlood:
Is their the possibility of getting some form of interface for this in future for exclusions (filtering) or support for modules to exclude themselves?

Issue #16982, opened 2024-10-23
Trait Toggling doesn't work on overriden strikes in the same item


MrVauxs:
When you have a choice between two `Strike`s, say, one using Charisma and one using Intelligence, and you later down the line want to use `AdjustStrike` to enhance it, `AdjustStrike` will only modify the first `Strike` element, EVEN if it is not turned on by the predicates not matching.

Here is an export of the feature this happens with:
[fvtt-Item-mystic-training-KfLxW2YtgvBic9k2.json](https://github.com/user-attachments/files/17494863/fvtt-Item-mystic-training-KfLxW2YtgvBic9k2.json)

And a video:
[chrome_7YXhqid8wQ.webm](https://github.com/user-attachments/assets/2890e237-f4c1-4bf9-bf56-4f211d479919)

CarlosFdez:
Can you make the simplest set of REs to repro this with, and can you test on the latest dev version to make sure the fix I just did has no effect? I know you made an actor, but it will be helpful to reduce how much im testing.

MrVauxs:
The export is the bare minimum. Just drag the class feature to any character.

CarlosFdez:
Can this not be reproduced without the choicesets though? Just two Strike RE's and an AdjustStrike?

MrVauxs:
It cannot be reproduced without ChoiceSets because the two Strikes override one another.

MrVauxs:
[fvtt-Item-mystic-training-KG4STe2rMICAAsiC.json](https://github.com/user-attachments/files/17528336/fvtt-Item-mystic-training-KG4STe2rMICAAsiC.json)



https://github.com/user-attachments/assets/7e87f084-cb9f-4470-9b86-55116cd3f625

CarlosFdez:
I would have preferred including the RE code as code blocks in comments, but I got it work.

I found the root cause but I don't know how to fix it in the current juncture. Basically trait-toggles.ts when running update(), finds a non-ignored strike rule element with a matching slug. The problem is that when strike B overrides strike A, strike A does not become ignored, so the lookup is always pulling up strike A.

CarlosFdez:
Simpler test case

```json
{
  "label": "Magical Blast",
  "img": "icons/magic/fire/beam-jet-stream-embers.webp",
  "damage": {
    "base": {
      "damageType": "fire",
      "dice": 1,
      "die": "d6"
    }
  },
  "key": "Strike",
  "slug": "test-override-override"
}
[...]

PR #16998, opened 2024-10-26
Fix trait toggles on strikes overriden within the same item

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/16982.

I was tempted on leveraging createBatchRuleElementUpdate() to tag every strike with the same slug across every item but I stopped myself. It might create a bit more stability to trait toggle behavior but I doubt it'll be necessary.

### DIFF
--- a/src/module/item/weapon/trait-toggles.ts
+++ b/src/module/item/weapon/trait-toggles.ts
@@ -1,6 +1,5 @@
 import type { ActorPF2e } from "@actor";
 import type { WeaponPF2e } from "@item";
-import type { StrikeRuleElement } from "@module/rules/rule-element/strike.ts";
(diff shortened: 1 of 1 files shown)


stwlam:
Can we hit the _right_ one rather that all of them?

CarlosFdez:
No clean way to do that. If you come up with something let me know, but the only other thing that comes to mind is storing a reference to the rule on the weapon itself.

I thought about it a bit more, and hitting all of them across every item is technically the best UX, because then the same traits and toggles (if it exists) will still be selected if the strike gets replaced due to the REs. Its also what we do for every other RE of this nature. I kinda chickened out and only did it to the same item.

stwlam:
I don't feel it should be an accommodated practice to plan a strike with variations by having multiple Strike REs.

Issue #17015, opened 2024-10-27
Change Shape RE gets stuck with identical suboptions from different sources


KaitoKuroba:
If the Myriad Forms Ancestry Feat is added to a Kitsune PC, the Change Shape Roll Option gets stuck at some point (typically after selecting Fox or Tailless).

This seems to be caused by identical suboptions from different sources. In this case the fox or taillless suboption, which is already added by the Heritage or Ancestry. Removing the duplicate suboption from the Myriad Forms RE fixes the problem for this PC.

I'll make a PR in a moment. It's just a little cleanup of the REs but it provides a workaround for Kitsune to this behavior.

PR #17016, opened 2024-10-27
provide a workaround for the Change Shape bug and a little RE cleanup

KaitoKuroba:
Partially addresses #17015

It's just a little unification for the Kitsune Change Shape REs. The Roll Option with both Suboptions is now part of the Ancestry while Heritages an Ancestry Feats enable these suboptions.

This also avoids the bug in #17015, as there are no more duplicate Suboptions.

### DIFF
--- a/packs/ancestries/kitsune.json
+++ b/packs/ancestries/kitsune.json
@@ -95,15 +95,20 @@
                 "label": "PF2E.NPCAbility.ChangeShape.Label",
                 "mergeable": true,
                 "option": "change-shape",
-                "predicate": [
-                    {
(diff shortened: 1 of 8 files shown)


TikaelSol:
I'd rather figure out the initial bug than rely on roll options like this.

KaitoKuroba:
I would prefer that too. But since I can't fix the bug in the code myself, I wanted to at least offer a workaround for this case. I couldn't find any open issues that also seem to be related to this bug, which would lower the priority of the actual fix. If such help is not desired, please feel free to close the PR.

PR #17017, opened 2024-10-27
Improve display of /act in chat for secret rolls

farling42:
When a `/act` with a secret roll is made, the display would include a space (or dot) before the action link, and would display "()" at the end of the string to players.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -466,7 +466,7 @@ class TextEditorPF2e extends TextEditor {
                 element.appendChild(document.createTextNode(" "));
 
                 const details = document.createElement("span");
-                if (dc && showDC && Number.isNumeric(dc)) {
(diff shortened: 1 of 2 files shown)


src/module/system/text-editor.ts
```diff
@@ -466,7 +466,7 @@ class TextEditorPF2e extends TextEditor {
                 element.appendChild(document.createTextNode(" "));
 
                 const details = document.createElement("span");
-                if (dc && showDC && Number.isNumeric(dc)) {
+                if (dc && showDC && Number.isNumeric(dc) && statistic) {
```
stwlam, line 469: This conditional branch is a little absurd at this point and should be refactored.

farling42: I was following the pattern of the following "else if" statement, which also contained unnecessary duplicate checks

PR #17024, opened 2024-10-27
fix "double vs" resistance requiring all entries to be true

websterguy:
Double Vs check was being treated as an "and" instead of an "or". Change the predicate test into "or".


src/module/actor/data/iwr.ts
```diff
@@ -369,7 +369,7 @@ class Resistance extends IWR<ResistanceType> implements ResistanceSource {
     /** Get the doubled value of this resistance if present and applicable to a given instance of damage */
     getDoubledValue(damageDescription: Set<string>): number {
         if (this.doubleVs.length === 0) return this.value;
-        const predicate = new Predicate(this.doubleVs.flatMap((d) => this.describe(d)));
+        const predicate = new Predicate({ or: this.doubleVs.flatMap((d) => this.describe(d)) });
```
stwlam, line 372: This would break any doubleVs with multiple statements

PR #17039, opened 2024-10-29
Implement remaster styled advanced alchemy and Double, Double

CarlosFdez:
The following occured refactor wise:
1) Crafting Ability now injects initial data during the active effects phase, so that AELike no longer needs to target the beforeDerived phase.
2) Crafting abilities with the same slug merge. The AELikes I added could have more CraftingAbility RE's, but I don't know if we want to incentivize that for abilities without craftable items.
3) Prepared items now support batch sizes. Each prepared item crafts the batch size in the craftable entry, and falls back to the batch size of the entire ability.
4) Items now have a timesTaken getter. The current implementation is inefficient if the getter gets used often. I'll refactor if we start using it a bunch.

This does not update every source of advanced alchemy. I am not touching alchemist class personally until all of this is done.

### DIFF
--- a/src/module/actor/character/crafting/ability.ts
+++ b/src/module/actor/character/crafting/ability.ts
@@ -8,6 +8,7 @@ import type {
 import { Predicate, RawPredicate } from "@system/predication.ts";
 import { ErrorPF2e } from "@util";
 import { UUIDUtils } from "@util/uuid.ts";
+import * as R from "remeda";
 import { CraftingFormula, PreparedFormula, PreparedFormulaData } from "./types.ts";
 
 class CraftingAbility implements CraftingAbilityData {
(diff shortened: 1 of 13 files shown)


src/module/rules/rule-element/crafting-ability.ts
```diff
@@ -61,25 +61,45 @@ class CraftingAbilityRuleElement extends RuleElementPF2e<CraftingAbilityRuleSche
...
-            batchSizes: this.batchSizes,
-            craftableItems: this.craftableItems,
-            maxItemLevel: Number(this.resolveValue(this.maxItemLevel)) || this.actor.level,
-            maxSlots: this.maxSlots,
-            preparedFormulaData: this.prepared,
-        };
+        const key = sluggify(slug, { camel: "dromedary" });
+        const maxSlots = this.maxSlots !== undefined ? Number(this.resolveValue(this.maxSlots)) : null;
+        const maxItemLevel = Number(this.resolveValue(this.maxItemLevel));
+        const existing = this.actor.system.crafting.entries[key];
```
stwlam, line 71: Nooo entries

CarlosFdez: We can't kill the entries till we migrate. AELikes unfortunately still target that.

static/templates/actors/character/tabs/crafting.hbs
```diff
@@ -199,6 +199,8 @@
                 <a class="adjust decrease" data-action="decrease-craft-quantity">&ndash;</a>
                 <input type="number" class="formula-number" data-craft-quantity value="{{formula.quantity}}" />
                 <a class="adjust increase" data-action="increase-craft-quantity">+</a>
+            {{else}}
+                x{{formula.quantity}}
```
stwlam, line 203: Let's localize this bit here

src/module/actor/character/crafting/ability.ts
```diff
@@ -128,31 +128,33 @@ class CraftingAbility implements CraftingAbilityData {
             return false;
         }
 
-        if (formula.item.level > this.maxItemLevel) {
+        // Check if it meets the entry requirements
+        const rollOptions = formula.item.getRollOptions("item");
+        const viable = this.craftableItems.find((c) => c.predicate.test(rollOptions));
```
stwlam, line 133: Confused by this: what is crafting viability?

CarlosFdez: Viable crafting definitions, but this is an artifact. viable is checked to find the best batch size, but we don't need that here. I tweaked it just now.

PR #17050, opened 2024-10-30
Add support for mythic feat slots

CarlosFdez:
This includes the variant introduced in a separate pdf. I put them in homebrew settings under campaign, but I'm almost tempted to put it under variant rules instead. Not sure.

This PR introduces the ability for a group to have a separate configurable limit in the feat header. So far the only other official use I can think of where it'd be handy is strength of thousands.

### DIFF
--- a/src/module/actor/character/feats/types.ts
+++ b/src/module/actor/character/feats/types.ts
@@ -6,6 +6,7 @@ import type { FeatGroup } from "./group.ts";
 
 /** Any document that is similar enough to a feat/feature to be used as a feat for the purposes of feat groups */
 interface FeatLike<TParent extends ActorPF2e | null = ActorPF2e | null> extends ItemPF2e<TParent> {
+    readonly level: number | null;
     category: string;
     group: FeatGroup<NonNullable<TParent>, this> | null;
     isFeat: boolean;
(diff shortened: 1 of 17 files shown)


src/module/actor/character/feats/types.ts
```diff
@@ -19,21 +20,57 @@ interface FeatLike<TParent extends ActorPF2e | null = ActorPF2e | null> extends
...
+    id?: string;
+    /** Short label for the feat slot, displayeed on the left. By default it is the level */
     label?: Maybe<string>;
+    /** The level of the feat that should go in this slot */
+    level?: Maybe<number>;
+    /**
+     * The limit value the group needs in order to display this feat.
+     * This can be different from level if the group is based on tiers.
+     */
+    rank?: Maybe<number>;
```
CarlosFdez, line 61: I'm shopping for a better variable name if anyone can think of anything.

stwlam: `notRank` would be slightly better than that

CarlosFdez: Renamed to `tier`. I was looking for a catch all term for the group's level, but its only ever going to tier or level anyways (even if its not necessarily the actor's level) so tier is fine.

PR #17060, opened 2024-10-31
Add support for tagging modifiers and damage dice

stwlam:
They are in turn included among modifier/dice roll options


CarlosFdez:
Neat!

PR #17061, opened 2024-10-31
Automate Grudge Strike

pedrogrullada:


packs/feats/grudge-strike.json
```diff
@@ -24,7 +24,46 @@
...
+                "damageType": "vitality",
+                "diceNumber": 2,
+                "dieSize": "d6",
+                "key": "DamageDice",
+                "predicate": [
+                    "grudge-strike",
+                    "target:negative-healing"
+                ],
+                "selector": "strike-damage"
+            }
```
TikaelSol, line 65: To keep the roll dialog cleaner I'd just make this void all the time then DamageAlteration on negative healing to override the damage type

pedrogrullada: Ah, good call, let me do that

PR #17072, opened 2024-10-31
Fix removing spell damage from corrupting heightening data

CarlosFdez:


src/module/item/spell/sheet.ts
```diff
@@ -174,7 +174,7 @@ export class SpellSheetPF2e extends ItemSheetPF2e<SpellPF2e> {
                 const key = anchor.dataset.id;
                 if (key) {
                     const values = { [`${baseKey}.damage.-=${key}`]: null };
-                    if (!overlayData) {
+                    if (!overlayData && this.item.system.heightening) {
```
stwlam, line 177:
```suggestion
                    if (!overlayData && this.item._source.system.heightening) {
```

PR #17073, opened 2024-10-31
Consider spirit bonuses to Strike damage in Energized Spark automation

pedrogrullada:


packs/feats/energized-spark.json
```diff
@@ -234,12 +234,34 @@
...
+                            {
+                                "and": [
+                                    "item:type:action",
+                                    "item:trait:transcendence"
+                                ]
+                            },
+                            {
+                                "and": [
+                                    "action:strike",
+                                    "damage:tag:exemplar-damage"
```
stwlam, line 255: _chews_ needs more `damage`

pedrogrullada: does look redundant now that you mention it huh

PR #17079, opened 2024-10-31
Internally represent special resources using dromedary

CarlosFdez:
I'm unsure about this and am only trying it out. It trades our current set of problems for a new set of problems.

One annoyance is that a feat that consumes a resource will use kebab case, and then will check using dromedary. So for example, if a feat consumes hero points, the slug will be `hero-points` and it will check `system.resources.heroPoints`.

### DIFF
--- a/src/module/rules/rule-element/special-resource.ts
+++ b/src/module/rules/rule-element/special-resource.ts
@@ -95,8 +95,9 @@ class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema> 
 
         if (this.itemUUID) {
             // For pre creation, we don't really know the true priority. Assume it is the better between this and existing
+            const key = sluggify(this.slug, { camel: "dromedary" });
             const thisMax = Number(this.resolveValue(this.max));
-            this.max = Math.floor(Math.max(thisMax, this.actor.system.resources[this.slug]?.max ?? 0));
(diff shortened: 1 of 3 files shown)


src/module/rules/rule-element/special-resource.ts
```diff
@@ -116,15 +117,16 @@ class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema>
...
         this.max = Number(this.resolveValue(this.max));
-        if (!(this.slug in this.actor.synthetics.resources)) {
-            this.actor.synthetics.resources[this.slug] = this;
-            this.actor.system.resources[this.slug] = fu.mergeObject(this.actor.system.resources[this.slug] ?? {}, {
+        if (!(key in this.actor.synthetics.resources)) {
+            this.actor.synthetics.resources[key] = this;
+            this.actor.system.resources[key] = fu.mergeObject(this.actor.system.resources[key] ?? {}, {
                 value: 0,
                 max: this.max,
             });
```
stwlam, line 127: Sorry, I forgot: what is the purpose of setting it in both synthetics and system data?

CarlosFdez: The system data is the true prepared data (and allows parity with true resources like hero points), and the synthetic allows us to perform updates.

PR #17110, opened 2024-11-01
Preserve token ring on extraction and fix Nahoa and Samo tokens

CarlosFdez:
We might need @SpartanCPA to make fallback tokens but they shouldn't look weird on the canvas anymore.

### DIFF
--- a/build/lib/extractor.ts
+++ b/build/lib/extractor.ts
@@ -467,12 +467,17 @@ class PackExtractor {
                                 img: ImageFilePath;
                                 prototypeToken: DeepPartial<foundry.data.PrototypeTokenSource>;
                             } = docSource;
+                            const ringSetting = docSource.prototypeToken.ring;
                             withToken.prototypeToken = { name: docSource.prototypeToken.name };
                             // Iconics have special tokens
                             if (withToken.img?.includes("iconics")) {
                                 withToken.prototypeToken.texture = {
                                     src: withToken.img.replace("Full", "") as ImageFilePath,
                                 };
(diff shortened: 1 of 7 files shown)


build/lib/extractor.ts
```diff
@@ -467,13 +467,18 @@ class PackExtractor {
                                 img: ImageFilePath;
                                 prototypeToken: DeepPartial<foundry.data.PrototypeTokenSource>;
                             } = docSource;
+                            const ringSetting = docSource.prototypeToken.ring;
```
stwlam, line 470:
```suggestion
```

build/lib/extractor.ts
```diff
@@ -467,13 +467,18 @@ class PackExtractor {
...
                             withToken.prototypeToken = { name: docSource.prototypeToken.name };
                             // Iconics have special tokens
                             if (withToken.img?.includes("iconics")) {
                                 withToken.prototypeToken.texture = {
                                     src: withToken.img.replace("Full", "") as ImageFilePath,
                                 };
                             }
+                            // If using dynamic ring, preserve enabled status and subject texture
+                            if (ringSetting?.enabled) {
```
stwlam, line 479:
```suggestion
                            const ringSetting = docSource.prototypeToken.ring;
                            if (ringSetting?.enabled) {
```

CarlosFdez: I've tried this, but line 471 clobbers the prototypeToken variable. withToken is actually not a destructuring, its a type narrow.

PR #17114, opened 2024-11-01
Fix Kashrishi climbing feats

CaptainWonders:
fixed kashrishi's scuttle-up and skilled climber feats to use selector, rather than type. Also made skilled climber change climb speed to movement speed when scuttle up is taken.

### DIFF
--- a/packs/feats/skilled-climber.json
+++ b/packs/feats/skilled-climber.json
@@ -31,8 +31,16 @@
         "rules": [
             {
                 "key": "BaseSpeed",
-                "type": "climb",
+                "selector": "climb",
                 "value": 10
+            },
+            {
+                "key": "BaseSpeed",
+                "predicate": [
+                    "feat:scuttle-up"
+                ],
+                "selector": "climb",
+                "value": "@actor.attributes.speed.total"
             }
         ],
         "traits": {
--- a/packs/feats/skilled-climber.json
+++ b/packs/feats/skilled-climber.json
@@ -44,7 +44,7 @@
             },
             {
                 "key": "RollOption",
-                "label": "Climbing Cavern Wall Or Tree",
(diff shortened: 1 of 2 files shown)


packs/feats/skilled-climber.json
```diff
@@ -31,8 +31,16 @@
...
+                "selector": "climb",
                 "value": 10
+            },
+            {
+                "key": "BaseSpeed",
+                "predicate": [
+                    "feat:scuttle-up"
+                ],
+                "selector": "climb",
+                "value": "@actor.attributes.speed.total"
```
pedrogrullada, line 43: This will need a toggle since it only applies when "climbing trees or cavern walls". I recommend you use a toggleable roll option predicated on `feat:scuttle-up` and then predicate this BaseSpeed on just that toggle.

CaptainWonders: Just committed a change for that

packs/feats/skilled-climber.json
```diff
@@ -31,8 +31,25 @@
...
+                "key": "BaseSpeed",
+                "predicate": [
+                    "climbing-cavern-wall-or-tree"
+                ],
+                "selector": "climb",
+                "value": "@actor.attributes.speed.total"
+            },
+            {
+                "key": "RollOption",
+                "label": "Climbing Cavern Wall Or Tree",
```
pedrogrullada, line 47: Please localize this in `re-en.json`, likely under something like `PF2E.SpecificRule.Kashrishi.SkilledClimber.ClimbToggle`

We try to keep plain text in rule elements as localization strings only.

TikaelSol, line 47:
```suggestion
                "label": "PF2E.SpecificRule.SkilledClimber.Label",
```

PR #17137, opened 2024-11-02
Fix non-alchemical items DC not showing in item identification dialog

reonZ:
Because the `getItemIdentificationDCs` always returns `{ crafting: dc }` for non-magical items, the else section in the `identify-item` template not only isn't necessary anymore but also doesn't display the DC of the item because it still expects `dcs.dc`.


reonZ:
Maybe you guys would still like to have the else section in the template though, if not, should the label use `PF2E.identification.IdentifyAlchemyDCs` or `PF2E.identification.IdentifyGenericDCs` ?

stwlam:
Conflicts!

reonZ:
I am not well versed with source control or git but i don't understand the conflict, it looks like it just want to put the part of the code i removed back.

static/templates/actors/identify-item.hbs
```diff
@@ -21,7 +21,7 @@
                 </tr>
             </tbody>
         </table>
-    {{else if isAlchemical}}
+    {{else}}
         <h1>{{localize "PF2E.identification.IdentifyAlchemyDCs"}}</h1>
```
stwlam, line 25: I don't think we'll want to show "Identify Alchemy DCs" for non-alchemical items.

reonZ: Right, i confused myself with the part after that, the `{{localize "PF2E.Skill.Crafting"}}`, because both were using the same label there.

PR #17159, opened 2024-11-03
Fix `DamageContext##findMatchingCheckContext`

In3luki:
This was caught by a Biome linter test run.

### DIFF
--- a/src/module/actor/roll-context/damage.ts
+++ b/src/module/actor/roll-context/damage.ts
@@ -40,7 +40,7 @@ class DamageContext<
             .find((message) => {
                 if (!message.rolls.some((r) => r instanceof CheckRoll)) return false;
                 if (message.actor?.uuid !== actor.uuid) return false;
-                if (target.token !== message.target?.token.object) return false;
+                if (target.token !== message.target?.token) return false;
 
                 const messageItem = message.item;
                 if (!messageItem?.isOfType("melee", "weapon")) return false;
@@ -49,7 +49,7 @@ class DamageContext<
 
                 return !!(
                     paramsItemSlug === messageItemSlug &&
-                    item.uuid === item.uuid &&
+                    item.uuid === messageItem.uuid &&
                     item.isMelee === messageItem.isMelee
                 );
             });
--- a/src/module/actor/roll-context/damage.ts
+++ b/src/module/actor/roll-context/damage.ts
@@ -47,7 +47,7 @@ class DamageContext<
                 const paramsItemSlug = item.slug ?? sluggify(item.name);
                 const messageItemSlug = messageItem.slug ?? sluggify(messageItem.name);
 
-                return !!(
+                return (
                     paramsItemSlug === messageItemSlug &&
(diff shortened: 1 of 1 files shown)


CarlosFdez:
So it seems the condition before was only same slug and kinda same usage (is melee). Is there a chance that it might to remove the uuid check. I'm wondering if there's a scenario where this fix might end up breaking something like RE strikes.

In3luki:
oh right, removing the check would probably be better as it didn't produce any noticeable bugs. so don't fix what's not broken.

stwlam:
Instead of removing the broken check, why not fix it?

src/module/actor/roll-context/damage.ts
```diff
@@ -47,11 +47,7 @@ class DamageContext<
                 const paramsItemSlug = item.slug ?? sluggify(item.name);
                 const messageItemSlug = messageItem.slug ?? sluggify(messageItem.name);
 
-                return !!(
-                    paramsItemSlug === messageItemSlug &&
-                    item.uuid === item.uuid &&
-                    item.isMelee === messageItem.isMelee
-                );
+                return !!(paramsItemSlug === messageItemSlug && item.isMelee === messageItem.isMelee);
```
stwlam, line 50:
```suggestion
                return !!(
                    paramsItemSlug === messageItemSlug &&
                    item.uuid === messageItem.uuid &&
                    item.isMelee === messageItem.isMelee
                );
```

stwlam (changes requested):

CarlosFdez:
More specifically I was hinting that maybe the uuid comparison should be tested with things like strikes from rule elements. Perhaps with strikes from battlezoo dragons that do some weird override thing.

src/module/actor/roll-context/damage.ts
```diff
@@ -16,8 +16,7 @@ class DamageContext<
     constructor(params: DamageContextConstructorParams<TSelf, TStatistic, TItem>) {
         super(params);
         // In case the user rolled damage from their sheet, try to fish out the check context from chat
-        params.checkContext ??= this.#findMatchingCheckContext();
-
+        params.checkContext ??= this.#findMatchingCheckContext(params.options);
         if (params.outcome) params.options.add(`check:outcome:${sluggify(params.outcome)}`);
```
In3luki, line 20: I'm wondering if this was supposed to be the value of `checkContext.outcome`.

In3luki:
Tested and works with:
- Melee Weapon
- Unarmed Attack
- Ranged Weapon
- Elemental Blast

src/module/actor/roll-context/damage.ts
stwlam: Likely not

src/module/actor/roll-context/damage.ts
```diff
@@ -49,7 +49,7 @@ class DamageContext<
 
                 return !!(
```
stwlam, line 50:
```suggestion
                return (
```

stwlam (approved):

PR #17170, opened 2024-11-04
Refactor Sneak Attack to be tag-based and add missing Avenger features

### LINKED ISSUE #17160: Avenger has missing features
Avenger should grant trained in medium armor, but doesn't yet. 
It also does not automate Sneak Attack with the deity weapon. It probably should grant it's own Sneak Attack feature. 
Ideally it should also remove the main Sneak Attack and the surprise attack, but apparently that's not an option in the system yet.

pedrogrullada:
Sneak Attack now gives its default weapons the `sneak-attack` tag and predicates its damage on that tag. Avenger and Ruffian tag their own weapons, thus qualifying them for Sneak Attack. Switched to tag-based predicate on Sly Striker as well.

Closes #17160

### DIFF
--- a/packs/classfeatures/avenger.json
+++ b/packs/classfeatures/avenger.json
@@ -115,7 +115,6 @@
             "keyOptions": [
                 "str"
             ],
-            "proficiencies": {},
             "suppressedFeatures": [
                 "Compendium.pf2e.classfeatures.Item.w6rMqmGzhUahdnA7"
             ]
--- a/packs/classfeatures/avenger.json
+++ b/packs/classfeatures/avenger.json
@@ -38,6 +38,10 @@
                 "value": 1
             },
             {
+                "key": "GrantItem",
+                "uuid": "Compendium.pf2e.actionspf2e.Item.Hunt Prey"
+            },
+            {
                 "choices": {
                     "filter": [
                         {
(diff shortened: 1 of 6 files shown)


packs/classfeatures/avenger.json
```diff
@@ -79,28 +79,36 @@
...
-                            "self:level",
-                            13
-                        ]
-                    }
+                    "target:condition:off-guard",
+                    "item:deity-favored"
                 ],
-                "value": "ternary(gte(@actor.system.proficiencies.defenses.light.rank,2),2,1)"
+                "selector": "strike-damage"
+            },
```
pedrogrullada, line 101: This is how the Ruffian racket does Sneak Attack, but we can probably do better now by tagging weapons that qualify for Sneak Attack, predicating the main feature off of that, and then features that add or remove to the list can simply add or remove the tag. Not sure if we want to do that *now* though.

ammalagonc: How hard of a "lift" would it be to make the change? 
I think this way (tagging the item) is cleaner and easier to maintain.

pedrogrullada: It's easy. I could probably do it in this PR, just not sure if there are stuff that would need migrating that I'm not considering.

ammalagonc: Fair

Issue #15961, opened 2024-08-09
CTRL+Drag to Measure and Drop Waypoints not functional when Drag Measurement is Active


aburley1234:
FoundryVTT v12.330
PF2E Version 6.2.1
Drag Measurement Setting - Always

Ctrl-Dragging while having the "Select Tokens" tool selected no longer starts a measurement.
Ctrl-Dragging from a token while having the "Select Tokens" tool selected does not put you into the state where you can drop waypoints

Khunkurisu:
I don't mean to be rude, but is feature request really the right label for this if the request is "don't disable a feature that has no reason to be disabled"? And if there's something about the implementation that requires it, that feels more like a bug or flaw in the implementation of the new ruler.

cuyima:
It's not actually just the Select Tokens tool. It's the Measure Distance one as well, making it impossible to use waypoints with any of the tools :/

Khunkurisu:
To be clear measurements while dragging tokens works fine. This is not related to that.

If you use the Foundry built in measurement alongside any popular tool module (Token Drag Ruler, Elevation Ruler, etc), you can access the measurement tool from the default select tool by holding the Control key and dragging the cursor.

This core functionality is disabled by the PF2e system’s ruler implementation, instead ctrl dragging has the same functionality as dragging with the select tool does with no modifiers. (Leading me to believe that the issue is likely just the implementation ignoring modifiers entirely and not passing the function to core.)

At any rate to ensure the folks working on the system have the cleanest experience possible PLEASE make separate issues for things that aren’t related to this one.

tyrielv:
It looks to me like the problem is [here](https://github.com/foundryvtt/pf2e/blob/515c1a062fc92b78887a2f0ff7f9026864ee0c8b/src/module/canvas/ruler.ts#L8C5-L10C6):
```
static override get canMeasure(): boolean {
        return this.#dragMeasurement ? game.activeTool === "ruler" : super.canMeasure;
    }
```
If the PF2E system's drag measurement is enabled (either set to "Always," or set to "In Combat" and it's in combat), then canMeasure will only return true if the active tool is the ruler. Super.canMeasure will also return true if Ctrl+Drag is used.

RisingStarPF2e:
This really needs to be fixed. This can't be intentional. This basically means you need to pick between using these timeless controls but have no deep module support because all of them utilize the base measure now OR use the modules but now you need to click on the ruler, double check your back on that token layer.

[...]

PR #17172, opened 2024-11-04
fix: Do not disable built-in ruler Ctrl+Drag to measure

tyrielv:
Fix for #15961. 

Examining the history of this method I suspect leaving this check was an oversight after checks for `token` were added elsewhere in the file.

### DIFF
--- a/src/module/canvas/ruler.ts
+++ b/src/module/canvas/ruler.ts
@@ -6,7 +6,7 @@ import type { TokenPF2e } from "./token/object.ts";
 
 class RulerPF2e<TToken extends TokenPF2e | null = TokenPF2e | null> extends Ruler<TToken, UserPF2e> {
     static override get canMeasure(): boolean {
-        return this.#dragMeasurement ? game.activeTool === "ruler" : super.canMeasure;
+        return super.canMeasure;
     }
 
     /** Whether drag measurement is enabled */


stwlam:
This would appear to introduce a new bug rather than fixing anything.

tyrielv:
> This would appear to introduce a new bug rather than fixing anything.


It definitely restores the built-in Ctrl+Drag behavior and doesn't seem to break the PF2E system token drag. What do you think it breaks?

stwlam:
I just tested it by removing the method (you don't need an override just to super up), and it does not enable the original ctrl-drag behavior.

tyrielv:
Ugh, the dreaded 'it works on my machine.' I'll try to document exactly what I'm doing/observing.

1. Foundry 12.331, Chrome 130.0.6723.92
2. Checkout commit 515c1a062fc92b78887a2f0ff7f9026864ee0c8b (from master)
3. Run `npm ci`, `npm run build`, `npm run watch` (with a Windows symbolic link set up instead of `npm run link`)
4. Create a new world
5. Create a new scene in the world
6. Use the `Token Controls,` `Select Tokens` tool:
image

7. With the PF2E system setting for drag measurement (Never/Always, being sure to click 'Save Changes' after updating it), hold left-CTRL and click+drag with left mouse button:
   Never: Works as expected
ctrl-drag-never-on-master

   Always: Does not work
ctrl-drag-always-on-master

8. Update `ruler.ts`, deleting `canMeasure()` as suggested, refresh browser:
   Never: Still works as expected
ctrl-drag-never-with-change

   Always: Now works as expected
ctrl-drag-always-with-change


I don't claim that removing the override is the right fix because I don't know what other situations `canMeasure()` is called in, but the change definitely fixes the observed problem.

stwlam:
Had a second person check this, and while it starts the initial drag with CTRL held down, it doesn't enable full standard CTRL measurement. A new PR is up to fully reintegrate the core feature with drag measurement.

PR #17177, opened 2024-11-04
Update Animist focus pool based on feats and features

Suldrun45:
Circle of Spirits and Dual Invocation (medium level 9) upgrade the focus pool to the number of attuned apparitions.
Third and Fourth apparitions increase the focus pool by 1.


pedrogrullada:
Keep in mind that these features you mention use the number of primary apparitions, not the number of attuned apparitions. It might be a bit premature to use AELikes as counters when we don't know what implementation will look like yet.

TikaelSol:
Let's hold off until we know what our animist implementation looks like.

PR #17201, opened 2024-11-05
Localize Quick Climb and Quick Swim notes and update the latter's value

### LINKED ISSUE #17199: Quick Climb and Quick Swim behave differently with -10ft Speed Penalty Armor
On 6.6.1
image
image
[fvtt-Actor-quickclimbswimdiscrepency-gtM07HEzktvooWCv(1).json](https://github.com/user-attachments/files/17636957/fvtt-Actor-quickclimbswimdiscrepency-gtM07HEzktvooWCv.1.json)

Hello.

Char has a Hell Knight Half Plate (un)equipped and both the Quick Climb and Quick Swim Feats. _Afaik_ these feats should behave identical. However, the full speed penalty of my armor is applied to the Swim Speed. (Unfortunatly, no hover for that one to show). The PC has enough strength to reduce the 10ft speed penalty down to 5tf.

Is this an erroneous automation if Quick Climber/Swim? Neither [Speed Penalty (aonprd)](https://2e.aonprd.com/Rules.aspx?ID=2174) nor [Aquatic Combat (aonprd)](https://2e.aonprd.com/Rules.aspx?ID=2438&Redirected=1) have texts indicating this behavior as intended.

Other boosts from Fleet, Fleetwind Centaur or Items like Boots of Bounding apply to all, as (I personally would) expect. (wrt this disussion https://github.com/foundryvtt/pf2e/issues/10887)

pedrogrullada:
Using brackets for the value made it so the armor penalties applied as if the character didn't meet the strength requirement. Perhaps the bracket resolving before that is taken into account? Either way, predicating the BaseSpeed rule element instead resolves the issue and is the better way to do it.
Closes #17199

### DIFF
--- a/packs/feats/quick-climb.json
+++ b/packs/feats/quick-climb.json
@@ -39,11 +39,15 @@
             },
             {
                 "key": "Note",
+                "outcome": [
+                    "success",
+                    "criticalSuccess"
+                ],
                 "predicate": [
                     "action:climb"
                 ],
                 "selector": "athletics",
-                "text": "You move 5 more feet on a success, and 10 more feet on a critical success.",
+                "text": "PF2E.SpecificRule.Feat.QuickClimb.Note",
                 "title": "{item|name}"
             }
         ],
--- a/packs/feats/quick-swim.json
+++ b/packs/feats/quick-swim.json
@@ -31,25 +31,23 @@
         "rules": [
             {
                 "key": "BaseSpeed",
+                "predicate": [
+                    "skill:athletics:rank:4"
+                ],
                 "selector": "swim",
-                "value": {
-                    "brackets": [
-                        {
-                            "end": 4,
-                            "start": 4,
-                            "value": "@actor.attributes.speed.total"
-                        }
-                    ],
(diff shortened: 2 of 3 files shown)


packs/feats/quick-climb.json
```diff
@@ -39,11 +39,15 @@
             },
             {
                 "key": "Note",
+                "outcome": [
+                    "success",
+                    "criticalSuccess"
+                ],
```
stwlam, line 45: This is a little complicated, but a creature with a climb speed can climb without rolling and automatically succeed, gaining all benefits that come with it. Usually that's nothing, but this very feat gives an extra five feet of movement.

pedrogrullada: I see what you're saying, but this notes only show up on rolls anyway. We could, in addition to this, append the description of these feats to the Climb and Swim action items via Item Alteration, which would remind players of the feats when they just post the action to chat rather than rolling anything.

stwlam: Eh, better to wait for refactor opportunities in an action overhaul

PR #17204, opened 2024-11-06
Update 2nd-level rollable tables

joeparis:
Update 2nd-Level Consumable Items and 2nd-Level Permanent Items tables with remaster changes, consolidating the treasure tables from the GM Core and Player Core 2 books, preserving the item order found in those tables. Weight is based on rarity:

* Common = 6
* Uncommon = 3
* Rare = 1

### DIFF
--- a/packs/equipment/cold-iron-buckler-high-grade.json
+++ b/packs/equipment/cold-iron-buckler-high-grade.json
@@ -1,0 +1,54 @@
+{
+    "_id": "lit8IyoUHuybenvU",
+    "img": "systems/pf2e/icons/equipment/shields/precious-material-shields/cold-Iron-buckler.webp",
+    "name": "Cold Iron Buckler (High-Grade)",
+    "system": {
+        "acBonus": 1,
+        "baseItem": "buckler",
+        "bulk": {
+            "value": 0.1
+        },
+        "containerId": null,
+        "description": {
+            "value": "<p>Cold iron shields don't typically have an additional effect, though when used for a shield bash, they are cold iron weapons.</p>\n<p>The shield has Hardness 8, HP 32, and BT 16.</p>\n<p><strong>Craft Requirements</strong> cold iron worth at least 2,500 gp</p>"
+        },
+        "hardness": 8,
+        "hp": {
+            "max": 32,
+            "value": 32
+        },
+        "level": {
+            "value": 15
+        },
+        "material": {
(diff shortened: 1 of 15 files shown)


packs/equipment/cold-iron-buckler-high-grade.json
```diff
@@ -0,0 +1,54 @@
+{
+    "_id": "lit8IyoUHuybenvU",
+    "img": "systems/pf2e/icons/equipment/shields/precious-material-shields/cold-Iron-buckler.webp",
+    "name": "Cold Iron Buckler (High-Grade)",
```
simonward, line 4: I'm not sure we want to add these items in, but I see that the links in the existing roll tables reference non-existent items, and therefore the links are broken. Any thoughts @TikaelSol?

joeparis:
The cold iron and silver shields/bucklers didn't come up when I searched for them in the compendium, I could only find them with the legacy module enabled which is why I made them. Should I have been able to find them with a search?

________________________________

packs/equipment/cold-iron-buckler-high-grade.json
TikaelSol: We don't want to add them. We removed them years ago

joeparis: Should I remove the entries from the table or make them text entries? And should I assume that any other entries that don't have corresponding items in the compendium should be handled the same way?

simonward: Maybe link to the base item?

joeparis: I believe the base item for all  is **Cold Iron Shield** GM Core p233, unless I am mistaken.

simonward: I was thinking of linking to Buckler, for example.

joeparis: Would Steel Shield be the base item for shields? (I swear I'm not trying to be difficult)

simonward: Yeah, that sounds reasonable.

joeparis: Dragging in buckler then editing the result details to match the books' tables breaks the item. They'll have to be text entries.

PR #17206, opened 2024-11-06
Add implementation for mythic points

CarlosFdez:
For NPCs it adds it to the sidebar under a resources subheader, alongside actual special resources.

Let me know where to put the `Resources` (npc sidebar) and `Special Resources` (character sidebar) localization.


src/module/actor/character/document.ts
```diff
@@ -369,15 +369,17 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
             value: resources.focus?.value || 0,
             max: 0,
             cap: 3,
         };
-
         resources.crafting = fu.mergeObject({ infusedReagents: { value: 0, max: 0 } }, resources.crafting ?? {});
         resources.crafting.infusedReagents.max = 0;
+        resources.mythicPoints = {
+            value: resources.mythicPoints?.value ?? 0,
+            max: this.itemTypes.feat.some((f) => f.system.traits.value.includes("calling")) ? 3 : 0,
```
stwlam, line 381: Can we check the (cached) setting here?

CarlosFdez: Sure! We may want to keep mythic points working on npcs in non-mythic games, but it makes perfect sense for pcs.

PR #17219, opened 2024-11-07
Update 1st-level rollable tables

joeparis:
Update 1st-Level Consumable Items and 1st-Level Permanent Items tables with remaster changes, consolidating the treasure tables from the GM Core and Player Core 2 books, preserving the item order found in those tables. Weight is based on rarity:

* Common = 6
* Uncommon = 3
* Rare = 1


joeparis:
I declared bankruptcy and started over.

PR #17223, opened 2024-11-07
Provide a subtitle with actions when using lores

xyzzy42:
Current it will show a subtitle like "Lie (PF2E.ActionsCheck.legal-lore)" when trying to use a skill that doesn't have a translation, which no lore will.

Detect the lack of translation and fall back to the "PF2E.ActionsCheck.x" translation format using the label of the skill.  This will result in "Lie (Legal Lore Check)" if the skill was labeled in that way.


src/module/system/action-macros/helpers.ts
```diff
@@ -54,11 +57,14 @@ class ActionMacroHelpers {
             default: {
                 const slug = sluggify(stat);
                 const property = `skills.${slug}`;
+                let subtitle = `PF2E.ActionsCheck.${stat}`;
```
stwlam, line 60: No `let`

src/module/system/action-macros/basic/escape.ts
```diff
@@ -86,7 +86,7 @@ function escapeCheckContext<ItemType extends ItemPF2e<ActorPF2e>>(
     );
 
     if (highest) {
-        const { checkType, stat: slug, subtitle } = ActionMacroHelpers.resolveStat(highest.statistic.slug);
+        const { checkType, stat: slug, subtitle } = ActionMacroHelpers.resolveStat(highest.statistic.slug, opts.actor);
```
stwlam, line 89: not sure how this is related?

xyzzy42: In order to get access to the skill label, I added the actor as an argument to `ActionMacroHelpers.resolveStat()`.  This is the one other call site of this function.

PR #17228, opened 2024-11-08
Update 5th-level rollable tables

joeparis:
Update 5th-Level Consumable Items and 5th-Level Permanent Items tables with remaster changes, consolidating the treasure tables from the GM Core and Player Core 2 books, preserving the item order found in those tables. Weight is based on rarity:

* Common = 6
* Uncommon = 3
* Rare = 1

For items that do not have a corresponding entry in the compendium, create a text entry in the table with the appropriate item name. If possible, set an appropriate icon from the system's icon set.

### DIFF
--- a/packs/rollable-tables/4th-level-consumables-items.json
+++ b/packs/rollable-tables/4th-level-consumables-items.json
@@ -22,273 +22,368 @@
             ],
             "text": "Climbing Bolt",
             "type": "pack",
-            "weight": 6
+            "weight": 6,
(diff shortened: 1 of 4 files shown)


TikaelSol:
This also includes changes to the 4th level tables, which appear to be different than the changes in the other PR

joeparis:
I had inadvertently included the non-updated 4th-level 4 tables. Upon review I found errors in the 5rh-level tables that I have now corrected.

PR #17273, opened 2024-11-10
Fix reroll using a hero point erroneously displaying for a mythic character

CarlosFdez:
The rerolll prompt checks if you have hero points before showing the option. This removes hero points from the pc. Even if the value is still in the data, the clamp will set it to zero during data preparation.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -365,9 +365,12 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         };
         system.proficiencies.spellcasting ??= { rank: 0 };
 
-        // Resources
+        // Resources (Mythic points replace hero points if the character is mythic)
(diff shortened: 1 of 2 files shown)


src/module/actor/character/document.ts
```diff
@@ -365,9 +365,12 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         };
         system.proficiencies.spellcasting ??= { rank: 0 };
 
-        // Resources
+        // Resources (Mythic points replace hero points if the character is mythic)
         const { resources } = this.system;
-        resources.heroPoints.max = 3;
+        const isMythic =
+            game.pf2e.settings.campaign.mythic !== "disabled" &&
+            this.itemTypes.feat.some((f) => f.system.traits.value.includes("calling"));
```
stwlam, line 372: Maybe merge mine first so you can look for the category

PR #17284, opened 2024-11-11
Convert loot system data to `TypeDataModel`

stwlam:


src/module/migration/migrations/885-convert-alignment-damage.ts
```diff
@@ -76,7 +79,7 @@ export class Migration885ConvertAlignmentDamage extends MigrationBase {
             }
         }
 
-        fu.mergeObject(source.system.attributes, iwr);
```
CarlosFdez, line 79: If the ?? occurs, there won't be any changes since its not being assigned to anything.

stwlam: image

stwlam: If the actor already (somehow) has no attributes, that's fine.

CarlosFdez: No, what I mean is that I think you have to do `source.system.attributes = fu.mergeObject(source.system.attributes ?? {}, iwr);`. If ?? {} occurs, its an entirely new object which is not assigned to anything. You might as well do an if statement on source.system.attributes since nothing is being mutated.

PR #17288, opened 2024-11-11
Improve aura rendering for tokens with multiple auras

In3luki:
As a bonus rerendering the same aura is now always cached even for single aura tokens.

Aura highlights are now always drawn from largest to smallest to ensure larger auras not painting over a smaller aura's squares. That could be reversed from small to large with skipping alredy highlighted squares but I'm not sure that the minor performance gain would warrant that.

I think I've got all reset cases covered expect when the scene changes the polygons of tokens in the old scene are not reset. They are reset when the scene changes back however, so it might not be necessary to do that.

### DIFF
--- a/src/module/canvas/token/aura/map.ts
+++ b/src/module/canvas/token/aura/map.ts
@@ -1,5 +1,7 @@
+import * as R from "remeda";
 import { TokenPF2e } from "../object.ts";
 import { AuraRenderer } from "./renderer.ts";
+import { createTestPolygons } from "./util.ts";
 
 export class AuraRenderers extends Map<string, AuraRenderer> {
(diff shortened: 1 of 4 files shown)


stwlam:
This is a lot of manual cache management for the relatively rare case of multiple auras; I'm not sure if a performance gain is worth the added maintenance. Is there no way to integrate it into canvas rendering workflows?

In3luki:
- Removed the polygon caching and reverted all changes tied to that
- The grouped polygons are now only created for tokens with more than one aura
- Moved the override polygons to the optional `AuraRenderer#polygons` property
- Removed the additional `polygons` argument from the `getAreaSquares` function

PR #17306, opened 2024-11-13
Batch create and delete items when daily crafting

CarlosFdez:
I find it odd that rest for the night *and* daily craft delete infused items, but I guess better safe than sorry.


stwlam:
Go on ...

Issue #17111, opened 2024-11-01
Foundry locks up with malformed `/act`


farling42:
The following text placed within a journal page will cause Foundry to lock up whilst attempting to enrich the HTML. (Note it only uses one "]" to terminate the `/act` command.)

```
[[/act bob statistic=society] @Check[what]{who}
```

Although the lockup occurs in a core foundry module, I haven't been able to replicate the lockup using the core foundry enrichers.

I came across this while converting lots of `@Check` to `/act`

PR #17309, opened 2024-11-13
Fix /act enricher regex to prevent catastrophic backtracking

Cuingamehtar:
Should resolve https://github.com/foundryvtt/pf2e/issues/17111
Also requires whitespace between slug and options.
Can add a check for a trailing whitespace after the slug, but I don't think it's necessary.

### DIFF
--- a/src/scripts/hooks/init.ts
+++ b/src/scripts/hooks/init.ts
@@ -143,7 +143,7 @@ export const Init = {
             });
 
             CONFIG.TextEditor.enrichers.push({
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I would need @nikolaj-a to test all uses of /act to make sure this isn't' breaky. I could do that for `@Damage` but I don't have any clue for /act

Cuingamehtar:
For what it's worth, it matches 197 cases (the same number as the old regex) of /act currently present (both in packs and in localization files

PR #17315, opened 2024-11-14
Fix drag measurement broadcast during combat

In3luki:


In3luki:
Works as expected after #17270 which I thought was already merged when I opened this.

PR #17316, opened 2024-11-14
Fix `RulerPF2e#_onMouseUp` not working for normal (non-drag) measurements

In3luki:


src/module/canvas/ruler.ts
```diff
@@ -268,8 +268,8 @@ class RulerPF2e<TToken extends TokenPF2e | null = TokenPF2e | null> extends Rule
             if (this.isDragMeasuring && game.activeTool === "ruler") {
                 return this._endMeasurement();
             }
```
stwlam, line 270:
```suggestion
```

PR #17336, opened 2024-11-16
Fix rollItemMacro to allow feat and action hotbar macros to function again

DocSchlock:
Resolves this issue present since the frequency tracking update.
image

Currently in release and master, only Self Effect and Frequency-defined hotbar macros of feats and actions will function.

To replicate the problem, drag any non-SelfEffect or Frequency-bearing feat or action to the macro hotbar and click on it, view the error in the console.

### DIFF
--- a/src/scripts/macros/hotbar.ts
+++ b/src/scripts/macros/hotbar.ts
@@ -22,7 +22,7 @@ export async function rollItemMacro(itemId: string, event: Event | null = null):
         return null;
     }
 
-    if (item.isOfType("action", "feat")) {
+    if (item.isOfType("action", "feat") && (item.system.selfEffect || item.system.frequency)) {
         return createUseActionMessage(item);
     }
 


CarlosFdez:
Perhaps the createUseActionMessage() function no longer needs the if statement at the very top? There is an item.toMessage() in there, which means that it should fallback cleanly enough, though perhaps it not receiving the event might lead to a problem?

DocSchlock:
> Perhaps the createUseActionMessage() function no longer needs the if statement at the very top? There is an item.toMessage() in there, which means that it should fallback cleanly enough, though perhaps it not receiving the event might lead to a problem? Bears testing.

I'll test it out today and see if that's a good route to go.

CarlosFdez:
Superseded by https://github.com/foundryvtt/pf2e/pull/17364. Thanks for bringing this to my attention.

PR #17343, opened 2024-11-16
Enforce crafting requirements when preparing a crafting formula

CarlosFdez:


src/module/actor/character/crafting/ability.ts
```diff
@@ -56,9 +56,20 @@ class CraftingAbility implements CraftingAbilityData {
...
 
+    /** Returns true if this ability cannot prepare more formulas */
+    get isFull(): boolean {
+        // Prepared with a resource can prepare infinitely, we validate when we try to craft
+        if (this.isPrepared && this.resource) {
+            return false;
+        }
+
+        return !this.maxSlots || this.preparedFormulaData.length >= this.maxSlots;
+    }
```
stwlam, line 71: No a value/max object of slots?

CarlosFdez: What would the property be called? Also we'd have to set max to infinity if its prepared with resources and that might be misleading.

PR #17347, opened 2024-11-17
Set label for item special resources and show in sidebar

CarlosFdez:


stwlam:
Screenshot?

CarlosFdez:
In retrospect, good call. Doing this means its singular, and we likely want plural. It might be best to enforce the label rather than determine by item uuid.

image

PR #17361, opened 2024-11-18
Add ability to simulate potency runes on Strike rule elements

kiptra:
This might make it easier to set up a Strike rule element on something like Spirit's Wrath, which gives a flat Item bonus to the granted Strike

https://2e.aonprd.com/Feats.aspx?ID=5858

image


stwlam:
We got the weapon potency RE for that.

kiptra:
Ohhh, like if you tie the rule elements together by slug?

PR #17364, opened 2024-11-18
Fix item macros for actions without use buttons and support modifier keys

CarlosFdez:
Supercedes https://github.com/foundryvtt/pf2e/pull/17336

### DIFF
--- a/src/scripts/macros/hotbar.ts
+++ b/src/scripts/macros/hotbar.ts
@@ -7,6 +7,7 @@ import { ChatMessagePF2e } from "@module/chat-message/document.ts";
 import { createUseActionMessage } from "@module/chat-message/helpers.ts";
 import { MacroPF2e } from "@module/macro.ts";
 import { objectHasKey } from "@util";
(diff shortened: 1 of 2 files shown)


src/scripts/macros/hotbar.ts
```diff
@@ -7,6 +7,7 @@ import { ChatMessagePF2e } from "@module/chat-message/document.ts";
 import { createUseActionMessage } from "@module/chat-message/helpers.ts";
 import { MacroPF2e } from "@module/macro.ts";
 import { objectHasKey } from "@util";
+import { eventToRollMode } from "@util/sheet.ts";
```
stwlam, line 10: Would these be importable from the above?

CarlosFdez: Interestingly enough, currently no. That said I think the plan is to move everything in @util/sheet.ts to @module/sheet/helpers.ts. Our sheet helpers are spread out and you preferred the second one.

PR #17387, opened 2024-11-20
Fix value display and restoration of item based special resources

CarlosFdez:


src/module/actor/character/crafting/crafting.ts
```diff
@@ -19,7 +19,7 @@ class CharacterCrafting {
 
         // Assemble all abilities. We check if label exists as a simple validation due to potential AELike tinkering
         const abilities = Object.values(actor.system.crafting.entries)
-            .filter((d): d is CraftingAbilityData => !!d?.label && !!d.slug)
+            .filter((d): d is CraftingAbilityData => !!d?.label && !!d.slug && !!d.craftableItems.length)
```
stwlam, line 22:
```suggestion
            .filter((d): d is CraftingAbilityData => !!d?.label && !!d.slug && d.craftableItems.length > 0)
```
semantically nicer :)

CarlosFdez: ok *fine*

PR #17400, opened 2024-11-20
Finish implementing remaster alchemist archetype including quick alchemy

CarlosFdez:


stwlam:
> Looking over the changes, this does not appear to fix Infused Reagents for Advanced Alchemy on the Crafting Tab. Setting the crafting to one per reagent instead of two (without feat progression) and changing the value from INT + LVL to 4 + INT. Is that planned for a different PR?

Infused reagents are no longer a mechanic in remastered advanced alchemy.

KakugaGoka:
> Infused reagents are no longer a mechanic in remastered advanced alchemy.
Sorry, I am speaking to the rules that replaced Infused Reagents by having you create 4 + INT number of items with the `infused` trait. I do not see these new rules accounted for, in this PR or in previous PRs. So I was wondering if I was misreading the system, as I am just now starting to look through this code base, or if it has not been accounted for as of yet. 
I can see that Versatile Vials is accounted for, though I also do not see where the 2 + INT value is being applied for those, which is why I am leaning toward that I am just not reading the implementation correctly. 
I am a C#/Swift engineer, so TypeScript/JavaScript is a little rusty for me.

packs/actions/quick-alchemy.json
```diff
@@ -18,7 +18,14 @@
             "remaster": true,
             "title": "Pathfinder Player Core 2"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "CraftingAbility",
+                "priority": 1,
```
stwlam, line 24: ?

packs/feats/alchemist-dedication.json
```diff
@@ -41,15 +41,40 @@
...
+                        {
+                            "end": 17,
+                            "start": 12,
+                            "value": 12
+                        },
+                        {
+                            "start": 18,
+                            "value": 18
+                        }
+                    ]
```
stwlam, line 67: no brackets

stwlam:
> Sorry, I am speaking to the rules that replaced Infused Reagents by having you create 4 + INT number of items with the `infused` trait.

The Alchemist archetype is initially limited to four versatile vials:
> You gain the Quick Alchemy benefits (page 174), creating up to 4 versatile vials during your daily preparations.

KakugaGoka:
> > Sorry, I am speaking to the rules that replaced Infused Reagents by having you create 4 + INT number of items with the `infused` trait.
> 
> The Alchemist archetype is initially limited to four versatile vials:
> 
> > You gain the Quick Alchemy benefits (page 174), creating up to 4 versatile vials during your daily preparations.

Yes, but my original question is about "Advanced Alchemy on the Crafting Tab". Seeing as in the current version, if you are an alchemist, your stats all read correctly, but the crafting page still behaves as if you are using the legacy alchemist. I understand the SRD. I am inquiring about the code and where values are implemented, if they are being implemented. 
I also do not see it being implemented in this PR: https://github.com/foundryvtt/pf2e/pull/17039/files

So my question of " Is that planned for a different PR?" still stands.

packs/actions/quick-alchemy.json
CarlosFdez: It doesn't rely on anything and its intent is to be expanded. This is the one on the ability itself.

packs/feats/alchemist-dedication.json
CarlosFdez: a 4 step ternary if is kinda scary to do with brackets and I probably have some mild dyslexia as once the brackets nest that deep I can no longer read it out. How would you recommend going about it? AELikes?

I had a thought for a cascading if else function, something like `ifElse(cond1, result, cond2, result, cond3, result, result)` but I wasn't sure how amenable people would be to it.

CarlosFdez:
Your question isn't very well stated if that's at all the intent, given you were asking about infused reagents and advanced alchemy, which this is neither.

This only touches the alchemist archetype as a proof of concept. The job of the alchemist class is better served to our data entry guys, as that's a more involved process and I trust them with it.

KakugaGoka:
> Your question isn't very well stated if that's at all the intent, given you were asking about infused reagents and advanced alchemy, which this is neither.
> 
> This only touches the alchemist archetype as a proof of concept. The job of the alchemist class is better served to our data entry guys (and won't require code changes), as that's a more involved process and I trust them with it.

Apologies for the question not being clear. Allow me to restate. 
The Crafting Tab in the Character Sheet uses different rules for the new Alchemist. As the amount crafted per formula is different and the source of this is different. The Class information appears to be filled in quite well, but the Crafting Tab itself does not respect Advanced Alchemy properly. Would this not be a system level change and not just data entry? If so, is there currently a system change planned to support this? 
I understand this is a volunteer project, and in no way am pressing or rushing. Was curious about the fact that this change is including the Alchemist Architype and Quick Alchemy, but is not updating the crafting page, as it seems like an extremely useful feature to focus on for the Alchemist class.

CarlosFdez:
Except for known formulas, the crafting tab's contents are generated by rule elements. The code here are changes to make those rule elements capable of new features that will be necessary for the new alchemist. You're asking about advanced alchemy which has nothing to do with this PR, but the `Advanced Alchemy` feat in the alchemist archetype has already been updated to the new rules (though not the class yet).

isAlchemical is the property in the existing alchemist class that enables the old style infused reagents behavior.

CarlosFdez:
Also, note that this doesn't close any issues.

packs/feats/alchemist-dedication.json
CarlosFdez: I ended up going with floorToValue(), its pushed up.

static/lang/re-en.json
```diff
@@ -5514,6 +5515,9 @@
             "VitalityManipulationStance": {
                 "Note": "When you use Stunning Fist, instead of saving against being stunned, you can choose to have the target make a @Check[fortitude|against:class] save against your class DC with the following results. You must choose to do this before the target rolls its save.</p><hr><p><strong>Critical Success</strong> The target is unaffected.</p><p><strong>Success</strong> The target's muscles become temporarily weaker. For 1 round, the target is @UUID[Compendium.pf2e.conditionitems.Item.MIRkyAjyBeXivMa7]{Enfeebled 2}, and any time it attempts to take an action with the manipulate trait, the action is disrupted unless it succeeds at a flat check with a DC equal to 5 + its enfeebled value.</p><p><strong>Failure</strong> As success, but the duration is 1 minute.</p><p><strong>Critical Failure</strong> As success, but the duration is 1 minute, the flat check is DC 11, and the target takes -10-foot status penalty to all Speeds for the duration.</p>",
                 "StrikeLabel": "Vitality Blast"
+            },
+            "VersatileVials": {
+
```
pedrogrullada, line 5520: Looks like a leftover from a previous iteration perhaps?

packs/actions/quick-alchemy.json
stwlam: I mean the priority. We should almost never use 1

CarlosFdez: Is the idea that it prevents someone from going first layter? It absolutely should always go first and I can't see a scenario where it does not, but if it makes you feel more confident, I can set it to 8, so it goes below all default priorities in the system and future proofs crafting ability being changed to default to 9 instead of 19 (which is something I'm still considering).

stwlam: It has a default of 100.

shemetz:
Thanks for all your hard work, I watched some demo gifs and this looks great!

One thing bugs me though -- from what I saw, the newly created items are Worn/Carried rather than Held in 1 Hand, which means that there's always an extra button click needed to draw them into your hand.  This is a little annoyance that seems easily solvable (if you haven't solved it already in a newer branch).

PR #17414, opened 2024-11-21
Fix damage rolls not dealing atleast 1 damage

### LINKED ISSUE #16484: Rolls ignoring minimum damage rules
[Player Core Pg. 406](https://2e.aonprd.com/Rules.aspx?ID=2301)

> If the combined penalties on an attack would reduce the damage to 0 or below, you still deal 1 damage. Sometimes there are other considerations, described below.

Enemies with negative damage modifiers are dealing 0 damage to targets without resistances. The intended minimum (before resistances) should be 1.

photo_2024-09-11_00-29-13

MiaOjala:
closes #16484


MiaOjala:
Any idea which direction this should be approached from then? check for it later or make it an option perhaps?

stwlam:
It should probably just ensure that it's at least zero--given that it's damage and not healing (I don't know of any healing abilities that could theoretically result in no healing, but maybe!). Then the IWR application helper can pre-bump the damage to 1 if the summed non-persistent-damage instances start at zero.

MiaOjala:
It seems that if IWR pre-bumps the damage to 1 it is only used when the actor has immunities/resistances to it. in the case of nothing and weaknesses the increase is ignored.

MiaOjala:
It seems someone has been making something larger that could also fix this problem, but the code seems unfinished so this would be a smaller fix

MiaOjala:
Hello! Could it be possible for someone to check this out?

CarlosFdez:
Sorry, the only person who can reasonably give you a good answer regarding IWR is stwlam, and I imagine he's very busy with V13 foundry dev at the current moment.

MiaOjala:
Alright thanks for the answer. I'll leave this here and make another fork for the more complicated way of approaching this.

PR #17444, opened 2024-11-23
Add the `UserTargets` class definition and use it for both `User#targets` & `UserPF2e#targets`

reonZ:
I know this doesn't have a direct impact on the current system needs but this bothered me.


types/foundry/client/pixi/core/interaction/targets.d.ts
```diff
@@ -0,0 +1,32 @@
...
+
+        /**
+         * Return the Token IDs which are user targets
+         */
+        get ids(): string[];
+
+        /**
+         * @override
+         */
+        add(token: TToken): this;
```
CarlosFdez, line 20: Can you do `override add` instead of the override jsdoc? `@override` is jsdoc nonesense cause javascript doesn't get those toys.

Same with the extends at the top.

PR #17447, opened 2024-11-23
Show skill replacements before medicine in treat wounds dialog

ceejii:
Reordering the skills dropdown in the treat wounds dialog, so that the skill options available from feats (Crafting for Chirurgeon or Nature for Natural Medicine) comes first in the list. It should be highly likely that the player wants to use the skill associated with the feat if they have the feat. For players that don't have a related feat, Medicine will be the only skill as before. For frequent healers this should be a nice quality of life change.


ceejii:
Is anything else needed from me for this PR to be actionable/mergable?

CarlosFdez:
Sorry for the delay, I got around to looking at PRs and it makes sense to me. Thank you for the PR.

PR #17483, opened 2024-11-25
fix memory leaks due to retained tagify, sortable or tooltipster instances

Codas:
This PR fixes memory leaks in character/npc/hazard/party/kingdom/item sheets and various other places.

Any application currently using `Sortable`, `tooltipster` or `Tagify` is currently leaking memory (some references + the entire detached application dom tree) because these utilities register global event handlers or other global references that survive the application elements being detached from the DOM.

The solution to this to use the respective `destroy` methods these libraries provide. Either `.destroy()` on the `Sortable` or `Tagify` instances or `$(element).tooltipster('destroy')` for tooltipster.
This is done by keeping references to the active instances and destroying and nulling them when the application is closed, new event handlers are registered or rule element forms are not used anymore.

The memory profiles in the coming examples were done the following way:
- Reload foundry
- perform activity once to trigger any lazy allocations that are done only once for each application
- trigger manual garbage collection in chrome memory profiler
- start memory profiling
- do the activity three more times, each time running manual garbage collection after the activity is performed
- stop memory profiling

Ideal results are: Three spikes in the allocation that reset to a flat line of exactly the same height as before the spike. 

These are the results:
<details>
<summary>NPC Sheet</summary> 

**before PR**
00-npc_sheet_before

**after PR**
00-npc_sheet_after

</details>

<details>
<summary>Character Sheet</summary> 

**before PR**
01-character_sheet_before

**after PR**

01-character_sheet_after

</details>

<details>
<summary>Hazard Sheet</summary> 
_Note: Hazard has to be in "unlocked" mode for tagify to be instantianted and triggere the memory leak_

**before PR**
02-hazard_sheet_before

**after PR**
02-hazard_sheet_after

</details>

<details>
<summary>Kingdom Sheet</summary> 
_Note: I was unable to test this one as I don't have access to the Kingmaker Premium Module._

**before PR**
N/A
**after PR**
N/A
</details>

<details>
<summary>Party Sheet</summary> 

**before PR**
03-party_sheet_before

**after PR**
03-party_sheet_after

</details>

<details>
<summary>IWR Editor</summary> 

**before PR**
04-iwr_editor_before

**after PR**
04-iwr_editor_after

</details>

<details>
<summary>Compendium Browser</summary> 
_Note: This requires the compendiom browser to be opened with a tab that has tagify instances. The Equipment tab for example._

**before PR**
05-compendium_browser_before

**after PR**
05-compendium_browser_after

</details>

<details>
<summary>Pick a Thing Prompt</summary> 
_Note: TODO describe which thing was picked_

**before PR**
06-pick_a_thing_before

**after PR**
06-pick_a_thing_after

</details>

<details>
<summary>Encounter Tracker</summary> 
_Note: This occured whenever the listeners were activated again. For testing, the encounter turn was advanced by one step._

**before PR**
07-encounter_tracker_before

**after PR**
07-encounter_tracker_after

</details>

_Note for Rule Elements: This was tested by creating the rule element for a standalone item, then deleting the rule element again without closing the item sheet. Even closing the item sheet however, memory was not freed._
<details>
<summary>Rule Element Forms: Actor Traits </summary> 
_Note: TODO - how to test?_

**before PR**
08-actor_traits_ref_before

**after PR**
08-actor_traits_ref_after

</details>

<details>
<summary>Rule Element Forms: Aura</summary> 

**before PR**
09-aura_ref_before

**after PR**
09-aura_ref_after

</details>

<details>
<summary>Rule Element Forms: Fast Healing </summary> 
_Note: Type has to be set to "regeneration" for memory leak to trigger_

**before PR**
10-fast_healing_ref_before

**after PR**
10-fast_healing_ref_after

</details>

<details>
<summary>Rule Element Forms: Roll Note </summary> 

**before PR**
11-roll_note_ref_before

**after PR**
11-roll_note_ref_after

</details>

<details>
<summary>Campaign Feature</summary> 

**before PR**
12-campaign_feat_sheet_before

**after PR**
12-campaign_feat_sheet_after

</details>

<details>
<summary>Deity Sheet</summary> 

**before PR**
13-deity_sheet_before

**after PR**
13-deity_sheet_after

</details>

<details>
<summary>Feat Sheet</summary> 

**before PR**
14-feat_sheet_before

**after PR**
14-feat_sheet_after

</details>

<details>
<summary>Spell Sheet</summary> 

**before PR**
15-spell_sheet_before

**after PR*
15-spell_sheet_after

</details>

<details>
<summary>Scene Config Sheet</summary> 

**before PR**
16-scene_config_before

**after PR**
16-scene_config_after

</details>

<details>
<summary>Homebrew Language Settings sheet</summary> 

**before PR**
17-homebrew_language_before

**after PR**
17-homebrew_language_after

</details>

<details>
<summary>Skill Check Prompt Macro</summary> 

**before PR**
18-request_check_macro_before

**after PR**
18-request_check_macro_after

</details>

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -79,6 +79,12 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
     /** Non-persisted tweaks to formula data */
     #formulaQuantities: Record<string, number> = {};
 
+    /** PC Sheet Tab manager instance. Must be destroyed to avoid memory leaks  */
+    #tabManager: PCSheetTabManager | null = null;
+
+    /** Elements with registered $.tooltipster instances. Have to be cleaned up to avoid memory leaks */
+    #tooltipsterElements: JQuery[] = [];
+
     static override get defaultOptions(): ActorSheetOptions {
         const options = super.defaultOptions;
         options.classes = [...options.classes, "character"];
@@ -685,16 +691,18 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             const side = hoverEl.dataset.tooltipSide
                 ?.split(",")
                 ?.filter((t): t is (typeof allSides)[number] => tupleHasValue(allSides, t)) ?? ["right", "bottom"];
-            $(hoverEl).tooltipster({
-                trigger: "click",
-                arrow: false,
-                contentAsHTML: true,
-                debug: BUILD_MODE === "development",
-                interactive: true,
-                side,
-                theme: "crb-hover",
-                minWidth: 120,
-            });
+            this.#tooltipsterElements.push(
+                $(hoverEl).tooltipster({
+                    trigger: "click",
+                    arrow: false,
+                    contentAsHTML: true,
+                    debug: BUILD_MODE === "development",
+                    interactive: true,
+                    side,
+                    theme: "crb-hover",
+                    minWidth: 120,
+                }),
+            );
         }
 
         // SPELLCASTING
@@ -762,6 +770,14 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         }
     }
 
+    protected override _resetListeners(): void {
+        super._resetListeners();
+        this.#tabManager?.destroy();
+        this.#tabManager = null;
+        this.#tooltipsterElements.forEach((element) => element.tooltipster("destroy"));
+        this.#tooltipsterElements = [];
+    }
+
     /** Activate listeners of main sheet navigation section */
     #activateNavListeners(html: HTMLElement): void {
         const sheetNavigation = htmlQuery(html, "nav.sheet-navigation");
@@ -774,7 +790,9 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
 
         navTitleArea.innerText = game.i18n.localize(activeTab.dataset.tooltip ?? "");
         const manageTabsAnchor = htmlQuery<HTMLAnchorElement>(sheetNavigation, ":scope > a[data-action=manage-tabs]");
-        if (manageTabsAnchor) PCSheetTabManager.initialize(this.actor, manageTabsAnchor);
+        if (manageTabsAnchor) {
+            this.#tabManager = PCSheetTabManager.initialize(this.actor, manageTabsAnchor);
+        }
 
         sheetNavigation.addEventListener("click", (event) => {
             const anchor = htmlClosest(event.target, "a[data-tab]");
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -685,7 +685,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             const side = hoverEl.dataset.tooltipSide
                 ?.split(",")
                 ?.filter((t): t is (typeof allSides)[number] => tupleHasValue(allSides, t)) ?? ["right", "bottom"];
-            this.tooltipsterElements.push(
+            this.ensureTooltipsterCleanup(
                 $(hoverEl).tooltipster({
                     trigger: "click",
                     arrow: false,
@@ -777,7 +777,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         navTitleArea.innerText = game.i18n.localize(activeTab.dataset.tooltip ?? "");
         const manageTabsAnchor = htmlQuery<HTMLAnchorElement>(sheetNavigation, ":scope > a[data-action=manage-tabs]");
         if (manageTabsAnchor) {
-            this.destroyables.push(PCSheetTabManager.initialize(this.actor, manageTabsAnchor));
+            this.ensureDestroyableCleanup(PCSheetTabManager.initialize(this.actor, manageTabsAnchor));
         }
 
         sheetNavigation.addEventListener("click", (event) => {
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -79,12 +79,6 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
     /** Non-persisted tweaks to formula data */
     #formulaQuantities: Record<string, number> = {};
 
-    /** PC Sheet Tab manager instance. Must be destroyed to avoid memory leaks  */
-    #tabManager: PCSheetTabManager | null = null;
-
-    /** Elements with registered $.tooltipster instances. Have to be cleaned up to avoid memory leaks */
-    #tooltipsterElements: JQuery[] = [];
-
     static override get defaultOptions(): ActorSheetOptions {
         const options = super.defaultOptions;
         options.classes = [...options.classes, "character"];
@@ -691,7 +685,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             const side = hoverEl.dataset.tooltipSide
                 ?.split(",")
                 ?.filter((t): t is (typeof allSides)[number] => tupleHasValue(allSides, t)) ?? ["right", "bottom"];
-            this.#tooltipsterElements.push(
+            this.tooltipsterElements.push(
                 $(hoverEl).tooltipster({
                     trigger: "click",
                     arrow: false,
@@ -770,14 +764,6 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         }
     }
 
-    protected override _resetListeners(): void {
-        super._resetListeners();
-        this.#tabManager?.destroy();
-        this.#tabManager = null;
-        this.#tooltipsterElements.forEach((element) => element.tooltipster("destroy"));
-        this.#tooltipsterElements = [];
-    }
-
     /** Activate listeners of main sheet navigation section */
     #activateNavListeners(html: HTMLElement): void {
         const sheetNavigation = htmlQuery(html, "nav.sheet-navigation");
@@ -791,7 +777,7 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         navTitleArea.innerText = game.i18n.localize(activeTab.dataset.tooltip ?? "");
         const manageTabsAnchor = htmlQuery<HTMLAnchorElement>(sheetNavigation, ":scope > a[data-action=manage-tabs]");
         if (manageTabsAnchor) {
-            this.#tabManager = PCSheetTabManager.initialize(this.actor, manageTabsAnchor);
+            this.destroyables.push(PCSheetTabManager.initialize(this.actor, manageTabsAnchor));
         }
 
         sheetNavigation.addEventListener("click", (event) => {
--- a/src/module/actor/hazard/sheet.ts
+++ b/src/module/actor/hazard/sheet.ts
@@ -5,12 +5,8 @@ import { HTMLTagifyTagsElement } from "@system/html-elements/tagify-tags.ts";
 import { htmlClosest, htmlQuery, tagify, traitSlugToObject } from "@util";
 import type { HazardPF2e } from "./document.ts";
 import { HazardActionSheetData, HazardSaveSheetData, HazardSheetData } from "./types.ts";
-import type Tagify from "@yaireo/tagify";
 
 export class HazardSheetPF2e extends ActorSheetPF2e<HazardPF2e> {
-    /** Active tagify instances. Have to be cleaned up to avoid memory leaks */
-    #tagifyInstances: Tagify<Record<"id" | "value", string>>[] = [];
-
     static override get defaultOptions(): ActorSheetOptions {
         const options = super.defaultOptions;
         return {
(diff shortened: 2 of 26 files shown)


CarlosFdez:
What should we be looking at in the profiler? I will be honest and admit that for larger performance concerns outside of time spent, these finer features of the debugger are lost on me.

Are you aware of the code/known issue/etc for all 3 of these libraries and can you link them? One of them being a leak wouldn't be a surprise, but all 3 is a bit wild. What's the likelihood that these global listeners aren't simply reused for all 3 once registered? Or that its memory that hasn't been gc'd yet?

Codas:
I am sorry, I should have been more precise in my original description.

What you should be looking at in the profiler (I have attached some screenshots of the before/after in the spoiler sections above) is mainly the green line, which represents the number of DOM nodes curently held in memory. Though The orange line (listeners currently in memory) in most cases perfectly correlates to that.
The Blue line in the above screenshots are plain "JS Heap", meaning pure javascript objects like strings, objects etc held in memory. That one is rapidly changing just from running the pixi event loop and is not a useful measurement here.

> and it seems like the heap is actually larger now overall

If by that you mean plain JS heap, that one is very volatile and just by running the event loop, some "garbage" will collect, be mainly very shortly lived and then be collected. I have attached screenshot of the memory profile over a 23.5s window. You should see some blue spikes that reset to (mostly) the same low as before.

<details>
<summary>Screenshot of the plain JS heap over time</summary> 
Bildschirmfoto 2024-11-26 um 06 24 17
</details>

Reason for the periodical increase is the pixi/foundry render loop, which allocates some objects, maybe strings, buffers, etc during the rendering of a frame. The drop part of the spike is when the garbage collector kicks in and all the objects that are not referenced anymore are collected. Looking at this in more details there might even be an issue with memory management in this core render loop somehow ("low" parts are ever increasing, even after a full garbage collection cycle), but that probably has nothing to do with the pf2e system.

### A more detailed analysis of the profile and testing

Lets look a the NPC sheet again. I'll create a new profile and concentrate on the important parts and provide some more details and actual values that are not available just looking at the screenshot. This is how I tested it (fresh reload not visible in the video)

**before**:
https://github.com/user-attachments/assets/c251c1af-496d-42d2-99de-575901ce07cb
Currently, after opening, then closing the NPC sheet, the number of DOM Nodes, and event listeners kept in memory is ever increasing. From 12,971 at the beginning to 18,767 at the end after a full garbage collecction.

**after**:
https://github.com/user-attachments/assets/9f88f075-d010-469e-96c6-1852e8288046
These tests were done with the codebase from this PR. In this case, the node count initially increases from 10,443 to 11,907 after closing the npc sheet and GC, but then after each opening, closing, GC cycle, the node count goes from 11,907 up to 14,148 an down again to 11,907 exactly, whithout a single node (or event listener) remaining in memory.

The important part here really is not how much memory is currently in use, how many nodes exactly are being allocated on a fresh reload etc. The important part here is that after opening, then closing a sheet, progressing in the encounter tracker etc, the node and listener count drops to exactly the value as before the operation. Everything else means some memory is being retained forever, ever increasing when of of these operations is being performed.  (memory leak)

### Regarding the libraries

> Are you aware of the code/known issue/etc for all 3 of these libraries and can you link them? 

There is no issue at all in the Sortable Library. The tagify Libray has some form of broken auto-cleanup which might warrant an issue or PR on that repository. But skimming over the documentation I would still go as far as to say they behave as intended. Tooltipster is slightly more complicated, but we'll get to that. Lets look at the libraries:

**Sortable**
`Sortable.js` for example keeps a reference of all active sortable instances in a local variable on module level, so instances that are created will never go out of scope (meaning there will always be a reference to that instance starting with the root GC node). [The line in question](https://github.com/SortableJS/Sortable/blob/62f0498ac3585982aa44f4492bbcfcb3e8776a80/src/Sortable.js#L438).
The [`destroy` function](https://github.com/SortableJS/Sortable/blob/master/src/Sortable.js#L1683-L1709) cleans up these references, making the sortable instance eligible for gargabe collection again.
I can find no reference either in the documentation or code that would suggest that some form of auto cleanup is intended when the DOM node sortable is registered for has become detached.


> What's the likelihood that these global listeners aren't simply reused for all 3 once registered?
They are not reused, it's not a question of likelyh

**Tagify**
At least [this line](https://github.com/yairEO/tagify/blob/master/src/parts/events.js#L95) creates a reference by creating a timer (setInterval) which continues running until the interval is manually destroyed. The [callback for this interval](https://github.com/yairEO/tagify/blob/master/src/parts/events.js#L709-L717) closes of the "this" variable, keeping it alive indefinitely. The check you see in the second code link (destroying the tagify instance if `parentNode` is not in the DOM anymore) is never triggered, since the direct parent of the tagify element is alwasys available. When replacing the HTML of the item sheet for example, only the root content node is actually not connected to the dom / has no parent anymore. All child elements are still connected in this detached DOM tree.
To make this library more resilient, the check could be changed to check if a `this.DOM.originalInput.closest('body')` evaluates to a truthy value... But it is also never documented anywhere that I can find that Tagify has this auto cleanup behavior, so I would argue manually calling `destroy` when the tagify instance is not needed anymore is the intended behavior of this library.

**Tooltipster**
I will not repeat the same analysis of what keeps references where for the tooltipster jQuery plugin, mainly because I'm unfamiliar with the jQuery plugin architecture... Fact is however that the situation with cleanup for tooltipster is a bit more complicated. It too provides a sort of `destroy` method by calling `$(element).tooltipster('destroy')`, the only issue being that this has to be called _before_ the element tooltipster has been used on is removed from the dom. Reason for that is that in normal cirumstances, removal or replacement of the HTML is done by jQuery via `containerElement.html(newHtml)` or `containerElement.replace(newContainerElement)`. Both function triger some sort of jQuery cleanup, removing data attributes set by jQuery if I understand correctly. This also removes the `tooltipster-ns` data attribute that tooltipster itself uses to verify that the instance is still active, throwing an error if that is not the case. Some global references to this tooltipster instance remain however if the element is removed by jQuery, leading to this unfortunate situation of having to destroy tooltipster before it is actually removed from the DOM.



**Finally**:
> ~~Or that its memory that hasn't been gc'd yet?~~ seems you accounted for this with manual gc in the profiler.

You already noticed, but just to clarify that part or my method a bit more. Yes, that is what the manual GC step (brush or trashcan icon in the performance profiler dev tools tab next to the memory checkbox) is used for. This triggeres a full Garbage collection major cycle (or apparently even two full cycles?). This means that both short lived and long lived objects are iterated over, removing ALL memory that is no longer referenced from the root GC node.



I hope I could answer at least some of your questions or concerns. Please do ask further question If something is still unclear :)

src/module/actor/party/kingdom/sheet.ts
```diff
@@ -374,16 +383,20 @@ class KingdomSheetPF2e extends ActorSheetPF2e<PartyPF2e> {
...
+                    trigger: "click",
+                    arrow: false,
+                    contentAsHTML: true,
+                    debug: BUILD_MODE === "development",
+                    interactive: true,
+                    side: ["right", "bottom"],
+                    theme: "crb-hover",
+                    minWidth: 120,
+                }),
+            );
```
stwlam, line 399:
```suggestion
        const $tooltipContent = $html.find("[data-tooltip-content]");
        if ($tooltipContent.length > 0) {
            this.#tooltipsterElements.push(
                $tooltipContent.tooltipster({
                    trigger: "click",
                    arrow: false,
                    contentAsHTML: true,
                    debug: BUILD_MODE === "development",
                    interactive: true,
                    side: ["right", "bottom"],
                    theme: "crb-hover",
                    minWidth: 120,
                }),
            );
        }
```

src/module/actor/sheet/base.ts
```diff
@@ -281,7 +284,7 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
         })();
         this.#activateInventoryDragDrop(inventoryPanel);
 
-        // Everything below here is only needed if the sheet is editable
+        // // Everything below here is only needed if the sheet is editable
```
stwlam, line 287:
```suggestion
        // Everything below here is only needed if the sheet is editable
```

src/module/actor/hazard/sheet.ts
```diff
@@ -5,8 +5,12 @@ import { HTMLTagifyTagsElement } from "@system/html-elements/tagify-tags.ts";
 import { htmlClosest, htmlQuery, tagify, traitSlugToObject } from "@util";
 import type { HazardPF2e } from "./document.ts";
 import { HazardActionSheetData, HazardSaveSheetData, HazardSheetData } from "./types.ts";
+import type Tagify from "@yaireo/tagify";
 
 export class HazardSheetPF2e extends ActorSheetPF2e<HazardPF2e> {
+    /** Active tagify instances. Have to be cleaned up to avoid memory leaks */
+    #tagifyInstances: Tagify<Record<"id" | "value", string>>[] = [];
```
stwlam, line 12: Any reason to not handle this from the base class?

src/module/actor/character/sheet.ts
```diff
@@ -79,6 +79,12 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
     /** Non-persisted tweaks to formula data */
     #formulaQuantities: Record<string, number> = {};
 
+    /** PC Sheet Tab manager instance. Must be destroyed to avoid memory leaks  */
+    #tabManager: PCSheetTabManager | null = null;
+
+    /** Elements with registered $.tooltipster instances. Have to be cleaned up to avoid memory leaks */
+    #tooltipsterElements: JQuery[] = [];
```
stwlam, line 86: Any reason to not handle this from the base class?

src/module/actor/party/kingdom/sheet.ts
```diff
@@ -67,6 +67,12 @@ class KingdomSheetPF2e extends ActorSheetPF2e<PartyPF2e> {
 
     #editingSettlements: Record<string, boolean> = {};
 
+    /** Elements with registered $.tooltipster instances. Have to be cleaned up to avoid memory leaks */
+    #tooltipsterElements: JQuery[] = [];
+
+    /** Active sortable instances. Have to be cleaned up to avoid memory leaks */
+    #sortables: Sortable[] = [];
```
stwlam, line 74: Any reason to not handle this from the base class?

src/module/actor/character/sheet.ts
Codas: No reason at all, should be better now

src/module/actor/party/kingdom/sheet.ts
```diff
@@ -718,6 +726,11 @@ class KingdomSheetPF2e extends ActorSheetPF2e<PartyPF2e> {
 
         return this.kingdom.update(data);
     }
+
+    override close(options?: { force?: boolean } | undefined): Promise<void> {
+        this._resetListeners();
+        return super.close(options);
+    }
```
stwlam, line 733:
```suggestion
```

src/module/actor/sheet/base.ts
```diff
@@ -462,6 +468,13 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
         }
     }
 
+    protected _resetListeners(): void {
```
stwlam, line 471: Since we're using typescript, underscore-leading method and property names generally aren't needed.

src/module/item/campaign-feature/sheet.ts
```diff
@@ -37,12 +40,20 @@ class CampaignFeatureSheetPF2e extends ItemSheetPF2e<CampaignFeaturePF2e> {
...
+            );
         }
     }
 
+    protected override _resetListeners(): void {
+        super._resetListeners();
+        this.#tagifyInstances.forEach((tagified) => tagified.destroy());
+        this.#tagifyInstances = [];
+    }
+
```
stwlam, line 56:
```suggestion
```

src/module/item/deity/sheet.ts
```diff
@@ -121,6 +132,12 @@ export class DeitySheetPF2e extends ItemSheetPF2e<DeityPF2e> {
         }
     }
 
+    protected override _resetListeners(): void {
+        super._resetListeners();
+        this.#tagifyInstances.forEach((tagifyInstance) => tagifyInstance?.destroy());
+        this.#tagifyInstances = [];
+    }
+
```
stwlam, line 140:
```suggestion
```

src/module/item/feat/sheet.ts
```diff
@@ -233,6 +238,12 @@ class FeatSheetPF2e extends ItemSheetPF2e<FeatPF2e> {
         this.#activateSuppressedFeaturesListeners(html);
     }
 
+    protected override _resetListeners(): void {
+        super._resetListeners();
+        this.#tagifyInstances.forEach((tagifyInstance) => tagifyInstance?.destroy());
+        this.#tagifyInstances = [];
+    }
+
```
stwlam, line 246:
```suggestion
```

src/module/item/spell/sheet.ts
```diff
@@ -330,6 +335,12 @@ export class SpellSheetPF2e extends ItemSheetPF2e<SpellPF2e> {
         });
     }
 
+    protected override _resetListeners(): void {
+        super._resetListeners();
+        this.#tagifyInstances.forEach((tagifyInstance) => tagifyInstance?.destroy());
+        this.#tagifyInstances = [];
+    }
+
```
stwlam, line 343:
```suggestion
```

stwlam (approved):
Looks great, thanks! We're going to wait for a data patch release but will merge soon after.

PR #17519, opened 2024-11-28
Attempt to fix a few type instantiation is excessively deep warnings

CarlosFdez:


src/module/actor/character/document.ts
```diff
@@ -1664,7 +1664,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         return flavor;
     }
 
-    consumeAmmo(weapon: WeaponPF2e<this>, params: RollParameters): boolean {
+    consumeAmmo(weapon: WeaponPF2e, params: RollParameters): boolean {
```
stwlam, line 1667: should probably still have a parent, especially since it's a public method

CarlosFdez: For one reason or another, it doesn't really care if there's an actor or not in the implementation, and we've been leaning to removing the generic if its annoying. But its not exploding if I set it to CharacterPF2e so maybe fine.

PR #17524, opened 2024-11-28
Update automation for Nonlethal Spell feat and add base automation to Wand of Mercy

ammalagonc:
Changed the name from level to rank


TikaelSol:
Not sure about doing the specific wands like this.

ammalagonc:
Fair - I'll yeet this into the trash can.

PR #17538, opened 2024-11-29
Update weapon-selecting spell effects to be compatible with NPCs

pedrogrullada:
There are some effects that are seemingly incompatible with NPCs outright, such as Runic Weapon, and Shillelagh, since they give weapons fundamental runes, so those may be best handled on a case by case basis.

I also omitted Draw the Lightning, since that one will require a bit more thought.

### DIFF
--- a/packs/spell-effects/spell-effect-blink-charge.json
+++ b/packs/spell-effects/spell-effect-blink-charge.json
@@ -23,16 +23,17 @@
         "rules": [
             {
                 "choices": {
-                    "ownedItems": true,
+                    "attacks": true,
                     "predicate": [
-                        "item:equipped"
-                    ],
-                    "types": [
-                        "weapon"
+                        "item:equipped",
+                        {
+                            "not": "item:trait:unarmed"
+                        }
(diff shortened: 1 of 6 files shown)


CarlosFdez:
I think I'd like @stwlam to see this one as its removing some ownedItems and they were his invention

pedrogrullada:
> I think I'd like @stwlam to see this one as its removing some ownedItems and they were his invention
Well, it's just an issue of NPC attacks having to interact with these spell effects too. `ownedItems` is fine otherwise.

CarlosFdez:
It implies that ownedItems is nonfunctional for effects that apply to weapons even if its explicitly codified to target weapons, because npcs use pretend weapons instead. That sounds like a feature gap more than it sounds like bad data entry, especially given that we're moving from AdjustStrike to ItemAlteration,

But while it feels that way to me, I can't really make the final call on that.

pedrogrullada:
Closing this, since this is just a band-aid to a more fundamental issue (which I've opened at #17726)

Issue #17546, opened 2024-11-29
[UX] Formulas in Crafting should have an "Add to Advanced Alchemy" button


tpendragon:
## Reasoning

When you're an alchemist, the known formulas section is very big. It's difficult to click a formula from known formulas, drag it, and scroll to the correct section. It would be much easier to handle if there were a button for each formula if you have the advanced alchemy feature that would add it to the list. I suspect it would be next to the "Craft" button, maybe an up arrow or a little potion bottle?

Screenshot of the UI I'm talking about:

<img width="498" alt="Screenshot 2024-11-29 at 3 49 42 PM" src="https://github.com/user-attachments/assets/f0e1e6f7-f250-4f3a-816e-7af609b1047e">

CarlosFdez:
There was a similar button, but it added to the first valid entry rather than a selected entry and thus was useless for this purpose.

This will likely become a popout in the future during the inevitable redesign.

tpendragon:
So if I worked this feature to add that button back, it wouldn't be welcome?

CarlosFdez:
It would not. That button was removed because it was awkward to update during all of the internal refactors, and it does not work with multiple crafting abilities very well at all.

If you want to try anything to make this easier, you can try the new sorting paradigm in the quick alchemy branch, where we sort in reverse level order, so that higher level formulas are at the top and closer to the crafting entries.

tpendragon:
You're saying if I PRed a reverse level sort in formulas you'd be open to that? Do you want a button to switch asc/desc?

Also if I can help with the other crafting stuff I'd be happy to do that.

tpendragon:
Another option is allowing items to be dropped into the advanced alchemy section from the compendium, is that also unreasonable?

CarlosFdez:
No toggle, descending all the time. `R.sortBy(something, (f) => -f.item.level)` will do it. You might have to do a conversion from a record to a list of level/formulas objects. Remember that all of this needs redesigns anyways.

No on dropping from compendium. Any sort of drop from compendium has the risk of hitting the wrong dropzone and adding the item to the inventory or something. We don't want to risk that.

tpendragon:
If I get some time tonight I'll take a look. Feel free to close this as wontfix.

tpendragon:
Okay reverse formula sort doesn't really get me there either, it's totally impossible to prepare crafts on a touchpad. How would you feel about a favorite button as an intermediary? Something to push a few up to the top.

[...]

PR #17558, opened 2024-11-30
Add searchable formulas.

tpendragon:
This is a stop-gap proposal until we have a more holistic crafting rewrite, and will improve the UX of preparing daily crafting from laptops especially.

I have a game in a little while and this'll make my alchemy life so much easier.

Refs #17546

Screenshot 2024-11-30 at 2 03 53 PM

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -69,12 +69,20 @@ import { ElementalBlast, ElementalBlastConfig } from "./elemental-blast.ts";
 import type { FeatBrowserFilterProps, FeatGroup } from "./feats/index.ts";
 import { PCSheetTabManager } from "./tab-manager.ts";
 import { CHARACTER_SHEET_TABS } from "./values.ts";
+import MiniSearch from "minisearch";
 
 class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e<TActor> {
     protected readonly actorConfigClass = CharacterConfig;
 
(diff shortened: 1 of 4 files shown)


tpendragon:
If I add one for inventory too can we get this through? I realize y'all are under a huge weight of expectation, but I think this will free up some of the pressure you're under to do the crafting rewrite specifically.

CarlosFdez:
The concern I have is that the popout formulas will likely exist next major release or the one after. Its too short of a time to include a stopgap (which would only be introduced during a major version release anyways). If it doesn't happen for one reason or another though then sure.

An inventory search filter wouldn't go amiss, but we would probably need more than a typeahead filter.

tpendragon:
So I'm hearing that there's no amount of work I can do to improve my alchemist experience other than fork and have my gm run that? Like, can I help you somehow?

I'm surprised this would be a major release, seems like a new feature rather than backwards incompatibility, but I believe you.

CarlosFdez:
Wrong terminology on my part. I meant minor releases. It wouldn't come out in a patch release. We're gearing up for a patch release currently.

I'll consider it if we're lining up towards a minor release and we're still not far enough ahead. I can't promise anything at this time, I just don't want you to do work that'll go to waste :(

tpendragon:
Alright. I'm @pendragont in discord, reach out and I'll donate whatever time you need for us to get alchemist crafting in a usable place. This only took like an hour, I'm happy and eager to help. Thanks!

stwlam:
Outside PRs are not being accepted for crafting features at this time, sorry.

PR #17588, opened 2024-12-03
Add alchemical tool and alchemical food consumable tags

pedrogrullada:
This would be useful to build the investigator's Quick Tincture and Wandering Chef's Quick Alchemy Crafting Ability rule elements.


stwlam:
Merged, but I may follow up and just unregulate consumable other-tags.

PR #17597, opened 2024-12-03
Ensure GM-only browser tabs are only viewable by GMs

In3luki:
There were still two ways to view GM-only tabs without GM persmission. Those should be adressed now.
Also includes minor cleanup and adresses some Svelte warnings in other files.

### DIFF
--- a/src/module/actor/character/apps/formula-picker/app.svelte
+++ b/src/module/actor/character/apps/formula-picker/app.svelte
@@ -66,7 +66,11 @@
                         <span>{formula.item.name}</span>
                         <ItemTraits traits={formula.item.traits} rarity={formula.item.rarity} />
                     </div>
-                    <button onclick={() => confirmSelection(formula.item.uuid)} data-tooltip="Confirm">
+                    <button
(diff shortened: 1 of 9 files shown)


src/module/actor/character/apps/formula-picker/app.svelte
```diff
@@ -66,7 +66,11 @@
                         <span>{formula.item.name}</span>
                         <ItemTraits traits={formula.item.traits} rarity={formula.item.rarity} />
                     </div>
-                    <button onclick={() => confirmSelection(formula.item.uuid)} data-tooltip="Confirm">
+                    <button
+                        onclick={() => confirmSelection(formula.item.uuid)}
+                        aria-label={game.i18n.localize("Confirm")}
```
CarlosFdez, line 71: I've been thinking we would have to cave, but would aria-describedby="tooltip" or aria-labelledby="tooltip" work instead? Foundry injects the former.

In3luki: labelledby works, describedby still produces the error.

PR #17598, opened 2024-12-03
Add Left-Hand Blood Effect

reyzor1991:


reyzor1991:
Yes, but I am not sure that it's correct because players could forget to to apply damage to himself

PR #17604, opened 2024-12-04
Remove/retire Aphorite and Ganzi versatile heritages

ammalagonc:
Now folded into the Nephilim feat


TikaelSol:
Update the remaster journal too

TikaelSol:
Did we ever redirect aasimar and tiefling? I think those got a migration, but we may want to redirect these ones.

ammalagonc:
Bah,
UUID redirect bombed out because it's in a folder - posted something on Discord to see if there's something that can be done.

ammalagonc:
OK, rerouting to folder didn't do much - I removed the redirects until I get some direction on how to do this.

PR #17622, opened 2024-12-05
Add support for aura effect resolvables and parent options

CarlosFdez:


CarlosFdez:
We could consider storing parent options on the effect data instead of the parent uuid, which would prevent creating a new parent options list per tested actor. Let me know if that sounds better.

src/module/actor/types.ts
```diff
@@ -67,6 +67,7 @@ interface AuraData {
 
 interface AuraEffectData {
     uuid: string;
+    parentUuid: string;
```
stwlam, line 70: why not just include the rule?

CarlosFdez: Doesn't look like it explodes, so sure.

PR #17633, opened 2024-12-06
Add applicable modifiers to battle form unarmed modifier comparison

### LINKED ISSUE #17563: Battle form modifiers are compared to unarmed attack modifier incorrectly when character has penalties
I have a player with a +24 unarmed modifier at base, casting rank 6 Dragon Form (+22 to hit). Normally, it works fine (they use their +24 to hit). But when they're Frightened 2 (or have any -2 or greater penalty to attacking normally), the battle form's unarmed attack modifier takes over and then applies the Frightened (or other penalty).

This means they jump from a +23 to hit (at Frightened 1, it's the normal modifier - 1) to a +20 to hit (at Frightened 2, using the battle form modifier - 2, because their normal (altered) modifier would be +22, which isn't higher than the battle form's unaltered modifier). AFAIK, the battle form override should only apply if their battle form modifier would actually be equal to or higher than their normal modifier (i.e., at Frightened 2, the battle form modifier would be +20, and their normal modifier would be +22, so they'd use their normal modifier)

Examples:
not frightened, no battle form
image

not frightened, battle form
image

frightened, no battle form
image

frightened, battle form
image

Expected behavior is that the battle form modifier used for comparison will take penalties (and presumably bonuses) into account when being compared to the character's current unarmed attack modifier.

adyoungblood:
Fixes #17563 

When comparing a battle form's static bonus to a character's unarmed attack bonus to choose which was higher, Foundry did not consider modifiers. This change grabs the applicable modifiers (status/circumstance bonuses, and penalties) and applies them in the comparison before setting the generated Strike's modifiers.

This will result in a slight (but acceptable) change in functionality. If a character's unarmed attack modifier is higher than the battle form's modifier, their modifiers will always be used, no matter what penalties are applied. This should work in reverse too (with bonuses and a lower unarmed attack modifier). **The total modifier won't change**, but the calculation will, since previously, the calculation would switch to the battle form's modifier once the character got enough penalties. This should be a low-risk change.

### DIFF
--- a/src/module/rules/rule-element/battle-form/rule-element.ts
+++ b/src/module/rules/rule-element/battle-form/rule-element.ts
@@ -2,7 +2,7 @@ import type { ActorType, CharacterPF2e } from "@actor";
 import { CharacterStrike } from "@actor/character/data.ts";
 import { SENSE_TYPES } from "@actor/creature/values.ts";
 import { ActorInitiative } from "@actor/initiative.ts";
-import { DamageDicePF2e, ModifierPF2e, StatisticModifier } from "@actor/modifiers.ts";
+import { DamageDicePF2e, ModifierPF2e, StatisticModifier, applyStackingRules } from "@actor/modifiers.ts";
 import { MOVEMENT_TYPES } from "@actor/values.ts";
 import { WeaponPF2e } from "@item";
 import { RollNotePF2e } from "@module/notes.ts";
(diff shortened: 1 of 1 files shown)


stwlam:
`applyStackingRules` mutates the modifier, so we shouldn't use that. Are stacking rules not already applied?

adyoungblood:
Oh, so they are! That makes things a lot easier

Issue #16534, opened 2024-09-14
Wounded Value incorrectly capped


MindlessReader:
There seems to be a cap to the Wounded condition value which should not exist. A character without diehard can only reach wounded 3 in the system right now, when there are abilities (such as Barbarian's Unstoppable Juggernaut or the Regenerate spell) which allow you to go past wounded 3. This is important because of abilities like Lifeblood's Call from Orc and Wild-Haired Fury from Yaksha which key off of your wounded value, as well as the fact that Regenerate has a conditional based on whether you have reached wounded 4 or not.

MiaOjala:
I guess it would be easy to just uncap the condition. The problem seems to be that the dying and wounded conditions are currently represented on the character sheet as circles that get coloured in as the condition is increased. So the wounded condition would still need some kind of cap for the new amount of circles or for the whole circle thing to be revamped.

PR #17638, opened 2024-12-07
Uncap the value of the wounded condition

MiaOjala:
addresses #16534 
uncaps the wounded value and replaces wounded pips with a number since there is no limit to the amount of pips there could theoretically be


ammalagonc:
How does it display on the character sheet?
Can you provide a screenshot?

MiaOjala:
image
image

ammalagonc:
Do you increase the value by clicking on it or by entering the number?

MiaOjala:
The previous functionality is still there so the condition can be raised and lowered by right and left clicking on the number or the condition tab

stwlam:
This issue was being sat on since it making a real difference is rare, and the main work would be coming up with better UI for it. I don't think I'd call this better.

PR #17648, opened 2024-12-08
Allow more characters to fit in transfer dialog quantity field

CarlosFdez:
6ch fits 5 characters. Padding seems to be about 1 character wide.


stwlam:
Isn't stack enabled by default?

PR #17658, opened 2024-12-09
Alter versatile vial traits with bomber class feature

manuel-hegner:
The acid, fire, electricity and cold trait are based on the damage type of the item. When the Bomber alchemist class feature changes the damage type of versatile vials it should also alter the damage trait of the item.

### DIFF
--- a/packs/classfeatures/bomber.json
+++ b/packs/classfeatures/bomber.json
@@ -77,6 +77,28 @@
                     "versatile-vial-damage"
                 ],
                 "value": "{item|flags.pf2e.rulesSelections.bomberFieldVials}"
+            },
+            {
+              "key": "ItemAlteration",
+              "mode": "add",
+              "itemType": "weapon",
+              "predicate": [
+                "bomber-field-vials",
+                "item:slug:versatile-vial"
+              ],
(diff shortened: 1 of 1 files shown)


pedrogrullada:
Do you have a rules reference for the vials trait changing? As far as I can tell all it says is that their damage type changes, nothing about their traits, even if I agree it makes sense.

manuel-hegner:
Sadly the rulebook is not very concrete about it. The damage type traits all have a definition in glossary like 
> Effects with this trait deal cold damage
This should at least make it clear that a bomber versatile vial should not have the acid trait anymore and kind of implies that everything dealing elemental damage should have the matching trait. But it is never stated outright, afaik.

pedrogrullada:
I purposefully avoided adding the trait back when I automated this since it's not something the ability says it does. I'm not strongly opposed to it, but I do think it's a bit of editorialising. Definitely deferring to the data leads on this one.

manuel-hegner:
I see how it is a little bit editorializing, but right now we have some strange interactions. E.g. the waterproof spell protects from versatile electricity vials, as they keep the acid trait.

PR #17680, opened 2024-12-11
Add tab support to svelte mixin

CarlosFdez:
Allows the use of tabGroups as a responsive object for the rendering.

### DIFF
--- a/types/foundry/client-esm/applications/_types.d.ts
+++ b/types/foundry/client-esm/applications/_types.d.ts
@@ -14,7 +14,7 @@ export interface ApplicationConfiguration {
     /** Click actions supported by the Application and their event handler functions */
     actions: Record<string, ApplicationClickAction>;
     /** Configuration used if the application top-level element is a form */
-    form?: ApplicationFormConfiguration;
+    form?: DeepPartial<ApplicationFormConfiguration>;
     /** Default positioning data for the application */
     position: Partial<ApplicationPosition>;
 }
--- a/types/foundry/client/apps/app.d.ts
+++ b/types/foundry/client/apps/app.d.ts
@@ -393,8 +393,8 @@ declare global {
     }
 
     interface ApplicationPosition {
-        width?: Maybe<number>;
-        height?: Maybe<string | number>;
+        width?: number | "auto";
+        height?: number | "auto";
         left?: Maybe<number>;
         top?: Maybe<number>;
         scale?: Maybe<number>;
--- a/src/module/sheet/mixin.svelte.ts
+++ b/src/module/sheet/mixin.svelte.ts
@@ -28,8 +28,26 @@ function SvelteApplicationMixin<
         /** State data tracked by the root component */
         protected $state: object = $state({});
 
+        /** Reactive version tabGroups that tabGroups itself will get replaced with on render */
+        #$tabGroups: Record<string, string> = $state({});
+
         /** The mounted root component, saved to be unmounted on application close */
         #mount: object = {};
+
+        override changeTab(tab: string, group: string, { force = false, updatePosition = true } = {}) {
+            if (!tab || !group) throw new Error("You must pass both the tab and tab group identifier");
+            if (this.tabGroups[group] === tab && !force) return; // No change necessary
+
+            // Update the tab group. This will trigger a svelte re-render if its looking for it
+            this.tabGroups[group] = tab;
+
+            // Update automatic width or height
+            if (!updatePosition) return;
+            const positionUpdate: { width?: "auto"; height?: "auto" } = {};
+            if (this.options.position.width === "auto") positionUpdate.width = "auto";
+            if (this.options.position.height === "auto") positionUpdate.height = "auto";
+            if (!foundry.utils.isEmpty(positionUpdate)) this.setPosition(positionUpdate);
+        }
 
         protected override async _renderHTML(
             context: SvelteApplicationRenderContext,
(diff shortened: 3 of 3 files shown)


CarlosFdez:
After continuing to work on porting the party sheet to svelte, I'm observing that while changeTab() in ApplicationV2 may alter the css of tabs without a reload, it does not actually initialize it to the default tab. The way we handle it in UserConfigPF2e is actually by generating the tabs programmatically. In light of this, I'm thinking of removing the dom manipulation code I did and mandating that we handle it all in svelte if that has not changed in v13. Agreed?

types/foundry/client/apps/app.d.ts
```diff
@@ -393,7 +393,7 @@ declare global {
     }
 
     interface ApplicationPosition {
-        width?: Maybe<number>;
+        width?: Maybe<string | number>;
         height?: Maybe<string | number>;
```
stwlam, line 397:
```suggestion
        width?: number | "auto";
        height?: number | "auto";
```

types/foundry/client-esm/applications/_types.d.ts
```diff
@@ -14,7 +14,7 @@ export interface ApplicationConfiguration {
     /** Click actions supported by the Application and their event handler functions */
     actions: Record<string, ApplicationClickAction>;
     /** Configuration used if the application top-level element is a form */
-    form?: ApplicationFormConfiguration;
+    form?: Partial<ApplicationFormConfiguration>;
```
stwlam, line 17:
```suggestion
    form?: DeepPartial<ApplicationFormConfiguration>;
```

stwlam (changes requested):

stwlam (changes requested):
Holding for use case demonstrating its necessity (or at least advantage)

CarlosFdez:
I can agree with waiting for the use case.

CarlosFdez:
I found one use case. If you don't make tab reactive on some degree, it breaks hmr.

image

Two tabs are active simultaneously, as the re-render on hmr is operating with stale data. This occurs even if you pass tabGroups into state, as hmr does not re-run application context.

Any sort of svelte re-render triggered in any way outside of the foundry lifecycle will cause this if we don't bind tabs to some reactive state.

types/foundry/client/apps/app.d.ts
```diff
@@ -393,8 +393,8 @@ declare global {
     }
 
     interface ApplicationPosition {
```
7H3LaughingMan, line 395: This was written before ApplicationV2 and was meant to represent the position of Application, there was no type/interface definition and whoever wrote the foundry types for PF2e ended up declaring their own. ApplicationV2 came along and did define a ApplicationPosition but it's not compatible with what Application uses, the main thing being that Application does not support using "auto" for it's width.

It would probably be better to remove the ApplicationPosition and properly declare it in the correct place so you aren't using ApplicationV2's position data for Application by mistake.

https://github.com/7H3LaughingMan/foundry-pf2e/commit/a9759b6695c71d71bce26906836966957df0e8e4#diff-8a61b31dd83b09ab3b6d7b2a0adaa14ded534c8e1fb6b54a54f783553d0e1280

CarlosFdez: I would have to check to be clear, and it might be part of a v13 research, but I did confirm that changing string to "auto" breaks any use of "auto" in the actual sheets. Typescript can't tell that its "auto" fsr:

image

7H3LaughingMan: I am not sure where PartySheetV2 is located but that looks like an issue with it's DEFAULT_OPTIONS.

The Foundry types are outdated mostly because the only things that are needed are updated. They also declare types/interfaces that aren't declared to try and "simplify" things, and over the months/years they have worked their way through a bunch of stuff and trying to correct it is turning into a mess.

When the Foundry types originally declared ApplicationPosition there was no actual ApplicationPosition as ApplicationV2 wasn't a thing, with ApplicationV2 it ended up declaring an ApplicationPosition for itself and now the Foundry types are trying to use it for both Application and ApplicationV2 but they aren't compatible with each other since they expect different things.

I am working on re-writing the Foundry types to get them up to date for module development since some modules use other parts of Foundry which aren't used by PF2e, this has caused some problems here and there due to how ingrained some of these types have become. But with the v13 work being done I can't push these fixes since I can fix the problem with v12 as they pop-up they will break things with v13 so I am going to wait and see about submitting one big PR once there is a v13 system release.

CarlosFdez: That might be wise. I don't wanna fix that distinction in this PR, but thanks for letting me know about it.

This was a PR to try to make the tab interface of appv2 compatible with svelte, and is more of a preview at this time. I'm coming up with alternative ideas as I continue, though I'm a bit blocked because I noticed that with our current implementation we may have to resort to prop drilling.

I experimented with this PR and putting foundryApp in the svelte context, and that actually seems to work! However now I have to prop drill `isEditable`, which is a DocumentSheet property, and having the mixin be aware of whether its a document sheet or not is above the intended scope.

So I'm experimenting with alternatives, and while I have a simpler way to handle tabGroups in my head now, it doesn't really solve prop drilling.

CarlosFdez:
Instead of this, I'll push up a draft pr for the in-dev party sheet soon. There are more problems to solve than just tabbing from my recent dive, and making every prop in reactive in ApplicationV2 and subclasses is a game of whack-a-mole.

PR #17685, opened 2024-12-11
Add tooltips to the item sheet traits

### LINKED ISSUE #17491: Feature Request: Hover over trait tags on item sheets to show description tooltip
Hovering over trait tags in chat cards or on expanded items on actor sheets displays a tooltip explaining what the trait does. For the sake of consistency, it'd probably make sense for this tooltip to also display when hovering trait tags on item sheets. (Screenshot attached in case this is insufficient explanation.) 

Use case: My group is a bunch of new players, and their lack of familiarity with the traits means they have to go look them all up when they're looking at new feats in the browser on level up. Having the trait descriptions accessible right from the item sheet would streamline the exploration process a lot.

Item sheet on the left is opened from an actor sheet. One on the right is from the feat browser.
image

Kulkodar:
Adjusted `createTagifyTraits` to include the tag description and adjusted the tag template to remove the browser tool tip from tags. Side effect tags in creature sheets also show their tootip.
Closes #17491

https://github.com/user-attachments/assets/a1026aaf-11e9-4aee-b223-22241a3910b1

TODO:
- [x] cleanup code
- [x] test with tags not defined in `CONFIG.PF2E.traitsDescriptions`
- [x] hide browser tool tip


src/util/tags.ts
```diff
@@ -66,6 +66,22 @@ function tagify(
...
+                    spellcheck='false' 
+                    tabIndex="${this.settings.a11y.focusableTags ? 0 : -1}"
+                    class="${this.settings.classNames.tag}"
+                    ${this.getAttributes(tagData)}> 
+                        <x title='' class="${this.settings.classNames.tagX}" role='button' aria-label='remove tag'></x>
+                        <div>
+                            <span class="${this.settings.classNames.tagText}">${tagData[this.settings.tagTextProp] || tagData.value}</span>
+                        </div>
+                    </tag>`;
+            },
```
stwlam, line 83: No inline HTML

Kulkodar:
i moved the html snipped into a getter function, i'm not sure if you meant that or that i should move it into a file.

CarlosFdez:
I assume shark means to use the dom functions to create elements. There's utility methods in src/util/dom.ts to reduce how verbose they are and make it play better with typescript as well.

stwlam:
A template would probably be best in this case.

src/util/tags.ts
Kulkodar: i might need help since i'm not that familiar with type script.
As far as i can see i cant use `renderTemplate` since an async function can't be assigned to `tag: string`.
I could use `createElements`, but then i have to either extend the function to allow custom tags and attributes or change the custom tags to html compliant tags and adjust the styling.

ricardoSlv:
@Kulkodar Still wanna finish this? I believe I help you if you want 🙋‍♂️

src/util/tags.ts
Kulkodar: @stwlam fixed now with the help of @ricardoSlv.

src/util/tags.ts
```diff
@@ -66,6 +67,43 @@ function tagify(
...
+                            ],
+                        }),
+                    ],
+                });
+
+                tag.contentEditable = "false";
+                tag.spellcheck = false;
+                tag.tabIndex = this.settings.a11y.focusableTags ? 0 : -1;
+
+                Object.entries(tagData).forEach(([key, value]) => {
```
CarlosFdez, line 100: prefer actual for loops

CarlosFdez:
Out of curiousity, before I check it out locally and start seeing if everything still works ok, why the CSS changes?

Kulkodar:
tagify has a css rule that apply on .tagify__tag>div that broke/changed the layout of the x button and tag itself since they are now divs instead of custom types.

CarlosFdez:
Is there an issue continuing to use that custom tag?

CarlosFdez:
I'd like to tag in @stwlam but if we don't extend it, you can use document.createElement().

I'd rather not chance inconsistencies or possible errors from changing the tag name. I think trying to keep the result as close as possible to what we already have is best.

CarlosFdez:
Alright, looks fine to me. There's some noise in the data (__isValid etc), but the original wasn't clean to begin with. Thanks for the PR.

Kulkodar:
Thanks for taking your time

PR #17687, opened 2024-12-11
Fix memory leaks and failed cleanups in destroyables Mutation Observer

Codas:
I really do like the Idea of using Mutation Observer for the otherwise errorprone task of cleanup up the `tooltipster`, `Sortable` and `Tagify` instances. I wish I had thought of that when authoring my PR.

However, I have identified some issues with the current implementation that this PR aims to address:

1. **Tooltipster cleanup timing**
Tooltipster instances normally to be destroyed **before** their associated DOM elements are removed. Tooltipster relies on jQuery’s .data attribute to store references to its instances, and these attributes are cleared when elements are removed via jQuery. If the destroy method is called after this happens, Tooltipster logs warnings or even throws errors.
Additionally, the instance itself is not automatically destroyed, and it retains a reference to the original HTML element. This alone isn’t enough to cause a memory leak, but Tooltipster's internal `setInterval` that closes over the tooltipster instance creates one.

2. **Handling removed nodes in MutationObserver**
The `removedNodes` property in the MutationObserver callback does not include the node it was registered for (the target). While the current implementation addresses this by observing a parent element (the closest .window-content element), the mutation handler mistakenly tries to use removed nodes as keys to access destroyable items. This approach will never succeed since the observed target is not included in `removedNodes`.

3. **Memory Leaks from unmanaged MutationObservers**
MutationObservers themselves can lead to memory leaks if not explicitly `disconnect()`ed. They will continue observing their target nodes even after those nodes are removed from the DOM. An explicit call to disconnect is necessary to release connected nodes and prevent memory leaks.

### Proposed Fixes
This PR resolves these issues by:

* Ensuring Tooltipster instances are properly destroyed. That part uses some internal details like `$(el).data('tooltipster-ns')` to access the tooltipster instance, but is unavoidable.
* Correctly managing the mapping and cleanup of destroyable objects.
* Explicitly disconnecting `MutationObserver`s to avoid residual references.

I have verified that after these changes, no memory leaks occur in NPC sheets, item/feat sheets, and character sheets. However, **IWR tooltips in the chat log still exhibit memory leaks**. This is because chat messages and their tooltips are rendered before being inserted into the popped-out chat log application, complicating the setup of `MutationObserver`s. Addressing this specific issue will require a separate investigation.

### DIFF
--- a/src/util/destroyables.ts
+++ b/src/util/destroyables.ts
@@ -2,11 +2,7 @@ import Sortable from "sortablejs";
 import { ErrorPF2e } from "./misc.ts";
 
 class DestroyableManager {
-    #bodyObserver: MutationObserver;
-
-    #appObservers = new Map<Node, MutationObserver>();
-
-    #destroyables = new Map<Node, Destroyable[]>();
+    #appObservers = new Map<Node, MutationObserverContext>();
 
     declare static instance: DestroyableManager;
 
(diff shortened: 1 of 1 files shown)


stwlam:
Do you have any documentation showing #3 is true? Everything I could find states they're cleaned up once the observation target is garbage-collected.

stwlam:
I'm also not clear on what is meant by #2: using the destroyables as keys was not a mistake.

Codas:
> Do you have any documentation showing 3 is true? Everything I could find states they're cleaned up once the observation target is garbage-collected.

No real documentation, just a closed (because duplicate) issue on the whatwg github https://github.com/whatwg/dom/issues/484 that then links to the issue of being able to disconnect single targets instead of all or nothing: https://github.com/whatwg/dom/issues/126

There are no details or measurements available in these issues unfortunately.

This prompted me to do my own testing though, and from my testing in chrome at least, MutationObserver does indeed keep a strong reference to the html element it is observing. 
This is the memory view of a profiling run with the code of this current PR:
Bildschirmfoto 2024-12-12 um 09 13 48


And this with the same code base, but just the following part removed:
```ts
context.observer.disconnect();
```
Bildschirmfoto 2024-12-12 um 09 11 28


The methodology is the same as in my previous PR: opening, then closing a sheet (character sheet in this example), then running manual garbage collection, repeating a few times.

stwlam:
I think the fix for the body observer is to simply set the keys to the application frame rather than the content-window element.

Codas:
That should work, yes.
But to still be able to fix the other issues, we would probably need two separate mutation observer handler functions, one for the global body observer and one for the local application content.

Codas:
This should be better now, using only a single observer to watch for changes of the `document.body` `childList`.

stwlam:
I think you can drop the `TooltipsterTarget` class entirely now as well.

Codas:
> I think you can drop the `TooltipsterTarget` class entirely now as well.

I'm not sure about that. I currently check for the inclusion of the html node linked to the `destroyable` instance. As far as I could tell, the plain Tooltipster instance value keeps no accessible reference to the html node, which makes the class still necessary.

We could as of now probably live without the `contains` check and just destroy every destroyable in the scope of the application that is being re-rendered, but I'm not sure I like that honestly. The contains check is reasonably cheap, only runs in very specific circumstances and gives some meassure of future proofing against partial html updates.

Codas:
Your are absolutely right, don't know why I missed that. Probably only looked at properties, not functions...

Codas:
The ad-hoc tooltipster instance type can also be removed as this is already available as `JQueryTooltipster.ITooltipsterInstance`.

PR #17693, opened 2024-12-12
Remove destroyable management from chat messages

stwlam:
Tooltipster usage outside application windows will be skipped pending their purgation


CarlosFdez:
kaboom

PR #17694, opened 2024-12-12
Automate Thaumaturge Regalia aura and other benefits

pedrogrullada:
~~Setting myself up for a conflict in the localization file, but I can deal with it when it comes~~ Dealt with


static/lang/re-en.json
```diff
@@ -5688,7 +5688,9 @@
                     "Chalice": "Chalice",
                     "Lantern": "Lantern",
                     "Mirror": "Mirror",
-                    "Regalia": "Regalia",
+                    "Regalia": {
+                        "UnflankableToggle": "Regalia — Origin is being flanked."
```
TikaelSol, line 5692:
```suggestion
                        "UnflankableToggle": "Regalia owner is being flanked."
```
Maybe this is more obvious as to what it means. Origin has a system meaning that may trip up some users, since it could refer to the origin of the strike too.

pedrogrullada: Yeah, I'm cool with this

PR #17709, opened 2024-12-14
Add alternate amp and effect for Mind's Light Circlet

pedrogrullada:


packs/equipment/minds-light-circlet.json
```diff
@@ -40,6 +40,67 @@
...
                 "type": "item",
                 "value": 2
+            },
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.Equipment.MindsLightCirclet.LightToggle",
+                "option": "minds-light-circlet-light",
+                "placement": "spellcasting",
+                "toggleable": true
+            },
```
pedrogrullada, line 50: Could've used a made up focus points-counting roll option here, but it felt hacky and I imagine users would like a way to turn off the light besides messing with their focus points or uninvesting the item anyway.

PR #17731, opened 2024-12-16
Support suboptions from config in roll option rule element

CarlosFdez:
Created to purify some frankly scary thaumaturge related rule elements

```json
{
  "key": "RollOption",
  "option": "super-skill",
  "toggleable": true,
  "phase": "beforeDerived",
  "suboptions": {
    "config": "skills",
    "predicate": [
      {
        "gt": [
          "skill:{choice|value}:rank",
          0
        ]
      }
    ]
  }
}
```


ottyn:
Just curious if this is being delayed until next major release or waiting for additional data entry?

PR #17747, opened 2024-12-18
Fix Divine Vessel spell effects

TikaelSol:


TikaelSol:
Bah. Of course NPCs have the spell.

PR #17755, opened 2024-12-19
Update select Archetype journal entries to use `@Embed` and rule elements on Ka Stone and Expeditious Advance feat

ammalagonc:


static/lang/re-en.json
```diff
@@ -3359,6 +3359,9 @@
                 "HumbugPocket": {
                     "ConcealAnObjectLabel": "Humbug Pocket (Conceal an Object in the Pocket)"
                 },
+                "KaStone": {
+                    "Label": "vs. Death and Void effects"
+                },
```
TikaelSol, line 3364: We have a generic section for this kind of thing

ammalagonc: Label and localization have been updated

PR #17782, opened 2024-12-21
Add search to other actor inventories

reyzor1991:
same as actor inventory search

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -62,7 +62,7 @@ import { IdentifyItemPopup } from "./popups/identify-popup.ts";
 import { ItemTransferDialog } from "./popups/item-transfer-dialog.ts";
 import { IWREditor } from "./popups/iwr-editor.ts";
 import { RemoveCoinsPopup } from "./popups/remove-coins-popup.ts";
-import {iterateAllItems} from "@actor/helpers.js";
+import { iterateAllItems } from "@actor/helpers.js";
 
 /**
  * Extend the basic ActorSheet class to do all the PF2e things!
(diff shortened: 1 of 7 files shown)


stwlam:
Definitely not going to settle for copy-pasting the same code over.

reyzor1991:
Moved filter to actor sheet

src/module/actor/sheet/base.ts
```diff
@@ -219,6 +231,10 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends ActorSheet<TActo
     }
 
     protected prepareInventoryItem(item: PhysicalItemPF2e): InventoryItem {
+        const items = [...iterateAllItems(this.actor)].filter((i) => i.isOfType("physical"));
+        this.#inventorySearchEngine.removeAll();
+        this.#inventorySearchEngine.addAll(items.map((i) => R.pick(i, ["id", "name"])));
```
CarlosFdez, line 236: This should be in prepareInventory(), not prepareInventoryItem(). prepareInventoryItem() is called per item.

static/templates/actors/party/sheet.hbs
```diff
@@ -56,6 +56,11 @@
 
             <section class="inventory">
                 {{> "systems/pf2e/templates/actors/partials/coinage.hbs" owner=@root.owner}}
+                <header class="inventory-header">
+                    <div class="search">
+                        <input type="search" spellcheck="false" placeholder="{{localize "PF2E.Actor.Stash.Search"}}" />
+                    </div>
+                </header>
```
CarlosFdez, line 63: if it will be part of every inventory, it might be wise to move this code to inventory.hbs, both for this and the one in character itself.

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/17786

CarlosFdez:
Alrighty, thank you!

PR #17804, opened 2024-12-23
Fix corruption in grant item data when jumping multiple levels

CarlosFdez:
There are 3 different bugfixes here, but as each one is useless on its own, I submitted them altogether
1) Move the AELike initializing adeptChoices and paragonChoices earlier. AELikes and Grant items resolve immediately during preCreate and ignore priority.
2) While granter flags are set in source data to be saved, AELikes referring to them were initially incorrect. We assign to the item during data preparation so that those AELikes function appropriately.
3) item updates performed in preCreate() immediately trigger actor data preparation and caused us to lose all temporary changes applied during preCreate. This was the real reason the array did not exist in the adept choice.

### DIFF
--- a/src/module/rules/rule-element/grant-item/rule-element.ts
+++ b/src/module/rules/rule-element/grant-item/rule-element.ts
@@ -1,7 +1,7 @@
 import type { ActorPF2e, ActorType } from "@actor";
 import { ConditionPF2e, ItemPF2e, ItemProxyPF2e } from "@item";
 import type { ItemSourcePF2e } from "@item/base/data/index.ts";
-import { ItemGrantDeleteAction, ItemSourceFlagsPF2e } from "@item/base/data/system.ts";
+import { ItemGrantDeleteAction, ItemGranterSource, ItemSourceFlagsPF2e } from "@item/base/data/system.ts";
 import { PHYSICAL_ITEM_TYPES } from "@item/physical/values.ts";
 import { SlugField, StrictArrayField } from "@system/schema-data-fields.ts";
 import { ErrorPF2e, isObject, setHasElement, sluggify, tupleHasValue } from "@util";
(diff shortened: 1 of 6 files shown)


stwlam:
This is going to need some serious testing with existing data (not recent thaumaturge things).

src/module/rules/rule-element/grant-item/rule-element.ts
```diff
@@ -321,12 +347,11 @@ class GrantItemRuleElement extends RuleElementPF2e<GrantItemSchema> {
                 (setHasElement(PHYSICAL_ITEM_TYPES, grantee.type) ? "detach" : "cascade"),
         };
 
-        if (grantee instanceof ItemPF2e) {
+        grantee.flags = fu.mergeObject(grantee.flags ?? {}, { pf2e: { grantedBy } });
+        if (grantee instanceof ItemPF2e && grantee._id) {
             // This is a previously granted item: update its grantedBy flag
             // Don't await since it will trigger a data reset, possibly wiping temporary roll options
```
stwlam, line 353: Why should this comment remain if there's nothing awaitable?

CarlosFdez: It should not, I failed to catch it in my victory lab when I was trying to find out what parts we actually needed.

src/module/rules/rule-element/grant-item/rule-element.ts
```diff
@@ -321,12 +347,11 @@ class GrantItemRuleElement extends RuleElementPF2e<GrantItemSchema> {
                 (setHasElement(PHYSICAL_ITEM_TYPES, grantee.type) ? "detach" : "cascade"),
         };
 
-        if (grantee instanceof ItemPF2e) {
+        grantee.flags = fu.mergeObject(grantee.flags ?? {}, { pf2e: { grantedBy } });
+        if (grantee instanceof ItemPF2e && grantee._id) {
```
stwlam, line 351: Check that the actor has the ID in its embedded collection rather than just confirming it has an ID.

CarlosFdez: I can do that, though the reason I did that check was actually so that typescript wouldn't complain. I figured if the original didn't check, the new one didn't, but there's harm in actually checking.

PR #17806, opened 2024-12-23
Don't await document create/delete in effect macros

LebombJames:
This should speed up the loops if multiple actors are selected


stwlam:
This revision looks extremely out of date. Mind rebasing?

LebombJames:
Rebased

PR #17813, opened 2024-12-23
Add inline healing toggleable upgrade for Chirurgeon's Greater Field Discovery

pedrogrullada:


packs/classfeatures/greater-field-discovery-chirurgeon.json
```diff
@@ -24,7 +24,106 @@
...
+                "property": "dice-number",
+                "selectors": [
+                    "inline-healing"
+                ],
+                "value": 0
+            },
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "chirurgeon-greater-field-discovery",
```
TikaelSol, line 60: Isn't this the item it's on?

pedrogrullada: This is the toggleable roll option I set up earlier

Issue #17807, opened 2024-12-23
Quick Alchemy create items do not stack/remove


manuel-hegner:
The following szenario:

1. alchemist with versatile vials
2. Use Quick Alchemy to craft an Alchemist's Fire
3. attack with the Alchemist's Fire
4. the alchemist now has a stack of 0 Alchemist's Fire
5. Use Quick Alchemy to craft an Alchemist's Fire
6. the alchemist now has a stack of 0 and a stack of 1 Alchemist's Fire
7. attack with the Alchemist's Fire
8. the alchemist now has two stacks of 0 Alchemist's Fire

Ideally, the stack would vanish completely after being used. That might be difficult to implement though, as generally stacks of size 0 are quite useful for players. The user friendliness would be improved considerably though, if Quick Alchemy would search for a stack of infused identical items not in a container and add to that before creating a new stack.

CarlosFdez:
Depleting size 0 stuff isn't possible now (rolling damage will stop working and the damage buttons will disappear even if we store it in the damage card), but stacking with a 0 quantity item with the same held state logically makes sense. at least in the short term.

manuel-hegner:
After some investigation there is already a handling for this. It fails right now because of https://github.com/foundryvtt/pf2e/blob/2ad0158cdbc5b564338137e56b99281c6515b681/src/module/item/physical/document.ts#L474

In the case of quick alchemy both items have their `isHeld` property set to `true`, which prevents stacking. I think this should only prevent stacking if both stacks are also non-empty.

CarlosFdez:
Not quite that, we don't want held items stacking with worn items. I saw the stacking before release, but we weren't aware of the exact solution we wanted at the time.

PR #17817, opened 2024-12-24
Allow held items to stack with other held items

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/17807

An alternative idea if we want to keep held items separate is to allow stacking held items if the quantity of any of them is 0, and to auto-delete if reduced to 0 and a stack exists, but that's significantly more complicated to solve and I don't think it ended up being worth it after all.

### DIFF
--- a/src/module/item/physical/document.ts
+++ b/src/module/item/physical/document.ts
@@ -464,14 +464,18 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         return super.getEmbeddedDocument(embeddedName, id, options);
     }
 
-    /** Can the provided item stack with this item? */
+    /**
+     * Can the provided item stack with this item?
+     * @param item an item we are trying to add to the inventory
+     */
     isStackableWith(item: PhysicalItemPF2e): boolean {
         const preCheck =
             this !== item &&
             this.type === item.type &&
             this.name === item.name &&
             this.isIdentified === item.isIdentified &&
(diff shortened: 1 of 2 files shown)


stwlam:
Yeah, let's only stack if quantity is zero.

src/module/item/physical/document.ts
```diff
@@ -464,14 +464,19 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
     isStackableWith(item: PhysicalItemPF2e): boolean {
         const preCheck =
             this !== item &&
             this.type === item.type &&
             this.name === item.name &&
             this.isIdentified === item.isIdentified &&
-            ![this, item].some((i) => i.isHeld || i.isOfType("backpack"));
+            this.isHeld === item.isHeld &&
+            !this.isOfType("backpack") &&
+            !item.isOfType("backpack");
```
stwlam, line 479: Move to override

CarlosFdez:
Sure, though that suggestion won't be clean unless we also delete at 0 if there's a possible merge, and to do that requires us to be able to strike even if the weapon is no longer in our inventory (since the chat message lookup will not find the item that was just deleted). This isn't an impossible task, though I probably won't do that by next release (maybe not even until weapons convert to statistics). It might solve this issue in most real world cases though, which might be good enough.

CarlosFdez:
Done. There is a quirk that can occur if you make the items before using them, but we can maybe solve that separately somehow.

PR #17818, opened 2024-12-24
Add effects for Prey Mutagen

pedrogrullada:


packs/actions/timely-dodge.json
```diff
@@ -0,0 +1,74 @@
...
+    "system": {
+        "actionType": {
+            "value": "reaction"
+        },
+        "actions": {
+            "value": null
+        },
+        "category": "defensive",
+        "description": {
+            "value": "<p><strong>Trigger</strong> You are targeted by a Strike</p><hr /><p><strong>Effect</strong> You dart aside, gaining a +1 circumstance bonus to AC, then Step after the Strike resolves.</p>"
```
TikaelSol, line 14: Feels like we should mention the scaling AC with the higher level versions

packs/equipment-effects/effect-prey-mutagen-lesser.json
```diff
@@ -0,0 +1,61 @@
+{
+    "_id": "HLn9nQY4SAvyhaom",
+    "img": "systems/pf2e/icons/equipment/alchemical-items/alchemical-elixirs/prey-mutagen.webp",
+    "name": "Effect: Prey Mutagen (Lesser)",
```
TikaelSol, line 4: Is there a way to condense these three effects? Maybe store the AC bonus value with AE likes then the action can pull its value from that data instead of using AdjustModifier?

pedrogrullada: The problem is the duration. The lesser one is 1 minute, the moderate is 10 minutes, and the rest are 1 hour.

packs/actions/timely-dodge.json
pedrogrullada: Each effect other than the lesser one has a description override with the right AC. Should I mention the scaling AC in the base item anyway?

TikaelSol: Ah, that's fine then

PR #17835, opened 2024-12-26
Condense Grudgestone equipment effects and update automation for Cloak of Elvenkind

ammalagonc:
- This removes the need for the effect on Cloak of Elvenkind


packs/equipment/cloak-of-elvenkind-greater.json
```diff
@@ -34,7 +34,23 @@
...
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "cloak-of-elvenkind-greater"
+                ],
+                "selector": "stealth",
+                "type": "item",
+                "value": 1
+            }
+        ],
```
TikaelSol, line 53: At the least it should be a +2, but also I don't know that the toggle is needed. I mean practically you'll just always have it on.

PR #17860, opened 2024-12-30
Add rarity tags to Archetype journals and add missing rarity to several entries

ammalagonc:
Adds rarity to several archetypes entries that were missing this and used rarity tags:

image


TikaelSol:
I think we can remove the hr after the tag

ammalagonc:
> I think we can remove the hr after the tag

Changes made

PR #17886, opened 2025-01-02
Add max-range to AdjustStrike RE

KaitoKuroba:
Currently, max-range on strikes can't be set dynamically with REs, which may cause limitations for homebrew or 3rd-party content. This PR adds the max-range property to the AdjustStrike RE.

The implementation is almost identical to the existing range-increment. However, it also checks whether a strike already has a range-increment. Changes to the max-range would otherwise mess up the calculation of the range-increment in these cases.


stwlam:
Sorry, but `AdjustStrike` isn't receiving further attention now that `ItemAlteration` exists.

PR #17898, opened 2025-01-03
Add range-increment and range-max to ItemAlteration

KaitoKuroba:
This is a follow up to #17886 and implements `range-increment` and `range-max` to `ItemAlteration`.

As discussed with @stwlam, melee weapons are filtered. This is done by checking if `item.system.range` or `item.system.maxRange` is set, which is only the case for ranged weapons.


src/module/rules/rule-element/item-alteration/alteration.ts
```diff
@@ -389,6 +392,24 @@ class ItemAlteration extends foundry.abstract.DataModel<RuleElementPF2e, ItemAlt
                 }
                 return;
             }
+            case "range-increment": {
+                const validator = ITEM_ALTERATION_VALIDATORS[this.property];
+                if (!validator.isValid(data)) return;
+                if (!(data.item instanceof WeaponPF2e) || !data.item.system.range) return;
```
CarlosFdez, line 398: Why do you need to check if its a weapon? It appears the other alterations don't need to perform this check and cast appropriately after running isValid().

src/module/rules/rule-element/item-alteration/alteration.ts
```diff
@@ -389,6 +392,24 @@ class ItemAlteration extends foundry.abstract.DataModel<RuleElementPF2e, ItemAlt
...
+                const rangeIncrement = data.item.system.range as number;
```
CarlosFdez, line 399:
```suggestion
                const rangeIncrement: number = data.item.system.range;
```

I'm assuming the casting is because range is a set of numbers, not a specific number, and modifying it might have caused issues.

KaitoKuroba:
Both the checks and the casting were not originally intended. In both cases, I got errors in VS Code without them. TBH, I don't know why it works now the second time, or what I missed the first time. I've now removed the parts.

CarlosFdez:
Looks good to me, thanks for creating the PR.

PR #17948, opened 2025-01-08
Fix razmiri mask, porcelain description

rectulo:
This is an apex item. My changes are the usual wording for such items (i.e. Necklace of allure).

I Fixed the number of required actions for the last activation (3 instead of 1 by the book)


rectulo:
image

PR #17961, opened 2025-01-09
Fix Versatile Vials having item bonus to attack rolls before level 4

Randommisha13:
Versatile Vials have a bug in automation that incorrectly applies item bonus to attack rolls from 4th level upgrade to lower level versions. As far as I can understand, the issue is in automation that scales them with actor level, the main problem lying in automation defaulting to +1 bonus for Versatile Vials below 12th level.

### DIFF
--- a/packs/equipment/versatile-vial.json
+++ b/packs/equipment/versatile-vial.json
@@ -55,7 +55,15 @@
                 "key": "FlatModifier",
                 "selector": "{item|id}-attack",
                 "type": "item",
-                "value": "ternary(gte(@item.level,18),3,ternary(gte(@item.level,12),2,ternary(gte(@item.level,4),1)))"
+                "predicate": [
+                    {
+                        "gte": [
+                            "parent:level",
+                            4
+                        ]
+                    }
+                ],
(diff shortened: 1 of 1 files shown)


packs/equipment/versatile-vial.json
```diff
@@ -55,7 +55,7 @@
                 "key": "FlatModifier",
                 "selector": "{item|id}-attack",
                 "type": "item",
-                "value": "ternary(gte(@item.level,18),3,ternary(gte(@item.level,12),2,1))"
+                "value": "ternary(gte(@item.level,18),3,ternary(gte(@item.level,12),2,ternary(gte(@item.level,4),1)))"
```
pedrogrullada, line 58: Good catch, but your ternary is missing arguments and will give warnings in the console. Consider leaving the formula intact and predicating the Flat Modifier on the following instead:
```json
{
      "gte": [
        "parent:level",
        4
      ]
}
```

Randommisha13:
Thank you for your help, I don't really know how to code rules yet, so I went for the most obvious solution

PR #17968, opened 2025-01-11
Add scaling to Rising Surf description

MrVauxs:


packs/spells/rising-surf.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p>You create a wave of water that you ride, banking around obstacles and surfing to higher ground. You move up to [[35 + (floor((@item.level - 1) / 2) * 5)]] feet, raising yourself up to [[5 + (floor((@item.level - 1) / 2) * 5)]] feet above the ground. (You can avoid many types of difficult terrain in this way.) You must end your movement on an unoccupied space where you have solid footing. This movement isn't a Stride, but you measure the distance in a similar way, and it still triggers reactions caused by movement. You can't transport anyone else with you.</p>\n<hr />\n<p><strong>Heightened (+2)</strong> Increase the distance you move by 5 feet and the maximum height traveled above the ground by 5 feet.</p>"
```
TikaelSol, line 14: Don't know that I like this one since it's not really a roll. We could get clever with a description alteration, but that means localizing the text of the feat and maybe that has a hidden snag. I can also definitely see a use for something that resolves to text directly, but that may have localization issues too.

PR #17994, opened 2025-01-14
Add support for registering spells to staves

CarlosFdez:
This is an initial version to get eyes on this feature before I to start to internally migrate all our existing data.

If the staff trait is removed, the data is removed when the sheet is closed. It is otherwise removed if the list of spells is empty.


CarlosFdez:
This should now work for codas as well, though having to check if the type is weapon/equipment explicitly rather than checking for the presence of the spell property bothers me a bit.

src/module/item/base/sheet/sheet.ts
```diff
@@ -499,6 +502,14 @@ class ItemSheetPF2e<TItem extends ItemPF2e> extends ItemSheet<TItem, ItemSheetOp
             });
         }
 
+        // Handle contenteditable fields
+        for (const input of htmlQueryAll<HTMLSpanElement>(html, "span[contenteditable][data-property]")) {
```
stwlam, line 506: If you have to assert anyway, just use `html.querySelectorAll`

CarlosFdez: Looks like the assert (is that actually called an assert? Seems very different from the usual cases I see) isn't necessary. Maybe it was necessary in a previous version of the thing and it carried on through.

stwlam: It's called a [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions), but maybe the DOM lib is wise to `${tagName}[${string}]`.

PR #18006, opened 2025-01-16
Skip item alteration roll option fetching if it doesn't match any items

CarlosFdez:
This is a very minor speedup on actors that have a few strikes and a lot of alterations of any kind. In the best case it removes around 3-4ms, but its simple and easy.


CarlosFdez:
I think the other tweaks have mostly made this unnecessary, but its simple and doesn't hurt.

PR #18012, opened 2025-01-17
Clone data before cloning actor in `CharacterPF2e#_preUpdate`

stwlam:
Property deletion is working better now.


stwlam:
No longer necessary!

PR #18014, opened 2025-01-17
Improve performance of mergeable sub options

CarlosFdez:
For an otherwise bare bones swashbuckler with 6 finisher options, it speeds up adjusting the toggle's selected suboption from 212ms to 27ms. Larger actors will lead to larger gulfs (which I've witnessed sometimes approach seconds).

You can test on a swashie with the sheet closed by selecting the token and running
`foundry.utils.benchmark(() => _token.actor.toggleRollOption("all", "finisher", null, true, _token.actor.synthetics.toggles.all.finisher.suboptions.find((s) => !s.selected).value), 50)`

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -766,29 +766,54 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
             }
         }
 
-        // If this item is of a certain type and belongs to a PC, change current HP along with any change to max
-        if (this.actor?.isOfType("character") && this.isOfType("ancestry", "background", "class", "feat", "heritage")) {
-            const actorClone = this.actor.clone();
-            const item = actorClone.items.get(this.id, { strict: true });
-            item.updateSource(changed, options);
-            actorClone.reset();
-
-            const hpMaxDifference = actorClone.hitPoints.max - this.actor.hitPoints.max;
-            if (hpMaxDifference !== 0) {
-                const newHitPoints = this.actor.hitPoints.value + hpMaxDifference;
-                await this.actor.update(
-                    { "system.attributes.hp.value": newHitPoints },
-                    { render: false, allowHPOverage: true },
-                );
-            }
-        }
-
         // Run preUpdateItem rule element callbacks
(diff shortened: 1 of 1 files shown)


src/module/item/base/document.ts
```diff
@@ -791,6 +774,48 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+    ): Promise<boolean | void> {
+        if ((await super._preUpdateOperation(documents, operation, user)) === false) {
+            return false;
+        }
+
+        // If this item is of a certain type and belongs to a PC, store current hp to be checked later
+        const actor = documents.find((d): d is ItemPF2e => d instanceof ItemPF2e)?.actor;
+        if (actor) {
+            operation.previousSystemData = fu.mergeObject(operation.previousSystemData ?? {}, {
+                maxHitPoints: actor.hitPoints?.max,
```
stwlam, line 791: prepared value?

src/module/item/base/document.ts
```diff
@@ -948,4 +973,9 @@ interface RefreshFromCompendiumParams {
     update?: boolean;
 }
 
+/** An extension of DatabaseUpdateOperation to pass on system specific data between phases */
+type ItemPF2eDatabaseUpdateOperation = DatabaseUpdateOperation<foundry.abstract.Document | null> & {
+    previousSystemData?: { maxHitPoints?: number };
```
stwlam, line 978: Maybe just call it `previous` or `prior` and include an outer `system` property.

CarlosFdez: We can omit the system property. I called it that to try to distinguish from core foundry properties, but then after sleeping I kinda realized that we'd just put pf2e in the name if we really want to do that. previous is fine!

src/module/item/base/document.ts
CarlosFdez: You mean to use `actor.system.attributes.hp.max` instead? Actor is a document, so it should be the prepared value already I think.

stwlam: I mean why are you using the prepared value and not the source value?

PR #18050, opened 2025-01-20
Update predicate for Bloodsoaked Dash feat

ammalagonc:
Current predicate wasn't working

### DIFF
--- a/packs/feats/bloodsoaked-dash.json
+++ b/packs/feats/bloodsoaked-dash.json
@@ -35,7 +35,7 @@
                         ]
                     }
                 ],
-                "selector": "all-speeds",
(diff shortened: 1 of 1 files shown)


packs/feats/bloodsoaked-dash.json
```diff
@@ -35,7 +35,7 @@
                         ]
                     }
                 ],
-                "selector": "all-speeds",
```
TikaelSol, line 38: If this isn't working then we have a problem. We shouldn't be using the `speed` selector

ammalagonc:
Whatever the issue was it got resolved
Putting this in the trash can

PR #18053, opened 2025-01-20
Fix damage of Spray of Stars

Redst0neFlux:
Spray Of Stars deals 2d4 fire, not 1d4. It dealt 1d4 in legacy


TikaelSol:
Thanks for the PR. As a note for suture submissions PR titles should be in imperative case, basically finishing the sentence `If accepted this PR will ___`. This is because when compiling release notes the PR titles will be turned into the initial line in the changelog. I changed the title for this one already.

Issue #18032, opened 2025-01-18
Deadly, Forceful, etc. doesn't respond to Damage Alteration


J-Dawe:
The Quickstart guide to rule elements says the following under Damage Alteration:

> This differs from simply adding a dice using a DamageDice rule because it increases the weapon's actual damage dice, similar to how a striking rune works. This will increase damage dealt by traits like deadly or fatal, or the damage from spells like gravity weapon. Unlike a striking rune or a Striking rule this will not add the magical trait to the weapon. The system uses this method to increase the damage of versatile vials, as an example.

However, from my testing this doesn't work. Increasing a rapier to three damage dice with Damage Alteration still results in it doing +1d8 damage on a critical hit instead of +2d8. A scimitar with multiple damage dice still only does +1/+2 damage with forceful. The Advantageous Assault feat only does +1 damage (or +3 with a two-handed weapon). There are probably other things that don't work but those were the ones I tested.

stwlam:
This is intentional. Increasing the damage of weapon should use an item alteration.

Issue #16714, opened 2024-09-29
DamageAlteration RE not working as expected {bug?}


CrackJackFlood:
@weapon.system.damage.dice oes not resolve additional damage dice added by the RE. Deadly trait works though.

stwlam:
The rule element doesn't have access to this information.

CrackJackFlood:
> The rule element doesn't have access to this information

If DamageAlteration is a way of modifying base damage dice so making a 1d8 into 2d8 etc How is it expected to work for RE's that rely on @weapon.system.damage.dice.

Logically @weapon.system.damage.dice should resolve an items damage dice *correctly* when its effected by DamageAlteration.



Can you possibly suggest workarounds for this if its not going to be fixed as its frustrating as its only half working as expected (deadly trait works for example) like for example gravity weapon and thaumaturges implements both go off bonus damage equal to the number of weapon dice

PR #18058, opened 2025-01-20
Extend ItemAlteration to be able to change base damage dice number

### LINKED ISSUE #18054: DamageAlteration RE does not affect effects that refer to `@weapon.system.damage.dice`
This issue had already been reported once in #16714, but back then, it was dismissed as `WONTFIX`. However, I believe Shark misunderstood what was being requested, and thought CrackJackFood was requesting that `@weapon.system.damage.dice` be available in `DamageAlteration`.

As CrackJackFood points out in the issue, the wiki _does_ refer to [the thing in the title being something DamageAlteration should do](https://github.com/foundryvtt/pf2e/wiki/Quickstart-guide-for-rule-elements#damage-alteration):

> This differs from simply adding a dice using a DamageDice rule because it increases the weapon's actual damage dice, similar to how a striking rune works. **This will increase** damage dealt by traits like deadly or fatal, or **the damage from spells like gravity weapon.** Unlike a striking rune or a Striking rule this will not add the magical trait to the weapon. The system uses this method to increase the damage of versatile vials, as an example.

_(Emphasis mine.)_

Despite that, `DamageAlteration` can only affect things like Fatal or Deadly as the wiki says.  
A quick glance at the code suggests that the cause of the difference is that `@weapon.system.damage.dice` is unprocessed damage -- the `DamageAlteration` happens way after the damage is processed. Additionally, `DamageAlteration` specifically goes over things like Fatal and Deadly to alter them.

Cerapter:
Sort of fixes #18054, #18032 and #16714?

While looking for DamageAlteration info, I noticed that in #18032, Shark mentioned that increasing the damage of weapons should be done in ItemAlteration. I suspect he meant damage faces, but I was curious if you could really do base weapon damage dice number increases that way, and turns out, yeah?

Additionally, I think having the base weapon damage dice number increased for weapons in ItemAlteration makes more sense than having it in DamageAlteration. This solution also "fixes" the problem that Gravity Weapon and the like have, where by referring to `@weapon.system.damage.dice`, they don't benefit from DamageAlteration increases.

### DIFF
--- a/src/module/rules/rule-element/item-alteration/alteration.ts
+++ b/src/module/rules/rule-element/item-alteration/alteration.ts
@@ -25,6 +25,7 @@ class ItemAlteration extends foundry.abstract.DataModel<RuleElementPF2e, ItemAlt
         "category",
         "check-penalty",
         "damage-dice-faces",
+        "damage-dice-number",
         "damage-type",
         "defense-passive",
         "description",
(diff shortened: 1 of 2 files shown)


src/module/rules/rule-element/item-alteration/alteration.ts
```diff
@@ -194,6 +195,23 @@ class ItemAlteration extends foundry.abstract.DataModel<RuleElementPF2e, ItemAlt
 
                 return;
             }
+            case "damage-dice-number": {
+                const validator = ITEM_ALTERATION_VALIDATORS[this.property];
+                if (!validator.isValid(data) || !(data.item instanceof ItemPF2e)) {
```
stwlam, line 200: I think this should be safe to run on source data. Are you getting an error without the `Item` instance check?

Cerapter: Ah, nope, seems like removing the Item instance check causes no compilation errors.

PR #18062, opened 2025-01-20
Add circumstance modifiers from climb and swim speeds to related actions

nikolaj-a:


src/module/system/action-macros/athletics/climb.ts
```diff
@@ -31,6 +40,15 @@ const action = new SingleCheckAction({
     cost: 1,
     description: `${PREFIX}.Description`,
     img: "icons/skills/movement/arrow-upward-blue.webp",
+    modifiers: [
```
CarlosFdez, line 43: I might be missing something, but why are these not derived from the actual modifiers?

nikolaj-a: What actual modifiers are you referring to? No modifiers come into this unless called by a macro or already present on the actor. These are action-specific modifiers that just happen to have a predictable predicate.

nikolaj-a: https://2e.aonprd.com/Rules.aspx?ID=2349

CarlosFdez: I mean that it looks like a repeat of what you did earlier, other than the modifier param which looks a bit strange.

nikolaj-a: Ah, now I understand what you mean. Remember that the action functions - `climb` and `swim` in this case - will go away at some point. So I put it in both places in anticipation of that.

The modifier param is just a concatenation to an optional array. Nothing fancy.

CarlosFdez: ooo, I see.

nikolaj-a: It's a bit hard to see in the GitHub diff without unfolding the unchanged lines, but it's actually two separate blocks. One is the (legacy) action function and the other is the action object.

Issue #18063, opened 2025-01-21
Vehicle Fortitude is not Functional


MelvinTheOdd:
The Vehicle Fortitude save does not function. Hover over  #the Fortitude Dice roller and the Dice does not spin. While you can click on it, it does not roll

PR #18066, opened 2025-01-21
Fix rolling vehicle fortitude saves

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/18063.

Is there a reason we don't return true in the else case?


static/templates/actors/vehicle/sidebar.hbs
```diff
@@ -124,4 +124,3 @@
         {{selectOptions emitsSoundOptions selected=data.attributes.emitsSound localize=true}}
     </select>
 </section>
-
```
stwlam, line 127: !

PR #18071, opened 2025-01-21
Add missing nature Ranger trained skill 

Najin-Alex:
Add missing nature trained skill for Ranger class,

### DIFF
--- a/packs/classes/ranger.json
+++ b/packs/classes/ranger.json
@@ -239,6 +239,7 @@
         "trainedSkills": {
             "additional": 4,
             "value": [
+                "nature",
                 "survival"
             ]
         },


ammalagonc:
https://github.com/foundryvtt/pf2e/pull/17919

Is the skill not being added when you add the class?

TMunsie:
I just created a new actor and added ranger as a class. Both nature and survival are already being correctly added. Can you please provide replication steps where this doesn’t happen?

Najin-Alex:
> #17919
> 
> Is the skill not being added when you add the class?



> I just created a new actor and added ranger as a class. Both nature and survival are already being correctly added. Can you please provide replication steps where this doesn’t happen?
it is not displayed visually in the class table.

https://github.com/user-attachments/assets/0c09e03c-1835-4ca5-8bd9-77b9c372ea8b

PR #18080, opened 2025-01-22
Correct monster warden circumstance bonus

### LINKED ISSUE #18079: Monster Warden effect applies wrong bonus from Monster Hunter
After the Fall Errata, the effects for Monster Warden and Monster Hunter were consolidated. Now, Monster Warden applies both bonuses.

However, it applies a +2 circumstance bonus to AC and saves (correct, from Monster Warden) and attacks (incorrect, original +1 from Monster Hunter should be used).

With Legendary Monster Hunter, both bonuses become +2 and the logic in the system right now is correct again.

dotadd:
Closes #18079

Separate the FlatModifier RE on the `Effect: Monster Warden` to only provide a +1 to attacks.

### DIFF
--- a/packs/feat-effects/effect-monster-warden.json
+++ b/packs/feat-effects/effect-monster-warden.json
@@ -4,7 +4,7 @@
     "name": "Effect: Monster Warden",
     "system": {
         "description": {
-            "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Monster Warden]</p>\n<p>You gain a +2 circumstance bonus to your next attack roll against that prey.</p>\n<p>You gain a +2 circumstance bonus to your next saving throw against that particular creature and to your AC against that creature's next attack against you.</p>"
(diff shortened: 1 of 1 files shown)


TikaelSol:
Should these all be +1?

dotadd:
So, this is actually a bit of a mess that my party ran into. As per the PC1 PDF and AoN, the bonuses are all +1 and become +2 with Legendary Monster Hunter. Then the Fall Errata changed the wording on Monster Hunter (attack rolls) and Monster Warden (ac and saves). But it also changed the value of the circumstance bonus, but only for ac and saves.

> Page 157: The original text of Monster Hunter could be read very narrowly, which wasn’t intended. Change the second to last sentence to “ You gain a +1 circumstance bonus to your next attack roll against that prey, and any ally you tell gains the same benefit. “

>Page 159: Update the text of Monster Warden to:
>”You understand how to defend against your prey. When you grant bonuses from Monster Hunter, each creature who benefits also gains a +2 circumstance bonus either to their AC the next time the creature attacks them or to their next saving throw against an effect from that particular creature (whichever comes first.)"

PR #18084, opened 2025-01-23
Update Favored terrain and other references to (premaster) Wild Stride

### LINKED ISSUE #16961: Replacement of Wild Stride by Unimpeded Journey
The Wild stride class ability has been remastered in Unimpeded Journey

I'm suggesting operate the following changes in :
- Effect: Favored Terrain (Increase Swim Speed), (Increase Climb Speed), (Gain Swim Speed), (Gain Climb Speed)
- Wild Strider feat (and the horizon walker archetype) 
- Ultimate Skirmisher (in the prerequisites)

dotadd:
Closes #16961
- Replace reference to (old) Wild Stride with Unimpeded Journey

Also:
- Add missing movement bonus effects to Favored Terrain
- Adjust these effects to remaster


TikaelSol:
These are automatable now, so we should actually mothball the effects and probably move the automation to the feature itself.

dotadd:
That sounds like a good idea. I'm just not clear on some of the details:

1) The terrain type should probably be a drop-down, because of things like [Terrain Master](https://2e.aonprd.com/Feats.aspx?ID=4885) being able to change it short-term, right? And a choice set would require removing and re-adding the feature.
2) How would we detect whether the actor is in that terrain? Just with toggle or anything more elaborate like regions?
3) Is ignoring difficult terrain from regions possibility?
4) I'm not sure how I would go about any part of this tbh: `Arctic: You need to eat and drink only one-tenth as much as usual, you aren’t affected by severe or extreme cold, and you can walk across ice and snow at full Speed without needing to Balance.`

TikaelSol:
Sorry, been away from merging for a bit.

1: We can do a drop down with terrain master to do the swap out, but we shouldn't make the initial choice a toggle like that. Basically a choice set and a separate roll option to set the terrain, predicate the roll option on not having terrain master. Then terrain master can have suboptions.

2: It would be with regions. `terrain:x` as the roll option already works.

3: Not currently. We need to figure out how we are going to handle difficult terrain in the system itself, which is waiting for some stuff in v13.

4: We wouldn't automate any of that

dotadd:
Thanks for the clarifications!

I'm on it, but running into an issue. I have to predicate the speeds on whether the actor already has the relevant speed type or not. So for example `["speed:swim"]` applied to a FlatModifier or a BaseSpeed respectively, but that doesn't seem to work. Here's the most simplified version of one such RE:
```
{
  "key": "FlatModifier",
  "selector": "swim-speed",
  "type": "status",
  "value": 10,
  "predicate": [
    "speed:swim"
  ]
}
```
Applied to an actor with a swim speed I set to 5 beforehand. And it just doesn't take effect. Nothing on the log either. Any idea what I'm missing here? How would I investigate this further? I'm not sure how I would "debug" this.

TikaelSol:
We would just have to ChoiceSet to ask if you have the speed.

dotadd:
Oh. I've been going at it completely differently, thinking that we weren't going to use an effect at all. And it took me a while trying complicated ways to integrate Terrain Master until I tried a simple way.

So how about this approach:
- Favored Terrain has all the REs, predicating on Terrain Master and Unimpeded Movement as needed
- Terrain is selected in a ChoiceSet
- Terrain Master unlocks a RollOption toggle with a dropdown to "overwrite" that choice (not actually, it's just predicates)
- The subcategories that have the option to gain or increase a speed get a drop-down to choose between those
- The actual FlatModifier and BaseSpeed REs predicate on that (or on the ones before them in the cases where there is no choice, like Sky increasing Fly Speed, but not granting one) as well as an `or` of `terrain:x` and a toggle. This is really just to not force GMs to consistently set terrain property regions on all maps. Cause I doubt most GMs even know those exist.

That works well, but before I tidy things up and push them, I'd like to check whether that's a desirable solution. For a character with all these feats/features and selecting one of the terrains that allow for both speed options, it looks like this. For characters with a subset of that, it thankfully takes up less space. Don't mind the exact labels for now.

dotadd:
This being the core of the logic (with another like it for climb speed).
```json
{
  "key": "RollOption",
  "option": "favored-terrain-swim-speed",
  "suboptions": [
    {
      "label": "Gain a swim speed",
      "value": "gain-swim-speed"
    },
    {
      "label": "Increase your swim speed",
      "value": "increase-swim-speed"
    }
  ],
  "toggleable": true,
  "alwaysActive": true,
  "value": true,
  "selection": "gain-swim-speed",
  "predicate": [
    "feature:unimpeded-journey",
    {
      "or": [
        {
          "and": [
            {
              "eq": [
                "{item|flags.pf2e.rulesSelections.favored-terrain}",
                "aquatic"
              ]
            },
            {
              "not": "favored-terrain-override"
            }
          ]
        },
        "favored-terrain-override:aquatic"
      ]
    }
  ],
  "priority": 101
}
```

It ends up being a lot of json for one feat. On the other hand, it works pretty well. So I'm not sure whether this is a good way of doing it. I'll appreciate any input.

dotadd:
@TikaelSol @ammalagonc This does work at least for all the cases I tested (including taking both feats after unimpeded movement). I just don't know whether this solution is desirable.

PR #18095, opened 2025-01-24
Add Mythic as an effect-only proficiency

kiptra:
## Context

Right now, the system does not have support for Mythic proficiency.

As a consequence of this, there is basically no automation surrounding Mythic abilities such as Rewrite Fate and Mythic Strike, which makes those abilities time-intensive to use due to needing to manually calculate the relevant bonuses and apply them individually.

## Changes

This PR adds Mythic as an effect-only proficiency, which means that players cannot select it from the dropdown of proficiencies, but instead they must be granted [temporary] Mythic proficiency with an effect (e.g. from Mythic Strike)

image

As for how this was accomplished, there were a few major points of implementation:
- First, a lot of proficiency details were centralized into `data.ts`; this enabled a bit of string de-duplication by centralizing the source of proficiencies
- Many usages of `ZeroToFour` were replaced with a new `ProficiencyRankNumber` to no longer assume the number of available proficiencies (and similarly, `ProficiencyRankString` has been created to cover `trained`/etc.)

If you could like to test this easily, I applied this to a random ability to grant Mythic Acrobatics

```json
{
  "key": "ActiveEffectLike",
  "mode": "upgrade",
  "path": "system.skills.acrobatics.rank",
  "value": 5
}
```

## Complications

The obvious complication is that a lot of things assumed the core game proficiencies of `U`/`T`/`E`/`M`/`L` would be the only five options. The less-obvious complication of this is that the static type-checking system of TS doesn't like the idea that Mythic can be enabled or disabled; a run-time configuration like that isn't well-liked by compile-time type-checking

Because of these factors, Mythic proficiency is always technically enabled/achievable through effects, though nothing that gives Mythic proficiency should really be in play unless Mythic rules are enabled (or someone makes a custom effect to get Mythic proficiency, but at that point they're doing whatever they want anyway)

This PR also does not concern itself with data entry; this is mostly to enable data entry later, though it could be worth implementing a system to spend resources like Mythic Points when such actions are performed (also not a part of this PR, though I would like to add that)

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -183,7 +183,7 @@ interface GamePF2e
                 pwol: {
                     enabled: boolean;
                     /** Modifiers for each proficiency rank */
-                    modifiers: [number, number, number, number, number];
+                    modifiers: [number, number, number, number, number, number];
                 };
                 /** Stamina */
                 stamina: boolean;
@@ -284,6 +284,7 @@ declare global {
         get(module: "pf2e", setting: "proficiencyExpertModifier"): number;
         get(module: "pf2e", setting: "proficiencyMasterModifier"): number;
         get(module: "pf2e", setting: "proficiencyLegendaryModifier"): number;
+        get(module: "pf2e", setting: "proficiencyMythicModifier"): number;
 
         get(module: "pf2e", setting: "metagame_partyVision"): boolean;
         get(module: "pf2e", setting: "metagame_secretCondition"): boolean;
--- a/src/module/actor/actions/base.ts
+++ b/src/module/actor/actions/base.ts
@@ -1,7 +1,6 @@
 import type { AbilityTrait } from "@item/ability/index.ts";
-import type { ProficiencyRank } from "@item/base/data/index.ts";
 import { ChatMessagePF2e } from "@module/chat-message/document.ts";
-import { PROFICIENCY_RANKS } from "@module/data.ts";
+import { getIndexedProficiencyOptions, ProficiencyRankString } from "@module/data.ts";
 import { getActionGlyph, sluggify } from "@util";
 import {
     Action,
@@ -26,20 +25,23 @@ interface BaseActionData<ActionVariantDataType extends BaseActionVariantData = B
     description: string;
     img?: string;
     name: string;
-    sampleTasks?: Partial<Record<ProficiencyRank, string>>;
+    sampleTasks?: Partial<Record<ProficiencyRankString, string>>;
     section?: ActionSection;
     slug?: string | null;
     traits?: AbilityTrait[];
     variants?: ActionVariantDataType | ActionVariantDataType[];
 }
 
-function labelSampleTasks(sampleTasks: Partial<Record<ProficiencyRank, string>>): { label: string; text: string }[] {
-    const unlabeled: { rank: ProficiencyRank; text: string }[] = [];
+function labelSampleTasks(
+    sampleTasks: Partial<Record<ProficiencyRankString, string>>,
+): { label: string; text: string }[] {
+    const unlabeled: { rank: ProficiencyRankString; text: string }[] = [];
     let rank: keyof typeof sampleTasks;
     for (rank in sampleTasks) {
         unlabeled.push({ rank, text: sampleTasks[rank]! });
     }
-    unlabeled.sort((t1, t2) => PROFICIENCY_RANKS.indexOf(t1.rank) - PROFICIENCY_RANKS.indexOf(t2.rank));
+    const indexedProficiencies = getIndexedProficiencyOptions();
+    unlabeled.sort((t1, t2) => indexedProficiencies[t1.rank]?.rank - indexedProficiencies[t2.rank]?.rank);
     return unlabeled.map((task) => ({ label: CONFIG.PF2E.proficiencyRanks[task.rank], text: task.text }));
 }
 
@@ -117,7 +119,7 @@ abstract class BaseAction<TData extends BaseActionVariantData, TAction extends B
     readonly description?: string;
     readonly img?: string;
     readonly name: string;
-    readonly sampleTasks?: Partial<Record<ProficiencyRank, string>>;
+    readonly sampleTasks?: Partial<Record<ProficiencyRankString, string>>;
     readonly section?: ActionSection;
     readonly slug: string;
     readonly traits: AbilityTrait[];
--- a/src/module/actor/actions/types.ts
+++ b/src/module/actor/actions/types.ts
@@ -1,8 +1,8 @@
 import type { ActorPF2e } from "@actor";
 import type { AbilityTrait } from "@item/ability/index.ts";
-import type { ProficiencyRank } from "@item/base/data/index.ts";
 import type { TokenPF2e } from "@module/canvas/index.ts";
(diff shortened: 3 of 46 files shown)


CarlosFdez:
I will read this soon, but my initial gut reaction isn't positive to this approach. I think I'd personally prefer either a proficiency type flat modifier or an adjust modifier that gets enabled by a roll option (and would show as a toggle in check dialogs as well). getProficiencyOptionByRank() is quite cumbersome for what it does as well.

Mythic proficiency is an overriding proficiency for a check, and not a base rank. Putting it part of the proficiency scaling risks it being selectable or saved to a sheet.

CarlosFdez:
I spoke it with shark. Mythic proficiency is not a rank you can obtain so I do not want them to be introduced as anything other than rule elements. Shark does not want "rank" to be a thing a rule element can said, and he shut down all discussion. Therefore the best we can is fake with rule elements, and I'm closing the PR.

stwlam:
Supe's explanation isn't quite on point, but it's sufficient to say he doesn't want proficiency ranks increased via AE-like.

CarlosFdez:
Ok, I read over the code more throughly. If you want to break apart certain refactors into their own thing so its much easier to see the shape of it, I'd appreciate it. Currently my concerns is the amount of places you have to account for mythic and some places you missed that you also need to account for. There might also be rule elements that operate on proficiency that might interfere. Perhaps there's a way to solve that which I haven't seen here yet.

So I think the getProficencyOptionByRank() is a good idea, but it needs a different implementation. `defaultModifier` should be `value`. Everything in that object is default, and its not the full modifier (missing level). rollOption should be part of the actual data, not a getter. It should also be created by a loop (value = index * 2) and renaming to getProficiencyDataByRank() is less ambiguous.

Is there a reason for the isMythic getter? I'm not opposed, but to make that correct you'll need to alternatively also check for the mythic trait.

kiptra:
Thanks for the comments @stwlam and @CarlosFdez! I'm happy to make changes or new PRs, but I think I'd like a hand figuring out how I can best help out here before I start throwing more PRs up

About the first few comments, are AE-likes not supposed to be able to modify proficiency? The Wiki for the repository specifically says `A common use of AE likes is increasing proficiency`, and so that's why I built off of that for mythic proficiency here (like in my test AE-like) and blocked the option in the dropdown (like in the above screenshots). I was imagining that one could make something like Mythic Strike grant mythic weapon proficiency for one strike with an AE-like, kind of like how Guidance works

https://github.com/foundryvtt/pf2e/wiki/Quickstart-guide-for-rule-elements#active-effect-like

Perhaps a better alternative would be to more carefully mark where mythic can and can't be used, and more selectively apply its existence?

> Ok, I read over the code more throughly. If you want to break apart certain refactors into their own thing so its much easier to see the shape of it, I'd appreciate it.

Yeah! Happy to split things out if it would help; would a PR for just centralizing proficiency rank information be a preferred starting point if mythic as a "standard-ish" proficiency isn't an outright-undesired outcome? Or would it be preferred to look into alternative ways to grant "mythic proficiency" that would be a better goal to work towards?

> Currently my chief concerns is the amount of places you have to account for mythic and some places you missed that you also need to account for. There might also be rule elements that operate on proficiency that might interfere. Perhaps there's a way to solve that which I haven't seen here yet.

May I please get some clarification on what's been missed? I covered everything I could find in the system based on the `ZeroToFour` types and a few other searches, but it's completely possible I missed something. If you're talking about things like Earn Income that are basically copy/pastes of Legendary rules, there is no table for Mythic Earn Income that I was able to find, so I defaulted to Legendary rules

> Is there a goal for the isMythic getter? I'm not opposed, but to make that correct you'll need to alternatively also check for the mythic trait, since that's how npcs are made mythic.

This was just moved from another spot in the code so that it would be more easily available for later checks (and to help with consistency for things like characters and NPCs having those different rules for seeing if they're mythic already); there is definitely room for improvement in the implementation, though I figured a simple lift was the least intrusive way of handling it in an otherwise-large PR. That said, I can change this (if I re-implement it) without too much trouble

https://github.com/foundryvtt/pf2e/pull/18095/files#diff-85a40a2b2e1aacf0d4cfefeb4d68a6ef0b9f377be6efdc5747b6a42048964729L361-L363

I am also curious about your thoughts on spending mythic points, though maybe this isn't the place to discuss that

CarlosFdez:
No worries! To be clear, a lot of my reaction is unfair. Paizo themselves seem to have treated Mythic almost as a throwaway thing. Whatever we do needs to be as invisible as possible with regards to its maintenance burden. This is why I think an iterative refactor to ensure mythic proficiency stuff is nearly invisible might be the play here.

> About the first few comments, are AE-likes not supposed to be able to modify proficiency? The Wiki for the repository specifically says `A common use of AE likes is increasing proficiency`

That is a reference to doing so semi-permanently, and in general we're moving away from that. Skills for example will likely stop using AELikes in the future to support tracking upgrades, and we have things like MartialProficiency for weapons. Also last I checked, AELike has issues being applied to specific items and specific targets.

I don't know what alternative people would be good with. Personally I'm good with anything that will show up in the check modifiers dialog, so upgrading FlatModifier to accept `rank` instead of `value` if `"type": "proficiency"` in the same way we accept `ability` if `"type": "ability"` is high up on my list, but I might be alone in that so don't jump the gun on it.

> May I please get some clarification on what's been missed?

At the time I wrote that I only caught one thing, but assumed there were more. That list has expanded:
1) The prof dropdown showing mythic should be read only, perhaps replacing the select with a span. It should be a full replacement if we show it on the sheet. An alternative is that it could revert to its real value when selected. Perhaps it shouldn't show on the sheet at all and only exist in rolls but I don't know how tenable that is.
2) Some abilities predicate on `proficiency:master` or `proficiency:legendary` like masterful flurry hunter. We may need proficiency options to be inclusive (master is trained), or convert all proficiency predication to rank gt predication. I'm not sure what's the right call here but the former would need a migration.
3) Data Entry says some of the math that assumes limit 4 (probably some min() functions) will break down, but I don't know any examples at this time.

> Yeah! Happy to split things out if it would help; would a PR for just centralizing proficiency rank information be a preferred starting point if mythic as a "standard-ish" proficiency isn't an outright-undesired outcome? 

It would yeah. Starting there would contribute towards the goal of having proficiency ranks be a bit more invisible. I don't know if we can cleanly or easily get to the final goal, but it would at least make follow up PRs look simpler.

> > Is there a goal for the isMythic getter? I'm not opposed, but to make that correct you'll need to alternatively also check for the mythic trait, since that's how npcs are made mythic.
> 
> This was just moved from another spot in the code so that it would be more easily available for later checks (and to help with consistency for things like characters and NPCs having those different rules for seeing if they're mythic already); 

I was a bit hesistant since we've been trying to avoid expanding the public interface as it often leads to maintenance burdens, but after having some time to think about it, I'm good for it. Make it is own PR, make sure the logic works for both characters and NPCs, and I'm good for it. 

> I am also curious about your thoughts on spending mythic points, though maybe this isn't the place to discuss that

Not in detail, but I can give a quick summary. We have [Use] buttons show up for self effects and frequencies. We can possibly add the ability to designate one or more resources as a cost for an ability that would also spawn the [Use] button. We would need spell.system.cast.focusPoints to interface with it or be migrated towards it though.

Issue #18089, opened 2025-01-23
Containers holding item objects with weight less than 1 Bulk are not calculating weight correctly


MatthewDF:
Items of a weight less than Bulk 1 are not behaving as they should within (certain?) containers on version 6.8.4

If you create a container and place it on a PC sheet, and then add lighter items, you'll notice odd behavior with the weigh the weight is counted where 10 L = 1 L (so 100 Light objects = Bulk 1)

When moving those items out of the container, the weight should be reflected normally on the PC's sheet

CarlosFdez:
I'll need a picture to verify. Are you sure you're not getting confused by backpacks reducing bulk, or spacious sack / bags of holding always being 1 bulk no matter what's inside?

MatthewDF:
Hey Carlos,

Image below. You'll notice this is on a Vehicle sheet, but recreatable on a PC sheet as well. To the best of my knowledge, there is nothing funny happening with the 'Gas Tank' container, including no Rules.

![Image](https://github.com/user-attachments/assets/739bd03e-0fd8-4601-b8f5-332d8c721a2d)

CarlosFdez:
That looks correct, other than total bulk not working. 22L is 2 bulk and 2 L. Total bulk should be 2 bulk, 4L.

MatthewDF:
I'll clarify that the behavior appears different between Vehicles and PC sheets now that I look again. See pictured here the strange behavior I described with 10 Light = 1 Light carried

![Image](https://github.com/user-attachments/assets/17423270-5bd9-4bbd-a23f-db47d8040fd1)

TMunsie:
The vehicle is large, the items you are placing in it are medium. It is performing the correct bulk conversion.

MatthewDF:
However, the Container is Medium. And what about the Medium PC? Should 10 L = 1 L in container in that instance?

CarlosFdez:
You omitted screenshotting the total bulk in the bottom. I suspect vehicle is showing total bulk and pc is showing unit bulk.

MatthewDF:
Here's the full PC Sheet: 

![Image](https://github.com/user-attachments/assets/908c4e00-0260-4dc5-8a5c-7cd7913fc871)

MatthewDF:
Here are several images showing bulk behavior on the vehicle that appear odd. All objects are Medium.

A weapon (2 bulk) and 23 L of Gas, but the total weight is 2 Bulk:
[...]

PR #18099, opened 2025-01-24
Fix stacks of items being bulk adjusted on vehicle actors

CarlosFdez:
This check also exists in PhysicalItemPF2e#bulk, but was inconsistent in its application. Closes https://github.com/foundryvtt/pf2e/issues/18089.

### DIFF
--- a/src/module/actor/inventory/bulk.ts
+++ b/src/module/actor/inventory/bulk.ts
@@ -1,7 +1,6 @@
 import { ActorPF2e } from "@actor";
 import { PhysicalItemPF2e } from "@item";
 import { Bulk } from "@item/physical/bulk.ts";
-import { Size } from "@module/data.ts";
 import { groupBy } from "@util";
 
 export class InventoryBulk {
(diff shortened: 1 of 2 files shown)


src/module/actor/inventory/bulk.ts
```diff
@@ -80,7 +80,12 @@ export class InventoryBulk {
...
     }
 
-    static computeTotalBulk(items: PhysicalItemPF2e[], actorSize: Size): Bulk {
+    static computeTotalBulk(items: PhysicalItemPF2e[], actorOrSize: ActorPF2e | Size | null): Bulk {
+        const [actor, actorSize]: [ActorPF2e | null, Size] = !actorOrSize
+            ? [null, "med"]
+            : typeof actorOrSize === "string"
+              ? [null, actorOrSize]
+              : [actorOrSize, actorOrSize.size];
         items = this.#flattenNonStowing(items);
```
stwlam, line 89: imo just pass the actor and an an array of items

CarlosFdez: Done

PR #18122, opened 2025-01-26
Update Crunch to use DamageAlteration

MrVauxs:


MrVauxs:
Turns out this doesn't solve the issue that spurred this PR (interaction between Animal Instinct (Wolf) and the Crunch feat overriding each other), as DamageAlteration just upgrades to a higher number regardless of whether the existing one is above or below.

TikaelSol:
I'm not entirely sure what the initial issue was exactly, but if it was having two jaws strikes then the likely fix is using tags to make sure the strikes are siloed away from each other. Still, the rules are also unclear there, like, can you even have two different jaws strikes?

Regardless, we should be moving to DamageAltrations.

stwlam:
This is mistaken: as the "weapon" itself is basing changed, it should use an item alteration rather than a damage alteration.

PR #18126, opened 2025-01-27
Extend AELikes to work on the parent item

Cerapter:
As the title says, extends the AELike Rule Element with an additional field called `target`. Defaults to `"actor"`, where it behaves just as it does now, but if set to `"item"`, it'll instead look for and modify the path on the RE's parent item.

Gonna be honest, this one is purely self-serving. I just wanna place flags on items without needing to use a macro.

### DIFF
--- a/src/module/rules/rule-element/ae-like.ts
+++ b/src/module/rules/rule-element/ae-like.ts
@@ -32,6 +32,12 @@ class AELikeRuleElement<TSchema extends AELikeSchema> extends RuleElementPF2e<TS
                 choices: fu.deepClone(this.PHASES),
                 initial: "applyAEs",
             }),
+            target: new fields.StringField({
+                required: false,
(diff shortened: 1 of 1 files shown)


CarlosFdez:
If I remember correctly, ItemAlteration was created specifically to avoid AELikes targetting items.

stwlam:
We're definitely trying to move _away_ from AE-likes. What problem does this attempt to solve? There is no corresponding issue, and the PR has no context.

Cerapter:
There is no corresponding issue because there is no problem it's trying to solve (as far as I'm aware), flags on items would just be really useful for homebrewing stuff.
However, I can vibe with doing it with ItemAlteration, and there *are* a few (actually useful to the project) things I wanna do with ItemAlteration, so I'll make a PR for that.

PR #18134, opened 2025-01-28
Adds default settlement data to "add-settlement" handler in kingdom sheet

MaximMukhanov:
Prevents kingdom sheet crash on adding settlement due to the lack of settlement data in "add-settlement" handler.

https://github.com/foundryvtt/foundryvtt-premium-content/issues/898 - issue in foundry premium modules repo that explains the problem.

### DIFF
--- a/src/module/actor/party/kingdom/sheet.ts
+++ b/src/module/actor/party/kingdom/sheet.ts
@@ -427,7 +427,28 @@ class KingdomSheetPF2e extends ActorSheetPF2e<PartyPF2e> {
             const id = fu.randomID();
             this.#editingSettlements[id] = true;
             this.focusElement = `[name="settlements.${id}.name"]`;
-            this.kingdom.update({ [`settlements.${id}`]: {} });
+            this.kingdom.update({
+                [`settlements.${id}`]: {
+                    name: "",
+                    type: "village",
+                    level: 1,
+                    overcrowded: false,
+                    description: "",
+                    sort: 0,
+                    consumption: {
+                        base: 1,
+                        reduction: 0,
+                        total: 0,
+                    },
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I'll dig into this soon and see if we can't solve this by using default values, but if I can't get something clean I'll simply merge this in. Thank you for looking into it.

MaximMukhanov:
No problem, thank you for working on system!

CarlosFdez:
From my initial investigation, this looks to be a bug with initial values of schema fields used in RecordField. Putting the schema field top level initializes all default values correctly, but it breaks apart entirely in the RecordField.

MaximMukhanov:
> From my initial investigation, this looks to be a bug with initial values of schema fields used in RecordField. Putting the schema field top level initializes all default values correctly, but it breaks apart entirely in the RecordField.
> 
> EDIT: this may be wrong, it may be data prep related
Also dug a bit on my lunch break today:D

It seems that "prepareBaseData" method incorrectly handles lack of settlement data provided by the handler. 
```ts
for (const settlement of settlements) {
            if (!settlement) continue; // this line seems odd
            const typeData = KINGDOM_SETTLEMENT_TYPE_DATA[settlement.type];
            settlement.consumption.base = typeData.consumption;
            settlement.consumption.total = Math.max(0, typeData.consumption - settlement.consumption.reduction);
        }
```

CarlosFdez:
stwlam did a fix last night, is this still an issue?

MaximMukhanov:
> I tested and the bug seems to be resolved now. Thanks a lot for attempting to fix this, and for bringing this bug to our attention.

Thank you for your work:D

Issue #17034, opened 2024-10-28
ItemAlteration missing certain target options like distance


7H3LaughingMan:
This is the same problem as #16698 where if you try to predicate an item alteration off of distance from the target it doesn't work. Trying to automate some stuff in regards to Monastic Archer Stance where monk feats/abilities are applied to ranged attacks made against a target within half of the first range increment.

```
{
  "itemId": "{item|_id}",
  "key": "ItemAlteration",
  "mode": "add",
  "predicate": [
    {
      "lte": [
        "target:distance",
        50
      ]
    }
  ],
  "property": "traits",
  "value": "magical"
}
```

CarlosFdez:
I could bless this as a feature request if you frame it as a request for more item alteration types and what you want them for. "Add the messing alterations" isn't really doable.

7H3LaughingMan:
> I could bless this as a feature request if you frame it as a request for specific item alteration types and what you want them for. "Add the messing alterations" isn't really doable.

This isn't regarding a missing alteration, it's regarding the fact that when you use item alteration that certain predicates related to the target are missing. In this instance it's for automating stuff like Monastic Archer Stance, while you are in this stance things that apply to melee attacks such as Stunning Blow also apply to ranged attacks if the target is within half your weapon's range increment.

CarlosFdez:
Ok I misread. What you want is for target roll options to be available in the contextual clone for item alterations?

7H3LaughingMan:
I believe that is correct? The idea is to be able to make alterations to an item based on your distance to a target.

PR #18145, opened 2025-01-29
Fix One-Inch Punch feat

### LINKED ISSUE #18106: Monastic Archer Stance in Combination with One-Inch Punch, additional Dice are not added on the Bow Attack
Additional dice from the One-Inch Punch are not taken into account for a bow. 

Image

ommad:
Fixes #18106
Changed the predicates in the "One-Inch Punch" feats to include the situation where the actor is using a shortbow, longbow or bow with the monk trait and has the Monastic Archer Stance active.


7H3LaughingMan:
If you are using One-Inch Punch with Monastic Archer Stance it only triggers when the target is within half of the first range increment (normally 50 feet for a longbow and 30 feet for a shortbow). This will actually make it trigger no matter the distance the target actually is.

ommad:
Ah shoot, you are totally correct. I'll keep an eye on [#17034](https://github.com/foundryvtt/pf2e/issues/17034) and update this if it gets resolved.

ommad:
Closing this for now as it's not achievable yet.

PR #18147, opened 2025-01-29
Update Alicorn Lance legacy item to match aon entry

### LINKED ISSUE #18036: Alicorn Lance weird text
The Alicorn is a legacy item that has a variant version that has this line:
> deal evil damage instead of good damage
When updated to remaster, the previous line got changed to this:
> deal spirit damage instead of spirit damage
Which is clearly nonsense, and should either have a note added to it, or just have that part removed.

ommad:
Fixes #18036
Updated the text to reflect the matching entry found here: https://2e.aonprd.com/Equipment.aspx?ID=1882


ammalagonc:
Hey there - the issue here is the item in question needs errata. 

The system now operates under Remaster rules, meaning there are no instances of alignment damage. The change of Alignment damage to spirit damage was done per Paizo's remaster guidance FAQ. 

Sadly the rest of the item text/abilities is janky until Paizo releases clarification/errata. That won't happen until the Treasure Vault remaster is released this June.

TikaelSol:
closing as this is an item that will be fixed in the TV remaster release later this year.

PR #18189, opened 2025-02-03
Adjust aura for Warsworn

SpartanCPA:


packs/pathfinder-bestiary-2/ravener.json
```diff
@@ -675,13 +675,17 @@
...
                 "rules": [
                     {
-                        "key": "Aura",
-                        "radius": 90,
-                        "slug": "cowering-fear",
-                        "traits": [
-                            "emotion",
-                            "fear",
-                            "mental"
+                        "itemType": "action",
```
simonward, line 678: Does the frightful presence ability need to be added?

Cuingamehtar:
While you are changing Ravener, could you fix  `damage} damage` in its Breath Weapon, please?

packs/pathfinder-bestiary-2/ravener.json
SpartanCPA: The creature already has frightful presence. This just adds Cowering Fear to the description of that feature, instead of drawing another circle for it.

image

SpartanCPA:
> While you are changing Ravener, could you fix `damage} damage` in its Breath Weapon, please?

Sure, easy enough fix.

PR #18191, opened 2025-02-03
added splash damage as resistance source

### LINKED ISSUE #17157: Missing Resistance (Splash Damage)
It's a simple issue: Splash is missing from the available resistance list. It's available as a weakness, and items such as the backfire mantles rely on setting up that resistance.

Kulkodar:
Added splash damage as a resistance source.

Manual tests done:
- Checked if untyped splash damage is reduced
- Checked if typed splash damage is reduced
- Checked that resistance against typed splash is correct if there is a resistance for both
- Checked double resistance
- ~~Checked exception~~ still working on that

Closes #17157

### DIFF
--- a/src/module/system/damage/iwr.ts
+++ b/src/module/system/damage/iwr.ts
@@ -178,6 +178,19 @@ function applyIWR(actor: ActorPF2e, roll: Rolled<DamageRoll>, rollOptions: Set<s
             );
             const applicableResistances = workingResistanceData.filter((r) => r.applicable);
             const criticalResistance = resistances.find((r) => r.type === "critical-hits");
+
+            const splashResistance = splashDamage
+                ? (resistances.find((r) => r.type === "splash-damage") ?? null)
+                : null;
+            if (splashResistance) {
+                applicableResistances.push(
+                    new WorkingResistanceData(splashResistance, {
+                        value: splashResistance.getDoubledValue(formalDescription),
+                        ignored: ignoredResistances.some((ir) => ir.test(formalDescription)),
+                    }),
+                );
+            }
+
             if (criticalResistance && isCriticalSuccess) {
                 const maxResistable = instanceTotal - critImmuneTotal;
                 if (maxResistable > 0) {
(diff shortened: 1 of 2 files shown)


Kulkodar:
I am not sure if the combined splash damage should be able to reduced by more than the splash damage e.g. if i have this damage instance "8 fire + 4 splash" and 5 resistance to splash damage does the damage reduce by 4 or 5?

Should i do the data entry fixes mentioned in the issues with this PR or open another one?

stwlam:
This appears to not add splash resistance but instead remove splash weakness. It also treats splash damage as a separate instance, which is incorrect.

Kulkodar:
As far as i can see it should work.
made with the latest commit of this pr:
https://github.com/user-attachments/assets/81bda048-dbae-40ba-9178-341a3ed3eb38

Kulkodar:
The predicate for the damage with splash check for the `damage:component:splash` statement which should now be added to the `formalDescription` of the damage instance if it is a splash.

stwlam:
Roll the following: `/r (1 + 1[splash])[fire]` and apply to an actor with a splash resistance of 1.

Kulkodar:
I think if there is something i should see i am too stupid to see it.
https://github.com/user-attachments/assets/4d8e933e-9fb4-42ae-a906-6b6ef009e2a6

src/module/system/damage/roll.ts
```diff
@@ -485,6 +488,7 @@ class DamageInstance extends AbstractDamageRoll {
                 `damage:type:${this.type}`,
                 typeCategory ? `damage:category:${typeCategory}` : [],
                 this.persistent ? "damage:category:persistent" : [],
+                this.splash ? "damage:component:splash" : [],
```
stwlam, line 491: This is marking the entire instance as splash damage.

Kulkodar: Just some question so i understand: 
Is splash damage it's own damage instance?
ksnip_20250219-210642 so that should mean in this example are 4 separate damage instances?

English is not my mother tongue so i might interpret the splash trait wrong; i interpret the splash trait as merging the damage and the splash together as one instance.

PR #18194, opened 2025-02-03
Update `gnoll` trait to `kholo` trait

### LINKED ISSUE #18193: Kholo Weapon Familiarity ancestry feat doesn't work
Foundry version **12.331**
PF2e version **6.8.5**

I have a Kholo with the Kholo Weapon Familiarity ancestry feat and he is using a Spirit Thresher (traits: Uncommon, Kholo, Sweep, Versatile S)

Image

In proficiencies you can see that he is trained in advanced Kholo weapons.

Image

But the attack box does not compute the proficiency value and reports that the weapon is untrained.

Image

The expected behavior would be to treat the weapon that has the Kholo and advanced traits as a martial weapon.

ammalagonc:
Closes #18193

This might require a migration - someone teach me how to do so and I will TCO.


TikaelSol:
We should just migrate instead

PR #18200, opened 2025-02-04
Add spell effect for Scintillating Safeguard

pedrogrullada:


TikaelSol:
Should also add bleed as a choice

pedrogrullada:
> Should also add bleed as a choice

The spell doesn't work on persistent damage

PR #18207, opened 2025-02-05
Improve the interaction between Lingering Composition and Courageous Anthem

ommad:
The "Lingering Composition" _spell_ Performance skill check now carries a new traits. This one is used to trigger a Note specific to the result of said Performance check. In case of Success or Critical success,  a new _spell-effect_ is provided. Once applied to the bard, it will use an ItemAlteration targeting the _spell_ "Courageous Anthem" so that this one now provide a new "Courageous Anthem" _spell-effect_ with an updated duration (3 for a success and 4 for a critical success).

The Performance skill check DC is also defaulted to Level DC based on the bard level. This is not exactly RAW. In theory the spell DC is based on the highest level character targeted by the bard. I could not find a way to use the target level and this seems like a correct approximation, assuming most character in a party will have a similar level.

### DIFF
--- a/packs/spell-effects/spell-effect-lingering-composition-critical-success.json
+++ b/packs/spell-effects/spell-effect-lingering-composition-critical-success.json
@@ -16,9 +16,9 @@
             "value": 1
         },
         "publication": {
-            "license": "OGL",
-            "remaster": false,
-            "title": ""
+            "license": "ORC",
+            "remaster": true,
+            "title": "Pathfinder Player Core"
         },
         "rules": [
             {
--- a/packs/spell-effects/spell-effect-lingering-composition-critical-success.json
+++ b/packs/spell-effects/spell-effect-lingering-composition-critical-success.json
@@ -1,0 +1,62 @@
+{
+    "_id": "pGWlbo549LtsXGg4",
+    "img": "systems/pf2e/icons/spells/lingering-composition.webp",
+    "name": "Spell Effect: Lingering Composition (Critical Success)",
+    "system": {
+        "description": {
+            "value": "<p>The next composition lasts 4 rounds.</p>"
+        },
+        "duration": {
+            "expiry": "turn-end",
+            "sustained": false,
+            "unit": "rounds",
+            "value": 1
+        },
+        "level": {
+            "value": 1
+        },
+        "publication": {
+            "license": "OGL",
+            "remaster": false,
+            "title": ""
+        },
+        "rules": [
+            {
+                "itemType": "spell",
+                "key": "ItemAlteration",
+                "mode": "override",
+                "predicate": [
+                    "item:slug:courageous-anthem"
+                ],
+                "property": "description",
+                "value": [
+                    {
+                        "text": "60-foot emanation",
+                        "title": "Area"
+                    },
+                    {
+                        "text": "4 round",
+                        "title": "Duration"
+                    },
+                    {
+                        "text": "You inspire yourself and your allies with words or tunes of encouragement. You and all allies in the area gain a +1 status bonus to attack rolls, damage rolls, and saves against fear effects."
+                    },
+                    {
+                        "text": "@UUID[Compendium.pf2e.spell-effects.Item.zot8JuAN0HXfWo3j]{Spell Effect: Courageous Anthem}"
+                    }
+                ]
+            }
+        ],
+        "start": {
+            "initiative": null,
+            "value": 0
+        },
+        "tokenIcon": {
+            "show": true
+        },
+        "traits": {
+            "value": []
+        }
+    },
+    "type": "effect"
+}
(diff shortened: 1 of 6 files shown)


packs/spells/lingering-composition.json
```diff
@@ -11,7 +11,7 @@
...
+            "value": "<p>You add a flourish to your composition to extend its benefits. If your next action is to cast a cantrip composition with a duration of 1 round, attempt a @Check[performance|options:spell-lingering-composition|dc:@self.level] check. The DC is usually a standard-difficulty DC of a level equal to the highest-level target of your composition, but the GM can assign a different DC based on the circumstances. The effect depends on the result of your check.</p><hr /><p><strong>Critical Success</strong> The composition lasts 4 rounds.</p>\n<p><strong>Success</strong> The composition lasts 3 rounds.</p>\n<p><strong>Failure</strong> The composition lasts 1 round, but you don't spend the Focus Point for casting this spell.</p>"
```
pedrogrullada, line 14: Your own level won't necessary be the highest level among targets.

packs/spells/lingering-composition.json
```diff
@@ -29,7 +29,63 @@
...
+                    "spell-lingering-composition",
+                    {
+                        "lt": [
+                            "check:total:delta",
+                            0
+                        ]
+                    }
+                ],
+                "selector": "performance-check",
+                "text": "The composition lasts 1 round, but you don't spend the Focus Point for casting this spell.",
```
pedrogrullada, line 45: Text needs to be localized.

packs/spell-effects/spell-effect-lingering-composition-critical-success.json
```diff
@@ -0,0 +1,62 @@
...
+            "value": 1
+        },
+        "level": {
+            "value": 1
+        },
+        "publication": {
+            "license": "OGL",
+            "remaster": false,
+            "title": ""
+        },
```
pedrogrullada, line 22:
```suggestion
        "publication": {
            "license": "ORC",
            "remaster": true,
            "title": "Pathfinder Player Core"
        },
```

packs/spell-effects/spell-effect-lingering-composition-critical-success.json
```diff
@@ -0,0 +1,62 @@
...
+                    },
+                    {
+                        "text": "You inspire yourself and your allies with words or tunes of encouragement. You and all allies in the area gain a +1 status bonus to attack rolls, damage rolls, and saves against fear effects."
+                    },
+                    {
+                        "text": "@UUID[Compendium.pf2e.spell-effects.Item.Spell Effect: Courageous Anthem (4 rounds)]"
+                    }
+                ]
+            }
+        ],
```
pedrogrullada, line 49: I don't see how Courageous Anthem needs its area and duration restated in this Item Alteration. It'd probably be better to link the effect using `add` rather than `override`. Not personally a fan of this multi-effect approach and would rather wait until it can be done better. Regardless, whatever text ends up being used needs to be localized.

packs/spells/lingering-composition.json
ommad: I tried to find a way around this but couldn't... From what I have read in the wiki, it is not possible to use Target data for now. I could revert this and just display in a single note the results for Failure, Success and Critical success. What do you think?

packs/spells/lingering-composition.json
ommad: Yes, I noticed this wasn't in place yet. Is there an example somewhere on how to set that up?

packs/spell-effects/spell-effect-lingering-composition-critical-success.json
pedrogrullada: With multiple durations, multiple effects is indeed the only way to go about it for now. I would personally hold off on doing any of this until there duration can be altered via Item Alteration.

I don't ultimately decide if this is merged though, so for now I would just suggest using `add` instead for the description alteration.

packs/spells/lingering-composition.json
pedrogrullada: I see you got help in the discord server, but yeah, you just need to add some entries to `re-en.json`, trying to pick some coherent paths for the things you're adding.

packs/spells/lingering-composition.json
pedrogrullada: A single Note would work for this approach, yeah.

packs/spells/lingering-composition.json
ommad: Done, it should all be localized or using {item|name}

packs/spells/lingering-composition.json
ommad: Done the single note when the roll is done.

TikaelSol:
I really don't want to add extra effects. I'd rather we wait and do it properly than have to clean up the mess later when we can.

PR #18221, opened 2025-02-07
Bring "Anima Invocation" spell effect in line with Remaster changes

### LINKED ISSUE #18195: [Request] Add Holy trait to the bonus Spirit Damage from Spell Effect: Anima Invocation
Originally, the _Anima Invocation (Modified)_ ritual added Good damage, with the specific enemy it was designed to deal with being weak to Good. In the remastering, the Good damage was changed to Spirit damage as expected, and the specific enemy was changed to being weak to Holy. This means the spell effect no longer triggers the weakness.

To best match the original use, the spell effect should be triggering that enemy's weakness to Holy.

There's also some elements that grants bonuses vs evil but they seem to be partially working?

ommad:
Added an Adjust-Strike weapon-trait holy.
Changed the save modifier rule elements predicates to unholy instead of evil.
Closes #18195

### DIFF
--- a/packs/spell-effects/spell-effect-anima-invocation.json
+++ b/packs/spell-effects/spell-effect-anima-invocation.json
@@ -45,10 +45,16 @@
                 "value": "ghost-touch"
             },
             {
+                "key": "AdjustStrike",
+                "mode": "add",
+                "property": "weapon-traits",
+                "value": "holy"
+            },
+            {
                 "key": "FlatModifier",
                 "predicate": [
                     "anima-invocation:critical-success",
-                    "origin:trait:evil"
+                    "origin:trait:unholy"
                 ],
                 "selector": "saving-throw",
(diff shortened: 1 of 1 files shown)


packs/spell-effects/spell-effect-anima-invocation.json
```diff
@@ -44,11 +44,17 @@
                 "property": "property-runes",
                 "value": "ghost-touch"
             },
+            {
+                "key": "AdjustStrike",
+                "mode": "add",
+                "property": "weapon-traits",
+                "value": "holy"
+            },
             {
```
TikaelSol, line 53:
```suggestion
            },
            {
```
The effect doesn't state this happens, so it shouldn't be added by us. We don't want to editorialize.

ommad:
Right, that makes sense, I'll close this PR. I'll comment the issue explaining the same.

PR #18224, opened 2025-02-07
Convert DamageDice damage type changes to DamageAlteration

SpartanCPA:


packs/equipment/ooze-ammunition-greater.json
```diff
@@ -38,11 +38,13 @@
...
-                    "damageType": "acid"
-                },
-                "selector": "{item|_id}-damage"
+                "key": "DamageAlteration",
+                "mode": "override",
+                "property": "damage-type",
+                "selectors": [
+                    "{item|_id}-damage"
+                ],
+                "value": "acid"
```
pedrogrullada, line 47: I'd echo the Item Alteration suggestion for ammunitions as well, so it only affects the base weapon damage as opposed to all damage from the Strike

packs/feats/agile-shield-grip.json
```diff
@@ -32,17 +32,17 @@
                 "toggleable": true
             },
             {
-                "key": "DamageDice",
-                "override": {
-                    "dieSize": "d4"
-                },
+                "key": "DamageAlteration",
+                "mode": "downgrade",
```
pedrogrullada, line 36: This rule element would downgrade all dice from a Shield Boss and Shield Spikes damage. Item Alteration would be better here as well.

pedrogrullada (commented):
For most Damage Alteration upgrades, downgrades, or damage type overrides, it'll usually be better to use Item Alteration so it only affects the base damage type. Damage Alteration can do this as well by using the `dice:slug:base` predicate.

PR #18240, opened 2025-02-08
Fix Slinger's Precision using description from Singular Expertise

Randommisha13:
I don't have a PDF for Guns & Gears Remastered, so I took the text from Demiplane.


ammalagonc:
I checked and text matches with what's on the PDF, so this one is good to go.

PR #18253, opened 2025-02-11
Apply `Token Mark` to Hunt Prey action and associated feats and effects

ammalagonc:
And one instance of lint.

### DIFF
--- a/packs/actions/hunt-prey.json
+++ b/packs/actions/hunt-prey.json
@@ -20,6 +20,13 @@
         },
         "rules": [
             {
+                "domain": "all",
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.HuntPrey.TargetHuntedPrey",
+                "option": "target:mark:hunted-prey",
+                "toggleable": true
+            },
+            {
                 "key": "FlatModifier",
                 "label": "PF2E.SpecificRule.HuntPrey.SeekHuntedPrey",
                 "predicate": [
--- a/packs/actions/hunt-prey.json
+++ b/packs/actions/hunt-prey.json
@@ -20,18 +20,11 @@
         },
         "rules": [
             {
-                "domain": "all",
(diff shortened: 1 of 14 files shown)


packs/feat-effects/effect-hunt-prey.json
```diff
@@ -0,0 +1,48 @@
...
+            "unit": "unlimited",
+            "value": -1
+        },
+        "level": {
+            "value": 1
+        },
+        "publication": {
+            "license": "OGL",
+            "remaster": false,
+            "title": ""
```
TikaelSol, line 21: Publication data

packs/actions/hunt-prey.json
```diff
@@ -19,19 +19,12 @@
             "title": "Pathfinder Player Core"
         },
         "rules": [
-            {
-                "domain": "all",
-                "key": "RollOption",
-                "label": "PF2E.SpecificRule.HuntPrey.TargetHuntedPrey",
-                "option": "hunted-prey",
```
TikaelSol, line 26: We may consider leaving this and making the option `target:mark:hunted-prey` instead, as a fallback for flurry / exploration mode.

packs/feat-effects/effect-hunters-edge-flurry.json
```diff
@@ -21,19 +21,23 @@
             "title": "Pathfinder Core Rulebook"
         },
         "rules": [
+            {
+                "key": "TokenMark",
+                "slug": "hunted-prey"
+            },
```
TikaelSol, line 27: This won't work for the MAP rule.

ammalagonc: It worked when I tested it out on my end. But looking at it now, it is redundant since the Token Mark isn't doing anything. Let me  revert.

PR #18269, opened 2025-02-13
Implement Slinger's Precision proficiency increase for some combination weapons

### LINKED ISSUE #18244: Slinger's Precision's proficiency with combination firearm's melee usage is not implemented
Getting ahead of the inevitable issue. This can't be done yet as rolls for the melee usage of a combination weapon lack any context for what its ranged usage is.

Randommisha13:
An attempt of implementing 3rd paragraph of Slinger's Precision.
Should resolve #18244

I don't know much about TypeScript and coding, so this probably contains some programminghorror material.

If I understand what I've done, this shoud add `item:rangedGroup` property to the `meleeUsage` of combination weapons, then if the character has **Slinger's Precision** feature it will bump proficiency for weapons that have `item:rangedGroup:firearm` or `item:rangedGroup:crossbow`.

### Known Issues:
- `rangedGroup` is manually assigned and not related to the actual weapon group of ranged usage
- ~~`rangedGroup` can't be assigned from GUI, only through editing packs directly~~
- `rangedGroup` assignment from GUI is not restricted to ranged weapon groups
- ~~Does not work with specific magic weapons, only base variants~~

### DIFF
--- a/packs/classfeatures/gunslinger-weapon-mastery.json
+++ b/packs/classfeatures/gunslinger-weapon-mastery.json
@@ -53,6 +53,24 @@
                         ]
                     }
                 ]
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "upgrade",
+                "path": "system.proficiencies.attacks.slingers-precision-combination-advanced.rank",
+                "value": 2
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "upgrade",
+                "path": "system.proficiencies.attacks.slingers-precision-combination-martial.rank",
+                "value": 3
+            },
+            {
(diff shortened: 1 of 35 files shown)


shemetz:
Storing the `rangedGroup` field inside the `meleeUsage` object is kinda smelly :(

Randommisha13:
Yes, it is bad, but it is the best solution I could get to work. It would be better to use existing data, but I don't know how.

pedrogrullada:
I believe this PR is superseded by #18301 and #18304, although we are still discussing some details on the exact implementation of Slinger's Precision, which will be coming soon. In any case, thank you for taking the time to submit this.

PR #18275, opened 2025-02-14
Add `preDelete` callback and add the `grantOnDelete` option to control when the granted item is created to `GrantItemRuleElement`

In3luki:
This adds a new option for `GrantItemRuleElement`s called `grantOnDelete` that controls whether the granted item should be created when the parent item is deleted instead.

Example rule element for `Effect: Rage` that creates the `Effect: Rage Temporary Hit Points Immunity` when the rage effect expires or is deleted:
```json
{
  "key": "GrantItem",
  "uuid": "Compendium.pf2e.feat-effects.Item.MaepXeSHiMoWDm7u",
  "allowDuplicate": false,
  "grantOnDelete": true
}
```


In3luki:
I've removed the `onExpire` callback as the derived nature of effect expiry makes it difficult to track if the item should be created.
With the setting to automatically remove expired effects it should still work for the use case.

PR #18276, opened 2025-02-14
Fix spell dc data of Dewey Daystar

RakuJa:
In all the other entities, in the spell dc fields (value, dc, mod) the values are written as int. This is the only case in which they are written as string.


TikaelSol:
I'm sure there's a few weird bits in the OoA actors. They were partly done with an experimental script to make the actors wholly from the PDF then having me double check the output, but I think you can see why we decided against doing that again.

PR #18293, opened 2025-02-16
Fix malformed Promises in `processPreUpdateActorHooks` helper

### LINKED ISSUE #18250: Using `"inMemoryOnly": true` in a `GrantItem` that grants a condition which also make use of `inMemoryOnly` will silently breaks the character sheet
I dunno to what extent this breaks a character or if it is just the sheet, but adding this RE on an effect will prevent the character sheet from being closed if it is opened while the effect is on the character.

The effect grants the condition `Paralyzed`, i have narrowed it down to ` "inMemoryOnly": true` being the culprit and it happens for other conditions (e.g. `Restrained`) that also grant themselves other conditions using `inMemoryOnly`.

There is no error in the console and once the sheet is opened, it cannot be closed even if you remove the effect after the fact. Also, this issue doesn't happen to NPC sheets.

```json
{
  "key": "GrantItem",
  "uuid": "Compendium.pf2e.conditionitems.Item.6uEgoh53GbXuHpTF",
  "onDeleteActions": {
    "grantee": "restrict"
  },
  "inMemoryOnly": true
}
```

In3luki:
The `new Promise(() => ({ create: [], delete: [] }))` resolved to `undefined` and that somehow prevented any new actor updates without throwing an error.

https://github.com/user-attachments/assets/46a3752c-a1e0-4b29-aa3a-babadef61087

https://github.com/user-attachments/assets/56b1522c-0e37-4199-90b4-32bf00a4eea0

Fixes #18250

### DIFF
--- a/src/module/rules/helpers.ts
+++ b/src/module/rules/helpers.ts
@@ -196,7 +196,7 @@ async function processPreUpdateActorHooks(
         await Promise.all(
             rules.map(
                 (r): Promise<{ create: ItemSourcePF2e[]; delete: string[] }> =>
-                    actor.items.has(r.item.id) ? r.preUpdateActor() : new Promise(() => ({ create: [], delete: [] })),
+                    actor.items.has(r.item.id) ? r.preUpdateActor() : Promise.resolve({ create: [], delete: [] }),
             ),
         )
     ).reduce(


stwlam:
Sorry for the misunderstanding!

PR #18301, opened 2025-02-17
Add some item roll options of original weapon usage to alternate usage clones

stwlam:


src/module/item/weapon/document.ts
```diff
@@ -502,6 +502,13 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
         const clone = super.clone(data, context);
         if (context?.altUsage && clone instanceof WeaponPF2e) {
             clone.altUsageType = context.altUsage;
+            const priorUsage = { thrown: "melee", melee: "ranged" }[context.altUsage];
```
CarlosFdez, line 505: would it be possible to store in a variable the current usage rather than inferring the opposite usage of the alt usage, or is that not reliable?

stwlam: It's reliable to infer the opposite usage of the alt usage.

CarlosFdez: It is *currently* reliable to infer the opposite usage, but if alt usage evolves in a year or two this can sneak in as a bug.

What I asked is if it was reliable to check the current usage. If it is, that sounds like that'd be better to me. But if its not, currently reliable is better than not reliable at all.

PR #18347, opened 2025-02-24
Update automation for Favored Terrain and its Unimpeded Journey upgrades

ammalagonc:
The idea is to move the effect bonuses to Unimpeded Journey, which would be predicated on the terrain selection made in the Favored Terrain feat.

Please review and provide feedback when you get a moment.


dotadd:
I think we had the same idea :D 

https://github.com/foundryvtt/pf2e/pull/18084
Didn't update in forever, sorry about that. Integrating Terrain Master was a bit of a pain. The current version works, but it's a lot. So if y'all end up closing my PR I won't be mad.

TikaelSol:
Not sure this is the right approach either. You could take the feat after getting unimpeded journey, so those rules wouldn't fire off correctly (this would mostly be relevant for retraining since people aren't likely to burn a 12th level feat slot on a level 2 feat).

ammalagonc:
Round 2 - Moved the increases from Unimpeded Journey into an effect.  This should cover the re-firing of the rules. 

For Favored Terrain retraining, would it make sense to leave a note in the feat saying if you need to retrain to delete and re-add it?

TikaelSol:
Nah, that's standard for retraining

ammalagonc:
Closing - this one requires some more thought through.

PR #18373, opened 2025-02-27
Fix `ItemAttacher` always selecting first option in select element

### LINKED ISSUE #18372: Attach item dialog always returns first item
Repro steps (6.9.0/dev on Foundry 12.331)
1) Create a blank character
2) Add 2 (or more) shields to inventory
3) Add a Shield Boss
4) Attempt to attach the Shield Boss to any shield except the first

The Shield Boss will be attached to the first item in the dialog

bpuk:
Fixes #18372

### DIFF
--- a/src/module/apps/pick-a-thing-prompt.ts
+++ b/src/module/apps/pick-a-thing-prompt.ts
@@ -49,7 +49,7 @@ abstract class PickAThingPrompt<TItem extends ItemPF2e, TThing extends string | 
             htmlClosest(event.target, ".content")?.querySelector<HTMLElement>("tag") ??
             htmlClosest(event.target, "button[data-action=pick]") ??
             htmlClosest(event.target, ".choice")?.querySelector("button[data-action=pick]");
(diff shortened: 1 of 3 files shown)


src/module/apps/pick-a-thing-prompt.ts
```diff
@@ -49,7 +49,7 @@ abstract class PickAThingPrompt<TItem extends ItemPF2e, TThing extends string |
             htmlClosest(event.target, ".content")?.querySelector<HTMLElement>("tag") ??
             htmlClosest(event.target, "button[data-action=pick]") ??
             htmlClosest(event.target, ".choice")?.querySelector("button[data-action=pick]");
-        const selectedIndex = valueElement?.getAttribute("value");
+        const selectedIndex = valueElement?.getAttribute("value") ?? valueElement?.getAttribute("data-choice");
```
stwlam, line 52: This looks like it's picking the wrong `valueElement`: a dataset entry for the button shouldn't be necessary.

bpuk: D'oh. Listener was updating 'data-choice' instead of 'value'

PR #18374, opened 2025-02-28
Automate Terror Bird's Tearing Clutch damage

ammalagonc:


packs/pathfinder-bestiary-3/terror-bird.json
```diff
@@ -155,6 +155,29 @@
...
+                    {
+                        "category": "persistent",
+                        "critical": false,
+                        "damageType": "bleed",
+                        "diceNumber": 1,
+                        "dieSize": "d4",
+                        "key": "DamageDice",
+                        "predicate": [
+                            "check:outcome:critical-success"
+                        ],
```
TikaelSol, line 179: Why not `"critical": true`?

ammalagonc: I didn't realize you could do it that way until you pointed it out :|

Change implemented.

TikaelSol: Flat modifier has something similar in the form

ammalagonc: I see that - 

image
image

I tested with all the options provided, but none remove the additional 1 damage on critical damage/hit.

Thoughts?

PR #18378, opened 2025-02-28
Fix talisman usages

bpuk:
Duplicate traits 'affixed-to-metal-weapon' and 'affixed-to-a-metal-weapon' existed, with both being in use on different items.

Removed affixed-to-metal-weapon and update affected items.
Updated affixed-to-a-melee-weapon to be consistent with the rest of the names.

Checked usages for the rest of the talismans, skipped Dragonbone Arrowhead as that's a unique and long usage


TikaelSol:
If we remove a usage we need to migrate away any in-world instances of that usage. But we aren't running migrations right now, so that will have to wait. Treasure Vault remaster is also coming, and with luck some of the near-identical usages will be condensed then as well.

bpuk:
Understood
Should I adjust this PR to only have the safe (non-removal) updates?
Appreciate some of these will need updating again when the treasure vault remaster comes out

TikaelSol:
Yeah, let's stick to the safe ones and table the others to revisit once we have Treasure Vault and know how that is going to shake out.

bpuk:
Took a couple of extra commits, but that should be the safe ones.

PR #18381, opened 2025-03-02
Add translation to item currency values and parsing

Cuingamehtar:
Instead of using hardcoded currency denominations, output localized names.
Also, when parsing item value, tries to convert it from localized name to English before parsing values

Example output (An item with value of 1 cp, 2 sp, 3 gp, 4 pp)
image
 
On item sheet (input visually truncated, but accepts both localized strings and English strings)
image

### DIFF
--- a/src/module/item/physical/coins.ts
+++ b/src/module/item/physical/coins.ts
@@ -99,7 +99,7 @@ class CoinsPF2e implements Coins {
                 const pattern = new RegExp(`(?!<\\p{L})${localizedDenomination}(?!\\p{L})`, "u");
                 return s.replace(pattern, denomination);
             },
-            String(coinString).trim().replace(/,/g, ""),
+            coinString.trim().replace(/,/g, ""),
         );
         return [...priceTag.matchAll(/(\d+)\s*([pgsc]p)/g)]
             .map((match) => {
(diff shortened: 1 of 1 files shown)


src/module/item/physical/coins.ts
```diff
@@ -91,7 +91,16 @@ class CoinsPF2e implements Coins {
...
-        const priceTag = String(coinString).trim().replace(/,/g, "");
+        const priceTag = DENOMINATIONS.reduce(
+            (s, denomination) => {
+                const localizedDenomination = game.i18n.localize(`PF2E.CurrencyAbbreviations.${denomination}`);
+                if (localizedDenomination === denomination) return s;
+                // This matches localized denomination if it's not followed or preceeded by an unicode letter character
+                const pattern = new RegExp(`(?!<\\p{L})${localizedDenomination}(?!\\p{L})`, "u");
+                return s.replace(pattern, denomination);
+            },
+            String(coinString).trim().replace(/,/g, ""),
```
stwlam, line 102:
```suggestion
            coinString.trim().replace(/,/g, ""),
```
Not sure why that was getting run through `String` before, but let's do without.

Cuingamehtar: Removing `String` seems to be safe, as the existing callers handle the type.
I've tested it and everything looks alright.

PR #18400, opened 2025-03-04
Type update: Allow optional ApplicationWindowConfiguration controls

DFreds:


stwlam:
Really these should all be required, with `DEFAULT_OPTIONS` the constructor options being partials.

DFreds:
> Really these should all be required, with `DEFAULT_OPTIONS` the constructor options being partials.

So, make `ApplicationConfiguration` have `window: Partial<ApplicationWindowConfiguration>`? I can make that change if so

stwlam:
Type updates frozen until V13

PR #18401, opened 2025-03-04
Type update: Add `isSuppressed` property to ClientBaseActiveEffect

DFreds:


stwlam:
Isn't this defined in `ActiveEffect`?

PR #18402, opened 2025-03-04
Type update: Fix user hasPermission method type to use string parameter

DFreds:


types/foundry/common/documents/user.d.ts
```diff
@@ -43,7 +43,7 @@ export default class BaseUser<TCharacter extends BaseActor<null> = BaseActor<nul
      * @param permission The permission name from USER_PERMISSIONS to test
      * @return Does the user have at least this permission
      */
-    hasPermission(permission: UserPermission): boolean;
+    hasPermission(permission: string): boolean;
```
stwlam, line 46: This method isn't defined in the class anymore.

DFreds: Can you explain what you mean by anymore?

I'm seeing in v12 
image

```js
    /**
     * Test whether the User has at least a specific permission
     * @param {string} permission    The permission name from USER_PERMISSIONS to test
     * @return {boolean}             Does the user have at least this permission
     */
    hasPermission(permission) {
      if ( this.isBanned ) return false;

      // CASE 1: The user has the permission set explicitly
      const explicit = this.permissions[permission];
      if (explicit !== undefined) return explicit;

      // CASE 2: Permission defined by the user's role
      const rolePerms = game.permissions[permission];
      return rolePerms ? rolePerms.includes(this.role) : false;
    }
```

CarlosFdez: Did you check v13? I haven't but it might be worth checking.

stwlam: I'm talking about V12

stwlam: Ope, sorry, was looking at `User` and `not `BaseUser`

stwlam, line 46:
```suggestion
    hasPermission(permission: UserPermissionString): boolean;
```

stwlam (changes requested):

stwlam (approved):

PR #18403, opened 2025-03-04
Type update: Fix string slugify definiton

DFreds:


stwlam:
Type updates frozen until V13

PR #18422, opened 2025-03-05
Allow removal of traits and prevent duplicate traits on action usage

nikolaj-a:


CarlosFdez:
I need to try what I'm doing for overrideTraits, once I do I can get back to you on this.

CarlosFdez:
I posted it in the discord but here it is again for good measure:

Me and stwlam talked it over and  decided to try to keep the API roughly the same. In other words, keep traits to adding new traits rather than replacing. Instead override-traits would be supported to replace the traits list with a new one. You don't have to support override-traits=true though (just override-traits=fire,water,cold. Even in @Check, overrideTraits now supports a list of traits as an alternative argument, and data entry should be doing that from now on.

Sorry for the incredible delay on that

PR #18482, opened 2025-03-09
Fix Kholo Weapon Familiarity

MrVauxs:
Also, there seems to be a "kholo" trait description in the translation data. Not sure if that's vestigial or the gnoll trait was actually meant to be renamed to kholo, as opposed to just having a different display value.

https://github.com/foundryvtt/pf2e/blob/376d59d743108cd77a3f44acf761b23c54a94e00/src/scripts/config/traits.ts#L1352


TikaelSol:
We will be renaming it fully to kholo once we can run migrations again.

PR #18485, opened 2025-03-09
Remove remaining rule element-based implementation of Apex items

pedrogrullada:
It looks like these dodged the migration in #9349 since it only looked for `equipment` physical items, and support for others such as `weapon` and `armor` was only added recently.


TikaelSol:
Several of these are intentional. The apex attribute is variable and the item gives a +2 not +1.

pedrogrullada:
> Several of these are intentional. The apex attribute is variable and the item gives a +2 not +1.

Are you talking about pre-remaster apex items increasing an ability score by 2? Because that'd be the equivalent of increasing an attribute by 1 in the remaster.

PR #18487, opened 2025-03-09
Update Bands of Force's Flat Modifier slug

### LINKED ISSUE #18479: Animal Skin and Bands of Force item bonuses do not stack
Animal Skin grants an item bonus to AC. The description says that this item bonus is cumulative with bands of force. Currently only the highest of those 2 bonuses is applied.
From what I understand, the Adjust Modifier RE on Animal Skin expects a modifier with the "bands-of-force" slug.
But the slug on the Flat Modifier of the bands of force is "bracers-of-armor". I presume this is a remnant of the premaster^^

pedrogrullada:
Closes #18479
and remove `bracers-of-armor` slug from Assassin's Bracers, which are seemingly the result of copy pasting rule elements from the Bands of Force.

### DIFF
--- a/packs/equipment/assassins-bracers-type-i.json
+++ b/packs/equipment/assassins-bracers-type-i.json
@@ -41,7 +41,6 @@
                     "ac",
                     "saving-throw"
                 ],
-                "slug": "bracers-of-armor",
                 "type": "item",
                 "value": 1
             },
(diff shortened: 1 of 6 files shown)


TikaelSol:
This may be intentional to handle Animal Skin's automation

pedrogrullada:
> This may be intentional to handle Animal Skin's automation

Animal Skin is looking for the `bands-of-force` slug

PR #18497, opened 2025-03-09
Add Aid degree of success adjustment to Headbands of Translocation

pedrogrullada:


packs/equipment/headbands-of-translocation.json
```diff
@@ -34,7 +34,36 @@
...
+            {
+                "adjustment": {
+                    "criticalFailure": "to-failure"
+                },
+                "key": "AdjustDegreeOfSuccess",
+                "predicate": [
+                    "action:aid",
+                    "headbands-of-translocation"
+                ],
+                "selector": "attack-roll"
```
TikaelSol, line 64: I don't think it is possible in the system to have an aid roll on an attack, since we can't do inlines. Maybe drop the aid on this since it's implicit in the other roll option?

Issue #18287, opened 2025-02-16
Request: Let effects be dragged from player sheet contain item origin when applied to another character


thejoester:
EDITED: The example below is one where when dragging from a character sheet description to another actor does not properly apply the effect with item origin to calculate certain effects like the Chirurgeon Field Discover feature. Below it has been pointed out that the way this should be applied is by sending it to chat and then dragging it out. 

I would suggest that either implement the ability to drag it from the character sheet or at least put in a little note indicating to send to chat and apply effect from chat card as this is not intuitive. 

If I should open a new issue and close this one instead of editing it let me know. 

> So with the [Chirurgeon Field Discovery (5th)](https://2e.aonprd.com/ResearchFields.aspx?ID=6) feature applying this affect to another actor after they consume an item that is tagged infused and healing, they should get temp hit points equal to the Alchemist's INT score. When applied to a character it is granting temp hp based on their own INT not the Alchemist who created the original item.

CrackJackFlood:
> So with the [Chirurgeon Field Discovery (5th)](https://2e.aonprd.com/ResearchFields.aspx?ID=6) feature applying this affect to another actor after they consume an item that is tagged infused and healing, they should get temp hit points equal to the Alchemist's INT score. When applied to a character it is granting temp hp based on their own INT not the Alchemist who created the original item.

You are likely applying the effect *wrong*.

https://github.com/user-attachments/assets/c9e53704-e54d-486c-b96a-1589c0f19057

The automation is working correctly.

neo2005:
> > So with the [Chirurgeon Field Discovery (5th)](https://2e.aonprd.com/ResearchFields.aspx?ID=6) feature applying this affect to another actor after they consume an item that is tagged infused and healing, they should get temp hit points equal to the Alchemist's INT score. When applied to a character it is granting temp hp based on their own INT not the Alchemist who created the original item.
> 
> You are likely applying the effect _wrong_.
> 
>  chir.mp4 
> The automation is working correctly.

The automation isn't working correctly

https://github.com/user-attachments/assets/5042ef5c-faef-40ca-aa40-19b9849a5f03

CrackJackFlood:
[...]

PR #18502, opened 2025-03-10
Get origin data when dragging effects from the actor sheet

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/18287


CarlosFdez:
I wasn't sure of throwing this in during the hotfixes but enough time has passed for this to get looked at.

PR #18518, opened 2025-03-11
Fix runelord from prompting for arcane thesis

CarlosFdez:
Should I also prevent it from being added to the actor at all? I suspect that's what we're actually going to do long term.


src/module/item/base/document.ts
```diff
@@ -650,9 +650,23 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+            // Our only goal is to prevent choice sets, which are not salvageable
+            const sourceId = item.sourceId;
+            if (sourceId && item.isOfType("feat")) {
+                const suppressed =
+                    items.some(
+                        (i) => i.isOfType("feat") && i.system.subfeatures.suppressedFeatures.includes(sourceId),
+                    ) || actor.itemTypes.feat.some((f) => f.system.subfeatures.suppressedFeatures.includes(sourceId));
+                if (suppressed) {
+                    item.suppressed = true;
+                }
```
stwlam, line 666:
```suggestion
                item.suppressed ||=
                    items.some(
                        (i) => i.isOfType("feat") && i.system.subfeatures.suppressedFeatures.includes(sourceId),
                    ) || actor.itemTypes.feat.some((f) => f.system.subfeatures.suppressedFeatures.includes(sourceId));
```

PR #18527, opened 2025-03-11
Suppress Arcane Thesis selection for `Runelord` Class Archetype

ammalagonc:


pedrogrullada:
It seems this issue is already being tackled by #18518

ammalagonc:
Didn't see that one
Thanks!

PR #18528, opened 2025-03-11
Update Class Archetype tags, add Flexible Spellcasting as a level 1 option for eligible classes and update Flexible Spellcasting journal entry

ammalagonc:
While working on Flexible Spellcasting, I noticed that the choice sets for cleric doctrines and arcane schools allow you to pick class archetypes on the second round of choices.

I added `x-class-archetypes` tags to the features and updated the choice sets for subclasses to pick up on these. I updated Elemental Magic to use these tags as well.

The subclass choice sets were also updated to account for whether or not it's the actual class or a dedication triggering the choice set to filter out Class Archetypes.

Let me know which parts of this PR we can keep and which ones to put on hold or scrap.

Thanks


packs/classfeatures/arcane-school.json
```diff
@@ -34,6 +53,11 @@
                 },
                 "flag": "arcaneSchool",
                 "key": "ChoiceSet",
+                "predicate": [
+                    {
+                        "not": "class:wizard"
+                    }
+                ],
```
TikaelSol, line 60: We can (and should) do this cleaner in a single ChoiceSet instead of having two with the same flag.

packs/classfeatures/avenger.json
```diff
@@ -134,7 +134,7 @@
         },
         "traits": {
             "otherTags": [
-                "rogue-racket"
+                "rogue-class-archetype"
             ],
```
TikaelSol, line 138: Instead, let's make two tags. One for racket and one that's just `class-archetype` then we can restrict the class archetype choices for dedications.

packs/classfeatures/bloodline.json
```diff
@@ -25,6 +25,25 @@
...
+                "adjustName": false,
+                "choices": {
+                    "filter": [
+                        {
+                            "or": [
+                                "item:tag:sorcerer-bloodline",
+                                "item:tag:sorcerer-class-archetype"
+                            ]
+                        }
+                    ]
```
TikaelSol, line 38: That would make this filter:
```json
{
    "or": [
        {
            "and": [
                "class:sorcerer",
                "item:tag:sorcerer-bloodline"
            ]
        },
        {
            "and": [
                "item:tag:sorcerer-bloodline",
                {
                    "not": "item:tag:class-archetype"
                }
            ]
        }
    ]
}
```
I think this should work, anyways.

TikaelSol:
Good catch on the multiclass stuff. System's getting too complicated 😓

PR #18530, opened 2025-03-11
Condense Fanged Rune and Potion of Electricity Resistance equipment effects

ammalagonc:


TikaelSol:
Does the fanged rune work with the note added by the system?

ammalagonc:
Ah, good question - I will test this and let you know.

Issue #18472, opened 2025-03-08
Reproduce Number Input behavior in the special resources field


XyrilP:
As of right now, Resources generated from the Special Resources RE does not increment or decrement the same way as HP or similar fields in the character sheet or interfaces.

CarlosFdez:
For edification, it should only work if the input is focused, like regular number inputs. It should not do this if unfocused.

Up/Down Arrow Keys I did attempt to bring back support for, but I couldn't find a good solution.

stwlam:
No scroll wheel listeners for inputs, sorry

CarlosFdez:
The PR Is incomplete, but they're asking for it to behave like a number input. I think that's fine. I was gonna do it, but I delayed it. I'm surprised someone asked about it this fast.

CarlosFdez:
Checking again, this is a bug. The number input does not have the features of a different number input.

PR #18531, opened 2025-03-12
Support delta inputs for health and shield hp in the character sheet

eryon:
Expands on the existing `input[data-allow-delta]` to support keyboard delta (ArrowUp/Down) and (eventually) mousewheel support, specifically for editable character sheet fields such as XP, special resources, and HP.

### DIFF
--- a/src/module/actor/sheet/helpers.ts
+++ b/src/module/actor/sheet/helpers.ts
@@ -83,10 +83,11 @@ function createAbilityViewData(item: AbilityItemPF2e | FeatPF2e): AbilityViewDat
 /**
  * Applies a delta change to the value of an input element
  *
- * @param input
- * @param delta
- * @param min
- * @param max
+ * @param input reference to the input field
+ * @param delta the amount of change to apply to the input's current value
+ * @param [min] minimum input value
+ * @param [max] maximum input value; if "0", do not limit the value
+ * @param [triggerChange] if true, dispatches a change event after this operation is complete
  * @returns the new value of the input field
  */
 function applyDeltaToInput(
--- a/src/module/actor/sheet/helpers.ts
+++ b/src/module/actor/sheet/helpers.ts
@@ -85,18 +85,11 @@ function createAbilityViewData(item: AbilityItemPF2e | FeatPF2e): AbilityViewDat
  *
  * @param input reference to the input field
  * @param delta the amount of change to apply to the input's current value
- * @param [min] minimum input value
- * @param [max] maximum input value; if "0", do not limit the value
- * @param [triggerChange] if true, dispatches a change event after this operation is complete
  * @returns the new value of the input field
  */
-function applyDeltaToInput(
-    input: HTMLInputElement,
-    delta: number,
-    min: number = 0,
-    max: number = 0,
-    triggerChange: boolean = true,
-): string {
+function applyDeltaToInput(input: HTMLInputElement, delta: number): string {
+    const min = Number(input.dataset.min) || 0;
+    const max = Number(input.dataset.max) || 0;
     const oldValue = Number(input.value) || min;
 
     if (max > 0) {
@@ -105,9 +98,7 @@ function applyDeltaToInput(
         input.value = String(Math.max(oldValue + delta, min));
     }
 
-    if (triggerChange) {
-        fu.debounce((el: HTMLInputElement) => el.dispatchEvent(new Event("change", { bubbles: true })), 0)(input);
-    }
+    fu.debounce((el: HTMLInputElement) => el.dispatchEvent(new Event("change", { bubbles: true })), 0)(input);
 
     return input.value;
 }
--- a/src/module/actor/sheet/helpers.ts
+++ b/src/module/actor/sheet/helpers.ts
@@ -81,13 +81,13 @@ function createAbilityViewData(item: AbilityItemPF2e | FeatPF2e): AbilityViewDat
 }
 
 /**
- * Applies a delta change to the value of an input element
+ * Applies a delta change to the numeric value of an input element. If the input field has a min or max data property,
+ * the delta change is limited to those values.
  *
  * @param input reference to the input field
  * @param delta the amount of change to apply to the input's current value
- * @returns the new value of the input field
  */
-function applyDeltaToInput(input: HTMLInputElement, delta: number): string {
+function applyDeltaToInput(input: HTMLInputElement, delta: number): void {
     const min = Number(input.dataset.min) || 0;
     const max = Number(input.dataset.max) || 0;
     const oldValue = Number(input.value) || min;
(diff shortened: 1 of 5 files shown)


eryon:
@CarlosFdez I believe you're the right person to ping on this.  It is a draft intended to show a possible solution for #18472.  I'm proposing to extend the `data-allow-delta` attribute already in place, and introduce the pattern of adding `min` and `max` to allow the generic delta utility to work safely.  If this makes sense, I can also add mouse-wheel deltas and apply it to the HP field.

src/util/misc.ts
```diff
@@ -65,6 +65,35 @@ function applyNTimes<T>(func: (val: T) => T, times: number, start: T): T {
...
+    const oldValue = Number(input.value) || min;
+
+    if (max > 0) {
+        input.value = String(Math.clamp(oldValue + delta, min, max));
+    } else {
+        input.value = String(Math.min(oldValue + delta, min));
+    }
+
+    if (triggerChange) {
+        $(input).trigger("change");
```
eryon, line 91: This trigger does not appear to be anywhere else in the system, but the base Item sheet does programmatically dispatch a change event ([here](https://github.com/foundryvtt/pf2e/blob/cfbf5d3a2e9f4035e94375750126b9e9cdc8f032/src/module/item/base/sheet/sheet.ts#L344-L345)). I want to clarify that it needs to be done this way - or a standard Event *with bubbling enabled* - because of the way jQuery is internally (in the Foundry form application) setup to catch the data submission and actually apply the value changes to the data model.

static/templates/actors/character/partials/header.hbs
```diff
@@ -34,7 +34,7 @@
         </div>
         <div class="exp-data">
             <div class="exp-input">
-                <input name="system.details.xp.value" type="text" value="{{data.details.xp.value}}" data-dtype="Number" data-allow-delta placeholder="0" size="4" />
+                <input name="system.details.xp.value" type="text" value="{{data.details.xp.value}}" data-dtype="Number" data-allow-delta data-min="0" data-max="{{data.details.xp.max}}" placeholder="0" size="4" />
```
CarlosFdez, line 37: I'm not 100% sure but there is probably a valid use case in not having a max, due to spillover.

CarlosFdez:
Repurposing data-allow-delta sounds fine to me

src/util/misc.ts
CarlosFdez: We have a sheet helpers you could probably stick this in, but if this is possible with a regular event ideally use that instead, we're trying to get off jquery (core will eventually do so as well).

I'm wondering how well throttling the change would work, so that repeated hits down submit immediately and it behaves "smoother". Maybe a future consideration.

eryon: > We have a sheet helpers you could probably stick this in, but if this is possible with a regular event ideally use that instead, we're trying to get off jquery (core will eventually do so as well).

I think I got it in the right place now in `src/module/actor/sheet/helpers.ts`

> I'm wondering how well throttling the change would work, so that repeated hits down submit immediately and it behaves "smoother". Maybe a future consideration.

I added a 50ms debounce to the event, but that was too slow for "small" fields like HP. A 0ms debounce, like I have shown now, is enough to push it to the bottom of the call stack and provides a smoother UX but is still fast enough to process changes correctly.

eryon:
@CarlosFdez I wasn't clear from the Discord convo if we should have mouse-wheel deltas supported. Can you confirm? And if we do want to support that, is it only when the field is in focus or also while the mouse is "just" hovered over the field (if that's even possible in the browser)?

CarlosFdez:
If you want to add it go for it. Make sure it doesn't cause re-rendering hiccups though. You can either throttle the update or have it on blur, whatever is most natural. Only when focused though, range inputs updating when you scroll wheel through the container is bad enough as it is.

eryon:
@CarlosFdez ready for your formal review, I believe.

My only gripe with this is that fields for embedded items, such as spellcasting slot inputs or those from SpecialResource REs lose focus after changing.  I believe this is happening because of the [change detection that calls `updateEmbeddedDocuments`](https://github.com/foundryvtt/pf2e/blob/bf5013e859a4f5fa110a964194dfa3e0f65a3b4c/src/module/actor/sheet/base.ts#L428) and causes that element to re-render, so I'm not aware of any way around this (I can imagine maybe a delayed call to refocus afterwards, but that feels unreliable).

CarlosFdez:
If possible, can you also grab shield hp? That's actually a common one where deltas would actually be useful (due to feats like shield warden).

I'll take a look at it locally later. Foundry core should be refocusing after an update, but perhaps its not working on that property because it doesn't have a name. We might need to reproduce the behavior on data-property stuff as well.

CarlosFdez:
Rebase when you get the chance. The data-resource and data-property fields should hopefully work a lot better with arrow inputs (hopefully).

src/module/actor/sheet/helpers.ts
```diff
@@ -80,4 +80,36 @@ function createAbilityViewData(item: AbilityItemPF2e | FeatPF2e): AbilityViewDat
...
+ * @param [triggerChange] if true, dispatches a change event after this operation is complete
+ * @returns the new value of the input field
+ */
+function applyDeltaToInput(
+    input: HTMLInputElement,
+    delta: number,
+    min: number = 0,
+    max: number = 0,
+    triggerChange: boolean = true,
+): string {
```
CarlosFdez, line 99: Given you're passing the input directly, I think you can pull min/max from this function directly. You don't use the triggerChange param either so you can probably omit it.

Not step though, since you're passing a delta and its part of the delta computation.

eryon: I can take care of it. With regards to step, you think I should leave the `delta` param in the helper but remove the `stepSize` bits from the base sheet, passing in `1` or `-1` directly?

CarlosFdez: Exactly that.

eryon:
And one more rebase...

src/module/actor/sheet/helpers.ts
```diff
@@ -80,4 +80,27 @@ function createAbilityViewData(item: AbilityItemPF2e | FeatPF2e): AbilityViewDat
...
+
+    if (max > 0) {
+        input.value = String(Math.clamp(oldValue + delta, min, max));
+    } else {
+        input.value = String(Math.max(oldValue + delta, min));
+    }
+
+    fu.debounce((el: HTMLInputElement) => el.dispatchEvent(new Event("change", { bubbles: true })), 0)(input);
+
+    return input.value;
```
CarlosFdez, line 103: Doesn't seem like you use the return value. You can omit it I think.

eryon: That's true, cleaned up!

CarlosFdez:
Tested it out, looks great, thanks for this! Last thing before the door and the V13 freeze.

PR #18557, opened 2025-03-15
Condense Insight Coffee equipment effects

ammalagonc:


packs/equipment-effects/effect-insight-coffee.json
```diff
@@ -0,0 +1,169 @@
...
+                "mode": "upgrade",
+                "predicate": [
+                    "certain-stratagem"
+                ],
+                "property": "dice-faces",
+                "selectors": [
+                    "inline-damage"
+                ],
+                "value": 8
+            },
```
TikaelSol, line 125: ok, this one is a holdover from the other effects, but I think this is wrong. It's too broad, since it would hit *all* inline damage. We should look at that and figure out what needs to be done to narrow it.

ammalagonc: Found the note in question -
`"Note": "You deal half of your @Damage[(floor((@actor.level+3)/4)d6[precision])[@item.system.damage.damageType]|traits:certain-stratagem]{Strategic Strike damage} to the target"`

It appears the inline in question is using a rollOption to narrow down this particular inline damage.

pedrogrullada: Now that I look at it, it *should* be wrong. Nevertheless, it does look like only the certain stratagem dice are upgraded, even if there's bonus dice added somehow.
image

Ideally we could have the ability to add tags to inline damage rolls, like we have with Flat Modifiers and Damage Dice. 

Anyway, that `traits:` should be changed with `options:` if you want to go ahead and do that in this PR as well.

PR #18569, opened 2025-03-16
Add damage note to the Confused condition

ammalagonc:
Feel free to provide feedback and suggestions to get this to the finish line.

Thanks!

### DIFF
--- a/packs/conditions/confused.json
+++ b/packs/conditions/confused.json
@@ -56,7 +56,7 @@
                   {
                     "or": [
                       "item:type:spell",
-                      "action:strike"
+                      "origin:action:trait:attack"
                     ]
                   }
                 ],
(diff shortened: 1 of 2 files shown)


packs/conditions/confused.json
```diff
@@ -47,7 +47,21 @@
...
+            },
+            {
+                "key": "Note",
+                "selector": "damage-received",
+                "text": "PF2E.condition.confused.note",
+                "predicate": [
+                  {
+                    "or": [
+                      "item:type:spell",
+                      "action:strike"
```
pedrogrullada, line 59: I don't think this roll option is present in `damage-received`, but we probably want something broader anyway. Non-strike attacks should also count for this.
```suggestion
                      "origin:action:trait:attack"
```

PR #18573, opened 2025-03-17
Update Deadly Strike to use Item Alteration over AdjustStrike

pedrogrullada:
~~It looks like Item Alteration doesn't remove lower deadly traits upon adding a higher one, so it resulted in quite a few more rule elements.~~


stwlam:
That's a bug; sit on this until it can be fixed (will prioritize it).

pedrogrullada:
The bug that was referenced here has been fixed since, allowing the involved rule elements to be simplified. This should be good to go.

PR #18598, opened 2025-03-21
Ruffian should allow critical specialization with advanced weapons

hollyvalenta:


pedrogrullada (commented):
Instead of doing this, replace the predicate in the existing Critical Specialization with this
```json
  "predicate": [
    "item:tag:sneak-attack",
    "target:condition:off-guard"
  ]
```
The `item:tag:sneak-attack` would be doing the heavy lifting here, since we have a rule element adding that tag to any ruffian weapon.

Additionally, please name your PR in imperative mood, as if completing the phrase "If merged, this PR will ...". This makes it easier for the person compiling the release to add your credit to the changelog. In your case, the title should be something like "Allow ruffian to use critical specialization with all qualifying weapons", or something along those lines.

PR #18629, opened 2025-03-23
Fix quantitiy item deduction

davidps92:
Deduct one from quantity as commented line says instead of the number of charges used. Which is not always 1.

### DIFF
--- a/src/module/item/consumable/document.ts
+++ b/src/module/item/consumable/document.ts
@@ -206,10 +206,17 @@ class ConsumablePF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
                 await this.delete();
             } else {
                 // Deduct one from quantity if this item has one charge or doesn't have charges
-                await this.update({
-                    "system.quantity": Math.max(quantityRemaining - 1, 0),
-                    "system.uses.value": uses.max,
-                });
+                if (uses.max > 1) {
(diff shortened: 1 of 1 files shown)


davidps92:
But deducting only one is the correct way. So the comment is correct, but the implementation is incorrect.
Example:
If you have an item with 10 max uses and quantity 20 if you consume 10, uses will reduce the quantity to 10, instead of 19.
The original script is wrong because it mixes uses deduction with quantity deduction.

CarlosFdez:
Double Barrel trait

CarlosFdez:
Ah, i see. The bugfix for double fire isn't properly handling magazine fire weapons.

Thinking more carefully about the problem, magazine ammo should only reduce by 1 (using part of the current magazine and part of the next magazine all at once doesn't make sense). Quantity reducing by more than one is for non-repeating ammo. It looks like this instead should have an else if to to check for uses.max.

CarlosFdez:
Make the comment reflect what you're doing and this works well enough

CarlosFdez:
Thanks for bringing this to our attention. We are close to a release and I wanted to at least test it a bit more throughly, so I ended up doing it myself since I didn't want to suddenly rush you into doing more things. I did the same general idea that was done here, though I future proofed for other potential bugs that may occur down the line.

PR #18634, opened 2025-03-24
Use more melee and ranged selectors instead of predicates where possible

websterguy:
Adjust to move `melee` and `ranged` out of predicates and use melee- and ranged- selectors where possible Changes some `melee` and `ranged` predicates that can't be moved into `item:melee` and `item:ranged`.

Change value on Vicious Swing feat to use a ternary instead of brackets while editing the feat. Same change made to Power Attack ability on NPC Simeek.

### DIFF
--- a/packs/abomination-vaults-bestiary/abomination-vaults-hardcover-compilation/urthagul.json
+++ b/packs/abomination-vaults-bestiary/abomination-vaults-hardcover-compilation/urthagul.json
@@ -133,16 +133,18 @@
                     {
                         "key": "FlatModifier",
                         "selector": "melee-strike-damage",
+                        "slug": "crimson-fulcrum-lens",
                         "type": "item",
                         "value": 2
                     },
                     {
-                        "key": "FlatModifier",
+                        "key": "AdjustModifier",
+                        "mode": "upgrade",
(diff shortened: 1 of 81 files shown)


packs/abomination-vaults-bestiary/abomination-vaults-hardcover-compilation/urthagul.json
```diff
@@ -132,17 +132,14 @@
...
                         "type": "item",
                         "value": 2
                     },
                     {
                         "key": "FlatModifier",
                         "predicate": [
-                            "melee"
+                            "item:melee"
                         ],
                         "selector": "jaws-damage",
```
TikaelSol, line 144: Feels like this one doesn't need the predicate at all, since jaws are kind of never not melee. Also, these are going to wind up with a slug conflict, we should probably adjust modifier instead to avoid duplicate modifiers from the item. Does the effect in compendium have these issues too?

websterguy: It is not the same as the normal Crimson Fulcrum Lens effect, no. The effect uses adjust mod
image

PR #18653, opened 2025-03-26
Add aux action for parry

CarlosFdez:
Fists and non-real strikes can gain the parry trait (ex: Spirit Warrior Dedication)

### DIFF
--- a/src/module/actor/character/helpers.ts
+++ b/src/module/actor/character/helpers.ts
@@ -239,8 +239,8 @@ class WeaponAuxiliaryAction {
             await actor.itemTypes.effect.find((e) => e.sourceId === COVER_UUID)?.delete();
         } else if (this.action === "parry") {
             // Apply Effect: Parry
-            const alreadyRaised = actor.itemTypes.effect.some((e) => e.slug === "parry");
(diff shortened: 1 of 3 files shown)


src/module/actor/character/helpers.ts
```diff
@@ -226,6 +237,14 @@ class WeaponAuxiliaryAction {
             }
         } else if (this.action === "end-cover") {
             await actor.itemTypes.effect.find((e) => e.sourceId === COVER_UUID)?.delete();
+        } else if (this.action === "parry") {
+            // Apply Effect: Parry
+            const alreadyRaised = actor.itemTypes.effect.some((e) => e.slug === "parry");
+            if (alreadyRaised) return;
```
stwlam, line 243:
```suggestion
            const alreadyParrying = actor.itemTypes.effect.some((e) => e.slug === "parry");
            if (alreadyParrying) return;
```

stwlam (dismissed):

src/module/actor/character/helpers.ts
CarlosFdez: Whoops!

PR #18673, opened 2025-03-28
Expand deactivation of in-memory conditions giving presence of others

stwlam:


src/module/item/condition/document.ts
```diff
@@ -241,20 +241,21 @@ class ConditionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
...
-                if (thisValue >= otherValue) deactivate(condition);
-            } else if (this.value === condition.value && this.inMemoryOnly) {
-                // Defer to other, identical condition if this one is in-memory-only
+                const otherValue = other.system.persistent?.expectedValue ?? 0;
+                if (thisValue >= otherValue) deactivate(other);
+            } else if (otherIsGTE && this.inMemoryOnly) {
+                // Defer to other, gte condition if this one is in-memory-only
                 deactivate(this);
-            } else if (this.value === condition.value && (!this.isLocked || condition.isLocked)) {
+            } else if (otherIsGTE && (!this.isLocked || other.isLocked)) {
```
CarlosFdez, line 253: This logic can lead to other getting deactivated if other is greater. Perhaps flipping the logic such that you check `!other.isLocked || this.isLocked` instead and then `deactivate(this)` might suffice?

Personally, I think it might be best to do a batch operation (group by then process each group) in the actor if you can swing the extra time. Trying to dedupe data on an item by item basis is pretty hard to reason around.

src/module/item/condition/document.ts
```diff
@@ -241,25 +241,27 @@ class ConditionPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
...
                 deactivate(this);
-            } else if (this.value === condition.value && (!this.isLocked || condition.isLocked)) {
+                return;
+            } else if (this.value === other.value && (!this.isLocked || other.isLocked)) {
                 // Deactivate other equal valued conditions if this condition isn't locked or both are locked
-                deactivate(condition);
-            } else if (this.value && condition.value && this.value > condition.value) {
+                deactivate(other);
+            } else if (!otherIsGTE) {
                 // Deactivate other conditions with a lower or equal value
```
CarlosFdez, line 258:
```suggestion
                // Deactivate other conditions with a lower value
```

PR #18703, opened 2025-03-31
Change Dragon Instinct pre-remaster dragons to arcane tradition

MrVauxs:
Currently the ItemAlteration RE breaks due to chromatic/metallic not being real traits.

Alternatively if this interpretation is not something the system wants to use, we should change the table to say N/A instead of Arcane.


pedrogrullada:
It's not quite the same situation, but for the record, this is what we do with legacy draconic sorcerers as well. Since arcane was their only option, that's what we give them.

PR #18714, opened 2025-04-01
Add description addenda for Irongut Goblin and Virtuous Tanuki when sickened

pedrogrullada:
Merging this will also
- Account for consumables with the ingested traits in sickened's Item Alteration (most of these are poisons, but a few are drugs with beneficial benefits that a character could conceivably be for a character's own use)

### DIFF
--- a/packs/heritages/goblin/irongut-goblin.json
+++ b/packs/heritages/goblin/irongut-goblin.json
@@ -19,15 +19,15 @@
         },
         "rules": [
             {
-                "key": "RollOption",
-                "label": "PF2E.SpecificRule.Goblin.IrongutGoblin.BonusToggle",
-                "option": "irongut-goblin-ingested",
-                "toggleable": true
-            },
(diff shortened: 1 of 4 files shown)


packs/heritages/goblin/irongut-goblin.json
```diff
@@ -18,37 +18,54 @@
...
-                "label": "Irongut Goblin (vs. afflictions & sickened)",
                 "predicate": [
-                    "ingested",
-                    {
-                        "or": [
-                            "affliction",
-                            "inflicts:sickened"
-                        ]
-                    }
+                    "irongut-goblin-ingested"
```
TikaelSol, line 30: I would rather stick with the above but made more proper, with `item:type:affliction` and `item:type:ingested`. I generally think these kind of rarer things don't need toggles, and can just be flipped on in the dialog until we can properly automate them.

PR #18718, opened 2025-04-01
Add jousting trait's circumstance bonus to damage to roll dialog

pedrogrullada:
Full disclosure, I just went off looking at how other traits do this, so please feel free to kick this one into shape if there's something wrong with it 🙃


pedrogrullada:
> Reconcile with https://github.com/foundryvtt/pf2e/blob/master/src/module/actor/character/helpers.ts#L33

Not sure if I understood correctly, but I've made a change to handle the die size override in the same place as the bonus. It's working in testing, but I'll recognise I might have bit more than I could chew with this one!

stwlam:
What's currently in place predates the "mounted" effect: it settles for assuming that if the jousting weapon is held in one hand, the character is mounted (since that's the only legal way to wield a jousting weapon one-handed). Note that in order to get the circumstance bonus, the jouster must have moved at least ten feet just prior to the attack.

pedrogrullada:
> What's currently in place predates the "mounted" effect: it settles for assuming that if the jousting weapon is held in one hand, the character is mounted (since that's the only legal way to wield a jousting weapon one-handed). Note that in order to get the circumstance bonus, the jouster must have moved at least ten feet just prior to the attack.

Ah, you're totally right, no point in trying to predicate on the effect then.

I've implemented the circumstance damage to be toggled off, so it's just something you can add in the roll dialog if the jouster does move the necessary 10 ft. mounted

PR #18740, opened 2025-04-04
Add Claws of the Tyrant feats to select Archetype Journal feats

ammalagonc:
Also adds missing feats from other sources, and adds the source publication for the feats.


packs/journals/archetypes.json
simonward, line 1: It looks like each new entry has `<p></p>` at the end. Is this intentional?

ammalagonc: Yes, otherwise, it ends up looking like this
image

PR #18745, opened 2025-04-04
Support description item alteration addendums in expandable self effect actions

CarlosFdez:
In the process of removing our dependency of the description getter, I saw a use case where we might want the full description.

image

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -1,7 +1,7 @@
 import { ActorPF2e } from "@actor/base.ts";
 import type { ContainerPF2e, PhysicalItemPF2e } from "@item";
 import { createConsumableFromSpell } from "@item/consumable/spell-consumables.ts";
-import { ItemChatData, itemIsOfType } from "@item/helpers.ts";
+import { itemIsOfType, markdownToHTML } from "@item/helpers.ts";
 import { ItemOriginFlag } from "@module/chat-message/data.ts";
 import { ChatMessagePF2e } from "@module/chat-message/document.ts";
 import { preImportJSON } from "@module/doc-helpers.ts";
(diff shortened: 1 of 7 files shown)


src/module/item/base/document.ts
```diff
@@ -453,7 +530,11 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         htmlOptions: EnrichmentOptionsPF2e = {},
         chatData: RawItemChatData,
     ): Promise<RawItemChatData> {
-        return new ItemChatData({ item: this, data: chatData, htmlOptions }).process();
+        const description = {
+            ...chatData.description,
+            ...(await this.getDescription(htmlOptions)),
+        };
```
stwlam, line 536: Both?

src/module/item/base/document.ts
```diff
@@ -453,7 +530,11 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         htmlOptions: EnrichmentOptionsPF2e = {},
         chatData: RawItemChatData,
     ): Promise<RawItemChatData> {
-        return new ItemChatData({ item: this, data: chatData, htmlOptions }).process();
```
stwlam, line 456: Is this class in use anymore?

src/module/item/base/document.ts
CarlosFdez: Its what the ItemChatData process function already did. I'm not sure why, maybe there's other props that chat data might slip in?

src/module/item/base/document.ts
CarlosFdez: Nope, I removed it here.

PR #18748, opened 2025-04-04
Set message sender name and portrait cursor type on chat messages to pointer

LebombJames:
Since they both have a click listener to control/pan to the token, their cursor type should both be pointer.


stwlam:
I think we need to make it conditional on whether the listener would do anything.

CarlosFdez:
Agreed, people who don't own the token wouldn't be able to click on it.

I've been wondering if there's a clean performant way to make the chat message portrait glow when the token is selected. I'm not sure how I'd go about doing it but it'd be pretty.

Issue #18723, opened 2025-04-02
Strike Modifying Activities


Ikaguia:
Strike Modifying activities like vicious strike, combat grab... should all be consolidated under a single roll option, and they should all have notes to display their success/failure effects on the roll.

![Image](https://github.com/user-attachments/assets/1816a6ee-3997-4509-b13b-eb4456ec89e2)

![Image](https://github.com/user-attachments/assets/3d2f8c66-fee2-40cc-ad6e-b791b70fb49f)

TheErikkuBlade:
Bumping this to say it would be a significant Quality of Life improvement. As the system acquires more and more ways to modify a Strike, the clutter for roll options becomes larger and larger, making the character sheet more and more cumbersome.

That said, this is just a minor nitpick in the example that was posted but I'm not sure if Reactive Strike is necessary as it doesn't modify the Strike at all (part from maybe the critical success disrupting actions), not enough to warrant its own Roll Option, maybe only the ones that currently have toggleable roll options.

CarlosFdez:
I can't speak for anyone else but I think the general trajectory is to add more support for usable actions. Using the action would probably prompt for the strike itself. I don't know a way to do that workflow particularly quickly though.

Ikaguia:
> Bumping this to say it would be a significant Quality of Life improvement. As the system acquires more and more ways to modify a Strike, the clutter for roll options becomes larger and larger, making the character sheet more and more cumbersome.
> 
> That said, this is just a minor nitpick in the example that was posted but I'm not sure if Reactive Strike is necessary as it doesn't modify the Strike at all (part from maybe the critical success disrupting actions), not enough to warrant its own Roll Option, maybe only the ones that currently have toggleable roll options.

I made reactive strike have it for my character because I always forget which types of actions it disrupts (is it manipulate? is it concentrate?) and it would add parity with other similar reactions that have different disruptions (thaumaturge and ranger interrupt everything that triggers it, vindicator disrupts spells, fighter's disrupting stance interrupts manipulate AND concentrate on a normal hit)

stwlam:
Action toggles are not going to be expanded in this way, sorry.

PR #18756, opened 2025-04-05
Implements "Strike Modifying Activities #18723"

Ikaguia:
Example of how an implementation of "Strike Modifying Activities #18723" would look like.
Only done on fighter feats so far.

### DIFF
--- a/packs/actions/reactive-strike.json
+++ b/packs/actions/reactive-strike.json
@@ -18,7 +18,56 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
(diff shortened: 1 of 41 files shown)


pedrogrullada:
Hello! We talked about your idea and this is not the direction the system is heading in for handling actions like this (see Supe's comment on the linked issue). One of the main issues is not having a clearly distinct in-game category to classify these feats in, the way things like spellshape and swashbuckler's finishers do. We would have to be making them up, and we'd rather avoid that and wait to develop a more elegant solution.

Thanks for the work you put on this, and in the future feel free to poke people with Data Entry roles in the community Discord so you can get some guidance on whether a big idea like this lines up with what we're doing.

PR #18767, opened 2025-04-06
Remove `Fused Polearm`, `Polearm Tricks` and `School Counterspell` Runelord feats

ammalagonc:
Rod of Rule -> Polearm Tricks 
Sin Counterspell -> School Counterspell


ammalagonc:
Should resolve once #18766 is merged

PR #18771, opened 2025-04-07
Extend Unstable note to actions that have the `unstable` trait and add notes RE to Inventor dedication

### LINKED ISSUE #18772: Explode is missing the Unstable alteration
The explode action does not include a flat check roll for the Unstable trait.

Tikael on Discord gave me the following solution

> Uh, looks like we missed the alteration on explode specifically 🙃
> On the Explode class feature (not the action) go to the rules tab and make a new rule then paste this in
> 
> {
>   "itemType": "action",
>   "key": "ItemAlteration",
>   "label": "PF2E.TraitUnstable",
>   "mode": "add",
>   "predicate": [
>     "item:trait:unstable",
>     {
>       "not": "self:effect:unstable-check-failure"
>     }
>   ],
>   "property": "description",
>   "value": [
>     {
>       "text": "PF2E.SpecificRule.Inventor.Unstable.Note"
>     }
>   ]
> }
> 
>

ammalagonc:
Closes #18772

### DIFF
--- a/packs/classfeatures/explode.json
+++ b/packs/classfeatures/explode.json
@@ -46,6 +46,24 @@
                         "text": "PF2E.SpecificRule.Inventor.Unstable.Note"
                     }
                 ]
+            },
+            {
+                "itemType": "action",
+                "key": "ItemAlteration",
+                "label": "PF2E.TraitUnstable",
+                "mode": "add",
+                "predicate": [
+                    "item:trait:unstable",
+                    {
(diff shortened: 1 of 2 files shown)


stwlam:
Put it with the other REs handling this, and cover any theoretical unstable action rather than just that one.

ammalagonc:
Already doing that - this is in the Explode Class Feature. Unless it needs to go somewhere else

PR #18790, opened 2025-04-10
Use @Embed enricher when creating spell scrolls and wands

Cuingamehtar:
Instead of copying the original description

### DIFF
--- a/src/module/item/consumable/spell-consumables.ts
+++ b/src/module/item/consumable/spell-consumables.ts
@@ -95,7 +95,6 @@ async function createConsumableFromSpell(
     traits.value.sort();
 
     consumableSource.name = getNameForSpellConsumable(type, spell.name, heightenedLevel);
-    const description = consumableSource.system.description.value;
 
     consumableSource.system.description.value = (() => {
         const paragraphElement = document.createElement("p");
@@ -104,7 +103,7 @@ async function createConsumableFromSpell(
         const containerElement = document.createElement("div");
         const hrElement = document.createElement("hr");
         containerElement.append(paragraphElement, hrElement);
-        hrElement.insertAdjacentHTML("afterend", description);
+        hrElement.insertAdjacentHTML("afterend", `@Embed[${consumableSource._stats.compendiumSource} inline]`);
 
         return containerElement.innerHTML;
     })();


CarlosFdez:
What's the goal here?

Cuingamehtar:
Reduce text duplication mostly. It's my understanding that it's there as a reference to how base item works, and it's not intended to be different from it (and it isn't, considering it's a copy).
The tradeoff would be that if a user wanted to modify this description, they would have to open the description of the base item.

But my motivation was wanting to reduce the amount of text that needed to be localized and kept track of on actors (if not by much). If this change was to be accepted, I intended to replace the descriptions in actor-owned items as well.

If these changes are not desired (at this moment or at all), I have no issues with closing this PR. It's not a big deal.

stwlam:
So long as scrolls and wands have full clones of spells embedded in them, we probably shouldn't embed the compendium text.

Cuingamehtar:
It's not the spell text that I'm suggesting embedding, it's the scroll/wand text.

CarlosFdez:
Yeah, it'd end up reducing how heavy each item is, with the cost of a database lookup when expanding an item description. I'm not opposed on principle, but knowing you I didn't think it was just because you wanted to reduce data size.

I'm guessing then that its easier to translate specific items, rather than the myriad of ways a scroll or wand can exist.

Cuingamehtar:
Yeah, I wanted to minimize the amount of repeated text to work with, instead of translating the scroll description every time.
Although this change would not completely remove the need to translate scroll descriptions because of the explicit spell link labels.

Database lookup would be done once per uuid. After the first lookup, the item object is cached in the pack object.
But the enricher would still have to find the description and replace the HTML every text render. 

However, now that I think about it, that might actually increase the number of unique text segments - instead of the scroll descriptions being different just by spell links, they would now be different by scroll embeds for different the spell ranks as well.
- - -
With all that in mind, I think it's ok to close the PR. I don't think the tradeoff is worth it.

CarlosFdez:
It should still be easier to translate the base scroll item and embed links, than trying to recreate a scroll's description but translated?

Cuingamehtar:
Well, maybe. But not really in our case.

The description is basically 99% identical to others in translation memory, so it's just a matter of replacing the spell link.

Other groups (de, fr, es) probably don't have difficulties either with the crowdin service they are using.

I'll probably just modify our babele converter to translate scrolls automatically (as their descriptions are very simple).

Thank you for this discussion. It helped me to figure out another way to achieving what I wanted to do without affecting everybody else.

PR #18817, opened 2025-04-15
Support editing staff effect html

CarlosFdez:
Turns out some staves (maybe just one?) need to use HTML. Because prosemirror doesn't fit, I removed some components that don't really make sense for single-paragraph effects. I can remove more if we want.

image

image

image


CarlosFdez:
There is one quirk, where if you edit an effect that is blank that has generated a default descripted, the prosemirror editor is blank. I'm considering showing the default text, and detecting if its the default text on submit to turn it back into empty string.

Issue #18813, opened 2025-04-14
Feature request: Template placement with click-drag-click


CatharsisFonts:
I love the 2e implementation in Foundry, but putting down a cone or line is a pain with a keyboard. Wouldn’t it be much more intuitive to click to place the origin, then track the mouse position with the cone/line until a second click locks it in place? Same for editing an existing cone/line. It would work equally well with mice and trackpads. 

Cheers, Christian

FolkvangrForgent:
Editing existing templates or placing using the button in the measurement section uses default FVTT interaction if I remember correctly. The improvement in regards to templates placed from chat is covered by #4618 .

CatharsisFonts:
Oof, that’s from 2022…?

Issue #18552, opened 2025-03-14
Slow rotate of template before placement uses wrong keybind on mac


JacksonBockus:
Per the core keybindings, slow rotation of an object should be with command-scroll on Mac. Instead it's mapped to control-scroll only when rotating a template before final placement.

PR #18819, opened 2025-04-15
Improve `MeasuredTemplate` preview rotation UX for cones and lines

### LINKED ISSUE #4618: Dragable Chat Template Placement Enchacements
I believe that that template placement from chat could be more intuitive to use. In particular I would like the see the following behavior:
- Clicking and then Dragging would allow placement and then rotation of the template towards cursor. (ideally with snapping angles like the template tool)
- Releasing while dragging would place the template.
- Right Clicking while dragging would free the template allowing you to adjust where you placed it.
- Right Clicking while not dragging would cancel the current placement as normal.
- Left Clicking quickly (i.e. not dragging) would place the template with it's current preview rotation.
- Mouse Wheel would keep it's current behavior allowing rotation without dragging.

In3luki:
This PR updates the rotation mechanism for cones and lines, replacing mouse wheel-based rotation with a more intuitive cursor position-based approach.

New Behavior:

- Left Click (1st click): Locks the template preview to the current (snapped) canvas position and enables rotation using the mouse cursor's position.
- Left Click (2nd click): Confirms the rotation and places the template.
- Right Click (while locked): Unlocks the preview without placing the template.
- Right Click (while unlocked): Cancels the operation entirely.

Additional Controls:

- Shift + Left Click:  Instantly places the template at the current position without locking or rotation.
- Holding Ctrl (while template is locked):  Allows free (unsnapped) rotation of cone templates for finer control.

[template-rotation.webm](https://github.com/user-attachments/assets/92dfffd2-d8d4-4560-98a2-7e67ff5496af)

Closes #4618 and #18813 and #18552

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -98,12 +98,10 @@ export class TemplateLayerPF2e<
                     const origin = preview.position;
                     const ray = new Ray(origin, destination);
                     if (ray.distance < canvas.grid.size / 4) return;
-                    if (preview.document.t === "cone" && !event.ctrlKey) {
-                        const snapAngle = Math.PI / 4;
+                    if (preview.document.t === "cone" && !(event.ctrlKey || event.metaKey)) {
+                        const snapAngle = Math.PI / (canvas.grid.isHexagonal ? 6 : 4);
                         preview.document.updateSource({
-                            direction: Math.toDegrees(
-                                Math.floor((ray.angle + Math.PI * 0.125) / snapAngle) * snapAngle,
-                            ),
+                            direction: Math.toDegrees(Math.floor(ray.angle / snapAngle + 0.5) * snapAngle),
                         });
                     } else {
                         preview.document.updateSource({ direction: Math.toDegrees(ray.angle) });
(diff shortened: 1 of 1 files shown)


FolkvangrForgent:
Awesome looking implementation! Any chance you could add the small amount of hex grid support in the form of the snap angle of 30 on hexagonal grid?
'const snapAngle = Math.PI / (canvas.grid.isHexagonal ? 6 : 4);` and `Math.floor((ray.angle + snapAngle / 2) / snapAngle) * snapAngle,` should be sufficient for that. Although I think `Math.floor(ray.angle / snapAngle + 0.5) * snapAngle,` is the expanded form.

In3luki:
> Edit: Out of curiosity was a workflow using dragging not possible? Dragging would be closest to default FVTT behavior but I can also see it being much harder to implement. I think this new workflow is already leagues better than current behavior.

As the preview template is not persisted until it is placed (and the canvas is unaware it exists) a drag event still triggers the creation of a new template. To prevent that we could add an override of `_onDragLeftStart` and expand the logic of `_onDragLeftMove`. But another problem *there* is that the methods expect an already created template which would lead to even more overrides to handle that.
It could probably be done with enough time spent but I think the solution in this PR is good enough for now.

FolkvangrForgent:
This should also make #18552 irrelevant?

In3luki:
Replaced by #19045

Issue #18587, opened 2025-03-19
Issue with AdjustStrike and adding Property Runes


Allalinor:
> Anytime any property runes are given via AdjustStrike, if you mess with the item even a little bit (for example, opening and closing the "Edit Item" or changing the item to a "Specific Item" and back) makes the runes granted permanent, which can make things a bit messy.
Using toggleable roll options for adding property runes tends to break things, often by making the runes permanent on the item even when turning off the roll option (like above).

This was mentioned before in https://github.com/foundryvtt/pf2e/issues/15372 but is still an issue in 6.10.2

I included an example item with Rule Elements to test with.

[fvtt-Item-+3-longsword-RcBNfaE2HAKqXj80.json](https://github.com/user-attachments/files/19339735/fvtt-Item-%2B3-longsword-RcBNfaE2HAKqXj80.json)

CarlosFdez:
It looks like the original issue was an omni issue and this specific issue got lost in there. I'll fix it momentarily.

PR #18852, opened 2025-04-19
Avoid persisting automated weapon property runes when closing the item sheet

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/18587

I also did some UI/UX stuff to make the adjusted colors behave more predictably with select, and made it themeable.

image

There is only one bug remaining: if you set it to empty input, the sheet will not re-render since there were no changes. This bug is a general problem with all of our handlebars sheets though, so I'm not really planning on fixing it here.

### DIFF
--- a/static/templates/items/weapon-details.hbs
+++ b/static/templates/items/weapon-details.hbs
@@ -96,7 +96,7 @@
                     name="system.runes.property.{{index}}"
                     id="{{../fieldIdPrefix}}runes-property-{{index}}"
                     {{disabled slot.disabled}}
-                    {{#if @root.item.isSpecific}}class="readonly"{{/if}}
+                    class="{{#if @root.item.isSpecific}}readonly{{/if}}{{#if slot.adjusted}} adjusted{{/if}}"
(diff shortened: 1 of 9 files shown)


static/templates/items/weapon-details.hbs
```diff
@@ -96,7 +96,7 @@
                     name="system.runes.property.{{index}}"
                     id="{{../fieldIdPrefix}}runes-property-{{index}}"
                     {{disabled slot.disabled}}
-                    {{#if @root.item.isSpecific}}class="readonly"{{/if}}
+                    class="{{#if @root.item.isSpecific}}readonly{{/if}}{{#if slot.adjusted}} adjusted{{/if}}"
```
stwlam, line 99: `readonly` isn't an attribute of select elements

CarlosFdez: It is a CSS class tho.

stwlam: Ope

PR #18861, opened 2025-04-20
Fix switching tabs in the Compendium Browser sometimes breaking the filter

In3luki:
See https://discord.com/channels/880968862240239708/880969241552113695/1363555342939848959

### DIFF
--- a/src/module/apps/compendium-browser/browser.ts
+++ b/src/module/apps/compendium-browser/browser.ts
@@ -3,7 +3,7 @@ import { ActionType, ItemType } from "@item/base/data/index.ts";
 import { PHYSICAL_ITEM_TYPES } from "@item/physical/values.ts";
 import { BaseSpellcastingEntry } from "@item/spellcasting-entry/index.ts";
 import { SvelteApplicationMixin } from "@module/sheet/mixin.svelte.ts";
-import { ErrorPF2e, setHasElement } from "@util";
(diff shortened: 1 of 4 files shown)


CarlosFdez:
Have you tried seeing if updating svelte fixes it? This error is odd, this shouldn't be necessary to do.

My other guess is that the issue might be that tab is complex object and reactivity isn't working.

In3luki:
I hadn't tried updating svelte but that didn't fix it either. I've tried making the result `$derived(browser.tabs[activeTabName],results)` which also didn't work.
The weird thing is that it works on subsequent renders... It only fails to update the result list on first render.
The underlying data is also correct as far as I can tell. Only the DOM update is not working.

CarlosFdez:
If it helps, I attempted to do that, and the idea of wrapping core foundry variables in reactive state wrappers had the reaction of nose scrunching. To handle tabs now, even though it downgrades tabs to second citizen, I'd replace changeTabs with a re-render, and leave it up to _prepareContext to pass the tab groups.

Something like what I did here https://github.com/foundryvtt/pf2e/pull/17702/files#diff-ef80b447bb09e777a1101473031c0f503dc2d4c651fd734b8c17a2307885eee1R40

stwlam:
Can close?

In3luki:
I could backport this but we can just say it's fixed in v13

PR #18867, opened 2025-04-21
Fill out v13 `InteractionLayer` and `PlaceablesLayer` types

In3luki:


types/foundry/client/canvas/layers/_types.d.mts
```diff
@@ -0,0 +1,100 @@
...
+    mask: PIXI.Graphics & {
+        preview: PIXI.Graphics;
+    };
+};
+
+/**
+ * The light part of {@link foundry.canvas.layers.types.CanvasVisionContainer}.
+ * The blend mode is MAX_COLOR.
+ */
+export type CanvasVisionContainerLight = PIXI.Graphics & _CanvasVisionContainerLight;
```
stwlam, line 76: These are defined this way in core because `@typedef` doesn't support interfaces (so they have to settle for intersections). Go ahead and turn them into interfaces extending the non-underscored type name.

PR #18886, opened 2025-04-23
Update Effects Panel to AppV2 and fix inline rolls styling

CarlosFdez:
Since Foundry V13 reduces borders by quite a bit, the current effects panel stands out. I changed it and themed the tooltips. Addendums have not been handled since item alterations are currently non-functional.

image

Also supports light mode interface

image


src/module/apps/effects-panel.ts
```diff
@@ -21,18 +20,21 @@ export class EffectsPanel extends appv1.api.Application {
      */
     refresh = fu.debounce(this.render.bind(this), 100);
 
-    static override get defaultOptions(): appv1.api.ApplicationV1Options {
-        return {
-            ...super.defaultOptions,
-            id: "pf2e-effects-panel",
-            popOut: false,
+    static override DEFAULT_OPTIONS: DeepPartial<fa.ApplicationConfiguration> = {
+        id: "pf2e-effects-panel",
```
stwlam, line 24:
```suggestion
        id: "effects-panel",
```
Set `root: true` and unwrap the template

PR #18902, opened 2025-04-26
Remove Speedy Rituals feat

ammalagonc:
The effects were folded into the Efficient Rituals feat


TikaelSol:
Add to the remaster journal?

PR #18919, opened 2025-04-27
Fix toggling of modifiers in `DamageModifierDialog` and `CheckModifiersDialog`

In3luki:
Looks like upstream is eating click events for elements with a `disabled` class.


src/module/apps/compendium-browser/components/filters/traits.svelte
```diff
@@ -135,7 +135,6 @@
         display: flex;
 
         input[type="radio"] {
-            --checkbox-size: var(--font-size-15);
             margin: 0 5px 0 3px;
```
stwlam, line 138:
```suggestion
            margin-top: 0;
```

PR #18920, opened 2025-04-27
Update `CompendiumBrowser` for v13

In3luki:
The new svelte version started to complain about altering the filter data inside the `ClearFilterButton` component. Now only a function is passed down to avoid having to bind the data multiple times.
Interestingly this apparently also fixed #18861


stwlam:
Looks good, but one thing that still needs doing is removing a good amount of the bespoke application styling in favor of core. V12 hadn't quite built that out for ApplicationV2, but there is much new in place now.

PR #18927, opened 2025-04-28
Remove `TinyMCE` dependency and add stubbed `TinyMCE` types

In3luki:
I used text search to find all `tinymce` imports as type checking is still a bit difficult in the dev branch. I think this covers all uses.
The format interfaces could probably be pared down some more if desired.

### DIFF
--- a/types/foundry/client/global.d.mts
+++ b/types/foundry/client/global.d.mts
@@ -5,7 +5,7 @@ import clipperlib from "js-angusj-clipper";
 import PixiJS from "pixi.js";
 import * as Showdown from "showdown";
 import * as SocketIO from "socket.io-client";
-import * as tinymce from "tinymce";
+import { TinyMCE as tinymce } from "./../tinymce-stub.mjs";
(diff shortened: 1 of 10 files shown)


stwlam (commented):
Let's have this live outside of types/foundry/client (alongside util.d.mts).

types/foundry/client/global.d.mts
```diff
@@ -14,7 +13,6 @@ declare global {
         export import CONST = constants;
         export import Hooks = foundry.helpers.Hooks;
         export import PIXI = PixiJS;
-        export import TinyMCE = tinymce;
```
stwlam, line 17: I think we can keep this global, right?

In3luki: Yeah probably. I wasn't sure if importing was preferred with the new file structure.

stwlam: so global or alias it as "tinymce"

In3luki: global works best I think.

PR #18965, opened 2025-05-01
Update Tengu Fan Item usage, description text and publication source

ammalagonc:
- Updated item to one-hand-held usage per PFS Clarification.
- Updated item text and publication source to Pathfinder Player Core 2
- Lint


ammalagonc:
Close to redo - this PR is borked

PR #18977, opened 2025-05-03
Convert AfflictionSystemData to DataModel and update sheet

CarlosFdez:
I added it back to the template, but omitting the data model in production and stripping it out in load should hopefully be enough of a disincentive.

### DIFF
--- a/src/module/item/affliction/data.ts
+++ b/src/module/item/affliction/data.ts
@@ -1,16 +1,244 @@
 import type { SaveType } from "@actor/types.ts";
-import type { ItemUUID } from "@client/documents/_module.d.mts";
-import {
-    AbstractEffectSystemData,
-    AbstractEffectSystemSource,
-    DurationData,
-    EffectAuraData,
-    EffectContextData,
-    TimeUnit,
-} from "@item/abstract-effect/index.ts";
-import type { BaseItemSourcePF2e, ItemFlagsPF2e } from "@item/base/data/system.ts";
+import { SAVE_TYPES } from "@actor/values.ts";
+import type { ModelPropsFromSchema, SourceFromSchema } from "@common/data/fields.mjs";
+import type { ItemUUID } from "@common/documents/_module.mjs";
+import { EffectAuraData, EffectExpiryType, TimeUnit } from "@item/abstract-effect/index.ts";
+import { EffectContextField } from "@item/abstract-effect/schema.ts";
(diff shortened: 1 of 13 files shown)


src/module/item/affliction/data.ts
```diff
@@ -1,90 +1,383 @@
...
+        >;
+    }>;
+    save: fields.SchemaField<{
+        type: fields.StringField<SaveType, SaveType, true, false, true>;
+        value: fields.NumberField<number, number, true, false, true>;
+    }>;
+    /** The current status of the affliction, including the current stage and whether it is currently in the onset phase */
+    status: fields.SchemaField<{
+        /** If true, this affliction is in the onset stage */
+        onset: fields.BooleanField<boolean, boolean, false, false, true>;
```
stwlam, line 171: Just make anything that has an initial value required

CarlosFdez: Done. The original onset type was optional, but in a data model world this doesn't make sense to be optional anyways.

src/module/item/affliction/data.ts
```diff
@@ -1,90 +1,383 @@
...
+};
+
+type EffectContextTargetSchema = {
+    actor: fields.DocumentUUIDField<ActorUUID, true, false>;
+    token: fields.DocumentUUIDField<TokenDocumentUUID, false, true, true>;
+};
+
+type EffectContextRollSchema = {
+    degreeOfSuccess: fields.NumberField<DegreeOfSuccessIndex, DegreeOfSuccessIndex, false, true, true>;
+    total: fields.NumberField<number, number, true, false>;
```
stwlam, line 350:
```suggestion
    total: fields.NumberField<number, number, true, false, true>;
```

There's a mistaken in `NumberField`: only if it is _nullable_ is there a default initial value.

CarlosFdez:
I ended up moving EffectContextField to abstract-effect/schema.ts. I see uses of schema.ts and model.ts so if you prefer the other let me know.

src/module/item/affliction/data.ts
stwlam: In v13-dev, setting `required` on a `Numberfield` to true has `initial` defaulting to have a value. Since it's omitted here, I assumed that was your intent. But if the opposite is your intent, fill it in that way.

CarlosFdez: The opposite was my intent, but your suggestion had set initial set to true when your words seemed to imply otherwise.
I changed it already though.

PR #18978, opened 2025-05-03
Reduce divider size in effects panel

CarlosFdez:
The hr size in v12 is very small, but in v13 its currently taller than even a single icon.

After tweak:
image

### DIFF
--- a/src/styles/system/_effects-panel.scss
+++ b/src/styles/system/_effects-panel.scss
@@ -121,7 +121,7 @@
     }
 
     > hr {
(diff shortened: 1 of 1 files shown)


stwlam:
Is it v13 styling making it faded?

CarlosFdez:
Yeah, v13 horizontal rules are a linear gradient instead of a hard white line (which is what v12 did). I kept it because its kinda pretty but we can change it.

stwlam:
It looks alright to me

PR #18985, opened 2025-05-03
Fix managed-by-rbv indicators in `SceneConfigPF2e`

stwlam:


src/module/apps/roll-inspector/app.ts
```diff
@@ -31,10 +31,6 @@ class RollInspector extends SvelteApplicationMixin(fa.api.ApplicationV2) {
         return this.options.message;
     }
 
-    constructor(options: Partial<RollInspectorConfiguration>) {
```
CarlosFdez, line 34: If this is no longer needed for the types to work, hell yeah!

stwlam: I was wondering why it was there. There's no objection from tsc when it's gone, so maybe the version bump fixed it?

CarlosFdez: yeah that was so calling new RollInspector with the additional args didn't cause it to choke.

PR #18988, opened 2025-05-03
Continue filling out and updating v13 types

In3luki:
I added defaults to the `Config` interface to allow importing to other files. The order changed because generics with defaults apparently can't be referenced before they are declared.


In3luki:
Partially superseded by #19002 and stuck in rebase hell. I'll merge the rest with the next PR.

PR #18991, opened 2025-05-04
Add expend field to weapons

CarlosFdez:
I can run the migration after. This also converts the SF2e playtest weapons and implements the Rowan Rifle


static/template.json
```diff
@@ -393,6 +393,7 @@
                 "value": 0
             },
             "range": null,
+            "expend": 0,
```
stwlam, line 396: This should be `null` by default since non-ammo consuming weapons don't have an expend (rather than getting free shots)

CarlosFdez: A normally ammo-consuming weapon with a 0 expend would become a non-ammo consuming weapon, and a melee weapon that uses batteries to attack (not a thing yet, its just per day) with would become an ammo consuming weapon. In that respect, expend 0 means "non-ammo consuming". There doesn't seem to be any distinction between 0 or null.

If that's not convincing I'll change it to null but that's why I went with expend 0. The distinction might then be that null means we won't show the text field at all, though we'll be caught off guard if an expending melee weapon comes out.

If we classify melee or thrown as null expend, would https://2e.aonprd.com/Equipment.aspx?ID=1185 be 0?

stwlam: https://2e.aonprd.com/Weapons.aspx?ID=348 ?

CarlosFdez: That one's instane! I can't even decide if that's null or 0. Perhaps star gun should be null then?

stwlam: Yeah it's not really fair to throw that one in the mix. But maybe we can think of a way to fit it in somehow.

CarlosFdez:
Added support for 0 expend ranged weapons properly as well as swapping to the default value when swapping between melee/ranged properly.

PR #18995, opened 2025-05-04
Update character and vehicle sheets for v13

CarlosFdez:
I recommend disabling white space diffs when viewing this one.

Specifically, this fixes the red underline header buttons and versatile damage button heights, the font of the coinage and wealth, the font color becoming white in the strike summary, and sets the primary font for characters sheets to be what it was in v12. The rest is a partial refactor to split vehicle css from characters. Once this gets in, I'll follow up with vehicles (which likely has similar breaks).


src/styles/actor/_coinage.scss
```diff
@@ -123,23 +123,19 @@
     justify-content: flex-end;
     padding: 0 0.25rem;
 
-    h3 {
+    .item-name {
+        color: var(--color-text-light-0);
```
stwlam, line 127: I'd prefer to shy away from appv1 variables if possible: how does `--color-light-1` fare here?

src/styles/actor/_index.scss
```diff
@@ -225,6 +225,10 @@
             grid-area: nav;
         }
 
+        button {
+            min-height: unset;
+        }
```
stwlam, line 230: ?

CarlosFdez: Foundry v13 adds min-height: var(--button-size); to the global styling. You can see the biggest break in the strikes list on a versatile weapon.

The more appropriate place might be reset.css, and I'm considering doing that if I see it fixes issues on other sheets.

stwlam: Just set `--button-size` instead of a height

src/styles/actor/_coinage.scss
CarlosFdez: Let me try some out, I didn't catch that it was using an old css variable. Its not quite the same but it looks fine to me:

image

stwlam: If we want to use whatever `--color-text-light-0` is, let's copy its value over to a system color variable

src/styles/actor/_index.scss
CarlosFdez: Should we be setting theme variables on a per element basis? I actually kinda regret doing that on the party sheet tags, since theme variables are a thing that should be set more globally to the sheet itself I think for best compatibility.

stwlam: It's not a theme variable (light and dark don't have different values).

CarlosFdez: A theme variable isn't necessarily for light vs dark, but also for theoretical third party themes in the future. We have theme variables in pf2e and we don't have a light vs dark.

Granted, button-size is kinda awkward to change globally, so even that's an odd case.

stwlam: I super duper promise that `--button-size` has nothing to do with themes.

CarlosFdez: I could set --button-size: auto in the reset.scss? I'd need to verify tomorrow to make sure everything is fine though and ensure its in a reset layer. If I set --button-size on each button, It'll take me a while as I comb over every button. Either or I'll have to tackle it tomorrow.

stwlam: I don't think you need to do anything with a reset file: just update the buttons that need it.

src/styles/actor/_coinage.scss
CarlosFdez: This change actually looks fine. Its pure white vs slightly off white. Might even be considered an improvement.

PR #19000, opened 2025-05-04
Add `system` CSS layer to Vite HMR

In3luki:
I've looked for a while and this seems to be the best way to do it.

image

### DIFF
--- a/src/styles/vite-hmr.scss
+++ b/src/styles/vite-hmr.scss
@@ -1,0 +1,4 @@
+// Wrap in the system layer for Vite HMR
+@layer system {
+    @import "main.scss";
+}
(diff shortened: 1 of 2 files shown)


CarlosFdez:
works for svelte?

In3luki:
now it does:

image

Svelte does not officially support layers yet and I found this solution in the issue discussing that.

stwlam:
Svelte CSS isn't injected into the layer on build, either.

In3luki:
The generated CSS is included in `pf2e.css`, so on build the layer works automatically for svelte.
I've added a condition to the transformation to only apply to HMR.

PR #19012, opened 2025-05-06
Restore initialization of `hotbarDrop` hook

### LINKED ISSUE #19004: [7.0.5-beta] Drag attack to macro bar not working
Dragging an attack from the character sheet onto the macro bar doesn't work. No macro is added to the bar. There is no error in logs.

foundry: 13.342
pf2e: 7.0.4-beta
no mods installed

Image

In3luki:
Closes #19004


CarlosFdez:
Do you know why the hotbar drop listener was removed before?

CarlosFdez:
I see. It used to be a hotbar subclass, but it became a hotbar hook later, and it was never added to listen. Well, I'll trust you on that I guess, but it'll need careful attention since it was probably never actually wired up for a reason.

In3luki:
Ah now it makes sense why I looked for the subclass when I started this. I thought I misremembered (and didn't check v12) when I found the hook.

PR #19025, opened 2025-05-07
Guns & Gears Iconics

mechamaya:
Adds Guns & Gears iconics actors with images extracted from the free pdfs. I didn't make a baked in token ring or dynamic token image though so that will still need to be done eventually.


TMunsie:
@SpartanCPA has the art for the tokens

SpartanCPA:
Data is good. I sent tokens over for inclusion.

mechamaya:
Art is in, but exportPacks iconics doesn't seem to be extracting the changes to the prototype token I made to link to the new token art and scale correctly.

SpartanCPA:
I'll merge, and see with I can do to tie them in another PR then

PR #19026, opened 2025-05-07
Add support for augmentation items

CarlosFdez:
A lot of the complexity is in the PhysicalItemSystemModel, but luckily a lot of that should carry over to future stuff. Handling max augments will be handled closer to release.

For now it can only be created if the playtest module is enabled. We should find a better way to do that later.


CarlosFdez:
We need more data on whether augmentation items are actually physical items, or if we start needing to create a category for "non-physical items that go in your inventory" and include tattoos. With the implanted usage split out though, SF2e players should have something for now, even if its not ideal.

PR #19029, opened 2025-05-08
Add automation for Fiery Body return damage

pehala:
* Add return damage automation through note
* Unify Fiery Body formatting for heightened variant
   * Use the same wording
   * Use the same paragraphs

PR #19030, opened 2025-05-08
Add automation for return damage to Furnace form

pehala:
Furnace Form feat (https://2e.aonprd.com/Feats.aspx?ID=4247) states:
```
You gain the benefits of the fiery body spell (except the ability to cast ignition) until the end of your next turn
```
However, in the actual effect, there is no mention of return damage, which combined with the fact the feat is just refering to the fiery body and not actually listing the effects itself can create a confusion what actually should be applied.

This PR fixes this issue by introducing automation for return damage.

### DIFF
--- a/packs/feat-effects/effect-furnace-form.json
+++ b/packs/feat-effects/effect-furnace-form.json
@@ -4,7 +4,7 @@
     "name": "Effect: Furnace Form",
     "system": {
         "description": {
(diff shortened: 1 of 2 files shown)


TikaelSol:
Like the other PR we don't add things to effect descriptions that the effects don't automate.

pehala:
v2:
  * Repurposed this PR to add automation, like in #19029

PR #19031, opened 2025-05-08
Address non-passive wheel event listener warning in actor sheet and add a debounce time to delta inputs from wheel events

In3luki:


stwlam:
Why not just refrain from calling `event.preventDefault()`? It's a passive event, so we can't do that.

In3luki:
I could change it to `passive: true` and remove the `preventDefault` but I wasn't sure of the side effects of doing that. I've explicitly set it to `passive: false` in this PR which also suppresses the warning.

stwlam:
I'd prefer to drop this part of the feature before expanding it like this. Try removing the `preventDefault()` call instead.

stwlam:
A debounce of 0 is usually just to send it to the bottom of the job queue (equivalent to `Promise.resolve().then(...)`). But, sure, try removing that.

In3luki:
Should be good now. The wheel still works and no more stuttering or rolling back values after fast scrolling.

In3luki:
I've been testing with the character sheet HP input `passive: true` doesn't prevent the sidebar from scrolling when the input is scrolled.

PR #19039, opened 2025-05-09
Move `Offer Story` to action description and remove `Raging Stories` effect

### LINKED ISSUE #19034: Raging Stories' effect is based on the recipient's proficiency instead of the origin's
Hello! 🖖🏽
Raging Stories is a feat from the Campfire Chronicler archetype from Rival Academies.

> When you Offer a Story about wild chances and fighting against all odds, you can wrap yourself in spectral flame in place of the normal benefits. The action then gains the fire trait. Your melee Strikes for the duration deal 2 additional fire damage. The creature you share a story with can choose this benefit even if you did not.

> The fire damage increases to 4 if you are a master of Religion and 6 if you are legendary.

I'm pretty sure the additional damage should be based on the origin's Religion proficiency, not the recipient, but as the title implies this currently isn't the case in Foundry

ammalagonc:
This way, the effect can be applied to others.

Closes #19034


TikaelSol:
Was raging stories ever linked anywhere?

ammalagonc:
No, it was sitting in the compendium all by it's lonesome

PR #19040, opened 2025-05-09
Fix data model issues in affliction

CarlosFdez:


src/module/item/abstract-effect/helpers.ts
```diff
@@ -1,9 +1,9 @@
+import type { AbstractEffectPF2e, AfflictionPF2e, ConditionPF2e, EffectPF2e } from "@item/index.ts";
 import { DurationData } from "./data.ts";
-import type { AbstractEffectPF2e } from "./document.ts";
 import { DURATION_UNITS } from "./values.ts";
 
 export function calculateRemainingDuration(
-    effect: AbstractEffectPF2e,
+    effect: AfflictionPF2e | ConditionPF2e | EffectPF2e | AbstractEffectPF2e,
```
CarlosFdez, line 6: as annoying as this was, this prevents the excessively deep error.

PR #19041, opened 2025-05-09
Convert EffectSystemData and ConditionSystemData to DataModel

CarlosFdez:
For now the typedatamodel lifecycle methods are unused, but I added the types


src/module/item/abstract-effect/schema.ts
```diff
@@ -126,5 +126,68 @@ type DurationDataSchema = {
...
+type EffectBadgeValueSource = SourceFromSchema<EffectBadgeValueSchema> & {
+    min?: never;
+    max?: never;
+};
+
+type EffectBadgeFormulaSource = Omit<SourceFromSchema<EffectBadgeFormulaSchema>, "evaluate"> & {
+    evaluate?: boolean;
+    min?: never;
+    max?: never;
+};
```
stwlam, line 182: Can these be interfaces instead of intersections?

CarlosFdez: Seems to work fine. I mostly went with the flow for extending schemas, which do use intersections instead of inheritance.

CarlosFdez:
Lumped condition into the mix.  Should be good to go now. There's some type errors in some places like grant item still but they did seem to be typescript bugs, and they're not new.

PR #19045, opened 2025-05-09
Improve `MeasuredTemplate` preview rotation UX for cones and lines

### LINKED ISSUE #4618: Dragable Chat Template Placement Enchacements

### LINKED ISSUE #18552: Slow rotate of template before placement uses wrong keybind on mac

### LINKED ISSUE #18813: Feature request: Template placement with click-drag-click

In3luki:
V13 version of #18819.

I've tried including the changes from #15552 but that lead to some weird behavior where the template position moved away from the initial square after a few rotations.

- Closes #4618 
- Closes #18552 
- Closes #18813

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -84,48 +84,41 @@ class TemplateLayerPF2e<TObject extends MeasuredTemplatePF2e = MeasuredTemplateP
     }
 
     #activatePreviewListeners(preview: TObject, initialLayer: InteractionLayer | null): void {
-        let lastMove = Date.now(); // Throttle 25ms
+        let lastMove = Date.now(); // Throttle 50ms
 
         const listeners: TemplatePreviewEventListeners = (this.#previewListeners = {
+            locked: false,
             mousemove: (event: PIXI.FederatedPointerEvent): void => {
                 event.stopPropagation();
                 const now = Date.now();
-                if (now - lastMove <= 25) return;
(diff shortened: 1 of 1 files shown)


CarlosFdez (approved):

In3luki:
I changed it to `lockedInPlace`.

CarlosFdez:
That works! I'll merge it tomorrow and play around with it to see how it feels, unless someone else wants to merge it before then.

Thanks for doing this. Laptop users will be pleased.

PR #19064, opened 2025-05-11
Correct treatment of Aquatic Combat Penalty

### LINKED ISSUE #18957: Amphibious creatures are incorrectly affected by the -2 circumstance penalty to B and S attacks from the Aquatic Combat Effect
Current Condition: When the Aquatic Combat Effect is applied to an actor with the Amphibious trait, the -2 circumstance penalty to bludgeoning and slashing attacks is still being applied.

Should be: The -2 circumstance penalty should not apply to actors with the Amphibious trait.

SpartanCPA:
Closes #18957

### DIFF
--- a/packs/other-effects/effect-aquatic-combat.json
+++ b/packs/other-effects/effect-aquatic-combat.json
@@ -49,7 +49,7 @@
                                     "self:trait:aquatic"
                                 ]
                             },
-                            "item:category:unarmed",
+                            "item:type:melee",
                             "self:type:npc"
                         ]
                     },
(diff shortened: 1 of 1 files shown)


TikaelSol:
The penalty does apply to ranged strikes though.

> You take a –2 circumstance penalty to your attack roll for bludgeoning or slashing attacks that pass through water.

We should lose the unarmed bit though

SpartanCPA:
> The penalty does apply to ranged strikes though.
> 
> > You take a –2 circumstance penalty to your attack roll for bludgeoning or slashing attacks that pass through water.
> 
> We should lose the unarmed bit though

Addressed

PR #19065, opened 2025-05-11
Convert ClassSystemData to TypeDataModel

CarlosFdez:


stwlam:
We can't safely convert this because of relatively recent breaking migrations (see migration 927).

stwlam:
If we're going to do this now, can we use the standard `DataModel.migrateData` method?

CarlosFdez:
927 is fine for class, it changes from abbreviations to longform, and that's temporarily accommodated for. But we definitely cannot do backgrounds, that had a structural change.

Do we bother with the migrateData() now for the skills, even though that's going to be handled? We might have to do it with background though.

stwlam:
Yeah, I guess we can live with this solution for now.

src/module/item/abc/data.ts
```diff
@@ -1,4 +1,7 @@
+import type { ImageFilePath } from "@common/constants.mjs";
+import type { ItemUUID } from "@common/documents/_module.mjs";
```
stwlam, line 2:
```suggestion
import type { ImageFilePath } from "@common/constants.d.mts";
import type { ItemUUID } from "@common/documents/_module.d.mts";
```

src/module/item/base/sheet/sheet.ts
```diff
@@ -1,6 +1,7 @@
 import type { FormSelectOption } from "@client/applications/forms/fields.d.mts";
 import type { ProseMirrorEditor } from "@client/applications/ux/_module.d.mts";
 import type { ApplicationV1HeaderButton, AppV1RenderOptions } from "@client/appv1/api/application-v1.d.mts";
+import type { SchemaField } from "@common/data/fields.mjs";
```
stwlam, line 4:
```suggestion
import type { SchemaField } from "@common/data/fields.d.mts";
```

src/module/item/class/data.ts
```diff
@@ -1,47 +1,201 @@
...
+        "occ",
+        "prf",
+        "rel",
+        "soc",
+        "ste",
+        "sur",
+        "thi",
+        ...Object.keys(CONFIG.PF2E.skills),
+    ] as unknown as SkillSlug[];
 }
```
stwlam, line 35: maybe we can save some lines here by defining the short names separately?

src/module/model.ts
```diff
@@ -8,6 +9,19 @@ class RarityField extends fields.StringField<Rarity, Rarity, true, false, true>
     }
 }
 
+class ProficiencyRankField extends fields.NumberField<ZeroToFour, ZeroToFour, true, false, true> {
+    constructor() {
+        super({
+            choices: R.mapValues({ ...CONFIG.PF2E.proficiencyLevels } as Record<ZeroToFour, string>, (v) =>
+                game.i18n.localize(v),
+            ),
```
stwlam, line 17: Can't localize because the field can be constructed before localization is initialized.

static/templates/items/class-details.hbs
```diff
@@ -24,69 +21,23 @@
...
-        <select name="system.attacks.unarmed" data-dtype="Number">
-            {{selectOptions proficiencyChoices selected=data.attacks.unarmed localize=true}}
-        </select>
-    </div>
+    {{formGroup systemFields.attacks.fields.simple value=data.attacks.simple}}
+    {{formGroup systemFields.attacks.fields.martial value=data.attacks.martial}}
+    {{formGroup systemFields.attacks.fields.advanced value=data.attacks.advanced}}
+    {{formGroup systemFields.attacks.fields.unarmed value=data.attacks.unarmed}}
 
     <div class="form-group other-weapons">
```
stwlam, line 42: Try `{{formInput}}`

static/templates/items/class-details.hbs
```diff
@@ -178,10 +100,9 @@
         </ul>
     </div>
 
-
     <div class="form-group additional-language">
         <label>{{localize "PF2E.AdditionalSkills"}}</label>
-        <input type="number" name="system.trainedSkills.additional" value="{{data.trainedSkills.additional}}">
+        {{formInput systemFields.trainedSkills.fields.additional value=data.trainedSkills.additional}}
         <span>{{localize "PF2E.AdditionalLanguagePlusInt"}}</span>
     </div>
```
stwlam, line 107: Isn't this just a standard form group?

src/module/model.ts
CarlosFdez: So go with localize=true in the proficiency inputs after all like I did initially?

stwlam: Localize them in the `getData` as needed

CarlosFdez: We need them localized in {{formGroup}} and {{formInput}} specifically. They're smart enough to render number fields as selects if a choices exists.

static/templates/items/class-details.hbs
CarlosFdez: For what? The form groups?

src/module/item/abc/data.ts
CarlosFdez: Done but we need a way to inform the editor to pick the right import somehow

src/module/item/class/data.ts
CarlosFdez: Done. Needed to be moved anyways since it seems deity needs the same treatment, and likely NPC as well.

src/module/model.ts
stwlam: They can (and should) be passed to `{{formGroup}}` already localized

CarlosFdez: for that to happen, don't they need to be localized in the data model itself?

stwlam: Like I said, do it in the `getData` method of the sheet.

CarlosFdez: Actually, at least in spanish as is, its working fine (note: labels are new fields so they're not translated yet)?

image

image

stwlam: Don't use `localize=true` because it'll double localize.

stwlam (changes requested):
Needs localization fix

stwlam (approved):

PR #19066, opened 2025-05-11
Change grid check location for aura

FolkvangrForgent:
## Description
I think I have found a work around for aura support on other grids (hex and gridless) however to avoid duplicating large sections of existing code I would still like to see the small change proposed in #16543 that moves the grid check from `checkAuras` to `containsToken`. I have not seen any adverse effect from this change in months of using my previous patch.
## Changes
- Move grid type check from `checkAuras` function to `containsToken`.

### DIFF
--- a/src/module/scene/helpers.ts
+++ b/src/module/scene/helpers.ts
@@ -7,7 +7,7 @@ let auraCheckLock = Promise.resolve();
 
 /** Check for auras containing newly-placed or moved tokens */
 const checkAuras = fu.debounce(async function (this: ScenePF2e): Promise<void> {
-    if (!(canvas.ready && this.isInFocus && this.grid.type === CONST.GRID_TYPES.SQUARE)) {
+    if (!(canvas.ready && this.isInFocus)) {
         return;
     }
 
--- a/src/module/scene/token-document/aura/index.ts
+++ b/src/module/scene/token-document/aura/index.ts
@@ -70,7 +70,13 @@ class TokenAura implements TokenAuraData {
     /** Does this aura overlap with (at least part of) a token? */
     containsToken(token: TokenDocumentPF2e): boolean {
         // If either token is hidden or not rendered, return false early
-        if (this.token.hidden || token.hidden || !this.token.object || !token.object) {
+        if (
+            this.token.hidden ||
+            token.hidden ||
+            !this.token.object ||
+            !token.object ||
+            this.scene.grid.type !== CONST.GRID_TYPES.SQUARE
+        ) {
             return false;
         }
 


stwlam:
Why would we want to check the grid type for every single token instead of once?

FolkvangrForgent:
Other than make my life much easier as a pf2e module developer, I believe that moving the check to `containsToken` is the better option as it is caller of square grid dependent functions (`measureDistanceCuboid` with a grid check and `getAreaSquares` without). `checkAuras` other than calling `containsToken` does not call functions that are square grid dependent. I would personally prefer to see the square grid check moved to `getAreaSquares` as that function is really just a square grid function. However due to the `PartyClownCar` use of that function I would want to modify the `PartyClownCar` logic to be grid independent (although gridless may require it's own code as my initial `getAdjacentOffsets` approach returns `[]` on gridless would also add support for large+ creatures while at it). Although with the systems apathy for other grids breaking `PartyClownCar` on non square grids may be acceptable? If the system raw doesn't care I can change the check to `getAreaSquares` as I can probably just override the clown car button in my module.

stwlam:
This isn't a sensible change, and I don't think we're going to reach a mutual understanding here.

FolkvangrForgent:
changing `getAreaSquares` to properly check grid type would look as such: https://github.com/foundryvtt/pf2e/commit/fef7c8b769218053b1b87a1ce1ea9f05e32ab61a

The main two changes of this without considering lowering the number of functions that I will have to patch are:
- Auras circle will render on non square grids
- Clown car deposit longer spreads tokens out on non square grids

If the second point is a deal breaker I can work on a MR that also adds a square/hex algorithm and a gridless algorithm (hopefully supporting all size categories) to the clown car although that may take more time. If the second point is not a deal breaker than I can open that branch as another MR.

CarlosFdez:
This is closest to the last invasive correct way, the issue is that its functionally worse for people who are not using hex grids. You have to remember that splash screens and gridless scenes are also a thing.

Is there a place you could add a isGridTypeSupported getter that would allow you to shortcircuit that you can simultaneously access outside of the module?

The thing to keep in mind is that whatever the end result is, it has to look as seamless as possible for the system itself.

CarlosFdez:
Once you find something I'll open it, but don't worry about party clown car. Dumping into adjacent spaces is a thing that can be solved on our own if someone puts effort into it. Its not like we're homebrewing pf2e.

FolkvangrForgent:
@CarlosFdez quick clarification are you referring to the code in this MR or the code in this MR https://github.com/foundryvtt/pf2e/pull/19074?

CarlosFdez:
This MR. The issue with this one is that checkAuras doesn't exit out sufficiently early.

CarlosFdez:
From my quick glance, what I mentioned could probably go in ScenePF2e. checkAuras is bolted onto the ScenePF2e class for the purposes of debouncing.

FolkvangrForgent:
With the flag I wouldn't the `containsToken` change was just there if the check was removed from the `checkAuras`

CarlosFdez:
The tricky thing is what to call the getter. canShowAura? isAuraSupported()? The isInFocus and canvas.ready check would likely be moved there, and AuraRenderers#showBordersHighlight would use that instead of the grid type and in focus check.

FolkvangrForgent:
Hmm MR not updating (probs my fault with git fixup and force push rather than another commit) but https://github.com/FolkvangrForgent/pf2e/commit/86d821cd5fab7d56eef090e23a26cc20d2bcd9bd is the new commit. This adds a getter `isAuraSupported` that has `(this.isInFocus && this.grid.type === CONST.GRID_TYPES.SQUARE)`. I left out `canvas.ready` as it isn't strictly necessary in the `AuraRenderers` function check.

CarlosFdez:
maybe its cause its closed, try again. You might have to set the remote.

FolkvangrForgent:
I feel like this may be a github feature to not update closed MRs given the link from this MR links to the branch with the correct commit.

CarlosFdez:
If you think so on canvas.ready go for it. Double checking canvas.ready actually doesn't bother me.

After some discussion and thinking, isAuraSupported() isn't correct if we're including can render checks. Please rename to canHaveAuras and only include the grid type check. in focus is something we may want more control over especially if we plan to address things like teleporter regions.

FolkvangrForgent:
Update name to `canHaveAuras` and moved the `isInFocus` back out. commit: https://github.com/FolkvangrForgent/pf2e/commit/43953801c8cec4d579fe308e79dcb7bc74fb9a9c

Also yeah I was being quite stupid and that's why I couldn't libwrapper it.

CarlosFdez:
Make a new PR. You are allowed this one last time for this topic!

PR #19072, opened 2025-05-11
Prevent normal drag operations when a `MeasuredTemplate` preview is active

In3luki:
Prevents the creation of a template when the preview is placed with a click-drag instead of a normal click.

### DIFF
--- a/src/module/canvas/layer/template.ts
+++ b/src/module/canvas/layer/template.ts
@@ -83,6 +83,18 @@ class TemplateLayerPF2e<TObject extends MeasuredTemplatePF2e = MeasuredTemplateP
         return template.rotate(template.document.direction + delta, snap);
     }
 
+    protected override _onDragLeftStart(event: PlaceablesLayerPointerEvent<TObject>): void {
+        // Prevent normal drag operations when a preview is active
+        if (this.#previewListeners) return;
+        return super._onDragLeftStart(event);
+    }
+
+    protected override _onDragLeftCancel(event: PlaceablesLayerPointerEvent<TObject>): void {
+        // Prevent normal drag operations when a preview is active
+        if (this.#previewListeners) return;
+        return super._onDragLeftCancel(event);
+    }
+
     #activatePreviewListeners(preview: TObject, initialLayer: InteractionLayer | null): void {
         let lastMove = Date.now(); // Throttle 50ms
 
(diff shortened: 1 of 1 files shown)


types/foundry/client/canvas/layers/base/placeables-layer.d.mts
```diff
@@ -381,7 +381,9 @@ export default class PlaceablesLayer<TObject extends PlaceableObject = Placeable
 
     protected override _onDragLeftDrop(event: PlaceablesLayerPointerEvent<TObject>): void;
 
-    protected override _onDragLeftCancel(event: PlaceablesLayerPointerEvent<TObject>): void;
+    protected override _onDragLeftCancel(
+        event: PIXI.FederatedPointerEvent | PlaceablesLayerPointerEvent<TObject>,
+    ): void;
```
stwlam, line 386: ?

In3luki: That's the v12 defintion. Probably for the same reason as now.
`canvas.stage.on` only takes `PIXI.FederatedPointerEvent`s so the preview listeners can't be `PlaceablesLayerPointerEvent<TObject>` and the call to `TemplateLayerPF2e#_onDragLeftCancel` in `#deactivatePreviewListeners` doesn't take the `PIXI.FederatedPointerEvent`.

Using a type assertion would also work:
```ts
    protected override _onDragLeftCancel(
        event: PIXI.FederatedPointerEvent | PlaceablesLayerPointerEvent<TObject>,
    ): void {
        // Prevent normal drag operations when a preview is active
        if (this.#previewListeners) return;
        return super._onDragLeftCancel(event as PlaceablesLayerPointerEvent<TObject>);
    }
```

stwlam: But a `PlaceablesLayerPointerEvent` _is_ a `PIXI.FederatedPointerEvent`. If a method won't accept a `PIXI.FederatedPointerEvent`, it also won't accept a `PlaceablesLayerPointerEvent`.

In3luki: yeah but typescript complains that `interactionData` could be missing in the `PIXI.FederatedPointerEvent` because it's not optional in `PlaceablesLayerPointerEvent`.
And I don't think the type of `canvas.stage.on` events can be changed as those are internal pixi types I believe.

PR #19074, opened 2025-05-11
Change grid check location to function that is actually grid dependant

FolkvangrForgent:
## Description
Moves a grid check to functions that actually contains grid dependent code. This is a sensible approach to limiting function behavior that assume a type of grid and improves readability in that regards. This limits `getAreaSquares` to square grid as that is the grid type the function is written for. The only downside of that is the `PartyClownCar` uses that function even on non square grids. In my opinion that function is deficient for use in `PartyClownCar` as it does not support Large, Huge, or Gargantuan creatures which will overlap other tokens. If the `PartyClownCar` support on non square grids is a deal breaker I can mark this MR as draft and implement a square/hex algorithm and a gridless algorithm for this purpose then unmark this MR as draft. 

#16543 - issue with MR was that system dev(s) did not want to expose Aura classes (I have worked around this)
#19066 - issue with MR was that system dev(s) mentioned the approach was not sensible

## Changes
- Remove grid type check from `#showBordersHighlights` and `checkAuras` functions
- Add grid type check to `getAreaSquares` function
## Effects
- Auras circle will render on non square grids
- Clown car deposit longer spreads tokens out on non square grids

### DIFF
--- a/src/module/canvas/token/aura/map.ts
+++ b/src/module/canvas/token/aura/map.ts
@@ -53,7 +53,7 @@ export class AuraRenderers extends Map<string, AuraRenderer> {
                 : this.token.combatant?.encounter.active);
 
         return (
-            canvas.scene?.grid.type === CONST.GRID_TYPES.SQUARE &&
+            canvas.scene !== null &&
             // Assume if token vision is disabled then the scene is not intended for play.
             canvas.scene.tokenVision &&
             // The scene must be active, or a GM must be the only user logged in.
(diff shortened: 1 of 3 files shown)


stwlam:
Stop making PRs for anything having to do with this.

FolkvangrForgent:
@stwlam please tell me what is wrong with this change. I don't understand why the system or you are so hostile to the smallest changes in regard to this area of code. The comparisons that this or the previous MR will add will be negligible in V8 especially once the function compiles down. I really enjoy the system but working with it especially when it comes to other grids is such a pain and even the smallest change in such regard is always met with extreme hostility.

PR #19083, opened 2025-05-12
Convert `WorldClockSettings` to `ApplicationV2`

stwlam:


CarlosFdez:
Kill worldCreatedOn!

stwlam:
? It got rolled into the setting object with the rest

CarlosFdez:
If a module wants to figure out the true world time, they need to get the world created on time and then add it to the world time. This is a system specific quirk we have, and it means any module that wants to do a system agnostic computation for time of day needs to have specific handling for pf2e. I think instead to make sure worldTime alone is correct, we outta update the world time when the world is first created, and use a worldClockInitialized var to remember that we did that.

This a complaint I received months ago in a DM that called my attention to it, but it was low priority and they had a solution in the meantime, but it nags on me every now and again. Now that you're doing a worldclock migration its the perfect time to fix this! If its fine to you, I can make a commit if you don't wanna deal with it?

stwlam:
Make a separate issue: the setting can't be removed without doing something instead to keep the world clock baseline functional.

CarlosFdez:
Done https://github.com/foundryvtt/pf2e/issues/19085

It can be this issue, or another migration. Yeah, I can imagine there might be more complications than refactoring the getters we created.

CarlosFdez:
The time to resolve what we do with the time api is going to be longer than desired for converting this, and im of the strong opinion that if its an improvement over the previous, then its fine.

PR #19084, opened 2025-05-12
Update v13 `ProseMirror` types

In3luki:
Most of the files in `prosemirror/schema` could probably be omitted.

### DIFF
--- a/types/foundry/common/prosemirror/_module.d.mts
+++ b/types/foundry/common/prosemirror/_module.d.mts
@@ -23,23 +23,23 @@ import { parseHTMLString, serializeHTMLString } from "./util.mjs";
 declare const dom: {
     parser: DOMParser;
     serializer: DOMSerializer;
-    parseString: parseHTMLString;
-    serializeString: serializeHTMLString;
+    parseString: typeof parseHTMLString;
+    serializeString: typeof serializeHTMLString;
 };
 
 declare const defaultPlugins: {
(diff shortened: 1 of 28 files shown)


types/foundry/common/prosemirror/_module.d.mts
```diff
@@ -23,23 +23,23 @@ import { parseHTMLString, serializeHTMLString } from "./util.mjs";
...
+    isDirty: Plugin;
+    clickHandler: Plugin;
+    pasteTransformer: Plugin;
+    baseKeyMap: Plugin;
+    dropCursor: Plugin;
+    gapCursor: Plugin;
+    history: Plugin;
+    columnResizing: Plugin;
+    tables: Plugin;
 };
```
stwlam, line 43: Are these prosemirror plugins or foundry plugin classes? Does the foundry one implement the prosemirror one?

In3luki: `ProseMirrorPlugin` is not an extension of `ProseMirror.Plugin`. The static `create` method builds a `ProseMirror.Plugin` and returns `ProseMirrorPlugin#render` (html) for the editor view.

`ProseMirrorMenu` looks like this:
```ts
  static build(schema, options={}) {
    return new Plugin({
      view: editorView => {
        return new this(schema, editorView, options).render();
      }
    });
  }
```

PR #19087, opened 2025-05-13
Support message item retrieval for damage from synthetic strikes

CarlosFdez:
Mostly an API bug fix right now, this allows message.item to resolve for unarmed attack damage (currently it is only supported for real items). There are also type fixes.

Should we rename ChatMessagePF2e's `_strike` getter to just `strike`? It needs to be public, and I doubt strike retrieval will change soon.


stwlam:
Public is fine, but it needs to be understood as non-contractual. Maybe mark it `@internal`? Or just warn Idle.

CarlosFdez:
I'll keep it as _strike then, since that can be used to understand it as non-contractual, but I added `@internal`. There's nothing really wrong persay with the strike getter, but more that strikes haven't jumped to the statistic api yet.

PR #19094, opened 2025-05-15
Update v13 `ui` types

In3luki:


types/foundry/client/ui.d.mts
```diff
@@ -26,4 +40,48 @@ export const pause: applications.ui.GamePause;
...
+    pause: typeof pause;
+    players: typeof players;
+    playlists: typeof playlists;
+    scenes: typeof scenes;
+    settings: typeof settings;
+    sidebar: typeof sidebar;
+    tables: typeof tables;
+    webrtc: typeof webrtc;
+    windows: typeof windows;
+}
```
stwlam, line 87: would prefer the types be here directly instead of ones derived from constants

PR #19096, opened 2025-05-15
Update `EnrichmentOptionsPF2e` to accept a function as `rollData`

In3luki:
I've renamed the `RollDataPF2e` interface in `text-editor.ts` to `RollDataContextPF2e` because of a collision with `RollDataPF2e` in `system/rolls.ts`.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -198,7 +198,8 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
 
     static processUserVisibility(content: string, options: EnrichmentOptionsPF2e): string {
         const html = createHTMLElement("div", { innerHTML: content });
-        const document = options.rollData?.actor ?? options.relativeTo;
+        const rollData = resolveRollData(options.rollData);
+        const document = rollData.actor ?? options.relativeTo;
         UserVisibilityPF2e.process(html, { document });
 
         return html.innerHTML;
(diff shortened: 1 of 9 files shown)


src/module/system/text-editor.ts
```diff
@@ -1013,14 +1015,17 @@ async function augmentInlineDamageRoll(
...
 interface EnrichmentOptionsPF2e extends EnrichmentOptions {
-    /** @todo as of v13 this should also support functions that produce roll data */
-    rollData?: RollDataPF2e;
+    rollData?: RollDataContextPF2e | (() => RollDataContextPF2e);
     /** Whether to run the enriched string through `UserVisibility.process` */
     processVisibility?: boolean;
 }
 
-interface RollDataPF2e {
+interface RollDataContextPF2e {
```
CarlosFdez, line 1028: Why the rename here?

In3luki: See PR description. There already is an exported `RollDataPF2e` in `system/rolls.ts`. We could keep both but I thought it'd be safest to rename one.

CarlosFdez: Given that the one in system/rolls extends dice.RollOptions, I think that one might be the one that needs renaming. The RollData here is actually called RollData within foundry itself.

In3luki: Yeah makes sense. I've renamed `RollDataPF2e` in `system/rolls.ts` to `BaseRollDataPF2e` and changed the one in `text-editor` back to `RollDataPF2e`.

PR #19100, opened 2025-05-16
Convert ContainerSystemData and EquipmentSystemData to TypeDataModel

CarlosFdez:
I tested with an old actor from the discord and the temporary source migration seems to be converting them correctly. This should allow us to start converting the less touched physical items, though weapons/armor/shields are still off limits right now.


CarlosFdez:
Rebased but since im currently busy, I'd prefer this to start going in this weekend at the earliest. Afterwards I can get started on ammo.

PR #19104, opened 2025-05-17
Fix display of NPC Shield health when 3 digits long

CarlosFdez:
Color are still a bit off but I'd rather not mess with colors outside of the V13 branch if possible

image


stwlam:
Can the base branch for this be v13-dev?

CarlosFdez:
We're doing one final V12 release then rebasing. Might as well get it in. The color thing looks like it wouldn't be easy to resolve in v13 anyways cause of the silly scoped input color variables.

PR #19113, opened 2025-05-17
Automate hurl at the horizon

### LINKED ISSUE #18160: Hurl at the Horizon doesn't add throw range to Ikon
As the title says, I've tried adding Hurl at the Horizon to Shadow Sheath loaded with a dagger and Barrow Blade dagger and it doesn't increase the throw range.

Ikaguia:
Closes #18160

### DIFF
--- a/packs/feats/class/exemplar/hurl-at-the-horizon.json
+++ b/packs/feats/class/exemplar/hurl-at-the-horizon.json
@@ -57,17 +57,20 @@
                 "value": "thrown-15"
             },
             {
-                "key": "Note",
-                "predicate": [
+                "key": "AdjustStrike",
+                "definition": [
+                    "item:thrown",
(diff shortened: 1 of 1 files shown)


TikaelSol:
Does this double dip on items without the thrown trait (add the base 15 then add 10 to that)?

pedrogrullada:
The tag that it's predicated on not having makes sure it only works for weapons that were already thrown.

I'm not sure if the range increment adjustment will work though. Was this tested?

Ikaguia:
Its the exact same predicate that was already there to add the note.
The `item:tag:thrown-from-hurl-at-the-horizon` tag is there to prevent double dipping.
I tested this on my character (not extensively but some) and it seemed to work fine, toggling divine spark on and off disabled it, and it only worked on the weapon with the spark.

pedrogrullada:
If it works then great. I had overlooked that AdjustStrike could do range increment originally.

PR #19147, opened 2025-05-20
Migrate `Signature Weapon` Yaoguai automation effects to feat

ammalagonc:
Don't believe this needs a UUID redirect for the effect, but let me know if it needs one


pedrogrullada:
What's the reason to move the automation? To me this is similar to something like Blessed Armaments where the weapon it's applied to can change with relative ease, and the effect makes that more convenient than having to delete the feat and re-add it, which is why it was implemented that way.

TikaelSol:
Restore the deleted effect to stop broken links from cropping up. We'll just be stuck with a stray effect.

PR #19158, opened 2025-05-20
Fix few RE localization keys

Cuingamehtar:


CarlosFdez:
NOW is when this comes in? 🙃

Cuingamehtar:
Sorry, I just noticed you preparing the release, and decided to check the data before going to bed. Guess I was a bit too late.

For the future, I really recommend `copy-json-path` extension for vscode.

CarlosFdez:
If possible (I know it won't always be) next time let me know to pause and I will do so.

Cuingamehtar:
Ok, I will.

Issue #16190, opened 2024-08-22
Custom Weakness Rule Elements proc for each damage type


JamzTheMan:
Given the below Custom Weakness RE, a Strike with property runes (say, cold, electricity, force) will add the Weakness 4 times.

```json
{
  "definition": [
    "action:strike"
  ],
  "key": "Weakness",
  "label": "Blessed Counterstrike Weakness",
  "type": "custom",
  "value": "7"
}
```

Given that Weaknesses never stack, these should be aggregated together and taken once. (granted there may be arguments for "instances of damage" so worse case an `applyOnce` or `applyAll` option could be added to let it be single or multi instance weakness?)

https://2e.aonprd.com/Rules.aspx?ID=2317
`If more than one weakness would apply to the same instance of damage, use only the highest applicable weakness value. This usually only happens when a creature is weak to both a type of damage and a material or trait, such as a cold iron axe cutting a monster that has weakness to cold iron and slashing.`

stwlam:
An instance of damage is basically a damage type. I agree there's a need for "contact" custom weaknesses similar to how water and salt work, so we'll look into adding an option for a custom weakness to identify itself as such.

pedrogrullada:
Closed by #19183

PR #19183, opened 2025-05-25
Allow custom non-damage weaknesses, fix Blessed Counterstrike weakness

xyzzy42:
Currently non-damage weaknesses are limited to a fixed set hard-coded in the
system with NON_DAMAGE_WEAKNESSES.

Add an optional field so the non-damage weakness type can be set.  If the field
is omitted, it will default to true if the weakness type is in
NON_DAMAGE_WEAKNESSES, false otherwise.

Address issue #16190

This allows creating custom non-damage weaknesses.  E.g., a definition of
"origin:signature:{item|origin.signature}" would create a weakness to strikes
from a specific player who is the origin of the effect with the weakness.  By
setting nonDamage to true, the weakness will only trigger once for a damage
roll, rather than triggering on every damage instance in the roll.

It also makes it possible to create weakness to spells/items with a trait or
traits that do not trigger multiple times on spells with multiple damage types.
This already exists for the magic traditions and a few other traits, which are
in the NON_DAMAGE_WEAKNESSES set, but not arbitrary traits.  E.g., there's no
way to make a weakness to fear spells that does not trigger twice on damage from
Invoke Spirits, which deals both mental and void damage.

### DIFF
--- a/packs/feat-effects/effect-blessed-counterstrike.json
+++ b/packs/feat-effects/effect-blessed-counterstrike.json
@@ -28,6 +28,7 @@
                 ],
                 "key": "Weakness",
                 "label": "PF2E.IWR.Custom.EnemyStrikes",
+                "applyOnce": true,
                 "type": "custom",
                 "value": "floor(@item.origin.level / 2)"
             }
--- a/src/module/actor/data/iwr.ts
+++ b/src/module/actor/data/iwr.ts
@@ -305,9 +305,12 @@ class Weakness extends IWR<WeaknessType> implements WeaknessSource {
 
     override value: number;
 
-    constructor(data: IWRConstructorData<WeaknessType> & { value: number }) {
+    readonly applyOnce: boolean;
+
+    constructor(data: IWRConstructorData<WeaknessType> & { value: number; applyOnce?: boolean }) {
         super(data);
         this.value = data.value;
+        this.applyOnce = data.applyOnce ?? APPLY_ONCE_WEAKNESSES.has(this.type);
     }
 
     override toObject(): Readonly<WeaknessDisplayData> {
@@ -322,6 +325,7 @@ type WeaknessDisplayData = IWRDisplayData<WeaknessType> & Pick<Weakness, "value"
 
 interface WeaknessSource extends IWRSource<WeaknessType> {
     value: number;
+    applyOnce?: boolean;
 }
 
 class Resistance extends IWR<ResistanceType> implements ResistanceSource {
@@ -371,7 +375,7 @@ interface ResistanceSource extends IWRSource<ResistanceType> {
 }
 
 /** Weaknesses to things that "[don't] normally deal damage, such as water": applied separately as untyped damage */
-const NON_DAMAGE_WEAKNESSES: Set<WeaknessType> = new Set([
+const APPLY_ONCE_WEAKNESSES: Set<WeaknessType> = new Set([
     ...MAGIC_TRADITIONS,
     "air",
     "earth",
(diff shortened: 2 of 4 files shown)


stwlam:
I think we're going to reframe this whole thing as apply-once weaknesses

packs/feat-effects/effect-blessed-counterstrike.json
```diff
@@ -28,6 +28,7 @@
                 ],
                 "key": "Weakness",
                 "label": "PF2E.IWR.Custom.EnemyStrikes",
+                "nonDamage": true,
```
stwlam: A bit strange to classify a strike as non-damaging.

src/module/actor/data/iwr.ts
```diff
@@ -301,9 +301,12 @@ class Weakness extends IWR<WeaknessType> implements WeaknessSource {
...
     override value: number;
 
-    constructor(data: IWRConstructorData<WeaknessType> & { value: number }) {
+    readonly nonDamage: boolean;
+
+    constructor(data: IWRConstructorData<WeaknessType> & { value: number; nonDamage?: boolean }) {
         super(data);
         this.value = data.value;
+        this.nonDamage = data.nonDamage ?? NON_DAMAGE_WEAKNESSES.has(this.type);
     }
```
stwlam, line 310: This needs to be part of the model definition

src/module/rules/rule-element/iwr/weakness.ts
```diff
@@ -12,6 +13,7 @@ class WeaknessRuleElement extends IWRRuleElement<WeaknessRuleSchema> {
             ...super.defineSchema(),
             value: new ResolvableValueField({ required: true, nullable: false, initial: undefined }),
             exceptions: this.createExceptionsField(this.dictionary),
+            nonDamage: new fields.BooleanField({ required: false, initial: undefined }),
```
stwlam: Besides reframing, this needs to be locked down such that only custom weaknesses and ones already considered single-application can be marked as single-application.

packs/feat-effects/effect-blessed-counterstrike.json
xyzzy42: I think of as it a weakness to the contact from being struck rather than as a weakness to the damage the strike does.

src/module/rules/rule-element/iwr/weakness.ts
xyzzy42: Should the new name be `applyOnce` or `singleApplication`?  

I can do that, but why is it necessary?  If someone wants to make an apply once weakness that is the same as an existing named weakness, why force it to be custom?

An example might be the existing `axe-vulnerability` weakness, which is I think supposed to be a apply-once weakness.  If someone wants the rule to be that way, why not let them change it?

xyzzy42: Should `applyOnce` of any value be rejected for non-custom weaknesses?  Or should it instead be required that the property, if present, must match the pre-defined apply onceness of the weakness.  So holy + applyOnce:true is ok, and fire + applyOnce:false is ok.

It'd be simpler, if there must be a restriction, to just only allow it for custom.  Otherwise you can't validate fail immediately since the values of either `type` or `applyOnce` might not be resolved.

xyzzy42:
I have renamed to `applyOnce`.

`applyOnce` can now only be specified for custom weaknesses.

Since the weakness type(s) can be an array, it doesn't make sense to also supply `applyOnce`, as it would not be able to match multiple types, some of which might be apply once and some not.  This is not a problem for custom weaknesses, as custom is already required to be a single weakness per RE.

Issue #19186, opened 2025-05-25
[7.0.6-beta] Effect Duration counter size is smaller size


ammalagonc:
V12 - 

![Image](https://github.com/user-attachments/assets/2b43a84c-59b9-4db4-838d-a12d314ee30e)

V13

![Image](https://github.com/user-attachments/assets/29363acd-63a8-4297-8fc9-1badbd2c1679)

PR #19195, opened 2025-05-26
Fix effect duration counter size

doatech:
https://github.com/foundryvtt/pf2e/issues/19186
Input size should be good enough now for 2 digits.
image

### DIFF
--- a/src/styles/item/_effect-sheet.scss
+++ b/src/styles/item/_effect-sheet.scss
@@ -7,7 +7,7 @@
             }
 
             select {
-                max-width: 9em;
+                width: 8em;
             }
         }
     }
--- a/src/styles/item/_effect-sheet.scss
+++ b/src/styles/item/_effect-sheet.scss
@@ -44,7 +44,7 @@
         display: flex;
         align-items: center;
         justify-content: space-between;
-        margin: 0.5rem 0 0.25rem 0;
+        margin: var(--space-8) 0 var(--space-4) 0;
     }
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
We're generally prefering rem units over em, em usually comes up for legacy stuff or minor variations. max-width might actually still make sense here.

Is there anything being cut off in the select when you do that?

src/styles/item/_effect-sheet.scss
```diff
@@ -3,19 +3,19 @@
         &.duration {
             label {
                 max-width: fit-content;
-                margin-right: 0.1em;
+                margin-right: 0.1rem;
```
CarlosFdez, line 6:
```suggestion
                margin-right: var(--space-2);
```

src/styles/item/_effect-sheet.scss
```diff
@@ -44,7 +44,7 @@
         display: flex;
         align-items: center;
         justify-content: space-between;
-        margin: 0.5em 0 0.25em 0;
+        margin: 0.5rem 0 0.25rem 0;
```
CarlosFdez, line 47:
```suggestion
        margin: var(--space-8) 0 var(--space-4) 0;
```

CarlosFdez:
Tyvm!

doatech:
ty for checking and for the additions @CarlosFdez, I will try to refactor more whenever I touch styles from now on 👍

PR #19204, opened 2025-05-27
Update types for multiple document source flags

CarlosFdez:
The actor/item ones are supplementary for correctness but this seems to resolve the UserPF2e not assignable to User thing. `DeepPartial<ModelPropFromDataField<DocumentFlagsField>>` isn't the same as `SourceFromDataField<DocumentFlagsField>`.

Feel free to suggest an alternative to WithPF2eFlags

### DIFF
--- a/src/module/actor/data/base.ts
+++ b/src/module/actor/data/base.ts
@@ -5,7 +5,7 @@ import type { InitiativeTraceData } from "@actor/initiative.ts";
 import type { StatisticModifier } from "@actor/modifiers.ts";
 import type { ActorAlliance, AttributeString, SkillSlug } from "@actor/types.ts";
 import type { Rolled } from "@client/dice/roll.d.mts";
-import type { DocumentFlags } from "@common/data/_module.d.mts";
+import type { DocumentFlags, DocumentFlagsSource } from "@common/data/_module.d.mts";
 import type { MeleePF2e, WeaponPF2e } from "@item";
 import type { ItemSourcePF2e } from "@item/base/data/index.ts";
 import type { MigrationRecord, Rarity, Size, ValueAndMaybeMax, ZeroToTwo } from "@module/data.ts";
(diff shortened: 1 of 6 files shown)


src/module/item/base/data/system.ts
```diff
@@ -51,7 +51,7 @@ type ItemFlagsPF2e = DocumentFlags & {
     };
 };
 
-type ItemSourceFlagsPF2e = DocumentFlags & {
+type ItemSourceFlagsPF2e = fields.SourceFromDataField<fields.DocumentFlagsField> & {
```
stwlam: Define an alias in the same file as `DocumentFlags`

src/module/actor/data/base.ts
```diff
@@ -22,19 +22,21 @@ type BaseActorSourcePF2e<
     TType extends ActorType,
     TSystemSource extends ActorSystemSource = ActorSystemSource,
 > = foundry.documents.ActorSource<TType, TSystemSource, ItemSourcePF2e> & {
-    flags: DeepPartial<ActorFlagsPF2e>;
+    flags: DeepPartial<WithPF2eFlags>;
```
stwlam: This definitely isn't correct: the flags don't only consist of the additional pf2e property

CarlosFdez: The & takes care of that (since we did not omit flags, the types merge). We can add the DocumentFlagsSource, but it becomes redundant. But perhaps that redundancy is fine? Let me know if you want the redundancy, there are some pros to that including clarity.

src/module/item/base/data/system.ts
CarlosFdez: I wasn't certain about adding a type that doesn't exist in the core types to the core types, but since you're fine with it I added it anyways.

src/module/actor/data/base.ts
stwlam: `ActorFlagsPF2e` is the intersection type, not `WithPF2eFlags`.

PR #19206, opened 2025-05-27
Add Expert Spellcasting to Expert Red Mantis Magic Feat

DocSchlock:
Does what it says: gives expert spellcasting prof to the Expert Red Mantis Magic feat as needed.

Also fixes the case in one of the Red Mantis focus spell targets.


DocSchlock:
Closes #19263

PR #19212, opened 2025-05-27
Show item name and UUID after a warning was emitted in `RuleElementPF2e#validate`

In3luki:
I don't think there is a better way to do this without replicating most of `DataModel#validate`.

image


CarlosFdez:
Perhaps if these errors occured in bulk for an actor somehow. If shark's fine with it I'm fine with it, perhaps its better than nothing or even fine temporarily, but I'd like to see if we're poking around here if there's a way to make the whole thing less noisy.

PR #19214, opened 2025-05-27
Remove non-effect action traits from existing aura rule elements

CarlosFdez:


src/module/migration/migrations/937-remove-aura-class-traits.ts
```diff
@@ -0,0 +1,55 @@
...
+                !rule ||
+                rule.key !== "Aura" ||
+                !("traits" in rule) ||
+                !Array.isArray(rule.traits) ||
+                rule.traits.length === 0
+            ) {
+                continue;
+            }
+
+            const newTraits = rule.traits.filter((t) => !classTraits.includes(t));
```
stwlam: Test for presence in effect traits, not for lack of presence in class traits (and import `effectTraits`).

CarlosFdez: We need it to not KO homebrew effect traits. What you suggest might be a bit aggressive, especially if certain modules got disabled while they're updating to V13.

EDIT: I could try to instead compile a list of ommissions, which might avoid that problem.

CarlosFdez:
After some testing, it works out fine. It will not alter non-system traits even if the module is disabled, which can be handy if someone isn't running a module.

src/module/migration/migrations/937-remove-invalid-aura-traits.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    static #_invalidTraits: Set<string> | null = null;
+
+    /**
+     * Computes and returns a list of all invalid traits.
+     * Doing it this way avoids us accidentally removing homebrew effect traits.
+     */
+    static get #invalidTraits(): Set<string> {
+        if (this.#_invalidTraits) return this.#_invalidTraits;
+
```
stwlam:
```suggestion
```
leave some whitespace for the sick children to eat

CarlosFdez: How wasteful of me

PR #19218, opened 2025-05-28
Fix `ItemAttacher` styling

In3luki:
attach-light
attach-dark

### DIFF
--- a/src/module/apps/item-attacher.ts
+++ b/src/module/apps/item-attacher.ts
@@ -8,6 +8,9 @@ import { ErrorPF2e } from "@util";
 class ItemAttacher<TItem extends PhysicalItemPF2e> extends PickAThingPrompt<TItem, PhysicalItemPF2e> {
     static override DEFAULT_OPTIONS: DeepPartial<fa.ApplicationConfiguration> = {
         id: "item-attacher",
+        window: {
+            contentClasses: ["standard-form"],
+        },
     };
 
     static override PARTS: Record<string, fa.api.HandlebarsTemplatePart> = {
--- a/src/module/apps/item-attacher.ts
+++ b/src/module/apps/item-attacher.ts
@@ -6,6 +6,10 @@ import { ErrorPF2e } from "@util";
 
 /** A prompt for the user to select an item to receive an attachment */
 class ItemAttacher<TItem extends PhysicalItemPF2e> extends PickAThingPrompt<TItem, PhysicalItemPF2e> {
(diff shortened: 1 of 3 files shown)


src/module/apps/item-attacher.ts
```diff
@@ -6,6 +6,12 @@ import { ErrorPF2e } from "@util";
 
 /** A prompt for the user to select an item to receive an attachment */
 class ItemAttacher<TItem extends PhysicalItemPF2e> extends PickAThingPrompt<TItem, PhysicalItemPF2e> {
+    static override DEFAULT_OPTIONS: DeepPartial<fa.ApplicationConfiguration> = {
+        window: {
+            contentClasses: ["item-attacher"],
+        },
+    };
```
stwlam: We only ever want one of these open, so just give it an id of "item-attacher"

In3luki: Done

src/module/apps/item-attacher.ts
```diff
@@ -6,6 +6,13 @@ import { ErrorPF2e } from "@util";
 
 /** A prompt for the user to select an item to receive an attachment */
 class ItemAttacher<TItem extends PhysicalItemPF2e> extends PickAThingPrompt<TItem, PhysicalItemPF2e> {
+    static override DEFAULT_OPTIONS: DeepPartial<fa.ApplicationConfiguration> = {
+        id: "item-attacher",
+        window: {
+            contentClasses: ["item-attacher"],
+        },
```
stwlam: Don't need the content class if it has the ID :)

src/styles/item/_item-attacher.scss
```diff
@@ -1,4 +1,4 @@
-#item-attacher .content {
+.item-attacher {
```
stwlam:
```suggestion
#item-attacher .window-content {
```

src/module/apps/item-attacher.ts
In3luki: Oh that makes sense yeah.

In3luki:
I just noticed there also was an `id` getter override. I removed that now that the id is set in the options.

src/styles/item/_item-attacher.scss
```diff
@@ -69,9 +69,13 @@
     }
 
     footer {
```
stwlam, line 67: Try a class of "form-footer" here along with a content-class of "standard-form"

In3luki:
Now looks like this:
image

PR #19230, opened 2025-05-30
Add `StatisticModifier#clone` and utilize it in the check roll dialog

### LINKED ISSUE #19222: Toggling a Substitute Roll With AdjustModifier caues it to be applied repeatedly
Hi,

I came across an issue today with a substitute roll when using an Adjust Modifier rule element with the mode - add.

The PC has Assurance and a rule element within an ancestry feature that adjusts their proficiency up roughly 1 every three+ levels. When you bring up the prompt to roll for the skill they have Assurance in, the toggle is at the top. When you toggle it on though, the prof bonus goes up 2. Toggle it back off and it goes up 2 again.

In3luki:
- Closes #19222


In3luki:
Doesn't work as expected in all circumstances.

PR #19231, opened 2025-05-30
Fix embedding items in journal sheets failing with a max depth warning

In3luki:
Passing the value instead of a reference avoids the recursion depth warning. Credit to `ChaosOS` from the Foundry discord.

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -1056,7 +1056,7 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
     ): Promise<HTMLCollection> {
         options.relativeTo = this;
         const container = document.createElement("div");
-        container.innerHTML = await TextEditorPF2e.enrichHTML(this.embedHTMLString(config, options), options);
+        container.innerHTML = await TextEditorPF2e.enrichHTML(this.embedHTMLString(config, options), { ...options });
         return container.children;
     }
 }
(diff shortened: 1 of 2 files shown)


farling42:
2nd parameter to enrichHtml is an options parameter. This PR passes all the fields of options to enrichHtml, which primarily means that any wanted to be passed as an option is actually simply ignored by the called function.

The better solution might be to make a copy of options and pass that in.

In3luki:
This PR spreads the original options paramter into a new object that is passed as `options` to `enrichHTML` which creates a deep copy of all top-level properties and a shallow copy of all nested properties of the original object.
There is another function call (that also takes the original `options` parameter as the second parameter) as the first parameter of `enrichHTML`. That confused me too at first glance.

In3luki:
The problem, as far as I understand it, is that passing around the same `options` reference means the same `_embeddedDepths` property is increased across all `enrichHTML` calls. By spreading it in a new object it's now referencing another memory address so each `_embeddedDepths` is now only increased by that one function call.

farling42:
Oh yes, I see the new object being created. Sorry, it was a misreading of the code by me.

CarlosFdez:
actor/item in the roll data are documents, so making a copy of those would lead to issues. Creating a shallow copy is better if a copy has to be made.

In3luki:
I've moved the fix to `TextEditor._embedContent` which should fix it for all `Document#toEmbed` instances.

PR #19232, opened 2025-05-30
Apply modifier adjustments to the original value

CarlosFdez:
This causes repeated modifier adjustments on the same modifier to lead to the same result each time.

Closes https://github.com/foundryvtt/pf2e/issues/19222


CarlosFdez:
Gonna draft until I see how it gets along with damage alterations.

CarlosFdez:
Damage alterations do not modify the value of a flat modifier, so its fine. If it ever did though, we'd have to combine the two operations.

PR #19239, opened 2025-06-02
Add roll options for SingleCheckAction previews

xyzzy42:
When it generates the previews is only uses the roll option(s) that are specific to the action, e.g. only "action:escape".  This doesn't match what will be used for the actual roll.  For instance, it won't include "armor:strength-requirement-met" and that results in the value being incorrect for nearly every str or dex based action.

### DIFF
--- a/src/module/actor/actions/single-check.ts
+++ b/src/module/actor/actions/single-check.ts
@@ -199,7 +199,11 @@ class SingleCheckActionVariant extends BaseActionVariant {
                     ...statistic.modifiers,
                     ...this.modifiers.map((modifier) => new ModifierPF2e(modifier)),
                 ];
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Point it at v13-dev. I don't know if it'll explode so I'll leave that to you.

xyzzy42:
W.r.t. the set creation, the first thing StatisticModifier will do is create a set from the roll options if it's not already one.  So creating the set here is effectively free.

https://github.com/foundryvtt/pf2e/blob/4cbdaa37d6c33e9519561bae2c59a23e0288cbce/src/module/actor/modifiers.ts#L508

CarlosFdez:
Seems you saw the old notification but I had observed that which is why I edited it.

PR #19240, opened 2025-06-02
Add Adjust DoS RE to Assured Identification

Cuingamehtar:


CarlosFdez:
wouldn't this end up telling the player that its cursed, rather than they somehow failed?

CarlosFdez:
Also no more merging into master

Cuingamehtar:
It's a secret check, so it shouldn't reveal any info to the player, I think.

Why not merge into master? I thought v13-dev was just script updates for v13 compatibility?

CarlosFdez:
If there is no way for the action to be rolled publically by accident and reveal important metagame knowledge, then sure.

v13-dev is a branch that was created to work on Foundry v13 while leaving v12 possible to update. If master is updated, we have to rebase. Since v13 is no longer in beta, we don't want to do those rebases anymore.

It just hasn't been merged yet because we're busy and there's something stwlam wants to do first.

Cuingamehtar:
Well, `@Check[arcana|options:action:identify-magic]` won't be secret, but if we consider the cases
- DC is not set: the player knows their roll result, but doesn't know the outcome. The note is present always and gives no additional info.
- DC is set: player knows the roll result and the outcome. The player already expects the success to reveal the item with or without a note (it's the normal outcome). The note is present on success as a reminder.

Either way the GM made a mistake by not making this roll secret, and the second case is metagaming even without a note.


If necessary, I'll just rebase this PR after you merge the branch. There is no hurry.

CarlosFdez:
The v13-dev thing is handled, that was a heads up for the future. Its pointing at v13-dev currently.

Roll notes do support `visibility: "gm"`, but I suppose "success but you learn nothing" already reveals that though. I'll leave it to data entry to decide how they feel about it.

EDIT: A lot of people were apparently not aware, so no worries.

Cuingamehtar:
Ah, I'm merging into the wrong branch. Github must have reused my previous PR and I didn't consider checking it.

CarlosFdez:
nono, it was handled cause I moved it. I put it back to v13-dev. It is correct now.

PR #19246, opened 2025-06-03
Avoid creating a new tooltip listener on every `EffectsPanel` render

### LINKED ISSUE #19244: Selecting, deselecting, and swapping tokens with effects eventually causes UI lag
A video to help explain the behavior:

https://github.com/user-attachments/assets/7f7b4921-9c99-4187-8611-b1e6a7ecd7d8

This seems to be contained to the effect icons themselves, as the remainder of foundry remains usable while the effect icons lag.

Note:
- It persists through scenes
- It resets on reload

Versions:
- Firefox 138.0
- v13.344
- PF2e 7.1.0

In3luki:
- Closes #19244

### DIFF
--- a/src/module/apps/effects-panel.ts
+++ b/src/module/apps/effects-panel.ts
@@ -1,4 +1,5 @@
 import type { ActorPF2e } from "@actor";
+import type { HandlebarsRenderOptions } from "@client/applications/api/handlebars-application.d.mts";
 import { AbstractEffectPF2e, AfflictionPF2e, ConditionPF2e, EffectPF2e } from "@item";
 import { PersistentDamageEditor } from "@item/condition/persistent-damage-editor.ts";
 import { createTooltipListener } from "@module/sheet/helpers.ts";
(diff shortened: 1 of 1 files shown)


xyzzy42:
Does not creating a listener only apply to the `onRender()` method, or does something that Hooks `render${doctype}` also need to avoid it to prevent building up listeners?

In3luki:
For the `EffectsPanel` the problem was that it is created and renderd on startup and then never closed. Each time a new token is controlled the `refresh` method re-renders the panel and so `_onRender` is called every time, always adding another listener on top of the existing ones.
Applications that are properly closed should destroy their DOM elements so you don't have to be as careful there.

CarlosFdez:
I suspect its because the listener was added to the root component, not a child. Probably a side effect of flattening the layout or a difference in the main element when re-rendering. Most child components would be destroyed and re-rendered in a handlebars layout.

In3luki:
Having the actual list be a child element with its own part id and only rendering that on refresh would probably also have worked as a fix, yeah.

PR #19247, opened 2025-06-03
Fix OutlineOverlayFilter for v13.344

oWave:
13.344 fixes OutlineOverlayFilter ignoring `knockout` and always showing only the outline (foundryvtt/foundryvtt#12721). Because of that, out-of-sight tokens now look like this: 
image

Setting knockout to true restore the previous appearance.


src/module/canvas/perception/modes.ts
```diff
@@ -73,7 +73,7 @@ class HearingDetectionMode extends fc.perception.DetectionMode {
     static override getDetectionFilter(): fc.rendering.filters.OutlineOverlayFilter {
         const filter = (this._detectionFilter ??= fc.rendering.filters.OutlineOverlayFilter.create({
             wave: true,
-            knockout: false,
+            knockout: true,
```
In3luki:
```suggestion

```
`true` is the default value.

PR #19248, opened 2025-06-03
Fix switching color theme for `EffectsPanel` tooltips

In3luki:
Retrieving the setting on every tooltip render isn't great but caching it doesn't really work there. I've only added the setting type as needed but can fill it out if you want.

### DIFF
--- a/src/module/sheet/helpers.ts
+++ b/src/module/sheet/helpers.ts
@@ -177,7 +177,7 @@ function createTooltipListener(
         render: (element: HTMLElement) => Promise<HTMLElement | null>;
     },
 ): void {
-    const tooltipOptions = R.pick(options, ["cssClass", "direction", "locked"]);
+    const tooltipOptions = R.pick(options, ["direction", "locked"]);
 
     element.addEventListener(
         "pointerenter",
@@ -190,14 +190,13 @@ function createTooltipListener(
             if (options.locked) {
                 game.tooltip.dismissLockedTooltips();
             }
-            if (options.themeGroup) {
-                const parts = tooltipOptions.cssClass?.split(" ") ?? [];
-                // Remove current theme if present
-                parts.findSplice((c) => c === "theme-dark" || c === "theme-light");
-                parts.push(getCurrentTheme(options.themeGroup));
-                tooltipOptions.cssClass = parts.join(" ");
-            }
-            game.tooltip.activate(target, { html, ...tooltipOptions });
+            game.tooltip.activate(target, {
+                html,
+                ...tooltipOptions,
+                cssClass: options.themeGroup
+                    ? `${(options.cssClass ?? "").trim()} ${getCurrentTheme(options.themeGroup)}`.trim()
+                    : options.cssClass,
+            });
 
             // A very crude implementation only designed for align top. Make it more flexible if we need to later
             if (options.align === "top") {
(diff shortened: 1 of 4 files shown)


src/module/sheet/helpers.ts
```diff
@@ -189,6 +190,13 @@ function createTooltipListener(
             if (options.locked) {
                 game.tooltip.dismissLockedTooltips();
             }
+            if (options.themeGroup) {
+                const parts = tooltipOptions.cssClass?.split(" ") ?? [];
+                // Remove current theme if present
+                parts.findSplice((c) => c === "theme-dark" || c === "theme-light");
+                parts.push(getCurrentTheme(options.themeGroup));
+                tooltipOptions.cssClass = parts.join(" ");
```
CarlosFdez: wouldn't you be able to avoid mutating the property if you assign it directly to game.tooltip.activate()?

src/module/sheet/helpers.ts
```diff
@@ -189,7 +190,15 @@ function createTooltipListener(
...
-            game.tooltip.activate(target, { html, ...tooltipOptions });
+            const cssClass = ((): string | undefined => {
+                if (options.themeGroup) {
+                    const parts = options.cssClass?.split(" ") ?? [];
+                    parts.push(getCurrentTheme(options.themeGroup));
+                    return parts.join(" ");
+                }
+                return options.cssClass;
+            })();
+            game.tooltip.activate(target, { html, cssClass, ...tooltipOptions });
```
CarlosFdez:
```suggestion
            game.tooltip.activate(target, { html, ...tooltipOptions, cssClass });
```

src/module/sheet/helpers.ts
In3luki: Updated to no longer mutate the original option. There might be a better way to do it but I couldn't think of one.

src/module/sheet/helpers.ts
In3luki: I've removed `cssClass` from `tooltipOptions` and have the IIFE return it if there is no `themeGroup` set so it shouldn't ever be overwritten by the spread.

CarlosFdez: Ah, I see, you did the pick. I still think its better to do that in order to avoid future bugs in case someone removes that pick. Less rope.

src/module/sheet/helpers.ts
In3luki: oh wait a sec. I don't need the splitting now.

src/module/sheet/helpers.ts
```diff
@@ -189,7 +190,13 @@ function createTooltipListener(
             if (options.locked) {
                 game.tooltip.dismissLockedTooltips();
             }
-            game.tooltip.activate(target, { html, ...tooltipOptions });
+            game.tooltip.activate(target, {
+                html,
+                ...tooltipOptions,
+                cssClass: options.themeGroup
+                    ? `${(options.cssClass ?? "").trim()} ${getCurrentTheme(options.themeGroup)}`.trim()
+                    : options.cssClass,
```
In3luki, line 198: It's a bit less clear but avoids the IIFE altogether.

CarlosFdez: That's fine, so long as its not a easy vector for possible future bugs

PR #19259, opened 2025-06-05
Set Mikazuki to reload 0

mechamaya:
Missed this in Treasure Vault Remastered data entry. Mikazuki should have reload 0 now.


pedrogrullada:
I have this change and a few source corrections in my other PR already, so we should probably avoid conflicts

mechamaya:
Ah damn, didn't see your PR, oh well 😓

PR #19264, opened 2025-06-05
Update various type imports

In3luki:
This started out as a type update PR but then I noticed the browser's `RollTable` methods stopped working.
The rest is all type fixes except for one typo in `src/module/system/action-macros/crafting/repair.ts`.

### DIFF
--- a/build/lib/compendium-pack.ts
+++ b/build/lib/compendium-pack.ts
@@ -1,6 +1,7 @@
 import type { ActorSourcePF2e } from "@actor/data/index.ts";
-import type { CompendiumActorUUID, CompendiumItemUUID, ItemUUID } from "@client/documents/abstract/_module.d.mts";
+import type { CompendiumActorUUID, CompendiumItemUUID, ItemUUID } from "@client/documents/_module.d.mts";
 import type { CompendiumDocumentType } from "@client/utils/_module.d.mts";
+import type { ImageFilePath } from "@common/constants.d.mts";
 import type { DocumentStatsData, DocumentStatsSchema, SourceFromSchema } from "@common/data/fields.d.mts";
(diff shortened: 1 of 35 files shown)


In3luki:
I'll pull the unrelated type changes out of this and create a separate PR for them later today.

src/global.ts
```diff
@@ -348,6 +348,10 @@ declare global {
         }
     }
 
+    interface Window {
+        AutomaticBonusProgression: typeof AutomaticBonusProgression;
+    }
+
```
stwlam: Why is it throwing an error now and not before?

In3luki: Good question. I didn't see the `const AutomaticaBonusProgression: typeof ABP` a few lines above and assumed that the definition was accidentally removed.

```ts
  // hooks/load.ts line 172+173

  // Make available immediately on load for module subclassing
  window.AutomaticBonusProgression = AutomaticBonusProgression;
```
This throws a type error now: `Property 'AutomaticBonusProgression' does not exist on type 'Window & typeof globalThis'.ts(2339)`

In3luki: Ok that was wrong. I forgot that the repo now defaults to `v13-dev`. The interface does exist in master and was removed somewhere in the v13 dev cycle.

src/global.ts
```diff
@@ -334,8 +334,6 @@ declare global {
             Hotbar<MacroPF2e>
         >;
 
-        const AutomaticBonusProgression: typeof ABP;
-
```
stwlam: This should stay

stwlam:
This is a bit large, but most of it is filling in missing imports. Can we cut it down to that for a quick merge and then look at the rest separately?

In3luki:
> This is a bit large, but most of it is filling in missing imports. Can we cut it down to that for a quick merge and then look at the rest separately?

Done

PR #19267, opened 2025-06-05
Handle actor size changes and token size updates in `ActorPF2e` descendant document lifecycle hooks

### LINKED ISSUE #19266: Token Scaling - v7.1.1
I am running Foundry V13.344 with the 7.1.1 pf2e system.  I create a new PC and make it a Minotaur.  The character is correctly tagged with the Large trait, but when I drag the token to the scene, it is only Medium-sized.  If I go to the settings of the token on the Appearance tab, I can unlock Dimensions (Unlink from Actor Size) and relock it (Link to Actor Size) and the token will resize to 2x2 (Large).

In3luki:
- Closes #19266 

The size update in `preparePrototypeToken` had a negative effect on this and wasn't doing anything useful otherwise I think.

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -2134,7 +2134,6 @@ interface RechargeOptions {
 
 /** Data stored before changes to the actor were made in descendant document `_pre` hooks for use in `_on` hooks */
 type OperationWithPreviousData<T = DatabaseOperation<Document>> = T & {
-    reevaluate?: boolean;
     previous?: {
         maxHitPoints?: number;
         size?: ActorSizePF2e;
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1797,12 +1797,12 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         parent: Document,
         collection: string,
         data: Record<string, unknown>[],
-        options: DatabaseCreateOperation<Document> & { size?: ActorSizePF2e },
+        options: DatabaseCreateOperation<Document> & { previous?: PrevoiusActorData },
         userId: string,
     ): void {
         super._preCreateDescendantDocuments(parent, collection, data, options, userId);
         if (parent === this && collection === "items") {
-            options.size = this.system.traits?.size;
+            options.previous = { size: this.system.traits?.size };
         }
     }
 
@@ -1819,12 +1819,12 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         collection: string,
         documents: Document<Document>[],
         changes: Record<string, unknown>[],
-        options: DatabaseCreateOperation<Document> & { size?: ActorSizePF2e },
+        options: DatabaseCreateOperation<Document> & { previous?: PrevoiusActorData },
         userId: string,
     ): void {
         super._onCreateDescendantDocuments(parent, collection, documents, changes, options, userId);
-        if (!options.size || this !== parent || collection !== "items") return;
-        this.#updateTokenSizes(options.size);
+        if (!options.previous?.size || this !== parent || collection !== "items") return;
+        this.#updateTokenSizes(options.previous.size);
     }
 
     /** Store certain data to be checked in _onUpdateDescendantDocuments */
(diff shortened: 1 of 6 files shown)


stwlam:
Would rather handle this in token creation than making a separate actor update with no assurance on atomicity or consistency.

In3luki:
Now adjusts the dimension in `TokenDocumentPF2e#_preCreate`. This made it necessary to add an additional adjustment of the drag preview in `ActorDirectoryPF2e#_onDragStart` to show the correct token size.

In3luki:
After the second issue was raised in #19266 I noticed that `ItemPF2e##updateTokenSizes` was supposed to also handle ancestry size changes. That is now fixed and works without the need to adjust the `TokenDocument` as a bandaid like this PR was attempting to do previously.
The second bug in that issue was caused by only checking the current item's rule element for `BattleForm` and `CreatureSize` rule elements. That doesn't work in cases where the rule element is on another item but predicated on some roll option from the item that is being deleted.
The only way to fix that, I believe, is to check every other item on the actor for  `BattleForm` and `CreatureSize` elements or skip that check altogether. Maybe there is a better way to do this.

src/module/item/base/document.ts
```diff
@@ -904,14 +905,15 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         data: this["_source"] | DeepPartial<this["_source"]>,
         action: DatabaseAction,
     ): Promise<boolean | void> {
+        const actor = this.actor;
+        if (!actor) return;
         // Look for any opportunity to abort early
         if (
             this.type !== "ancestry" &&
-            !data.system?.rules?.some((r) => ["BattleForm", "CreatureSize"].includes(r.key ?? ""))
+            !data.system?.rules?.some((r) => ["BattleForm", "CreatureSize"].includes(r.key ?? "")) &&
```
CarlosFdez: This can probably be omitted if you're checking all the items anyways?

CarlosFdez:
If you have a merged toggleable sub-option, which updates multiple items at the same time, and the actor happens to have any battleform or creaturesize on them, would this cause a bunch of actor clones to spawn, once per item per merge?

Might some of these checks be a candidate for moving to Actor#_preUpdateDescendantDocuments() ?

src/module/item/base/document.ts
In3luki: As this can be called from `_preCreate` the triggering item could not exist in the item collection yet.

In3luki:
Third title change. Now completely avoids creating actor clones.

Tested:
- Spell Effect: Enlarge ✔️
- Spell Effect: Dinosaur Form (Tyrannosaurus) ✔️
- Any non-medium ancestry ✔️
- `CreatureSize` RE predicated on another effect ✔️

CarlosFdez:
How does it look like in the prototype token editor compared to the regular token editor in these instances?

src/module/actor/base.ts
```diff
@@ -1806,12 +1838,13 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
         collection: string,
         changes: Record<string, unknown>[],
-        options: DatabaseUpdateOperation<Document> & { previous?: object },
+        options: DatabaseUpdateOperation<Document> & { previous?: object; size?: ActorSizePF2e },
         userId: string,
     ): void {
         super._preUpdateDescendantDocuments(parent, collection, changes, options, userId);
         if (parent === this && collection === "items") {
             options.previous = { maxHitPoints: this.hitPoints?.max };
+            options.size = this.system.traits?.size;
```
CarlosFdez: This should go in previous. Previous is a property added in order to reduce how much we're adding to the root level. options.size is also ambiguous if its the size we're updating to or from.

In3luki: Done

In3luki:
For the prototype token sheet we need to add a subclass that replaces the appearance tab template like the non-prototype sheet is doing.
It might be enough to duplicate the existing `TokenSheetPF2e` class but I think I remember there being subtle differences in the source data.
I can take a look later today.

src/module/actor/base.ts
```diff
@@ -2042,6 +2160,14 @@ interface RechargeOptions {
     commit?: boolean;
 }
 
+/** Data stored before changes to the actor were made in descendant document `_pre` hooks for use in `_on` hooks */
+interface PrevoiusActorData {
+    previous?: {
+        maxHitPoints?: number;
+        size?: ActorSizePF2e;
+    };
+}
```
In3luki: The interface name isn't great but I couldn't think of anything better.

CarlosFdez: PreviousActorData would be a start!

You could have maxHitPoints and size be top level, and then do `DatabaseDeleteOperation<Document> & { previous?: PreviousActorData }` maybe. Or subclass DatabaseDeleteOperation, which is what I think I did originally?

CarlosFdez:
If there is no regression on the prototype token sheet then don't worry about it for this PR. I only brought it up since I saw you removed some prototype token code.

PR #19277, opened 2025-06-06
Fix Greater Advancing rune effect, TV wands rank, Anylength Rope description

Cuingamehtar:


Cuingamehtar:
Also, some items like Library Robes and Reef Heart still have spell _levels_ in their description in PDF. I assume it's a mistake by Paizo.

PR #19278, opened 2025-06-06
Fix creating `RollTable`s from the compendium browser and fix crafting action

### LINKED ISSUE #19299: V13 Compendium Browser 'Create Roll Table' does not observe entered weights and 'Add to Roll Table' does not work at all
V13 Compendium Browser 'Create Roll Table' does not observe entered weights and 'Add to Roll Table' does not work at all.

Foundry v13.345
PF2E 7.1.1
Running as server

In3luki:
- Closes #19299


CarlosFdez:
Awesome, thanks!

PR #19279, opened 2025-06-06
Fix Earn Income Macro always trying to use Armor Check Penalty

### LINKED ISSUE #17141: Earn Income macro includes armor check penalty even if the strength requirement is met
image

DocSchlock:
Closes #17141

Issue was the Earn Income Macro was not including the rollOption domains required to collect armor-requirements:met.


CarlosFdez:
Is there a reason we can't roll actor.skills[skill] instead? If so, it may be worth improving that api.

DocSchlock:
> Is there a reason we can't roll actor.skills[skill] instead? If so, it may be worth improving that api.

Ah, was going by least-disturbance-to-resolve (plus had no idea that was a thing).

I'll look into trying to change it - that definitely seems like a better avenue.

CarlosFdez:
Looks decent enough to me, at least for now. Thanks!

PR #19293, opened 2025-06-07
Add Bracers of Devotion effect

pedrogrullada:
Roll options default to true on purpose since a champion will more often be raising their blessed shield, or have their blessed armament targeted by a Disarm, than not.

Also update text on Second Blessing, which was somehow still pre-remaster


packs/equipment-effects/effect-bracers-of-devotion.json
```diff
@@ -0,0 +1,141 @@
...
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "action:disarm",
+                    "bracers-of-devotion-disarm-bonus"
+                ],
+                "selector": "reflex-dc",
+                "type": "item",
+                "value": 2
+            },
```
TikaelSol, line 85: Why the toggle?

pedrogrullada: It only applies to Disarm the character of their blessed armament, not any other item

PR #19295, opened 2025-06-07
Change Disarm, Trip, Grapple, and Shove to use best equipped weapon bonus

### LINKED ISSUE #17947: Disarm/Grapple/Shove/Trip - Not using best weapon for item bonus
It looks like the item bonus you get from having an equipped weapon with the relevant trait isn't coming from the best one, but instead the one last in the list. For example if you give someone a +2 Flail and then a +1 Flail and they both have them equipped it's only giving a +1 item bonus and not a +2 item bonus. As soon as you unequip the +1 Flail it decides to give a +1 item bonus, or if you do it in reverse order where you give them the +1 Flail first and then a +2 Flail.

DocSchlock:
Closes #17947 

Obviously, this still has limitations and exceptions. Putting this up as an improvement, however minor, to the current process.

### DIFF
--- a/src/module/system/action-macros/helpers.ts
+++ b/src/module/system/action-macros/helpers.ts
@@ -4,7 +4,7 @@ import type { StrikeData } from "@actor/data/base.ts";
 import { getRangeIncrement } from "@actor/helpers.ts";
 import { CheckModifier, ModifierPF2e, ensureProficiencyOption } from "@actor/modifiers.ts";
 import type { RollOrigin, RollTarget } from "@actor/roll-context/types.ts";
-import { ItemPF2e, WeaponPF2e } from "@item";
+import type { ItemPF2e, WeaponPF2e } from "@item";
 import type { AbilityTrait } from "@item/ability/types.ts";
 import type { WeaponTrait } from "@item/weapon/types.ts";
 import { RollNotePF2e } from "@module/notes.ts";
--- a/src/module/system/action-macros/helpers.ts
+++ b/src/module/system/action-macros/helpers.ts
@@ -1,10 +1,10 @@
-import type { ActorPF2e } from "@actor";
+import { ActorPF2e } from "@actor";
 import { AutomaticBonusProgression } from "@actor/character/automatic-bonus-progression.ts";
 import type { StrikeData } from "@actor/data/base.ts";
 import { getRangeIncrement } from "@actor/helpers.ts";
 import { CheckModifier, ModifierPF2e, ensureProficiencyOption } from "@actor/modifiers.ts";
 import type { RollOrigin, RollTarget } from "@actor/roll-context/types.ts";
-import type { ItemPF2e, WeaponPF2e } from "@item";
+import { ItemPF2e, WeaponPF2e } from "@item";
 import type { AbilityTrait } from "@item/ability/types.ts";
 import type { WeaponTrait } from "@item/weapon/types.ts";
 import { RollNotePF2e } from "@module/notes.ts";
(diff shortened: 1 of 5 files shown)


nikolaj-a:
Looks good to me! Maybe remove the old helper method if it's no longer used?

DocSchlock:
> Looks good to me! Maybe remove the old helper method if it's no longer used?

If you are talking about getApplicableEquippedWeapons(), that's still used in the new Best function to grab the list of weapons that qualify. I left it because it's cleaner to call that with the forEach, but I can remove it if desired.

src/module/system/action-macros/helpers.ts
```diff
@@ -343,6 +343,27 @@ class ActionMacroHelpers {
...
     }
 
+    static getBestEquippedItemForAction(
+        actor: ActorPF2e,
+        traits: WeaponTrait[],
+        selector: string,
+    ): WeaponPF2e<ActorPF2e> | null {
+        const items = [] as WeaponPF2e<ActorPF2e>[];
+        traits.forEach((x) => items.push(...ActionMacroHelpers.getApplicableEquippedWeapons(actor, x)));
+
```
CarlosFdez:
```suggestion
        const items = traits.flatMap((t) => ActionMacroHelpers.getApplicableEquippedWeapons(actor, t));
```

src/module/system/action-macros/helpers.ts
```diff
@@ -1,10 +1,10 @@
-import type { ActorPF2e } from "@actor";
+import { ActorPF2e } from "@actor";
 import { AutomaticBonusProgression } from "@actor/character/automatic-bonus-progression.ts";
 import type { StrikeData } from "@actor/data/base.ts";
 import { getRangeIncrement } from "@actor/helpers.ts";
 import { CheckModifier, ModifierPF2e, ensureProficiencyOption } from "@actor/modifiers.ts";
 import type { RollOrigin, RollTarget } from "@actor/roll-context/types.ts";
-import type { ItemPF2e, WeaponPF2e } from "@item";
+import { ItemPF2e, WeaponPF2e } from "@item";
```
CarlosFdez: It doesn't look like you're actually using the instances though? Why the removal of type imports?

DocSchlock: I think this got removed by accident by my IDE - I didn't intentionally remove it. Probably when I was fiddling around with something.

Added back.

nikolaj-a:
> If you are talking about getApplicableEquippedWeapons(), that's still used in the new Best function to grab the list of weapons that qualify. I left it because it's cleaner to call that with the forEach, but I can remove it if desired.

Oh, I see it. Consider making it module private instead then. I'm trying to prevent accidentally usage elsewhere that will create yet another dependency we need to untangle later.

CarlosFdez:
Its already exposed so that doesn't really matter too much, but sure.

src/module/system/action-macros/helpers.ts
```diff
@@ -1,4 +1,4 @@
-import type { ActorPF2e } from "@actor";
+import { ActorPF2e } from "@actor";
```
CarlosFdez:
```suggestion
import type { ActorPF2e } from "@actor";
```

DocSchlock:
@CarlosFdez Is there anything else you'd like changed for this?

CarlosFdez:
None, though us KOing a helper that is already exposed did concern me.

xyzzy42:
> None, though us KOing a helper that is already exposed did concern me.

I think the entire class, `ActionMacroHelpers`, is not exposed.  No objects of that type are ever created, the methods are all static, and the class isn't in anywhere in `game.`, etc.

It does provide an interface for something I've been wanting to improve in BAM.

The problem with choosing the weapon with the best potency modifier is that it isn't necessarily correct.  It doesn't take agile and MAP into account nor does it consider reach.  Or more simply, the player may have a good reason to want a specific item other than the one with the highest potency modifier.

Right now BAM has two buttons for each attack action, one that is agile and one that isn't.  I think a better way to do it would be to present choices of which item to use.  Likely fewer buttons and fixes the problem of combining the agile trait from the unarmed attack with the potency modifier of a non-agile weapon.

And so we must get a list of what items can be used with the action.  Which the action code already finds.  But there's no way to get it.

Maybe the preview method could do that?

CarlosFdez:
It seems it needs to look at reach when filtering, and support selecting a specific weapon (for aux button use).

xyzzy42:
Easy enough to add `item` to SingleCheckActionUseOptions and have it override the auto-selected item.  I remember I tried to do this, and there was a problem.

I think there's an issue around when CheckMacroContext is called.

If it's going to select an item to use, then it really should pull in *all* the roll options and injectable properties associated with the item, and not just the potency bonus.  And to do that, you need to know the item much earlier in the roll prep than the point CheckMacroContext is called.

PR #19304, opened 2025-06-08
Add `PrototypeTokenConfigPF2e`

### LINKED ISSUE #19152: V13: Prototype Token and Token dialogs differ
Image

Latest v13 branch, latest foundry.

The screenshot above shows the prototype token dialog (left) and the token dialog (right). Note that many settings differ. I tried to change the width and height of the token on the prototype token which does not seem to work. My guess is that the dimension is locked to the token size and therefore resets back to the predefined values. That's fine but you can't undo the locking because the Size section is not present in the prototype token dialog.

In3luki:
I've refactored the token configuration to reduce code duplication between `TokenConfigPF2e` and the now added `PrototypeTokenConfigPF2e`.

### Problems
The two config classes share a significant amount of logic, but can't use a common base class due to structural differences:
- `PrototypeTokenConfig` uses `TokenApplicationMixin(ApplicationV2)` instead of `TokenApplicationMixin(DocumentSheetV2),` which makes it incompatible with the base class used by `TokenConfigPF2e`.
- Additionally, `PrototypeTokenConfig#token` is a `PrototypeToken,` not a `TokenDocument`.

### Changes
- I've extracted the shared logic from `TokenConfigPF2e` into a new helper class called `TokenConfigHelper`.
- `TokenConfigHelper` is now used by both `TokenConfigPF2e` and `PrototypeTokenConfigPF2e`, keeping their unique behavior isolated.
___
- Closes #19152


CarlosFdez:
Might be worth introducing a `get actor()` I think to reduce some of those diffs (and line lengths)

stwlam:
Why not a mixin?

In3luki:
> Why not a mixin?

I hadn't tought of that but totally makes sense in this case. I'll see what I can do.

In3luki:
Now uses a mixin. I had to use the return type of the base `TokenApplicationMixin` (which is used by both configs) to stop Typescript from complaining. The resulting type isn't 100% correct but it works.
That's the reason why `_processFormData` can't be defined in the mixin itself.

PR #19310, opened 2025-06-09
Localize most rule element fields using LOCALIZATION_PREFIXES

CarlosFdez:
It'd be nice if stuff like predicate resolved regardless how deep it was, but it is what it is.

### DIFF
--- a/src/module/rules/rule-element/aura.ts
+++ b/src/module/rules/rule-element/aura.ts
@@ -20,6 +20,8 @@ import fields = foundry.data.fields;
 
 /** A Pathfinder 2e aura, capable of transmitting effects and with a visual representation on the canvas */
 class AuraRuleElement extends RuleElementPF2e<AuraSchema> {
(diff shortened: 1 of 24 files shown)


src/module/rules/rule-element/aura.ts
```diff
@@ -20,6 +20,8 @@ import fields = foundry.data.fields;
 
 /** A Pathfinder 2e aura, capable of transmitting effects and with a visual representation on the canvas */
 class AuraRuleElement extends RuleElementPF2e<AuraSchema> {
+    static override LOCALIZATION_PREFIXES = ["PF2E.RuleEditor.RuleElement", "PF2E.RuleEditor.Aura"];
+
```
stwlam, line 24: It is untypical that a data model would have its prefixes in an application's namespace

CarlosFdez: It is. Looking at it again, I think I ran into an issue since PF2E.RuleElement is already taken for the keys. We need a new parent or to move localization for the keys and hope nothing explodes if we convert strings into objects.

stwlam: How about `RULE_ELEMENT`?

PR #19313, opened 2025-06-09
Enable `AdjustModifier` predication on derived values and pass more options to the `Statistic` created for `@Check`

In3luki:
- As the predicate for `AdjustModifier` was immediately constructed in `beforePrepareData` it wasn't possible to resolve dervied values.
- Passing the item in `checkDC` makes it possible to predicate on item roll options.

This PR would enable automation for feats like `Powerful Alchemy`:
````
Alchemical items you infuse are particularly potent. When you create an infused alchemical item
that allows a saving throw, you can change its DC to your class DC.
````

```json
{
  "key": "AdjustModifier",
  "mode": "override",
  "value": "@actor.classDCs.alchemist.mod",
  "selector": "inline-dc",
  "predicate": [
    "item:trait:infused",
    {
      "gt": [
        "{actor|classDCs.alchemist.dc.value}",
        "inline-dc:value"
      ]
    }
  ]
}
```
(That should probably be a toggle but it's what I've tested with)

### DIFF
--- a/src/module/rules/rule-element/adjust-modifier.ts
+++ b/src/module/rules/rule-element/adjust-modifier.ts
@@ -74,6 +74,7 @@ class AdjustModifierRuleElement extends RuleElementPF2e<AdjustModifierSchema> {
         const adjustment: ModifierAdjustment = {
             slug: this.slug,
             test: (options): boolean => {
+                // Lazy load the predicate and avoid constructing it again every time the adjustment is tested
                 predicate ??= new Predicate(this.resolveInjectedProperties(fu.deepClone([...this.predicate])));
                 return predicate.test([...options, ...this.item.getRollOptions("parent")]);
             },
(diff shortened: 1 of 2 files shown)


src/module/rules/rule-element/adjust-modifier.ts
```diff
@@ -69,11 +69,12 @@ class AdjustModifierRuleElement extends RuleElementPF2e<AdjustModifierSchema> {
     override beforePrepareData(): void {
         if (this.ignored) return;
 
-        const predicate = new Predicate(this.resolveInjectedProperties(fu.deepClone([...this.predicate])));
```
CarlosFdez, line 72: It doesn't seem like this is used anywhere else in the function. Is there a reason this is like that? Optimization? If so it could maybe use a comment, perhaps `Lazy load the predicate but hold onto it to avoid superfluous processing`?

In3luki: Yeah, this way the predicate is only construced once per data prep cycle. Without that it would be construced every time the adjustment is tested.
I've added a comment.

PR #19316, opened 2025-06-09
Automate Bellflower Tiller's Scarecrow

### LINKED ISSUE #19315: Bellflower Tiller's Scarecrow automation
Bellflower Tiller's Scarecrow should have a RE for its extra damage effect:
> When you hit a creature that is adjacent to two or more members of your crop with a weapon or unarmed attack, you gain a circumstance bonus to your damage roll equal to the number of weapon damage dice.

Ikaguia:
Closes #19315

### DIFF
--- a/packs/feats/archetype/bellflower-tiller/scarecrow.json
+++ b/packs/feats/archetype/bellflower-tiller/scarecrow.json
@@ -29,7 +29,18 @@
             "remaster": false,
             "title": "Pathfinder #147: Tomorrow Must Burn"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "label": "Scarecrow",
+                "selector": "weapon-damage",
+                "value": "@weapon.system.damage.dice",
+                "type": "circumstance",
+                "predicate": [
+                    "bellflower-scarecrow"
+                ]
+            }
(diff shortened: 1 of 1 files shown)


packs/feats/archetype/bellflower-tiller/scarecrow.json
```diff
@@ -29,7 +29,18 @@
...
+                "key": "FlatModifier",
+                "label": "Scarecrow",
+                "selector": "weapon-damage",
+                "value": "@weapon.system.damage.dice",
+                "type": "circumstance",
+                "predicate": [
+                    "bellflower-scarecrow"
+                ]
+            }
+        ],
```
pedrogrullada: We can just omit the label since it's just the name of the feat, but keep in mind for the future that plain text needs to be localized. Your RE properties should be sorted in alphabetical order as well, which the extractor will do for you, among other things, so it's good practice to make changes within Foundry and then extract.
```suggestion
        "rules": [
            {
                "key": "FlatModifier",
                "predicate": [
                    "bellflower-scarecrow"
                ],
                "selector": "weapon-damage",
                "type": "circumstance",
                "value": "@weapon.system.damage.dice"
            }
        ],
```

PR #19321, opened 2025-06-10
Don't set Dark Mode colors on windows with `theme-light` override

jfn4th:
I've only found chat popouts to be affected by this. If you know of any other AppV2's that might have this override, let me know and I'll take a look!

Screenshot 2025-06-09 at 8 16 18 PM
Screenshot 2025-06-09 at 8 16 29 PM

### DIFF
--- a/src/styles/_colors.scss
+++ b/src/styles/_colors.scss
@@ -374,7 +374,7 @@ body,
     }
 }
 
-body.theme-dark .application,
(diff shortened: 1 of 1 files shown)


CarlosFdez:
How about targeting `body.theme-dark .application:not(.themed)` instead? While still targeting `.themed.theme-dark` of course.

jfn4th:
Sure, that works too!

PR #19329, opened 2025-06-11
Add support for spell rank alterations

CarlosFdez:
This attempts to solve it by adding a further divide between slot rank and cast rank, with slot rank being the slot rank. It gets a bit weird for spell.cast() on items though, but we can interpret those as some sort of source rank. If neither cast nor slot rank are provided, we assume the spell item's rank is a slot rank and apply alterations (if available). We could relax this, but then it means that ability.cast(fireball) would not apply rank alterations.


CarlosFdez:
It is no longer needed and also there are a lot of problems and risks, even if some things in the game do use this. I won't be able to finish it, so I will close it.

PR #19340, opened 2025-06-11
Preserve certain changes when chaining spell.loadVariant()

CarlosFdez:
Doesn't fix any known bugs, but improves its consistency. Prerequisite for something else.


src/module/item/spell/document.ts
```diff
@@ -471,7 +471,8 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
     loadVariant(options: SpellVariantOptions = {}): SpellPF2e | null {
         if (this.original) {
             const entryId = this.system.location.value;
-            return this.original.loadVariant({ entryId, ...options });
+            const overlayIds = Array.from(this.appliedOverlays?.values() ?? []);
```
stwlam:
```suggestion
            const overlayIds = this.appliedOverlays?.values().toArray() ?? [];
```

PR #19342, opened 2025-06-12
Update Hexing Jar RE with remastered patrons

Cuingamehtar:


static/lang/re-en.json
```diff
@@ -3552,6 +3552,26 @@
...
+                        "SilenceInSnow": "Silence in Snow",
+                        "SpinnerOfThreads": "Spinner of Threads",
+                        "StarlessShadow": "Starless Shadow",
+                        "TheInscribedOne": "The Inscribed One",
+                        "TheResentment": "The Resentment",
+                        "TheUnseenBroker": "The Unseen Broker",
+                        "WhisperOfWings": "Whisper of Wings",
+                        "WildingSteward": "Wilding Steward"
+                    }
+                },
```
pedrogrullada: Since these patron labels may be used for other witch rules in the future, maybe moving them to be under `PF2E.SpecificRule.Witch.Patron` would be better?

PR #19345, opened 2025-06-12
Add REs to Walking Cauldron

### LINKED ISSUE #5064: Data entry: walking cauldrons weight should be modular
image
what i would like to see is if it is invested then it means theres no weight as its following you. but if it isnt invested then its the full weight. also when its invested it should become a container up to 2 bulk. and when not invested no container. i dont know if that is possible. but i thought id point this out for a systemic fix.

DocSchlock:
Closes #5064

Allows players to toggle the Bulk of the Cauldron to represent it walking. The container stuff isn't doable at this time.

### DIFF
--- a/packs/equipment/walking-cauldron.json
+++ b/packs/equipment/walking-cauldron.json
@@ -39,7 +39,8 @@
                 "key": "RollOption",
                 "option": "walking-cauldron-walking",
                 "toggleable": true,
-                "value": true
+                "value": true,
+                "label": "PF2E.SpecificRule.Equipment.WalkingCauldron.ToggleLabel"
             },
             {
                 "itemType": "equipment",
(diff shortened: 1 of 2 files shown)


packs/equipment/walking-cauldron.json
```diff
@@ -34,7 +34,25 @@
...
         },
         "quantity": 1,
-        "rules": [],
+        "rules": [
+            {
+                "key": "RollOption",
+                "option": "walking-cauldron-walking",
+                "toggleable": true,
+                "value": true
+            },
```
TikaelSol: Since this toggle wouldn't be obvious what its doing we probably want a label here.

DocSchlock: Done!

static/lang/re-en.json
```diff
@@ -3656,6 +3656,9 @@
                 "UnholyWater": {
                     "Note": "Can damage only creatures with the holy trait."
                 },
+                "WalkingCauldron": {
+                    "ToggleLabel": "Walking Cauldron - Animated"
```
```suggestion
                    "ToggleLabel": "Walking Cauldron – Animated"
```

PR #19353, opened 2025-06-13
Fix hidden check outcome vanishing when player rerolls

### LINKED ISSUE #17420: Metagaming setting causes rerolls to loose target context
When the metagame information setting "Show check outcomes" is disabled, and a player rerolls an attack roll, the new roll looses all target context information on the GM side (like the degree of success, target name and hit info). If I reroll it as a GM, the context info remains there.

https://github.com/user-attachments/assets/7aaaede1-e462-40ab-afbe-f7467058d26e

If "Show check outcomes" is enabled then this problem seems to go away.

DocSchlock:
Closes #17420

Issue: When a player re-rolls a check with Show Check Outcomes disabled, the check outcome GM box vanishes from the re-roll.

Cause: when the re-roll is generated from the player's end, the gm-visible box isn't present in the HTML for the original roll. Since the elements aren't present, the outcome update code fails to run, removing the box completely.

Resolution: test if the check outcome box is in the old flavor. If it is, use the original workflow to replace the check outcomes. If not, insert a new check outcome element from the reroll after the action section.

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -570,7 +570,10 @@ class CheckPF2e {
                   const opposer = context.target?.actor === actor.uuid ? context.origin : context.target;
                   const targetFlavor = await this.#createResultFlavor({ degree, self, opposer, targeting });
                   if (targetFlavor) {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I plan to look at this in more detail at a future date.

PR #19367, opened 2025-06-15
Also copy over dependent tokens in actor clone during item preCreate

CarlosFdez:
Fixes Token Marks Regression

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -424,6 +424,20 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
         return 0;
     }
 
+    /**
+     * @inheritdoc
+     * Overriden to also clone dependent tokens if keepId is true
+     */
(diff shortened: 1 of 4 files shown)


src/module/item/base/document.ts
```diff
@@ -688,6 +688,9 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
 
         // Convert all non-kit sources to item objects, and recursively extract the simple grants from ABC items
         const actorClone = actor.clone({}, { keepId: true });
+        for (const [key, value] of actor._dependentTokens.entries()) {
+            actorClone._dependentTokens.set(key, value);
+        }
```
stwlam: Any reason to not do this in a `clone` override?

CarlosFdez: Caution before a release. If you're feeling lucky then lets do it.

CarlosFdez: Done!

PR #19371, opened 2025-06-15
Add build check for localization keys correctness in rule elements

Cuingamehtar:
This will iterate over all the fields in RE data, find those matching /^PF2E\.[^{}]+$/, and check if those are present in the static files.
The curvy braces makes it ignore those dynamic keys like class keys on some feats.
In addition, fixed some incorrect keys I found with this approach. Divine Vessel keys are not fixable as the traits no longer exist. Not sure what to do with those. Maybe have a list of exceptions.

Important: the flattenObject I used was based on core Foundry's, but it's modified to be able to handle arrays (Foundry implementation considers arrays unflattable). I exported it together with other foundry-utils functions, but I'm not sure that's a correct thing to do.


Cuingamehtar:
Judging by the fact that my typo https://github.com/foundryvtt/pf2e/pull/19376 slipped by, the checked pattern might not be sufficient.

PR #19380, opened 2025-06-16
Fix scroll wheel on delta inputs

### LINKED ISSUE #19331: Mouse wheel doesn't work on delta inputs
Some fields of the character sheet don't trigger any update (until the sheet  is closed)..

https://github.com/user-attachments/assets/655f0d76-51ba-4ddc-bcc6-d232d50e7f8a

The video shows that modifying the `HP`, `Temp HP` or `Shield HP` fields do not trigger any update while other fields like `Stamina` or `Resolve` do.

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/19331

This also refactors delta inputs it to work standalone even if not targeting actor properties, which is helpful for direct ammo manipulation (which will be useful for abilities that expend ammo that are not automated).

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -60,7 +60,7 @@ import type {
     InventoryItem,
     SheetInventory,
 } from "./data-types.ts";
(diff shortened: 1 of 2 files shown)


stwlam:
We shouldn't support scroll wheel for any of this. :)

PR #19389, opened 2025-06-17
Remove actor clone from inventory preparation

### LINKED ISSUE #19349: Inventory Tab on Character Sheet does not update after Prototype Token is updated until page refresh
Foundry Build: 345
Pf2e Version: 7.1.1
Modules Enabled: 0

Reproduction steps:

1. Open an existing or new actor
2. Go to Inventory tab and add a new item from the item browser. Inventory Tab updates to show the item.
3. Open the Prototype Token config from header
4. Don't have to make changes, but hit Update Token to close the Prototype Token config
5. Try to add item to inventory again. Inventory Tab will not update to show the item (but it will appear in the Actions tab if it's a weapon, etc)
6. Refresh (F5) and the previously added items will now correctly appear in the Inventory Tab

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/19349 though the root problem remains

Removes what seems to be a not necessary or no longer necessary actor clone during data prep. The incorrect data is due to a core bug, but I think we're better without the clone anyways unless there's something I'm missing.

EDIT: Yeah without, a somewhat modified Jirelle's re-render goes from 90ms to 50ms on my machine. Appreciable speedup, though its a bit concerning how a single clone is that expensive. It likely doesn't matter in most scenarios at least.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -191,7 +191,8 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
     }
 
     protected prepareInventory(): SheetInventory {
-        const items = [...iterateAllItems(this.actor)].filter((i) => i.isOfType("physical"));
+        const actor = this.actor;
+        const items = [...iterateAllItems(actor)].filter((i) => i.isOfType("physical"));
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Since the bug this was working around is a core bug anyways, this is more of a performance fix, updated appropriately.

PR #19402, opened 2025-06-18
Add automation to `Fortune's Coin` items

ammalagonc:


TikaelSol:
Since this doesn't have a duration we should probably just have the rules on the items and ditch the effect by using a toggle

PR #19406, opened 2025-06-18
Adjust DC of invested items with Intensify Investiture

pedrogrullada:


packs/feats/class/thaumaturge/intensify-investiture.json
```diff
@@ -29,7 +29,29 @@
...
-        "rules": [],
+        "rules": [
+            {
+                "key": "RollOption",
+                "option": "intensify-investiture",
+                "toggleable": true
+            },
+            {
+                "key": "AdjustModifier",
+                "mode": "override",
```
TikaelSol:
```suggestion
                "mode": "upgrade",
```
Does upgrade work? Do we actually need the toggle? While you could use the base DC there really isn't a situation where you would.

pedrogrullada: It might, I will check tomorrow. It needs to be a toggle because it's a limited use action.

TikaelSol: Ah, then yeah just see if upgrade works. Also same for the alchemist PR.

PR #19420, opened 2025-06-19
Exclude flat checks from adjusting `inline-dc`

In3luki:
Required to exclude some check types (e.g. flat check) from being adjusted by rule elements.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -875,7 +875,9 @@ function getCheckDC({
         const idDomain = item ? `${item.id}-inline-dc` : null;
         const slugDomain = `${sluggify(name)}-inline-dc`;
         const domains =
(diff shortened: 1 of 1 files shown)


In3luki:
Now excludes all flat checks from being modified by the `inline-dc` statistic. I've kept the type because that might still be useful but it can be removed if necessary.

CarlosFdez:
We might still want a statistic for it, but with a more specific selector. We should make sure nothing is relying on it though.

I'd prefer to avoid adding the extra option since they're hard to take-backsie, but I don't feel too strongly about that.

In3luki:
Updated to give inline flat checks the `flat-check` domain exclusively. Should that be `inline-flat-check`?
Also I found no rule elements that modify the flat check dc value in the system at the moment.
Also removed the `type` roll option. I think it's not really required anymore at this point.

PR #19434, opened 2025-06-21
Fix Dullahan Keen Note outcome and localization key

Cuingamehtar:
As it was, the Note was never displayed, as the result is already promoted by the time this Note checks it.

But should this Note even be here? Dullahan's Strikes all add Head Hunter as an additional effect, and that one has a link to the Keen rune in it's description.


Cuingamehtar:
I tried to change the Note priority to be before the Adjust Degree of Success RE, but it still doesn't show with `success` outcome. Seems like promotion happens before the Note RE anyway.

PR #19469, opened 2025-06-23
Add name and group alterations to ItemAlteration

mechamaya:
This adds `group` and `name` alterations to ItemAlteration. Group validation checks to make sure if the itemType is `weapon` then it checks for valid weapon groups, and same for `armor` and armor groups. Also includes validation to prevent changing "ranged only" and "melee only" weapons into incompatible groups, to prevent alterations that create invalid weapon states.

Name alteration works but if we want to restrict it to just weapon and armor names for now I can add that, since that's all that's needed at the moment.

----

These were added because I'm working on some Starfinder Mechanic playtest data entry and being able to alter weapon groups and rename weapons was useful to selecting different turret weapon arrays, which changes an existing turret weapon item to have different properties. And it's useful to use a weapon item rather than a generated strike, because this weapon can have upgrades and improvements applied to it like any other weapon.

### DIFF
--- a/src/module/rules/rule-element/item-alteration/schemas.ts
+++ b/src/module/rules/rule-element/item-alteration/schemas.ts
@@ -301,6 +301,38 @@ const ITEM_ALTERATION_VALIDATORS = {
         }),
         value: new StrictNumberField({ required: true, nullable: false, integer: true, initial: undefined } as const),
     }),
+    "group": new ItemAlterationValidator(
+        {
+            itemType: new fields.StringField({ required: true, choices: ["armor", "weapon"] }),
+            mode: new fields.StringField({ required: true, choices: ["override"] }),
+            value: new fields.StringField({
+                required: true,
+                nullable: false,
+                initial: undefined,
+            } as const),
+        },
+        {
+            validateForItem: (item, alteration): validation.DataModelValidationFailure | void => {
+                const value = alteration.value;
+                if (item.type === "armor") {
+                    const validTraits = CONFIG.PF2E.armorGroups;
+                    if (typeof value !== "string" || !(value in validTraits)) {
+                        return new validation.DataModelValidationFailure({
+                            message: `${alteration.value} is not a valid armor group`
+                        });
+                    }
+                }
+                else if (item.type ==="weapon") {
+                    const validTraits = CONFIG.PF2E.weaponGroups;
+                    if (typeof value !== "string" || !(value in validTraits)) {
+                        return new validation.DataModelValidationFailure({
+                            message: `${alteration.value} is not a valid weapon group`
+                        });
+                    }
+                }
+            },
+        }
+    ),
     hardness: new ItemAlterationValidator({
         itemType: new fields.StringField({ required: true, choices: Array.from(PHYSICAL_ITEM_TYPES) }),
         mode: new fields.StringField({
@@ -465,6 +497,11 @@ const ITEM_ALTERATION_VALIDATORS = {
             blank: false,
             initial: undefined,
         } as const),
+    }),
+    "name": new ItemAlterationValidator({
+        itemType: new fields.StringField({required: true}),
+        mode: new fields.StringField({required: true, choices: ["override"],}),
+        value: new fields.StringField({required: true,nullable: false,blank: false,} as const),
     }),
     "speed-penalty": new ItemAlterationValidator({
         itemType: new fields.StringField({
--- a/src/module/rules/rule-element/item-alteration/schemas.ts
+++ b/src/module/rules/rule-element/item-alteration/schemas.ts
@@ -377,6 +377,16 @@ const ITEM_ALTERATION_VALIDATORS = {
             initial: undefined,
         } as const),
     }),
+    img: new ItemAlterationValidator({
+        itemType: new fields.StringField({ required: true }),
+        mode: new fields.StringField({ required: true, choices: ["override"] }),
+        value: new fields.FilePathField({
+            required: true,
+            nullable: false,
+            blank: false,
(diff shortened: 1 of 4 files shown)


CarlosFdez:
what happens if you change from a melee weapon group to a mandatory ranged, or vice versa?

It might be good to validate that you're not turning a melee to a ranged or vice versa, at least for now?

src/module/rules/rule-element/item-alteration/schemas.ts
```diff
@@ -301,6 +301,37 @@ const ITEM_ALTERATION_VALIDATORS = {
...
+                nullable: false,
+                initial: undefined,
+            } as const),
+        },
+        {
+            validateForItem: (item, alteration): validation.DataModelValidationFailure | void => {
+                const value = alteration.value;
+                if (item.type === "armor") {
+                    const validTraits = CONFIG.PF2E.armorGroups;
+                    if (typeof value !== "string" || !(value in validTraits)) {
```
CarlosFdez: You should be able to replace this with `if(objectHasKey(CONFIG.PF2E.armorGroups, value))`, which was added to make these checks simpler.

mechamaya:
> what happens if you change from a melee weapon group to a mandatory ranged, or vice versa?
> 
> It might be good to validate that you're not turning a melee to a ranged or vice versa, at least for now?

Yeah, I'll test this more tomorrow when I get a chance, if it creates problems I'll see if some safeguards can be put in place.  Though when I was testing I was able to use an alteration to put a ranged weapon into the sword group and nothing bad happened yet.

CarlosFdez:
I can imagine that some of the problems would be sneaky. There's a non-zero chance it'll cause issues as well once I finish adding ammo data. Unless there's a need I think the extra safeguard for now might be helpful. It can always be removed later if we need to, with whatever modifications is needed as well.

CarlosFdez:
I just remembered that you may also need to prevent changes to the item sheet from overwriting the actual weapon group with the prepared one. The usual approach is to use data-property. The file is weapon-details.hbs

mechamaya:
I tested it some more today and the most obvious problem was I was once able to glitch an Arquebus into locking itself into the bow group, even after removing the rule element. The sheet kept trying to change it back to Bow even without a rule element, and even with other rule elements coming in and trying to override it back to firearm. I'm not sure if that was the sheet trying to change it back, or a rogue rule element that persisted on the object even after refreshing. Only deleting the weapon stopped it from running. But I haven't been able to reproduce that particular issue more than once so far.

Another test I did was using the Item Alteration to set it to the sword group, manually adjusting the range to 0ft (melee) in the item editor, then back to some ranged range, then removing the rule element. This leaves it as sword (doesn't return it back to Firearm).

I definitely worry about the ammo changes coming, especially if they are tied to group. I'm starting to look into preventing the ranged/melee crossover during validation.

(as this is my first code PR to the system, thanks for entertaining this)

CarlosFdez:
The reason it didn't change back was the issue I mentioned about needing to prevent changes. If you make any form change on an item sheet, it will do a submit with all data currently on the form. Since the prepared value is the one in the form, it is the one being submit, thus becoming permanent. This is the reason for all the recurring "a thing changed by a rule element became permanent" bugs. In weapon-details.hbs, look at the precious material selection or system.damage.dice for examples on how to fix it (replacing `name` with `data-property` is usually enough).

TBH, some of those sheet fields are based on whether it is a melee or ranged weapon, so given that someone could start fiddling with the item data, it is not safe to allow changes that can cause users to edit in a corrupted state. A lot of validation code would have to start switching to using source data instead of prepared data. A very involved project and not at all needed for this playtest.

An no problem, thanks a lot for putting the effort in trying to figure it all out. At times the codebase can be a very complicated game of jenga if you're not careful. Typescript can only do so much.

mechamaya:
For sure, it's definitely a bigger issue than I thought but I'm still going to give it an honest try. Definitely overkill for the playtest for sure, but if I can get it working it could be beneficial for other data entry projects in the future. Starfinder doing some weird stuff and who knows what rules we might need.

CarlosFdez:
I need to try to test this soon. I'm curious how this interacts with +2 greater striking localization, especially considering that localization modules can override that.

mechamaya:
I've done some testing on the name changes with adding fundamental runes and whatnot, and it seems to override those changes, which I would say is the expected behavior. But further testing wouldn't hurt.

I should also mention I've added the `img` property to the list of alterations as well, but if either img or `name` end up being too problematic I can take those out. The `group` change is really the big one that I'd like to try to push if possible, the rest are nice to haves. But with all three together we can basically turn any ranged weapon into any other type of ranged weapon while still supporting attachments and runes and the like. Being able to change the item image and the name is quite handy, as long as we don't find any weird side effects that this creates.

CarlosFdez:
Overriding the name changes sounds reasonable to me. It not doing that was my concern.

CarlosFdez:
The hotfix is released so it should be possible to circle back. I'm wondering if "group" is correct or if it should be called "weapon-group" and "armor-group". I genuinely don't know if item alterations should lean towards the rules or towards the implementation.

We have a "category", which is pretty generic, so maybe its fine. But I'll leave it that comment here in case anyone has a different opinion.

stwlam:
Group is fine so long as it fully validates for either a weapon or armor and its corresponding group options. I don't think we can do img without preventing it from saving upon closure of the item sheet.

mechamaya:
I haven't run into issues with the img saving on sheet close. But an issue I am encountering is when you open the img file picker and it auto-fills the path to the overridden image rather than the source image. So I'm pulling the img alteration out of this PR and putting it aside as a separate branch until I sort that issue out.

PR #19503, opened 2025-06-25
Restore wheel rotation for unlocked measured template previews and add a tooltip explaining the new placement workflow

In3luki:
image

image

### DIFF
--- a/static/templates/chat/spell-card.hbs
+++ b/static/templates/chat/spell-card.hbs
@@ -54,8 +54,15 @@
                         </div>
                     {{/if}}
                     {{#if (and data.area (not data.isAura))}}
-                        <div class="spell-button">
-                            <button type="button" data-action="spell-template" data-visibility="owner">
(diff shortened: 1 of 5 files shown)


static/templates/chat/spell-card.hbs
```diff
@@ -54,8 +54,16 @@
                         </div>
                     {{/if}}
                     {{#if (and data.area (not data.isAura))}}
-                        <div class="spell-button">
-                            <button type="button" data-action="spell-template" data-visibility="owner">
+                    {{log this}}
```
CarlosFdez:
```suggestion
```

In3luki: Oops

CarlosFdez:
I was not aware of the abort controller, what an intriguing little toy. I tested it locally and it seems to perform well. I'm wondering if a manual rotate should make it place immediately afterwards, but we can wait for more user testing to be sure, this is still a direct improvement.

PR #19506, opened 2025-06-25
Fix the level and price of Dr. Ushernacht's Astonishing Ink (Major)

HavocsCall:
According to Archives of Nethys: https://2e.aonprd.com/Equipment.aspx?ID=3564, this should be level 17 and 2250 gold. The modifier and will save are already correct.

### DIFF
--- a/packs/equipment/dr-ushernachts-astonishing-ink-major.json
+++ b/packs/equipment/dr-ushernachts-astonishing-ink-major.json
@@ -19,7 +19,7 @@
             "value": 0
         },
         "level": {
-            "value": 10
+            "value": 17
         },
         "material": {
             "grade": null,
@@ -27,7 +27,7 @@
         },
         "price": {
             "value": {
-                "gp": 160
+                "gp": 2250
             }
         },
         "publication": {


pedrogrullada:
We prefer if things are confirmed in the actual source, but of course not everyone has access to it. I've confirmed in the PDF.
image

HavocsCall:
I do not have the PDF, but will look into getting it. Thanks for the heads up.

pedrogrullada:
> I do not have the PDF, but will look into getting it. Thanks for the heads up.

Yeah, it's fine since someone here usually has the book to check, so don't worry, you don't have to get the book on our account :laughing:  Thanks for contributing!

HavocsCall:
If I don't get it, is there another way to confirm without making you guys look it up?

Although, I have a feeling even if someone says they looked it up in the PDF, you'd still check, rightfully so.

pedrogrullada:
> If I don't get it, is there another way to confirm without making you guys look it up?
> 
> Although, I have a feeling even if someone says they looked it up in the PDF, you'd still check, rightfully so.

I don't think there is another way, but it's fine. If you mention in the PR description that you haven't confirmed in the PDF that should be enough of a prompt for the reviewer to take a peek or for them to ask someone who has the book to. If you want to make sure before putting the work on a PR, you could always ask someone in the #paizo-games channel on our community Discord to confirm it for you.

PR #19529, opened 2025-06-28
Support homebrew class traits and use in feat subfeatures

CarlosFdez:
Only via module flags for now. Class traits added will show up in feat subfeatures.
image

If the module becomes disabled, rather than exploding, it will warn in the console and make the invalid traits deletable

chrome_2025-07-10_06-36-33

### DIFF
--- a/src/module/item/feat/sheet.ts
+++ b/src/module/item/feat/sheet.ts
@@ -108,6 +108,8 @@ class FeatSheetPF2e extends ItemSheetPF2e<FeatPF2e> {
         const feat = this.item;
         const localize = localizer("PF2E.Actor.Character");
         const selectedIncreases = feat.system.subfeatures.proficiencies;
+        const sourceData = this.item.system._source.subfeatures.proficiencies;
+        const invalidKeys = Object.keys(sourceData).filter((k) => !(k in selectedIncreases));
 
         return {
             other: {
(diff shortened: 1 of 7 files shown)


CarlosFdez:
The easy alternative is to not validate proficiency keys in the schema, and to handle it in the feat system data during prepareBaseData(). This would remove the need for a LaxRecordField at all. I'd be down to do so if preferred.

As for alternative solutions for the LaxRecordField, while there were simpler solutions for removing the invalid keys, I could not find a different solution that also preserved them in the source object to allow removal. I would love alternatives, though ideally one that's tested.

src/module/item/feat/sheet.ts
```diff
@@ -432,7 +445,7 @@ interface FeatSheetData extends ItemSheetDataPF2e<FeatPF2e> {
     hasSenses: boolean;
     languages: LanguageOptions;
     mandatoryTakeOnce: boolean;
-    maxTakableOptions: FormSelectOption[];
+    maxTakableOptions: foundry.applications.fields.FormSelectOption[];
```
stwlam:
```suggestion
    maxTakableOptions: fa.fields.FormSelectOption[];
```

stwlam:
Do something more than just make invalid entries red, such as using `notification error` CSS (or just the icon if that's too much).

CarlosFdez:
Done. notification error looks pretty good once the min-height is set.

PR #19542, opened 2025-06-29
Add ItemAlteration Description to Extended Kinesis

MrVauxs:
Also add a missing period to the description.

### DIFF
--- a/packs/feats/class/kineticist/extended-kinesis.json
+++ b/packs/feats/class/kineticist/extended-kinesis.json
@@ -40,7 +40,8 @@
               "property": "description",
               "value": [
                 {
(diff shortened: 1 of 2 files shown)


packs/feats/class/kineticist/extended-kinesis.json
pedrogrullada, line 1: Remember to localize text. ~~Also, you can use a single Item Alteration rule and predicate each paragraph object inside the value array.~~ (Ignore this, you *were* using a single rule, my bad.)

There is a `"title"` key you can use for the bolded titles (Proliferate, Regulate and Sculpt).

PR #19553, opened 2025-06-30
Fix superfluous rerender happening sometimes when there wasn't a change

CarlosFdez:
This was at its worst when changing turns in initiative.

### DIFF
--- a/src/module/actor/helpers.ts
+++ b/src/module/actor/helpers.ts
@@ -818,12 +818,13 @@ async function applyActorGroupUpdate(
     data: Partial<ActorGroupUpdate>,
     { render = true }: { render?: boolean } = {},
 ): Promise<void> {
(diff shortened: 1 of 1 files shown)


src/module/actor/helpers.ts
```diff
@@ -835,7 +836,7 @@ async function applyActorGroupUpdate(
         await actor.deleteEmbeddedDocuments("Item", itemDeletes, { render: false });
     }
 
-    const changed = data.actorUpdates || itemCreates.length || itemUpdates.length || itemDeletes.length;
+    const changed = !!actorUpdates || itemCreates.length || itemUpdates.length || itemDeletes.length;
```
stwlam, line 839: Why !! when the rest are all numbers :o)

CarlosFdez: No reason, can be removed or added to all. Late night bug fix quirks

PR #19592, opened 2025-07-05
Fix handling of reach value of zero in CreatureSize RE

### LINKED ISSUE #19370: Large PCs incorrectly affected by Applereed Mutagen
### Description

To my understanding Applereed Mutagen increases a creature's size by 1 step, but does not provide any additional benefits apart from the increase in space (e.g. no extra reach is granted). Per RAW: "...your larger size does not have any additional effects beyond the potential change in space." In Foundry, this behaves correctly on Medium PCs, but on Large PCs an additional 5ft reach is given for an unknown reason when increasing their size to Huge.

I used the following macro to test this:
```
for (let token of canvas.tokens.controlled) {
  console.log(token.actor.getReach({action: 'attack'}))
}
```

### Steps to reproduce
1. Create a new script macro using the above code.
2. Create an Actor of type "Player Character"
   2.1. Add the "Elf" ancestry to the actor (or any other Medium ancestry)
   2.2. Place the actor into a scene
   2.3. Run the macro, and observe that the console outputs "5" **(expected)**
3. Add an Applereed Mutagen item to the actor's inventory
   3.1. Click on Applereed Mutagen, then drag the Effect onto the the token.
   3.2. Run the macro, and observe that the console outputs "5" **(expected)**
   3.3 Remove the Applereed Mutagen effect
4. Change the actor's ancestry to "Centaur" or "Minotaur" (both Large size ancestries)
   4.1. ~(potentially related?) Observe https://github.com/foundryvtt/pf2e/issues/19361#issue-3146443596~
   4.2. Run the macro, and observe that the console outputs "5" **(expected)**
   4.3. Click on Applereed Mutage, then drag the Effect onto the token
   4.4. Run the macro, and observe that the console outputs "10" **(unexpected, should be "5")**

### Notes:
- During step 4, creating a new actor still produces the unexpected behavior.
- A potentially similar issue occurs with a 2nd level "Enlarge" spell effect. RAW, this has no effect on large or larger creatures, but a Large PC still gains +5ft reach and clumsy 1 with this effect. However note that a heightened (4th level) "Enlarge" **does** work properly, giving +10ft reach and enlarging to Huge, which does work on Large PCs.

### Version Info:
PF2E System: 7.1.1, 7.2.0
Foundry VTT: Version 13 Build 345
Modules Installed: None

stwlam:
- Closes #19370

### DIFF
--- a/packs/equipment-effects/effect-applereed-mutagen-greater.json
+++ b/packs/equipment-effects/effect-applereed-mutagen-greater.json
@@ -39,6 +39,9 @@
             },
             {
                 "key": "CreatureSize",
+                "reach": {
+                    "add": 0
+                },
                 "value": 1
             },
             {
--- a/packs/equipment-effects/effect-applereed-mutagen-lesser.json
+++ b/packs/equipment-effects/effect-applereed-mutagen-lesser.json
@@ -47,6 +47,9 @@
             },
             {
                 "key": "CreatureSize",
+                "reach": {
+                    "add": 0
+                },
                 "value": 1
             }
         ],
--- a/packs/equipment-effects/effect-applereed-mutagen-moderate.json
+++ b/packs/equipment-effects/effect-applereed-mutagen-moderate.json
@@ -47,6 +47,9 @@
             },
             {
                 "key": "CreatureSize",
+                "reach": {
+                    "add": 0
+                },
                 "value": 1
             }
         ],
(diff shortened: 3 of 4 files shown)


CarlosFdez:
If I'm reading this correctly, applereed will also need a reach that adds 0 in order to actually close that issue right?

EDIT: resolved

PR #19614, opened 2025-07-07
Improve search performance of ABC picker and fix clearing query

CarlosFdez:
By clearing query, I mean clicking the little X icon on the input.

### DIFF
--- a/src/module/actor/character/apps/abc-picker/app.svelte
+++ b/src/module/actor/character/apps/abc-picker/app.svelte
@@ -1,9 +1,23 @@
 <script lang="ts">
     import { ErrorPF2e } from "@util";
-    import type { KeyboardEventHandler, MouseEventHandler } from "svelte/elements";
+    import type { MouseEventHandler } from "svelte/elements";
     import type { ABCPickerContext } from "./app.ts";
 
     const { actor, foundryApp, state: data }: ABCPickerContext = $props();
+    let searchQuery = $state("");
+    const filteredItems = $derived.by(() => {
+        const query = searchQuery.trim();
+        if (!query.length) return data.items;
+
+        const regexp = new RegExp(RegExp.escape(query), "i");
+        return data.items.filter((item) => {
+            if (searchQuery.length === 0) return true;
+            const name = item.name;
+            const originalName = item.originalName;
+            return regexp.test(name) || (originalName && regexp.test(originalName));
+        });
+    });
+
     const typePlural = game.i18n.localize(`PF2E.Item.${data.itemType.capitalize()}.Plural`);
     const searchPlaceholder = game.i18n.format("PF2E.Actor.Character.ABCPicker.SearchPlaceholder", {
         items: typePlural,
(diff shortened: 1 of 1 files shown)


src/module/actor/character/apps/abc-picker/app.svelte
```diff
@@ -13,7 +27,7 @@
     const viewItemSheet: MouseEventHandler<HTMLButtonElement> = async (event): Promise<void> => {
         const uuid = event.currentTarget.closest("li")?.dataset?.uuid ?? "";
         const item = await fromUuid(uuid);
-        item?.sheet.render(true);
+        item?.sheet?.render(true);
```
stwlam:
```suggestion
        item?.sheet.render(true);
```

src/module/actor/character/apps/abc-picker/app.svelte
```diff
@@ -26,26 +40,15 @@
...
-            const originalName = row.dataset.originalName;
-            row.hidden = !regexp.test(name) && !(originalName && regexp.test(originalName));
-        }
-    }, 200);
 </script>
 
 <search>
     <i class="fa-solid fa-search"></i>
-    <input type="search" spellcheck="false" placeholder={searchPlaceholder} onkeyup={searchItems} />
+    <input type="search" spellcheck="false" placeholder={searchPlaceholder} bind:value={searchQuery} />
```
stwlam, line 47: This doesn't respond to keydown events, right?

src/module/actor/character/apps/abc-picker/app.svelte
CarlosFdez: image

stwlam: As in every case, fix the type rather than dancing around it in runtime unless necessary. `Item#sheet` is never `null`.

src/module/actor/character/apps/abc-picker/app.svelte
CarlosFdez: It does update as I type, so its not just blur. I see registration of a generic event handler on both keyup/keydown in the settings. Probably a whole bunch of things.

src/module/actor/character/apps/abc-picker/app.svelte
CarlosFdez: It is likely because fromUuid is returning a ClientDocument rather than necessarily an item. The alternative is instanceof Item. Still a check either way.

CarlosFdez: Ah, I could do a generic which ends up casting it to an item.

src/module/actor/character/apps/abc-picker/app.svelte
stwlam: Not focus/blur, keyup/keydown

CarlosFdez: Ah, you mean specifically keydown as opposed to just keyup, rather than compared to blur. Yes, this will do the search as soon as the key is down. Same way the formula picker currently does it. That said before it was using a debounce, so the keyup served to increase the debounce window. For svelte we don't need that.

CarlosFdez:
This is how it works after these changes


https://github.com/user-attachments/assets/f0c0a7ce-53dc-47a2-a24f-327a847a3e17

PR #19615, opened 2025-07-07
Automate Instinct Crown

### LINKED ISSUE #18377: Instinct Crown doesn't have rule elements or an effect
If you play an Animal Barbarian and get your hands on an Instinct Crown, you can't actually use the Rage action from the crown.
It would make your unarmed attacks Deadly D8, but you can't even manually fix it by giving your handwraps the Deadly D8 trait yourself.
Handwraps will never roll Deadly D8. Other weapons do, when manually given the trait.

pedrogrullada:
Closes #18377

Dragon effect *could* be done with an Ephemeral Effect note, but it would have wrong origin context and no way to retrieve the correct damage type. A Note on damage taken could work as well, but isn't really right with the rules of the item, so I opted to leave it without automation until we can do better.


packs/equipment-effects/effect-instinct-crown-quick-tempered.json
```diff
@@ -0,0 +1,53 @@
+{
+    "_id": "ZzsUW4Jp2LuPTT8P",
+    "img": "icons/equipment/head/crown-feather-brown.webp",
+    "name": "Effect: Instinct Crown (Quick-Tempered)",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Instinct Crown (Animal)], @UUID[Compendium.pf2e.equipment-srd.Item.Instinct Crown (Dragon)], @UUID[Compendium.pf2e.equipment-srd.Item.Instinct Crown (Giant)], @UUID[Compendium.pf2e.equipment-srd.Item.Instinct Crown (Spirit)], @UUID[Compendium.pf2e.equipment-srd.Item.Instinct Crown (Superstition)]</p>\n<p>You Rage and gain an additional 10 temporary Hit Points from that Rage.</p>"
```
TikaelSol: Should this be an effect or just a toggle? It's once per day but dragging two effects onto the actor seems odd

pedrogrullada: I liked the idea of granting rage with it, but a toggle is probably less awkward in actual play. I'll do just that.

PR #19619, opened 2025-07-08
Add special "magic" immunity to Will-O'-Wisp

stwlam:
"A will-o'-wisp is immune to all spells except force barrage, quandary, and revealing light."


pedrogrullada:
Late to the party, but is there a reason for this one to not be a custom exception of the recently defined magic immunity instead of an entirely custom resistance? I'm guessing it's functionally the same, but just wondering.

PR #19624, opened 2025-07-08
Fix chat token images blocking saves posted to chat once more

alexbrault:
Restores a fix from #17756 that was lost in the v13 migration


stwlam:
Make it inert in the img tag rather than using CSS

PR #19646, opened 2025-07-13
Update automation to Conspirator Dragon's `Conjure Disguise`

ammalagonc:


packs/pathfinder-monster-core/conspirator-dragon-adult-spellcaster.json
```diff
@@ -2251,7 +2251,7 @@
         },
         {
             "_id": "OpdWoyPywj7o3f4N",
-            "img": "systems/pf2e/icons/default-icons/action.svg",
+            "img": "systems/pf2e/icons/actions/Reaction.webp",
```
TikaelSol, line 2254: We don't need to set the icons on NPC abilities on actors. It's fine, it's just not needed when they're on the actors.

PR #19648, opened 2025-07-13
Replace collapsed actions with general minHeight based solution

CarlosFdez:
This allows any description or card-content to become auto-previewable if it has the `data-auto-collapse` property. The preview is now based on min-height rather than text only as well.

I went for an approach that had minimal dom manipulation with no wrapper elements, so I needed to do something a bit weird with the shadow element since otherwise the mask-image would have made it invisible. Hopefully that's ok.

After this it should be easy to leverage on things like spells.

https://github.com/user-attachments/assets/70491c26-735b-4879-80e1-df74e69ecabe

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -243,6 +243,19 @@ class ChatMessagePF2e extends ChatMessage {
             }
         }
 
+        // If the description has auto-collapse, collapse if text exceeds a certain length
+        // Its not possible to check the actual size if its not in the DOM yet
+        const collapsableElement = html.querySelector<HTMLElement>(
+            ".description[data-auto-collapse], .card-content[data-auto-collapse]",
+        );
+        if (collapsableElement && collapsableElement.innerText.length > 250) {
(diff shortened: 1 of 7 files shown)


src/module/chat-message/document.ts
```diff
@@ -243,6 +243,19 @@ class ChatMessagePF2e extends ChatMessage {
             }
         }
 
+        // If the description has auto-collapse, collapse if text exceeds a certain length
+        // Its not possible to check the actual size if its not in the DOM yet
+        const collapsableElement = html.querySelector(
+            ".description[data-auto-collapse], .card-content[data-auto-collapse]",
+        );
+        if (collapsableElement instanceof HTMLElement && collapsableElement.innerText.length > 250) {
```
stwlam:
```suggestion
        const collapsableElement = html.querySelector<HTMLElement>(
            ".description[data-auto-collapse], .card-content[data-auto-collapse]",
        );
        if (collapsableElement && collapsableElement.innerText.length > 250) {
```

PR #19681, opened 2025-07-18
Stop adding invested trait to armor potency alterations

CarlosFdez:


CarlosFdez:
This also fixes combining the potency alteration with ac-bonus, which hasn't happened yet but it was only a matter of time.

stwlam:
This should be an option since item alterations don't just come from consumables and can be permanent in one way or another.

CarlosFdez:
I had forgotten to check isInvested, so that's done. If it already requires investment then it shouldn't just do the change.

Between being able to add an invested trait via alteration as well as such an effect (maybe) not existing yet, I don't know if we should add the option. But if that's not sufficiently convincing, how would I go about it?

edit: perhaps options objectfield that individual alterations can define a schema for?

PR #19682, opened 2025-07-18
Implement rerolling checks with a mythic point

In3luki:
[Recording 2025-07-18 at 12.34.43.webm](https://github.com/user-attachments/assets/66bff17c-97dd-4c49-9818-15bb36d31a54)

Rerolling with a hero point still looks the same as before:
image

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -39,9 +39,9 @@ import { CheckRoll, CheckRollDataPF2e } from "./roll.ts";
 import { CheckCheckContext } from "./types.ts";
 
 interface RerollOptions {
-    /** @deprecated Use `usepoint: "hero"` instead. */
+    /** @deprecated Use `resource: "hero-points"` instead. */
     heroPoint?: boolean;
-    usePoint?: "hero" | "mythic";
+    resource?: "hero-points" | "mythic-points";
     keep?: "new" | "higher" | "lower";
 }
 
@@ -445,7 +445,7 @@ class CheckPF2e {
     /** Reroll a rolled check given a chat message. */
     static async rerollFromMessage(
         message: ChatMessagePF2e,
-        { heroPoint = false, keep = "new", usePoint }: RerollOptions = {},
+        { heroPoint = false, keep = "new", resource }: RerollOptions = {},
     ): Promise<void> {
         if (!(message.isAuthor || game.user.isGM)) {
             ui.notifications.error(game.i18n.localize("PF2E.RerollMenu.ErrorCantDelete"));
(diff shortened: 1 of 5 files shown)


src/module/system/check/check.ts
```diff
@@ -39,7 +39,9 @@ import { CheckRoll, CheckRollDataPF2e } from "./roll.ts";
 import { CheckCheckContext } from "./types.ts";
 
 interface RerollOptions {
+    /** @deprecated Use `usepoint: "hero"` instead. */
     heroPoint?: boolean;
+    usePoint?: "hero" | "mythic";
```
CarlosFdez: Perhaps rename to resource.

src/module/system/check/check.ts
```diff
@@ -457,26 +460,48 @@ class CheckPF2e {
...
+                                rerollFlavor: "PF2E.RerollMenu.MessageMythicPoint",
+                            };
+                        default:
+                            return null;
+                    }
+                })();
+                if (resource) {
+                    const count = resource.data.value;
+                    if (count > 0) {
+                        await rerollingActor.updateResource(resource.name, Math.clamp(count - 1, 0, resource.data.max));
```
CarlosFdez: Between the count > 0 and updateResource() already clamping, I don't think you need to clamp again.

src/module/system/check/check.ts
```diff
@@ -457,26 +460,48 @@ class CheckPF2e {
...
-                        ),
-                    });
-                    rerollFlavor = game.i18n.format("PF2E.RerollMenu.MessageHeroPoint", { name: rerollingActor.name });
-                } else {
-                    ui.notifications.warn(
-                        game.i18n.format("PF2E.RerollMenu.WarnNoHeroPoint", { name: rerollingActor.name }),
-                    );
-                    return;
+                const resource = (() => {
+                    switch (usePoint) {
```
CarlosFdez: You can likely simplify this logic by using a format string. The plural name of the resource is in the resource data, which is in turn localized from PF2E.Actor.Resource (which regrettably, I should have keyed by either property name or slug)

You would do something like rename usePoint to resource and then `resource = resource ?? (hero-points ? "hero-points" : null)`. You can then get the resource data from `actor.getResource(slug)`. Make sure you use the slugs, even if camel case is technically supported.

In3luki: That is indeed much simpler and cleaner! Updated.

src/module/system/check/check.ts
```diff
@@ -457,24 +460,31 @@ class CheckPF2e {
...
+                const points = rerollingActor.getResource(resource);
+                if (points && points.value > 0) {
+                    await rerollingActor.updateResource(resource, points.value - 1);
+                    rerollFlavor = game.i18n.localize(
+                        `PF2E.RerollMenu.Message${sluggify(resource, { camel: "bactrian" })}`,
+                    );
                 } else {
                     ui.notifications.warn(
-                        game.i18n.format("PF2E.RerollMenu.WarnNoHeroPoint", { name: rerollingActor.name }),
+                        game.i18n.format("PF2E.RerollMenu.WarnNoResource", {
```
CarlosFdez: I think ui.notifications.warn accepts a format parameter now.

In3luki: Oh nice, I must have missed that. Updated.

src/module/system/check/check.ts
```diff
@@ -488,27 +499,42 @@ class CheckPF2e {
         context.skipDialog = true;
         context.isReroll = true;
         context.options.push("check:reroll");
-        if (heroPoint) context.options.push("check:hero-point");
+        if (resource) context.options.push(`check:${resource.slice(0, resource.length - 1)}`);
```
CarlosFdez: I don't think I'm a fan of presuming plural ending with s and converting to signature. I'd actually rather a ternary. Or introduce check:resource:X using the resource's slug. I'm not sure, I'm willing to receive input on this one.

src/module/system/check/check.ts
```diff
@@ -637,6 +677,9 @@ class CheckPF2e {
         // Remove the buttons if this is the discarded roll
         if (isOld) element.querySelector(".message-buttons")?.remove();
 
+        // Add mythic reroll class to dice formula element if necessary
+        if (resource === "mythic-points") element.querySelector(".dice-formula")?.classList.add("mythic");
```
CarlosFdez: Given that you're using mythic to handle visibility, what are your thoughts on instead including hidden by default and omitting it if its mythic instead. This will show all old re-rolled formulas, but that's not really that bad, and it is more semantic.

src/module/system/check/check.ts
In3luki: Yeah I wasn't sure about that either. Probably best to keep `check:hero-point` for compatibility for a while and add `check:resource:${resource.slug}` going forward.

src/module/system/check/check.ts
In3luki: I'll give it a try.

src/module/system/check/check.ts
CarlosFdez: Someone might confuse that with a check that requires a resource to perform in the first place rather than a check that was re-rolled using a resource. That's why I said I wasn't sure.

In3luki: Maybe `check:reroll:${resource}`?

CarlosFdez: that sounds good to me!

src/module/system/check/check.ts
```diff
@@ -627,13 +666,26 @@ class CheckPF2e {
...
         if (typeof die?.total !== "number") throw ErrorPF2e("Unexpected error inspecting d20 term");
 
         const html = await roll.render();
         const element = parseHTML(`<div>${html}</div>`);
 
+        // Handle roll formula visibility
+        const formulaElement = htmlQuery(element, ".dice-formula");
+        if (formulaElement && !isOld && resource !== "mythic-points") {
+            // Hide rerolled formula for everything except mythic point rerolls
+            formulaElement.classList.add("hidden");
```
In3luki: I tried using `formulaElement.hidden = true` but that was stripped somewhere along the ways. The CSS class seems to work without problems.

PR #19684, opened 2025-07-19
Add effect for Extravagant Parry and brush up Dueling Parry

ammalagonc:
Looking at the two feats, they do things a bit differently, so they might as well have separate effects.

Dueling Parry now has a predicate on hands-free - let me know if we should remove that.

### DIFF
--- a/packs/feat-effects/effect-dueling-parry.json
+++ b/packs/feat-effects/effect-dueling-parry.json
@@ -4,7 +4,7 @@
     "name": "Effect: Dueling Parry",
     "system": {
         "description": {
-            "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Extravagant Parry] or @UUID[Compendium.pf2e.feats-srd.Item.Dueling Parry]</p>\n<p>You gain a circumstance bonus to AC.</p>"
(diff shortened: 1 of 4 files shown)


packs/feat-effects/effect-dueling-parry.json
```diff
@@ -23,6 +23,9 @@
         "rules": [
             {
                 "key": "FlatModifier",
+                "predicate": [
+                    "hands-free:but-really:1"
```
TikaelSol: This is the wrong hands free roll option to use to use, since it counts free-hand weapons as taking up a hand (It's basically only for Laughing Shadow Magus, who specifically calls out free-hand stuff as still taking up a hand).

ammalagonc:
Closing to redo -

packs/feat-effects/effect-dueling-parry.json
```diff
@@ -23,6 +23,9 @@
         "rules": [
             {
                 "key": "FlatModifier",
+                "predicate": [
+                    "hands-free:1"
```
TikaelSol: Make this gte 1, since we now have 4+ armed ancestries

PR #19687, opened 2025-07-21
Strip special keys when migrating to avoid errors

CarlosFdez:
In v12 these special keys were harmless, but in v13 they explode if they're malformed if you attempt to save them back into the document. Saved special keys don't really make sense, and are usually a problem anyways.

This is causing an issue for some pf2e hud users updating from v12. It is difficult for Idle to fix it on our behalf since foundry disables modules when updating to v13, but it probably shouldn't be so explodey anyways.

Related to but not the same as https://github.com/foundryvtt/foundryvtt/issues/13201


stwlam:
Link to the core issue in a comment so we can remove this whenever someone notices the issue is closed.

PR #19688, opened 2025-07-21
Add support for ammo types

CarlosFdez:
Even though this is a draft and the migration/localization is missing, I need approval before I can finish it. Let me know if there is a standalone thing here that I can or should split into a new PR to reduce the complexity of this one.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -1411,7 +1411,8 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         }
 
         // Show the ammo list if the weapon requires ammo
-        if (weapon.system.expend) {
+        // Built in ammo will require special handling, and isn't supported yet
+        if (weapon.system.expend && !weapon.system.ammo?.builtIn) {
             const compatible = ammos
                 .filter((a) => a.isAmmoFor(weapon))
                 .map((a) => ({ id: a.id, label: `${a.name} (${a.quantity})` }));
@@ -1472,7 +1473,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                         ? configuredAmmo.uses.value
                         : configuredAmmo.quantity
                     : (configuredAmmo?.quantity ?? 0);
-                params.consumeAmmo ??= expend > 0;
+                params.consumeAmmo = weapon.system.ammo?.builtIn ? false : (params.consumeAmmo ?? expend > 0);
 
                 if (params.consumeAmmo && expend > ammoRemaining) {
                     ui.notifications.warn(
--- a/src/module/item/consumable/document.ts
+++ b/src/module/item/consumable/document.ts
@@ -7,13 +7,14 @@ import { RawItemChatData } from "@item/base/data/index.ts";
 import { performLatePreparation } from "@item/helpers.ts";
 import { TrickMagicItemEntry } from "@item/spellcasting-entry/trick.ts";
 import type { SpellcastingEntry } from "@item/spellcasting-entry/types.ts";
+import { PickAThingPrompt } from "@module/apps/pick-a-thing-prompt/app.ts";
 import type { ValueAndMax } from "@module/data.ts";
 import { DamageRoll } from "@system/damage/roll.ts";
 import type { EnrichmentOptionsPF2e } from "@system/text-editor.ts";
 import { ErrorPF2e, setHasElement } from "@util";
 import * as R from "remeda";
 import type { ConsumableSource, ConsumableSystemData } from "./data.ts";
-import type { ConsumableCategory, ConsumableTrait, OtherConsumableTag } from "./types.ts";
+import { AmmoType, type ConsumableCategory, type ConsumableTrait, type OtherConsumableTag } from "./types.ts";
 import { DAMAGE_ONLY_CONSUMABLE_CATEGORIES, DAMAGE_OR_HEALING_CONSUMABLE_CATEGORIES } from "./values.ts";
 
 class ConsumablePF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends PhysicalItemPF2e<TParent> {
@@ -73,9 +74,13 @@ class ConsumablePF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
         this._embeddedSpell = undefined;
 
         this.system.uses.max ||= 1;
+        this.prepareAmmoData();
+    }
+
+    private prepareAmmoData() {
+        if (!this.isAmmo) return;
 
         // Refuse to serve rule elements if this item is ammunition and has types that perform writes
-        if (!this.isAmmo) return;
         for (const rule of this.system.rules) {
             if (rule.key === "RollOption" && "toggleable" in rule && !!rule.toggleable) {
                 console.warn("Toggleable RollOption rule elements may not be added to ammunition");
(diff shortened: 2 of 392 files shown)


src/module/apps/select-dialog/app.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ApplicationV2 } from "@client/applications/api/_module.mjs";
+import { SvelteApplicationMixin, SvelteApplicationRenderContext } from "@module/sheet/mixin.svelte.ts";
+import Root from "./app.svelte";
+
+interface SelectDialogConfiguration<TThing extends OptionType = OptionType> extends ApplicationConfiguration {
+    prompt: string;
+    options: (OptionGroup<TThing> | Option<TThing>)[];
+}
+
+class SelectDialog<TThing extends OptionType> extends SvelteApplicationMixin<
```
stwlam: Is this a reinvention of the `PickAThingPrompt`?

CarlosFdez: Unfortunately yes. The tagify select doesn't support option groups. I could maybe consider porting the current pick a thing to svelte as a standalone PR, though I don't know how to maintain the inheritance in svelte yet. We'll see.

CarlosFdez: Yeah I'm leaning on doing that first. Let me know if you believe otherwise but I'm going to try to port our current prompts into svelte.

src/module/item/consumable/document.ts
```diff
@@ -74,18 +75,33 @@ class ConsumablePF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extend
...
-        for (const rule of this.system.rules) {
-            if (rule.key === "RollOption" && "toggleable" in rule && !!rule.toggleable) {
-                console.warn("Toggleable RollOption rule elements may not be added to ammunition");
-                this.system.rules = [];
-                break;
-            } else if (["GrantItem", "ChoiceSet"].includes(String(rule.key))) {
-                console.warn(`${rule.key} rule elements may not be added to ammunition`);
-                this.system.rules = [];
-                break;
+        if (this.isAmmo) {
```
stwlam: Return early if it isn't ammo to spare the nesting level

src/module/item/weapon/data.ts
```diff
@@ -148,11 +162,10 @@ interface WeaponSystemData
     expend: number | null;
     reload: {
         value: WeaponReloadTime | null;
-        /** Whether the ammunition (or the weapon itself, if thrown) should be consumed upon firing */
-        consume: boolean | null;
         /** A display label for use in any view */
         label: string | null;
     };
+    ammunition: WeaponAmmunitionData | null;
```
stwlam: Let's go with `ammo` where possible to save on some ink

src/module/item/weapon/document.ts
```diff
@@ -228,7 +228,13 @@ class WeaponPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
...
     /** Whether this weapon can serve as ammunition for another weapon */
     isAmmoFor(weapon: WeaponPF2e): boolean {
-        return this.system.usage.canBeAmmo && !weapon.system.traits.value.includes("repeating");
+        const weaponAmmoTypeSlug = weapon.system.ammunition?.baseType;
+        const ammoTypeData = weaponAmmoTypeSlug ? CONFIG.PF2E.ammoTypes[weaponAmmoTypeSlug] : null;
+        if (!this.system.usage.canBeAmmo || ammoTypeData?.category !== "weapon") {
+            return false;
+        }
+
+        return tupleHasValue([null, this.baseType], ammoTypeData.weapon);
```
stwlam:
```suggestion
        if (!this.system.usage.canBeAmmo) return false;
        const ammoType = weapon.system.ammunition?.baseType;
        const typeData = ammoType ? CONFIG.PF2E.ammoTypes[ammoType] : null;
        return typeData?.category === "weapon" && tupleHasValue([null, this.baseType], typeData.weapon);
```

src/scripts/config/index.ts
```diff
@@ -312,6 +314,173 @@ const grades = {
...
 
+const ammoCategories = {
+    ...R.mapToObj(
+        ["arrow", "blowgun-dart", "bolt", "cutlery", "round", "sling-bullet", "spray-pellet", "wooden-taw"] as const,
+        (slug) => [
+            slug,
+            {
+                label: `PF2E.Item.Consumable.AmmoCategory.${slug}`,
+                behavior: "single",
+            },
```
stwlam:
```suggestion
            { behavior: "single", label: `PF2E.Item.Consumable.AmmoCategory.${slug}` },
```

src/module/item/weapon/data.ts
CarlosFdez: Sure. Ammo was my preference, I only called it ammunition cause it seemed more official.

src/module/item/consumable/document.ts
CarlosFdez: My concern with doing that is that doing an early return of that sort on a more generic function for a specific use case can cause a potential issue where we don't notice the early return when we add more to it. Its a mistake we've done before, although in the sheets.

Perhaps a private function for prepare ammo data?

src/module/actor/character/document.ts
```diff
@@ -1472,7 +1473,11 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                         ? configuredAmmo.uses.value
                         : configuredAmmo.quantity
                     : (configuredAmmo?.quantity ?? 0);
-                params.consumeAmmo ??= expend > 0;
+                if (!weapon.system.ammo?.builtIn) {
+                    params.consumeAmmo = false;
+                } else {
+                    params.consumeAmmo ??= expend > 0;
+                }
```
stwlam: Consume only if builtin?

src/module/item/consumable/document.ts
stwlam: Secondary method is fine

src/module/actor/character/document.ts
CarlosFdez: Likely a mistake I introduced when swapping built in to a variable, thanks for catching.

src/module/item/consumable/document.ts
CarlosFdez: Done, though untested. If I remember, using actual private with our proxy stuff didn't work.

src/module/item/weapon/sheet.ts
```diff
@@ -90,11 +90,23 @@ export class WeaponSheetPF2e extends PhysicalItemSheetPF2e<WeaponPF2e> {
...
+            })),
+            R.sortBy((o) => o.label),
+        );
 
         return {
             ...sheetData,
             abpEnabled,
             adjustedDiceHint,
+            isAmmoLocked: this.item.system.traits.value.includes("repeating"),
+            ammoTypes: [{ value: "", label: "Any Basic Ammo" }, ...ammoTypes],
```
stwlam: localize

src/module/actor/character/document.ts
```diff
@@ -1472,7 +1473,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                         ? configuredAmmo.uses.value
                         : configuredAmmo.quantity
                     : (configuredAmmo?.quantity ?? 0);
-                params.consumeAmmo ??= expend > 0;
+                params.consumeAmmo = !weapon.system.ammo?.builtIn ? (params.consumeAmmo ?? expend > 0) : false;
```
stwlam:
```suggestion
                params.consumeAmmo = weapon.system.ammo?.builtIn ? false : (params.consumeAmmo ?? expend > 0);
```

CarlosFdez:
You asked me to change weapon.system.ammunition to weapon.system.ammo. Do we do the same thing for consumables?

PR #19702, opened 2025-07-25
Support implanted usage as distinct from worn

CarlosFdez:
If there's a better icon let me know

image

### DIFF
--- a/src/module/item/base/data/values.ts
+++ b/src/module/item/base/data/values.ts
@@ -20,7 +20,7 @@ class MystifiedTraits {
     }
 }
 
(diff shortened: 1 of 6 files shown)


stwlam:
I don't think augmentations are physical items. Do you have information indicating otherwise?

CarlosFdez:
We talked separately but this pr isn't for augmentations, but for the implanted usage, which grafts (howl of the wild) currently make use of.

The main concern is that prothesis items are worn, regardless of whether you replaced a limb or not.

We may need to migrate to avoid unequipping grafts from existing items?

PR #19704, opened 2025-07-25
Add installed usage and sf2e weapon/armor upgrades

CarlosFdez:
Adds support for the installed usage and attaching them to weapons and armor. The attach/detach logic was pulled from the ammo dev work but works fine with existing attached items.

The one feature missing are the rule.parent.parentItem options, or what we decided to call "host options". We don't need those for starfinder player core, so I'd prefer to push them back to a later date to figure out the how.

Doing it this way (instead of rule element donation) allows choice sets to run when the attachable item is first created and avoids conflicts (useful for the energy shielding module, which locks a damage type when created), and allows item alterations to modify the attachable's description. A weapon can add extra damage by doing something like the following:

```json
{
  "key": "DamageDice",
  "damageType": "cold",
  "diceNumber": 1,
  "dieSize": "d6",
  "selector": "{item|parentItem.id}-damage"
}
```


src/module/actor/sheet/base.ts
```diff
@@ -707,7 +707,7 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
             },
             "detach-subitem": (event) => {
                 const subitem = inventoryItemFromDOM(event);
-                return detachSubitem(subitem, event.ctrlKey);
+                return subitem.detach({ skipConfirm: event.ctrlKey });
```
stwlam: Not a new bug, but this should be wise to MacOS

src/module/item/physical/document.ts
```diff
@@ -91,6 +93,13 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
     }
 
     get isEquipped(): boolean {
+        // Items with embed-type usages are equipped if they have a parent that is equipped
+        // @todo do the same for non-weapon attachments and affixed.
+        // subitem carryType is overriden to match the parent during prep, and weapons rely on that behavior.
+        if (this.system.usage.type === "installed") {
+            return !!this.parentItem?.isEquipped;
+        }
+
```
```suggestion
        if (this.system.usage.type === "installed") return !!this.parentItem?.isEquipped;
```

stwlam (approved):

PR #19708, opened 2025-07-25
Turn `TokenConfigMixinPF2e.PARTS` into a getter

### LINKED ISSUE #19707: `TokenConfigMixinPF2e.PARTS` override makes it impossible for modules to add parts to the TokenConfig in a system-agnostic way
There is no way for a module to add a tab to the TokenConfig on the PF2e system due to the `TokenConfigMixinPF2e.PARTS` being run at launch rather than being a getter, unless it specifically tagets the TokenConfigPF2e.

How to reproduce:
```js
Hooks.on("init", () => {
  // Add a tab
  TokenConfig.TABS.sheet.tabs.push({ id: "foo", label: "Foo" });
  // Remove the Footer, otherwise it gets rendered before the content
  const footerPart = TokenConfig.PARTS.footer;
  delete TokenConfig.PARTS.footer;
  // Add the tab
  TokenConfig.PARTS.foo = { template: "foo.html" };
  // Add the footer back in
  TokenConfig.PARTS.footer = footerPart;
});

Hooks.on("ready", () => {
  // These are equal
  console.log(TokenConfig.TABS.sheet.tabs);
  console.log(CONFIG.Token.sheetClasses.base["pf2e.TokenConfigPF2e"].cls.TABS.sheet.tabs);
  
  // These are not
  console.log(TokenConfig.PARTS);
  console.log(CONFIG.Token.sheetClasses.base["pf2e.TokenConfigPF2e"].cls..PARTS);
});
```

What a module needs to do to fix this:
```js
Hooks.on("ready", () => {
	if (game.system.id === "pf2e") {
		const cls = CONFIG.Token.sheetClasses.base["pf2e.TokenConfigPF2e"].cls;
		const footerPart = cls.PARTS.footer;
		delete cls.PARTS.footer;
		cls.PARTS.aboutFace = TokenConfig.PARTS.aboutFace;
		cls.PARTS.footer = footerPart;
	}
});
```

What the system needs to do to fix this:
```diff
- static override PARTS = (() => {
+ static override get PARTS() {
  const parts = super.PARTS;
  parts["appearance"].template = "systems/pf2e/templates/scene/token/appearance.hbs";
  return parts;
-})()
+}
```

mclemente:
Closes #19707.

### DIFF
--- a/src/module/scene/token-document/sheets/mixin.ts
+++ b/src/module/scene/token-document/sheets/mixin.ts
@@ -26,11 +26,11 @@ function TokenConfigMixinPF2e<TBase extends ReturnType<typeof TokenApplicationMi
             },
         };
 
-        static override PARTS = (() => {
+        static override get PARTS() {
             const parts = { ...super.PARTS };
             parts["appearance"].template = "systems/pf2e/templates/scene/token/appearance.hbs";
             return parts;
-        })();
+        }
 
         get linkToActorSize(): boolean {
             return !!this.token.flags.pf2e.linkToActorSize;


CarlosFdez:
Something here doesn't sit right. Nothing in the core code makes this a getter. Why is it an issue for PF2e but not core? If there's a reason for that, perhaps fixing it might be more ideal.

EDIT: ok, I'm getting more info on what's going on. Let me dig after this meeting.

CarlosFdez:
In the case of the prototype sheet, I think its automatic, changing CONFIG.Token.prototypeSheetClass should work. However it looks like TokenConfig is where it gets tricky. That's set in `foundry.applications.sheets.TokenConfig`, which I don't know if that's intended to be replaceable.

PR #19711, opened 2025-07-26
Check consumable trait instead when checking crafting quantity

CarlosFdez:
Existing checks were kept out of caution but I don't need much convincing to remove them.


stwlam:
Replace them all if you're feeling lucky

CarlosFdez:
I am!

PR #19714, opened 2025-07-26
Extract lore slug generation to LorePF2e class

Cuingamehtar:
Localizes Lore slug generation in LorePF2e class.

~Intent: that way Lore slug generation can be monkeypatched by localization modules to make them work with REs that target Lore slugs with English names as selectors/predicates (`CONFIG.PF2E.Item.documentClasses.lore.prototype.generateSlug` would be exposed).~
I'm still going to use it to override lore slugs in translation.

Just want to put it in advance of Battlecry release where Warfare Lore is particularly important.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -886,10 +886,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         });
 
         // Assemble lore items, key'd by a normalized slug
-        const loreItems = R.mapToObj(this.itemTypes.lore, (loreItem) => {
-            const rawLoreSlug = sluggify(loreItem.name);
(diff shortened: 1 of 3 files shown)


stwlam:
Monkey-patching will never be intentionally facilitated in the system, but it'd be okay to override the `slug` getter of lore items.

Cuingamehtar:
Changed the method into slug getter.
I didn't even consider that I could achieve what I wanted without libwrapper, just by subclassing the Lore on init.

stwlam:
This should start with the source slug if it's available, only using the name as a fallback.

Cuingamehtar:
Should the initial slug be "lorefied" or left as is?
Ex. `item.system.slug` is "warfare" (without "Lore" part). 
Should `item.slug` be "warfare-lore" or just "warfare"?

stwlam:
Start with the source slug and put it through the same motions as a sluggified name.

PR #19721, opened 2025-07-27
Add support for weapon, armor, and shield grade

CarlosFdez:
Shields have a grade but is not currently implemented. On Tikael's suggestion, we are errating kingmaker ones into starfinder ones.


CarlosFdez:
A confirmation dialog has been added when adding/removing the traits from weapons and armor. The dialog only shows when the weapon/armor has runes or is non-commercial. The dialog also adds a button to swap to the opposite tech trait, in case the intent was to change from an analog to tech item or vice versa.

src/module/item/physical/document.ts
```diff
@@ -739,6 +739,55 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
...
+                ("potency" in runes && runes.potency) ||
+                ("striking" in runes && runes.striking) ||
+                ("resilient" in runes && runes.resilient) ||
+                runes.property.length > 0;
+            const hasGrade = this.system.grade && this.system.grade !== "commercial";
+            if (wasSF2e !== becomingSF2e && (hasRunes || hasGrade)) {
+                const key = `PF2E.Item.Physical.ChangeEquipmentSystem.${becomingSF2e ? "ToStarfinder" : "ToPathfinder"}`;
+                const removedTrait = beforeSF2eTraits.find((t) => !newTraits.includes(t));
+                const otherTrait = removedTrait === "tech" ? "analog" : "tech";
+                const newTrait = newTraits.find((t) => t === "tech" || t === "analog");
```
```suggestion
                const newTrait = newTraits.find((t) => ["tech", "analog"].includes(t));
```
Or was there a reason for expanding this out?

CarlosFdez:
I added shield support and fixed a bug with the conversion dialog that I caught while working on it. Shield item name generation doesn't work yet, but that's because the sf2e module doesn't (or can't) add base shield types, which is a separate thing to resolve.

src/module/item/armor/document.ts
```diff
@@ -92,9 +92,18 @@ class ArmorPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Phy
         const maxPropertySlots = getPropertyRuneSlots(this);
         this.system.runes.property.length = Math.min(this.system.runes.property.length, maxPropertySlots);
 
-        // Add traits from fundamental runes
         const abpEnabled = ABP.isEnabled(this.actor);
         const baseTraits = this.system.traits.value;
+
+        // Determine if this armor uses runes or CTASEUP and clear the opposing data
+        const isSF2e = baseTraits.some((v) => v === "tech" || v === "analog");
```
stwlam:
```suggestion
        const isSF2e = baseTraits.some((v) => ["tech", "analog"].includes(v));
```
It really is magical

src/module/item/armor/sheet.ts
```diff
@@ -2,9 +2,10 @@ import { AutomaticBonusProgression as ABP } from "@actor/character/automatic-bon
 import { ItemSheetOptions } from "@item/base/sheet/sheet.ts";
 import {
     CoinsPF2e,
+    type Grade,
     MATERIAL_DATA,
-    MaterialSheetData,
-    PhysicalItemSheetData,
+    type MaterialSheetData,
+    type PhysicalItemSheetData,
```
stwlam: This isn't really necessary if it's a mix: esbuild just needs to know if it can drop the imported file entirely.

src/module/item/physical/document.ts
CarlosFdez: This one is outdated, but the reason was stricter typing (and marginally better perf doesn't hurt). I can swap it over since that doesn't matter too much.

PR #19728, opened 2025-07-28
Update effect links in the static language file

Cuingamehtar:
Remove the outdated effect links from static language file. Since those only exist as part of a generated message, I think it should be safe to remove the redirect as well.

Not sure if I should actually keep or update the labels.


pedrogrullada:
One thing to keep in mind with undoing redirects is that it's possible some text containing the effects wasn't always localized. From a quick git blame it looks like it's just Glue Bombs (previously known as Tanglefoot Bags) that had previously unlocalized text, which was localized 3 years ago, for example.

I don't think a copy *that* old of Tanglefoot Bags in someone's world is worth worrying about, but good to know.

PR #19734, opened 2025-07-29
Fix type error in rule element creation

CarlosFdez:
parent became optional during the v13 beta cycle in the base interface. The change in effect was to make it match the super class's method signature.


stwlam:
Why'd we make it optional?

CarlosFdez:
I don't think it was intentional, it looks like the end result of a lot of type upheaval in the core system:
https://github.com/foundryvtt/pf2e/commit/d1f629871651f3d2db7fcada3452e0701ec82607

It used to extend ParentedDataModelConstructionOptions, which no longer exists.

PR #19758, opened 2025-08-01
Add Roll Option to parry trait items, add effect for Extravagant Parry

ammalagonc:
Redo of #19684

### DIFF
--- a/packs/equipment/acrobats-staff.json
+++ b/packs/equipment/acrobats-staff.json
@@ -52,7 +52,12 @@
         "reload": {
             "value": null
         },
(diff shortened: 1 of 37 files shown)


pedrogrullada:
This will not work with anything that adds the parry trait via Item Alteration, which is a non-trivial number of things. Best to hold off on trying to hack this until we can do better. The Aldori Dueling Sword could be handled this way since it only affected that one item.

ammalagonc:
Good point - let me kill this one and restore the OG. That one prompts the user to choose if they have the trait or not.

PR #19769, opened 2025-08-01
Add traits.value list to commander and guardian classes.

abrhoda:
Hey all, I maintain a local sql database with pf2e content for myself and my local group. Saw that commander and guardian classes got added so ran some scripts to rebuild the dataset and noticed it was breaking on the new classes. Appears that other classes have a `value` list that is empty under their `traits` tag but this is missing from the commander and guardian classes. To avoid data drift and arbitrary uniqueness among that classes, this is simply adding that empty list to those 2 new classes. 

Realistically, this list is always empty for all classes. I'm not familiar with the entire codebase itself to know about removing this `traits.value` list from all the classes. However, the commander and guardian classes function fine without it. If that's the case and it is always an empty list for all the classes, removing the key/value might be a good call.

### DIFF
--- a/packs/classes/commander.json
+++ b/packs/classes/commander.json
@@ -219,7 +219,8 @@
             ]
         },
         "traits": {
-            "rarity": "common"
+            "rarity": "common",
+            "value": []
         }
     },
     "type": "class"
--- a/packs/classes/guardian.json
+++ b/packs/classes/guardian.json
@@ -219,7 +219,8 @@
             ]
         },
         "traits": {
-            "rarity": "common"
+            "rarity": "common",
+            "value": []
         }
     },
     "type": "class"


CarlosFdez:
This is likely a result of the extract script or the run migrations. I don't know exactly why it occured this time, but we shouldn't be manually altering these fields to satisfy external providers. In fact, classes aren't supposed to have traits, so its likely the other ones are in error. I can at least give you a heads up on how to satisfy your needs though.

Data that is in these files undergo pruning, but this is usually on manually provided properties. All information here is merged on top of a set of default properties. These default properties were traditionally handled by `template.json`, but they have been slowly shifting towards using foundry's data models instead. I would recommend creating your own local version of base files that you merge pf2e files on top of based on the initial properties of a created document. That should satisfy your use case.

stwlam:
The now-deleted template.json entry had a `traits.value` in it.

abrhoda:
Ah yes, that makes sense about the `template.json`! Thanks all.

PR #19809, opened 2025-08-04
Condense various Parry feat effects

ammalagonc:
Feats tackled in this PR:
- Defensive Armaments (thanks @Mikkidez)
- Dueling Parry
- Extravagant Parry
- Twin Parry

I did not delete and redirect the effects, as those might need a migration. If that is not the case, let me know.

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -5254,6 +5254,9 @@
             "PackAttack": {
                 "Label": "Target is Within Range of At Least Two Allies"
             },
+            "Parry": {
+                "Prompt": "Weapon has the parry trait?"
+            },
             "PathfinderAgentDedication": {
                 "ThoroughReports": "Thorough Reports"
             },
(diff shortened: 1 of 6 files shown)


Mikkidez:
I apologize for bothering you, but would it be possible to also look at the parry effect of this gunslinger feat? Thank you very much in advance!

https://2e.aonprd.com/Feats.aspx?ID=3160

pedrogrullada:
This approach would exclude integration with 3PP. It'd be good to discuss with a lead about what to do, but I think a toggle in those feats to increase or alter the parry bonus would be preferable to baking it into the parry effect.

static/lang/re-en.json
```diff
@@ -5254,6 +5254,9 @@
             "PackAttack": {
                 "Label": "Target is Within Range of At Least Two Allies"
             },
+            "Parry": {
+                "Prompt": "Weapon has the parry trait?"
```
```suggestion
                "Prompt": "Are you wielding a weapon with the parry trait?"
```

PR #19816, opened 2025-08-05
Fix special resource sheet prep type error

CarlosFdez:
value allowed undefined before in order to allow proper initialization later on, but we can do that by checking the source value

### DIFF
--- a/src/module/rules/rule-element/special-resource.ts
+++ b/src/module/rules/rule-element/special-resource.ts
@@ -36,12 +36,15 @@ class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema> 
         if (this.level !== null && !this.itemUUID) {
             this.failValidation("level can only be set if itemUUID is set");
         }
+
+        // If unset, set to 0 for now. It'll gain an actual value during beforeCreateData()
+        this.value ??= 0;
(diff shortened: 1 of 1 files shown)


src/module/rules/rule-element/special-resource.ts
```diff
@@ -36,12 +36,15 @@ class SpecialResourceRuleElement extends RuleElementPF2e<SpecialResourceSchema>
...
+
+        // If unset, set to 0 for now. It'll gain an actual value during beforeCreateData()
+        this.value ??= 0;
     }
 
     static override defineSchema(): SpecialResourceSchema {
         return {
             ...super.defineSchema(),
-            value: new fields.NumberField({ required: false, nullable: false, initial: undefined }),
+            value: new fields.NumberField({ required: true, nullable: true, initial: null }),
```
stwlam, line 47: nullable?

CarlosFdez: Yes. We need to distinguish between depleted and first time initialization during data preparation.

CarlosFdez: I can see why it'd look weird though, it was using undefined to do that before.

Issue #19722, opened 2025-07-27
Token Image rule element issue


bwiz4k:
Hello I've found a bug with the Token Image rule element.

When the effect is applied it only shows if the token is owned by the player and by the GM.

All other players who do not own the token being changed cannot see the token image change, unless they refresh or the token is redropped.

Tested on PF2e system version 7.2.3 with no modules.

Thank you!

websterguy:
Seems that Token Light also does not update immediately.

PR #19830, opened 2025-08-06
Fix tokens not updating for non-owners

websterguy:
I don't know the complete reasoning behind change from #19500, but this `isOwner` check prevents updates from TokenImage and TokenLight showing to any player who doesn't own the token until they refresh their browser or the scene is reloaded.


websterguy:
Fixes #19722

stwlam:
Superseded by #19837

PR #19841, opened 2025-08-07
Automate Advanced Shooter proficiencies in most common cases

pedrogrullada:
This isn't perfect, as this automation will only work with the proficiencies granted by the gunslinger class and archetype. For cases other than those, such as a different source of firearms/crossbow proficiency, the feat will remain without automation.

### DIFF
--- a/packs/pf2e/feats/class/gunslinger/level-6/advanced-shooter.json
+++ b/packs/pf2e/feats/class/gunslinger/level-6/advanced-shooter.json
@@ -25,7 +25,49 @@
             "remaster": true,
             "title": "Pathfinder Guns & Gears"
         },
-        "rules": [],
+        "rules": [
+            {
+                "adjustName": false,
+                "choices": [
+                    {
+                        "label": "PF2E.WeaponGroupCrossbow",
+                        "value": "crossbow"
+                    },
+                    {
+                        "label": "PF2E.WeaponGroupFirearm",
+                        "value": "firearm"
+                    }
+                ],
+                "flag": "group",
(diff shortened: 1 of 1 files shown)


7H3LaughingMan:
Quick question, what is the choice set for? My understanding of this feat is that you treat both advanced firearms and advanced crossbows as if they were martial. With the choice set it is only going to apply one of them.

stwlam:
or

pedrogrullada:
> Quick question, what is the choice set for? My understanding of this feat is that you treat both advanced firearms and advanced crossbows as if they were martial. With the choice set it is only going to apply one of them.

As shark mentioned, it's a choice between firearms and crossbows

7H3LaughingMan:
> > Quick question, what is the choice set for? My understanding of this feat is that you treat both advanced firearms and advanced crossbows as if they were martial. With the choice set it is only going to apply one of them.
> 
> As shark mentioned, it's a choice between firearms and crossbows

I don't think shark was stating that it's a choice between firearms and crossbows, I think he was stating that you should make it an or statement so it applies to both. And I was asking why you made it a choice when the feat doesn't give you a choice, you get them both.

Acrobat_6FR3YYOUab

pedrogrullada:
Apparently this changed in the remaster of G&G and Foundry still had the old text, which is what I was going off of.
image
Easy fix at least.

7H3LaughingMan:
Okay, that explains it then. I thought I was going crazy for a second but turns out I didn't realize that the description in Foundry wasn't correct and I was just reading what was in the books.

PR #19858, opened 2025-08-09
Implement tracking trait

CarlosFdez:
Missed a spot.

As a bonus, it is now possible to use item alterations to remove the magical trait for weapons that have potency runes.


src/module/actor/character/document.ts
```diff
@@ -1169,44 +1170,66 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
...
+            // Handle weapon attack from the tracking trait (which isn't specific to tech/analog)
+            // If found, filter out inferior tracking traits from the weapon in case there are duplicates
+            const trackingTraits = ["tracking-1", "tracking-2", "tracking-3"] as const;
+            const tracking = trackingTraits.findLastIndex((t) => weapon.system.traits.value.includes(t)) + 1;
+            weapon.flags.pf2e.tracking = tracking;
+            if (tracking > 0) {
+                weapon.system.traits.value = weapon.system.traits.value.filter(
+                    (t) => t === `tracking-${tracking}` || !t.startsWith("tracking-"),
+                );
+            }
```
stwlam: Not crazy about using a flag for a core trait, and storing the index doesn't seem very useful, either.

CarlosFdez: The flag is for rule elements to use. If we're willing to say item bonus = tracking, then we can drop it! I'll remove it since I wasn't a fan of that. If a soldier wants to drink a pf2e mutagen, yeah sure go for it.

EDIT: though it could be a getter or a data prep only property as well.

CarlosFdez: As for the index, the index + 1 is equal to the tracking value. -1 + 1 = 0. 0 + 1 = 1, etc.

CarlosFdez: For PR posterity we plan on putting it in `weapon.system.traits.tracking` or `weapon.system.traits.data.tracking` but I think that's better suited for a followup PR so I removed it from here.

PR #19865, opened 2025-08-09
Remove Hunt Prey toggle and update automation on several Ranger features

pedrogrullada:
- Remove Hunted Prey toggle since it was interfering with the Token Mark roll option, overriding it to be false even when targeting the prey. Most automation works off targeting the hunted prey. The exceptions are skill checks lacking target data, such as to Seek, Track, and Recall Knowledge. The hope is that we can delegate these to be toggled via the roll dialog toggle instead.
- Clean up implementation of hunter's edges:
  - Use mergeable suboptions in Precision to switch between first, second, and third hit against the hunted prey.
  - Use AdjustModifier rule elements in Outwit instead of duplicating Flat Modifiers, and fix an issue with its AC bonus predicating off `target:` rather than `origin:`.
  - Predicate effect's bonus and spell alteration on Vindication more precisely, and account for sharing its benefits via Warden's Boon et. al.
- Update rule elements on Masterful Hunter features. Ideally these rule elements should be in their respective hunter's edge features, but removing these would need a migration most likely.
- Update effects to share Hunter's Edge benefits, and link them in relevant feats (see `hunters-edge-sharing` tag) via an Item Alteration on their respective hunter's edge features, making the implementation expandable for other edges, such as those by 3PP.
  - Rename effects associated with these feats to `Effect: Share Hunter's Edge (<Name>)` and update their automation to be in line with the aforementioned changes to hunter's edge benefits.
- Use Token Mark for Monster Hunter and Monster Warden. In the future, they can hopefully use the existing Hunted Prey Token Mark, but for now, this should suffice. Use tags to implement Legendary Monster Hunter.
- Further automate Hunter's Aim and remove unlocalized text.
- Add Flat Modifier and nature-based `[[/act]]` links in Vindicator's Dedication
- A few minor formatting and miscellaneous fixes.

GitHub shows some deleted effects, but they're just renames.


TikaelSol:
How does this affect existing rangers? Will we need to migrate?

pedrogrullada:
> How does this affect existing rangers? Will we need to migrate?

~~Existing rangers would have to do the dance to work nicely with the masterful hunter and share hunt prey reworks, or we'd need to migrate yeah. There wouldn't be any errors, but some automation involving those features wouldn't work.~~

~~I know we're not at a great spot to do migrations, so let me know if you want me to just carve out the Hunt Prey toggle thing and leave the rest of this stuff for later.~~

After further testing and some changes, there will actually be no need for a migration.

PR #19866, opened 2025-08-09
Extend modular logic to allow for more configuration setups

Cerapter:
This PR extends the usage of modular beyond just changing the damage type to B, P, or S.
- The existing B-P-S switcher modular trait is kept as-is so that no migration is necessary.
- Several parts of the code where it is assumed that the modular selection can only be a damage type is changed to be just a string (which should be the key of the modular config, see below).
- A new PF2E config is introduced, called `modularConfigs`, that matches traits to sets of modular configs, which they themselves match a string key to an optional damage type (optional because SF2's shock truncheon), an array of traits (that can be empty), and a label that is effectively the localisation of the key of the config.
- Added the modular trait to Switchscythe, with the Switchscythe-specifics.

Though not present here, I've tested this with a few setups, like having a thrown trait in the modular trait, or with combination weapons.
Here's a visual demonstration:

https://github.com/user-attachments/assets/be670cd4-111e-48bc-8f78-b273c5749224

There are a few spots where I'm not sure if I approached this correctly -- `WeaponTraitToggles`'s `applyChanges()` and `WeaponPF2e`'s `toThrownUsage()`, I'm not *too* sure about, the latter I think might benefit more from a reorganisation from when data is handled? But I don't know enough about the system to feel out how it would best be reorganised, so consider this a "half-draft" PR.

### DIFF
--- a/packs/equipment/switchscythe.json
+++ b/packs/equipment/switchscythe.json
@@ -66,7 +66,8 @@
             "rarity": "uncommon",
             "value": [
                 "fatal-d10",
-                "gnome"
+                "gnome",
+                "modular-p-grapple-s-sweep"
             ]
         },
         "usage": {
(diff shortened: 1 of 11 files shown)


Cerapter:
Closed, see #20576 instead.

PR #19889, opened 2025-08-12
Create trait value data

CarlosFdez:
If accepted, the tracking pr will change to using `weapon.system.traits.data.tracking`

### DIFF
--- a/src/module/item/base/data/system.ts
+++ b/src/module/item/base/data/system.ts
@@ -20,26 +20,35 @@ interface ActionCost {
     value: OneToThree | null;
 }
 
+interface ValuedTraits {
+    volley?: number;
+    [key: string]: number | undefined;
+}
+
 interface ItemTraits<T extends ItemTrait = ItemTrait> {
     value: T[];
     rarity: Rarity;
     otherTags: string[];
+    config?: ValuedTraits;
 }
 
 interface ItemTraitsNoRarity<T extends ItemTrait = ItemTrait> extends Omit<ItemTraits<T>, "rarity"> {
     rarity?: never;
+    config?: ValuedTraits;
 }
 
 interface RarityTraitAndOtherTags {
-    readonly value?: never;
+    value?: never;
     rarity: Rarity;
     otherTags: string[];
+    config?: never;
 }
 
 interface OtherTagsOnly {
-    readonly value?: never;
+    value?: never;
     rarity?: never;
     otherTags: string[];
+    config?: never;
 }
 
 type ItemFlagsPF2e = DocumentFlags & {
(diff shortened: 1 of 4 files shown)


src/module/item/base/data/system.ts
```diff
@@ -20,26 +20,35 @@ interface ActionCost {
...
+interface ValuedTraits {
+    volley?: number;
+    [key: string]: number | undefined;
+}
+
 interface ItemTraits<T extends ItemTrait = ItemTrait> {
     value: T[];
     rarity: Rarity;
     otherTags: string[];
+    data?: ValuedTraits;
```
stwlam: Can we call it `config` :o)

src/module/item/helpers.ts
```diff
@@ -74,4 +82,91 @@ function markdownToHTML(markdown: string): string {
...
+ * Wraps over item data to extend for the purposes of system data.
+ * @todo once all items use data models, replace with data model implementation
+ */
+function wrapItemTraits<T extends ItemTraits | ItemTraitsNoRarity | RarityTraitAndOtherTags | OtherTagsOnly>(
+    item: ItemPF2e,
+    traits: T,
+): T {
+    if (traits.value) {
+        traits.value = traits.value.filter((t) => t in item.constructor.validTraits);
+    }
```
stwlam:
```suggestion
    traits.value &&= traits.value.filter((t) => t in item.constructor.validTraits);
```

src/module/item/helpers.ts
```diff
@@ -74,4 +82,91 @@ function markdownToHTML(markdown: string): string {
         .innerHTML.trim();
 }
 
-export { itemIsOfType, markdownToHTML, performLatePreparation, reduceItemName };
+const valueRegex = /^([\w-]+)-(\d{1,3})$/;
+
+function evaluateTraitData(traits?: string[]): ValuedTraits {
```
stwlam:
```suggestion
function evaluateTraitConfiguration(traits?: string[]): ValuedTraits {
```
Or `evaluateTraitConfig`?

src/module/item/helpers.ts
CarlosFdez: image

Some of the possible item traits don't have a value.

stwlam: That's not what the error says

CarlosFdez: Its not, but that's why it happens.

but I think I can convert `readonly value?: never;` to `value?: never` and it'll be fine.

stwlam: Those should really be equivalent, but that's on typescript. And maybe we should iterate through them and splice out bad ones? It's not often that any are going to need to be removed.

PR #19907, opened 2025-08-14
Add click handlers to `ChatPopout` on render

### LINKED ISSUE #19793: Damage buttons don't work in popped out chat messages
Image

To reproduce:
1. make an enemy roll some damage, causing a damage roll message appear in the chat log
2. select some token
3. click the damage button on that damage roll message, to see that it works (damage gets applied, a new "X takes Y damage" message appears)
4. right-click the damage roll message and click Pop Out Message
5. in the popped out message, click the damage button, and see that **nothing happens**

(tested in pf2e v7.3.1 with no modules enabled)

stwlam:
- Closes #19793

### DIFF
--- a/src/module/apps/sidebar/chat-log.ts
+++ b/src/module/apps/sidebar/chat-log.ts
@@ -1,6 +1,7 @@
 import { ActorPF2e } from "@actor";
 import { Kingdom } from "@actor/party/kingdom/model.ts";
-import type { ApplicationRenderContext } from "@client/applications/_types.d.mts";
+import type { ApplicationRenderContext, ApplicationRenderOptions } from "@client/applications/_types.d.mts";
+import type ChatPopout from "@client/applications/sidebar/apps/chat-popout.d.mts";
 import type { ContextMenuCondition, ContextMenuEntry } from "@client/applications/ux/context-menu.d.mts";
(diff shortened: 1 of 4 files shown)


src/module/apps/sidebar/chat-log.ts
```diff
@@ -105,7 +106,25 @@ class ChatLogPF2e extends fa.sidebar.tabs.ChatLog {
...
+        if (!options.isFirstRender) return;
+        const actions = popout.options.actions;
+        Object.assign(actions, {
+            applyDamage: ChatLogPF2e.#onClickApplyDamage,
+            applyEffect: ChatLogPF2e.#onClickApplyEffect,
+            findToken: ChatLogPF2e.#onClickFindToken,
+            revertDamage: ChatLogPF2e.#onClickRevertDamage,
+            recoverPersistentDamage: ChatLogPF2e.#onClickRecoverPersistent,
+            setAsInitiative: ChatLogPF2e.#onClickSetAsInitiative,
+            shieldBlock: ChatLogPF2e.#onClickShieldBlock,
```
CarlosFdez: Maybe we can assign this listener map to another static property?

stwlam: how about dat

CarlosFdez (approved):

PR #19918, opened 2025-08-15
Add effect to `Trudd's Strength` filigree

### LINKED ISSUE #19898: No effect for Trudd's Strength
There is no effect for Trudd's Strength or Trudd's Strength (Greater)

ammalagonc:
Closes #19898

### DIFF
--- a/packs/equipment-effects/effect-trudds-strength.json
+++ b/packs/equipment-effects/effect-trudds-strength.json
@@ -1,0 +1,89 @@
+{
+    "_id": "cB5nrsadZxDaxFjz",
+    "img": "icons/commodities/treasure/brooch-shield-grey.webp",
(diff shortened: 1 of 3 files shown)


pedrogrullada:
I don't think this is the way we handle runes, which is essentially what these are.

ammalagonc:
I was going off the Fanged Rune, which has an effect on it.

pedrogrullada:
I'm honestly not a fan of that either since it's not like runes are expected to be held in the inventory for any effects on them to be accessible. But putting that to one side, the rule elements on the rune itself wouldn't be of much use for that same reason.

ammalagonc:
I can remove the RE's - With the upgrades introduced in SF2e, I wonder if this is the way forward with runes. But that's another chat for another day.

PR #19932, opened 2025-08-16
Cleanup Book 3 FOTRP NPCs

DocSchlock:
This is a creature-by-creature audit of the Book 3 AP NPCs for Fists of the Ruby Phoenix.

Upgrades some automations but mostly focused on fixing errors and adding missing data.

I also refreshed inventory items where applicable where some of the older versions were broken or lacked data. Didn't touch any spells because I know those need to stay as they are.

Includes a couple updates to Book 2 Hazards as well.

A note: Shino has a "potion of haste" in the book, which was never an item to begin with. I added a potion of quickness to her inventory based on the intent there.


TikaelSol:
Not sure about refreshing the thunderstones for the same reason as refreshing spells.

DocSchlock:
> Not sure about refreshing the thunderstones for the same reason as refreshing spells.

Reverted

PR #19933, opened 2025-08-16
Fix regression in automation addition of invested trait to magic armor

stwlam:
Dispensing with `traits.config` and the proxy for now

### DIFF
--- a/src/module/item/base/document.ts
+++ b/src/module/item/base/document.ts
@@ -15,7 +15,7 @@ import type {
 import type { ImageFilePath, RollMode } from "@common/constants.d.mts";
 import type { ContainerPF2e, PhysicalItemPF2e } from "@item";
 import { createConsumableFromSpell } from "@item/consumable/spell-consumables.ts";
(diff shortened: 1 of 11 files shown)


src/module/item/base/document.ts
```diff
@@ -279,9 +280,9 @@ class ItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         this.system.description.initialized = false;
 
         const traits = this.system.traits;
-        traits.value?.forEach((trait, index) => {
-            if (!(trait in this.constructor.validTraits)) traits.value.splice(index, 1);
-        });
+        for (const trait of [...(traits.value ?? [])]) {
+            if (!(trait in this.constructor.validTraits)) traits.value?.splice(traits.value.indexOf(trait), 1);
+        }
```
CarlosFdez: Because we have to create a new array here *anyways* for this to work, you might as well do `this.system.traits.value &&= this.system.traits.value.filter((t) => t in this.constructor.validTraits);`

I think if we want to rely on this being not needed often as an optimization, an index based for loop that iterates in reverse would allow you to use splice without making a new array.

PR #19938, opened 2025-08-17
Add `highest` property to actor resistance and weakness arrays

stwlam:
Iffy about the implementation; might be better as just an `Array` subclass

### DIFF
--- a/src/module/actor/hazard/data.ts
+++ b/src/module/actor/hazard/data.ts
@@ -283,8 +283,8 @@ interface HazardAttributes extends ActorAttributes, Omit<HazardAttributesSource,
     hasHealth: boolean;
     stealth: HazardStealthTraceData;
     immunities: Immunity[];
-    weaknesses: Weakness[];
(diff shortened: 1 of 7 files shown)


src/module/actor/hazard/data.ts
```diff
@@ -283,8 +283,8 @@ interface HazardAttributes extends ActorAttributes, Omit<HazardAttributesSource,
     hasHealth: boolean;
     stealth: HazardStealthTraceData;
     immunities: Immunity[];
-    weaknesses: Weakness[];
-    resistances: Resistance[];
+    weaknesses: Weakness[] & { get highest(): Weakness & { type: null; value: 0 } };
+    resistances: Resistance[] & { get highest(): Resistance & { type: null; value: 0 } };
```
CarlosFdez: Was this meant to be | instead of &?

stwlam: nuh uh

PR #19940, opened 2025-08-17
Fix attaching to nested items and detaching double nested subitems

CarlosFdez:


src/module/apps/item-attacher.ts
```diff
@@ -21,10 +22,10 @@ class ItemAttacher<TItem extends PhysicalItemPF2e> extends PickAThingPrompt<TIte
         if (!item.isAttachable) {
             throw ErrorPF2e("Not an attachable item");
         }
-        const collection =
-            item.actor?.inventory.contents ??
-            game.items.filter((i): i is PhysicalItemPF2e<null> => i.isOfType("physical"));
+        const actor = item.actor;
+        const collection: ItemPF2e[] = actor ? [...iterateAllItems(actor)] : game.items.contents;
```
stwlam, line 26: We have too many loose helpers :slightly_frowning_face:

PR #19943, opened 2025-08-17
Fix regression with editing non-physical items

CarlosFdez:
This also adds the stuff I was missing on sub items like chat messages. This should work a bit better than my last attempt.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -595,11 +595,33 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
 
     /** Sheet-wide click listeners for elements selectable as `a[data-action]` */
     protected activateClickListener(html: HTMLElement): SheetClickActionHandlers {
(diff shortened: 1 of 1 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -595,11 +595,33 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
...
+            const itemUuid = htmlClosest(event.target, "[data-uuid]")?.dataset.uuid;
+            const itemEl = htmlClosest(event.target, "[data-item-id]");
+            const itemId = itemEl?.dataset.itemId;
+            const collectionId = itemEl?.dataset.entryId;
+            const collection = collectionId
+                ? (actor.spellcasting?.collections.get(collectionId, { strict: true }) ?? null)
+                : null;
+            if (collection) {
+                return collection.get(itemId, { strict: true });
+            }
```
stwlam: $5 for the whitespace

stwlam:
```suggestion
            if (collection) return collection.get(itemId, { strict: true });
```

PR #19948, opened 2025-08-17
Fix first-parameter type of `Document#_preCreate`

stwlam:


CarlosFdez:
What's the bugfix?

stwlam:
A couple `_preCreates` are theoretically explody because they assume complete source data.

PR #19968, opened 2025-08-19
Don't auto adjust price using grade for specific magic weapons

CarlosFdez:
Eventually we'll figure out what to do with specific magic X items exactly once there's more content, but it should behave more consistently for now.


stwlam:
There's a good chance that specific magic weapons and armor aren't going to be a thing.

PR #19974, opened 2025-08-19
Apply reach checks to player looting and door operation

stwlam:

### DIFF
--- a/src/module/canvas/token/object.ts
+++ b/src/module/canvas/token/object.ts
@@ -615,6 +615,34 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
         return super._onHoverOut(event);
     }
 
+    /** Require that a loot actor or dead creature is in reach for a player to view its sheet. */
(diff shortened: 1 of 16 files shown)


src/module/canvas/token/object.ts
```diff
@@ -615,6 +615,34 @@ class TokenPF2e<TDocument extends TokenDocumentPF2e = TokenDocumentPF2e> extends
...
+        );
+        if (!requiresReach || this.document.isOwner || !actor?.isLootableBy(game.user) || !canvas.grid.isSquare) {
+            return super._onClickLeft2(event);
+        }
+        const isInReach = R.unique(
+            [canvas.tokens.controlled, game.user.character?.getActiveTokens(true, false) ?? []].flat(),
+        ).some(
+            (t) =>
+                t.actor?.isOwner &&
+                t.actor.isOfType("creature") &&
```
shemetz, line 632: this prevents Party tokens from counting, which seems bad for a group that uses a single shared Party token to move around a dungeon and open doors etc until they find an encounter.

stwlam: Write an issue.

PR #20025, opened 2025-08-24
Cleanup NPC `system.skills` data entry/migration mistakes

Daomephsta:
Mainly to prevent potential future issues during migrations, but some of these are causing issues now (e.g Asper Hajeri 9-10 likely has an incorrect Occultism modifier).
I've left notes on a few of these changes, as someone needs to double check the sourcebook for them.

### DIFF
--- a/packs/curtain-call-bestiary/book-3-bring-the-house-down/prince-of-propaganda.json
+++ b/packs/curtain-call-bestiary/book-3-bring-the-house-down/prince-of-propaganda.json
@@ -1339,7 +1339,7 @@
             "diplomacy": {
                 "base": 38
             },
-            "intimidate": {
+            "intimidation": {
                 "base": 40
             },
             "religion": {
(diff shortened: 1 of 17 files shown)


packs/curtain-call-bestiary/book-3-bring-the-house-down/prince-of-propaganda.json
```diff
@@ -1339,7 +1339,7 @@
             "diplomacy": {
                 "base": 38
             },
-            "intimidate": {
+            "intimidation": {
```
Daomephsta, line 1342: 99% sure this change is correct, but would like someone to check the creature is supposed to have Intimidation, rather than a bonus to Intimidate or something else

packs/pfs-season-5-bestiary/5-20/asper-hajeri-9-10.json
```diff
@@ -2003,10 +2003,7 @@
                 "base": 24
             },
             "occultism": {
-                "base": 20
-            },
-            "occultism+23": {
-                "base": null
+                "base": 23
```
Daomephsta, line 2006: This needs to be checked against the source. 
I've guessed that the occultism bonus is supposed to be 23, rather than the 20 it was, since every other skill modifier is higher than the base version of Asper Hajeri.

PR #20029, opened 2025-08-24
Correct Animate Dream's spell list and resistances (Gatewalkers version)

ammalagonc:
Fear should be heightened to 3rd rank and sleep to 4th rank per the PDF.


simonward:
Consider updating its resistance exception to spirit rather than void.
Screen Shot 2025-08-25 at 2 25 37 PM

PR #20037, opened 2025-08-25
Implement action glyph display for token ruler segments

In3luki:
~~Only supports the `walk` action for now.~~
Now only glpyhs. No more colors.

Land speed 30ft:
Screenshot 2025-08-31 152816

Difficult terrain:
Screenshot 2025-08-31 154211


stwlam:
We should slap down those labels where the cost increases, using waypoints only if they are in the same square.

In3luki:
Adding custom labels is more difficult than I anticipated.
`TokenRuler#refresh` is called on every tick where it rerenders all waypoints and replaces the children of the `#hud #measurement #${rulerId}` div element with the newly rendered labels.

The problem is that `TokenRuler#refresh` is synchronous but it calls the private async `TokenRuler##renderWaypointLabels` which is not awaited. So adding custom ~tokens~ labels after `super.refresh()` doesn't work because they are immediately replaced once `TokenRuler##renderWaypointLabels` is finished.

`TokenRuler#refresh` also computes the final path and cost for all segments using two other private methods `TokenRuler##getNonintermediatePath` and `TokenRuler##getSnappedIntermediatePath` which in turn call another private method `TokenRuler##measurePathAndLinkWaypoints`.
So it looks like getting the final ruler path is impossible without reimplementing those three methods.

I have experimented with adding a new container div element and handling refresh and destruction of that, which works somewhat. The labels are added to the new div element in `TokenRulerPF2e#_getWaypointLabelContext` and show up on the canvas but I haven't found a way to limit that to just the first time the action cost increases. The lable either floats along with the token preview or doesn't show up at all. I believe there may be a race condition or some double render going on where the second `refresh` call resets my `lastActionLabelPlaced` property before there was time to render it for the first time.

I have also experimented with adding the action cost as a `PIXI.Texture` with a transformation matrix to the gird highlight which also works, but also is either rendering the texture for every waypoint after a cost increase or not at all.

In3luki:
Rebase and cleaned up. I still haven't found a way to add custom labels to specific points along the ruler's path.

In3luki:
Superseded by #20092

PR #20041, opened 2025-08-25
Update `EnvironmentFeatureBehaviorType` to work with the new token ruler

In3luki:
This enables difficult terrain calculations while keeping the core `Modify Movement Cost` functional (if someone wants to use that for whatever reason).
Only for the `walk` action for now. It can't handle elevation yet.

r


In3luki:
Superseded by #20044

Issue #20034, opened 2025-08-24
BattleForm cannot resolve properties


MrVauxs:
Attempting to use either `@actor` or `{actor|...}` notation in battleforms ends with them being unable to resolve them or the Rule Element outright failing validation due to forcing the values to be integers.

Example Rule Element:
```json
{
  "key": "BattleForm",
  "overrides": {
    "armorClass": {
      "modifier": "@item.actor.system.attributes.spellDC.value"
    },
    "strikes": {
      "primary": {
        "label": "Primary",
        "ability": "str",
        "category": "unarmed",
        "damage": {
          "damageType": "bludgeoning",
          "dice": 1,
          "die": "d8",
          "modifier": "@item.level * 2"
        },
        "group": "brawling",
        "modifier": "@item.actor.system.attributes.spellDC.value",
        "traits": [
          "unarmed",
          "versatile-p",
          "versatile-s"
        ]
      }
    },
    "tempHP": "@item.level * 2"
  }
}
```

PR #20048, opened 2025-08-26
Fix filtering `SpellcastingEntry`s in `Actor#prepareData`

In3luki:
`proficency.slug` is an empty string in every spellcasting entry I've checked.

Addresses resolving `@actor.system.attributes.spellDC.value` in #20034

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -31,7 +31,6 @@ import { createDisintegrateEffect } from "@item/effect/helpers.ts";
 import { itemIsOfType } from "@item/helpers.ts";
 import { CoinsPF2e } from "@item/physical/coins.ts";
 import { getDefaultEquipStatus } from "@item/physical/helpers.ts";
-import { MAGIC_TRADITIONS } from "@item/spell/values.ts";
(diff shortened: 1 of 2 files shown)


CarlosFdez:
I suggest holding. This split was to fix deviant ability special statistics, and if they're working right now, we don't need the split any longer and can always have spellcasting entries run after. Maybe some remaster followup work fixed it at the root.

CarlosFdez:
It is confirmed that this does fix deviant abilities as well. However I think you should instead change the check to check for a falsy proficiency.slug. The reason it checked for tradition's is because that's how spellcasting entries worked before remaster. If you look at `SpellcastingEntryPF2e#prepareStatistic()` and you look for the baseStat, you can see that it proficiency.slug refers to the proficiency that is being extended, and if its blank that means its "base-spellcasting". You could maybe set proficiency.slug to default to base-spellcasting in prepareBaseData() and check that if you're feeling lucky.

In3luki:
I believe this is what you meant. I didn't extensively test this as I'm out of time again. I can follow up in a few hours if this is not it.

PR #20051, opened 2025-08-27
Add support for adding skill choices to Treat Wounds

websterguy:
Creates an array in flags to hold alternate skills to use for treat wounds and changes the treat wounds macro to add skills from there to its menu. This will generalize for any possible future features and homebrew additions.

May need explicit support for Chirurgeon and Natural Medicine added back to the macro unless existing features could be migrated to add the rule element to them.

image

### DIFF
--- a/src/scripts/macros/treat-wounds.ts
+++ b/src/scripts/macros/treat-wounds.ts
@@ -43,11 +43,16 @@ async function treatWounds(options: ActionDefaultOptions): Promise<void> {
         return;
     }
 
+    const skills = ["medicine"].concat(...(<[]>actor.flags.pf2e.treatWoundsSkills)).sort();
     const medicineName = game.i18n.localize("PF2E.Skill.Medicine");
-    const chirurgeon = CheckFeat(actor, "chirurgeon");
-    const naturalMedicine = CheckFeat(actor, "natural-medicine");
     const domIdAppend = fu.randomID(); // Attached to element id attributes for DOM uniqueness
     const riskySurgeryChecked = actor.getRollOptions(["medicine"]).includes("risky-surgery") ? " checked" : "";
+    const skillSelectOptions = skills
+        .map(
+            (skill) =>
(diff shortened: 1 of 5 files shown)


src/scripts/macros/treat-wounds.ts
```diff
@@ -43,11 +43,16 @@ async function treatWounds(options: ActionDefaultOptions): Promise<void> {
         return;
     }
 
+    const skills = new Set(["medicine"].concat(...(<[]>actor.flags.pf2e.treatWoundsSkills)).sort());
```
stwlam, line 46: Don't assert, and don't use concat (or at least not concat _and_ a spread).

src/scripts/macros/treat-wounds.ts
```diff
@@ -43,11 +43,16 @@ async function treatWounds(options: ActionDefaultOptions): Promise<void> {
...
-    const chirurgeon = CheckFeat(actor, "chirurgeon");
-    const naturalMedicine = CheckFeat(actor, "natural-medicine");
     const domIdAppend = fu.randomID(); // Attached to element id attributes for DOM uniqueness
     const riskySurgeryChecked = actor.getRollOptions(["medicine"]).includes("risky-surgery") ? " checked" : "";
+    const skillSelectOptions = [...skills]
+        .map(
+            (skill) =>
+                `<option value="${skill}">${game.i18n.localize("PF2E.Skill." + game.pf2e.system.sluggify(skill, { camel: "bactrian" }))}</option>`,
+        )
+        .join();
```
stwlam, line 55: At this point we should probably be using a template.

stwlam (changes requested):
This needs a migration, doesn't it?

src/scripts/macros/treat-wounds.ts
websterguy: I can definitely restructure the line, but having an issue that the assert solved. Getting error that `actor.flags.pf2e.treatWoundsSkills` is of type unkown. Not sure how to fix that specifically

image

websterguy:
> This needs a migration, doesn't it?

Yeah, I mentioned that it would need a migration unless I put back the explicit Natural Medicine and Chirurgeon checks.

PR #20070, opened 2025-08-30
Add support for resolvable value traits

CarlosFdez:
This is part of the work necessary to implement the Widen Area feat from Starfinder 2e. We still need a way to access the target item's existing area value inside ItemAlteration. Please recommend a way to avoid value.value. value.config instead?

The change to the data union validation allows it to report if value.trait or value.value are missing, but has the error message state that it must be a string if value is omitted instead of an object.

```json
{
  "key": "ItemAlteration",
  "mode": "add",
  "property": "traits",
  "itemType": "weapon",
  "value": {
    "trait": "area-burst",
    "annotation": 10
  }
}
```

https://2e.aonsrd.com/feats/633-widen-area

### DIFF
--- a/src/module/rules/rule-element/item-alteration/handlers.ts
+++ b/src/module/rules/rule-element/item-alteration/handlers.ts
@@ -1,7 +1,6 @@
 import type { DataFieldOptions } from "@common/data/_types.d.mts";
 import { ItemPF2e, WeaponPF2e } from "@item";
 import type { ItemSourcePF2e, ItemType } from "@item/base/data/index.ts";
-import type { ItemTrait } from "@item/base/types.ts";
 import { PersistentDamageValueSchema } from "@item/condition/data.ts";
 import { itemIsOfType } from "@item/helpers.ts";
 import { prepareBulkData } from "@item/physical/helpers.ts";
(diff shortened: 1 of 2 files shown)


stwlam:
Since we're going with `annotation` elsewhere, we might as well here, too.

CarlosFdez:
Makes sense to me, updated!

src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -966,29 +977,47 @@ const ITEM_ALTERATION_HANDLERS = {
...
+                [
+                    new fields.SchemaField({
+                        trait: new fields.StringField({ required: true, nullable: false }),
+                        annotation: new ResolvableValueField({ required: true, nullable: false }),
+                    }) satisfies TraitsValueConfigField,
+                    new StrictStringField<ItemTrait, ItemTrait, true, false, false>({
+                        required: true,
+                        nullable: false,
+                        initial: undefined,
+                    }),
```
stwlam: This is just a `StrictStringField` of `string`, right?

CarlosFdez: Ya. It was a StringField for string that claimed it was for Item trait before. I think the original intent was that the validateForItem will validate that it's an ItemTrait

CarlosFdez: If we remove the ItemTrait that'll let me upgrade this to remove "volley" without having to specify for what, but we'd lose validation for subtract and we'd need to double validate add so not sure about that

CarlosFdez: The validation is deferred. Once validated, the type is ItemTrait and the handler requires it to be that. This isn't a change from before either really.

stwlam: Yeah, but the source value isn't necessarily an `ItemTrait`, which is what this claims it is.

CarlosFdez: Isn't that true for every single stringfield ever? EDIT: At least for rule elements

stwlam: Only resolvable ones, and they are marked accordingly as just `string`s at the source level.

PR #20073, opened 2025-08-30
Unify handling of battle-form omission of some REs

### LINKED ISSUE #20062: Amphisbaena Handwraps showing Versatile P in Battle Form attacks
as title, adding Versatile P surely counts as changing the statistics of the form

stwlam:
Also add `ItemAlteration` to the mix

- Closes #20062


src/module/rules/rule-element/item-alteration/rule-element.ts
```diff
@@ -34,6 +34,7 @@ class ItemAlterationRuleElement extends RuleElementPF2e<ItemAlterationRuleSchema
                 choices: itemTypeChoices,
                 initial: undefined,
             }),
+            battleForm: new fields.BooleanField({ required: false }),
```
CarlosFdez: Weren't we not supposed to use required: false ?

PR #20079, opened 2025-08-31
Add trait annotations logging

CarlosFdez:
This resurrects the traits config with a different implementation. This will (in combination with #20070) allow the implementation of soldier's Widen Area as well as a correct implementation of Area Fire itself.

I considered allowing removeTrait() to support unannotated traits (allowing "volley" to remove any volley), however we would need to relax validation of item alteration to do that. If we want to keep that validation we would need to register a list of annotate-able traits somewhere.

### DIFF
--- a/src/module/item/base/data/system.ts
+++ b/src/module/item/base/data/system.ts
@@ -20,10 +20,17 @@ interface ActionCost {
     value: OneToThree | null;
 }
 
+interface TraitConfig {
+    volley?: number;
+    tracking?: number;
+    [key: string]: number | undefined;
+}
+
 interface ItemTraits<T extends ItemTrait = ItemTrait> {
     value: T[];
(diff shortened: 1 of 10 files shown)


src/module/item/base/data/system.ts
```diff
@@ -20,10 +20,17 @@ interface ActionCost {
...
+    volley?: number;
+    tracking?: number;
+    [key: string]: number | undefined;
+}
+
 interface ItemTraits<T extends ItemTrait = ItemTrait> {
     value: T[];
     rarity: Rarity;
     otherTags: string[];
+    annotations?: TraitAnnotations;
```
stwlam: `config` still

CarlosFdez: Ok fine!

PR #20092, opened 2025-09-01
Show action glyphs on ruler-waypoint labels and intermediate waypoints

stwlam:
This is in dire need of attendance to edge cases, but I'd like to get it merged before doing so enlarges the commit's diff too much.

[Screencast from 08-31-2025 10:53:47 PM.webm](https://github.com/user-attachments/assets/fedcfc4b-31cb-4463-98fd-17884367f82e)


CarlosFdez:
This is way too intense for me to review normally. Is there something in specific you want me to look at? Otherwise I'm more inclined to think about this conceptually, and other than characters might need to have a configurable crawl speed in the future (nimble crawl...) the idea itself is fine by me.

CarlosFdez (approved):

In3luki:
Looks good to me. Using the token layer to extract the ruler path is something I never would have thought of. The use of a mutation observer to sidestep the async rendering of the labels is also very nice. I need to remember that use case for the future.

PR #20151, opened 2025-09-10
Fix Runic Body and Runic Weapon not working in ABP

CarlosFdez:


src/module/rules/rule-element/item-alteration/alteration.ts
```diff
@@ -21,6 +21,11 @@ class ItemAlteration extends foundry.abstract.DataModel<RuleElementPF2e, ItemAlt
                 initial: undefined,
                 choices: R.keys(ITEM_ALTERATION_HANDLERS),
             }),
+            fromEquipment: new fields.BooleanField({
+                required: true,
+                nullable: false,
+                initial: true,
```
stwlam, line 27:
```suggestion
                initial: false,
```

CarlosFdez: This will cause flat modifiers to behave differently. Flat modifiers default to always negating item bonuses in ABP. In the PR ItemAlteration defaults to always negating potency/resilient/striking in ABP. Flipping that will cause it to not negate unless we tell it to. That's ok?

CarlosFdez: One thought I had was having the alteration default to false and the rule element default to true, but I wasn't sure about that.

stwlam: I don't think it will cause them to behave differently since this is the ItemAlteration RE

CarlosFdez: It will cause item alterations to behave differently from flat modifiers, if I structure it less vaguely.

CarlosFdez: Done. `fromEquipment` has no bearing on other types of alterations anyways. But at least now they will be forced false even on the alterations that don't check it. This does add some degree of hardcoding though. Every handler that will check fromEquipment needs to be added to the exception list.

EDIT: The reason FlatModifier does that check is because ABP suppresses flat modifiers that are fromEquipment regardless of the bonus type (see `FlatModifierPF2e#beforePrepareData()`). Item Alterations suppression instead occurs in the property handler, so the check is kinda implicitly already done.

PR #20159, opened 2025-09-12
Fix behavior of SF2e weapons and armor when using ABP

CarlosFdez:
Disable whitespace when reading this one.

### DIFF
--- a/static/templates/items/weapon-details.hbs
+++ b/static/templates/items/weapon-details.hbs
@@ -77,29 +77,31 @@
             </select>
         </div>
 
-        {{#if data.grade}}
-            <div class="form-group">
-                <label for="{{fieldIdPrefix}}grade">{{localize "PF2E.Item.Physical.FIELDS.grade.label"}}</label>
-                <select data-property="system.grade" id="{{fieldIdPrefix}}grade">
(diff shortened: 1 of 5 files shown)


static/templates/items/weapon-details.hbs
```diff
@@ -77,28 +77,30 @@
...
-                </select>
-            </div>
-            <div class="form-group">
-                <label for="{{fieldIdPrefix}}rune-striking">{{localize "PF2E.StrikingRuneLabel"}}</label>
-                <select data-property="system.runes.striking" id="{{fieldIdPrefix}}rune-striking" data-dtype="Number">
-                    <option value="0"></option>
-                    {{selectOptions (omit runeTypes.striking 0) selected=data.runes.striking labelAttr="name" localize=true}}
-                </select>
-            </div>
+        {{#if (not abpEnabled)}}
```
stwlam: Maybe flip these branches so it's not "if-not, else"

CarlosFdez: There's no else branch, the else is for the grade (SF2e vs PF2e). I can replace it with an `#unless` though?

stwlam: That would be a little saner, sure

PR #20160, opened 2025-09-12
Localization improvements "Cast" term and the Compendium Browser's "Abilities" label

### LINKED ISSUE #19833: Request: change Compendium Browser's Abilities label
Title is vague because it would be super long to explain in there, sorry.

Issues:

1. The Abilities tab is labeled that way because it includes the passive abilities of NPCs and Familiars, but it also includes all the Actions.
2. The localization string for the tab (`PF2E.Item.Ability.Plural`) isn't specific to the Compendium Browser, so the string can't be too different from the original meaning of "Abilities", which may sound weird to localize.

My suggestions:

1. Consider changing the label to "Actions/Abilities", similar to the "Feats/Features" tab.
2. Add a specific localization string for the tab.

I believe (2) is simple enough that nobody would object it, but I think (1) is contentious so I've opened this issue for discussion instead of just making a PR.

mclemente:
Both changes improve localization by solving cases where a single string was being used in different contexts.
- Split the `PF2E.Item.Spell.Cast` localization string ("Cast") between Noun and Verbal string.
- ~~Created a bespoke string for the Compendium Browser's Abilities tab.~~ Updated the unused `TabAction` string to "Abilities" and set it to "Abilities" tab. Closes #19833
- Created bespoke strings for the Equipment, Hazards and Spells tabs.
- Fixed the Compendium Browser's settings displaying "Actions" instead of "Abilities" for the first tab's name.


stwlam:
For what reason are the other browser tabs not in need of the same treatment? Is the reason incidental to Portuguese but perhaps not shared by other languages?

mclemente:
> For what reason are the other browser tabs not in need of the same treatment? Is the reason incidental to Portuguese but perhaps not shared by other languages?

The other tabs' names already represent their content even though they don't use unique strings (Equipment, Hazards, Spells), so I didn't see a reason to change it. I could also change them if you'd like.

IMO "Abilities" isn't a good fit for the tab's name (at least in Portuguese) and the string not being unique means I can't change its name without affecting other places where the string is used.

stwlam:
So the reason is incidental to Portuguese. We should replace all tab names, in that case.

src/module/apps/compendium-browser/tabs/action.ts
```diff
@@ -7,7 +7,7 @@ import { ActionFilters, CompendiumBrowserIndexData } from "./data.ts";
 
 export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
     tabName: ContentTabName = "action";
-    tabLabel = "PF2E.Item.Ability.Plural";
+    tabLabel = "PF2E.CompendiumBrowser.TabAction";
```
stwlam: These are in fact _abilities_, not just actions. Quite a few are passive.

mclemente: That string already exists on the system, I've changed it to "Abilities" [here](https://github.com/foundryvtt/pf2e/pull/20160/files#diff-488d88d8d2c417a9ebb2273d262f1f90a447232007d4a2ee5635d6fba79da8b5R1057).

This is meant to avoid having an orphan `TabAction` string and keeping the naming pattern for the strings.

stwlam: Replace it with `TabAbilities`

PR #20178, opened 2025-09-13
Migrate legacy speed path in rule elements and non-NPC system data

stwlam:
Migration result waiting behind approval

### DIFF
--- a/src/module/actor/creature/document.ts
+++ b/src/module/actor/creature/document.ts
@@ -374,30 +374,20 @@ abstract class CreaturePF2e<
         // Set IWR guaranteed by traits
         setImmunitiesFromTraits(this);
 
-        // Set difficult terrain roll options
-        if (game.ready && game.scenes.active) {
-            const tokens = this.getActiveTokens(true, true);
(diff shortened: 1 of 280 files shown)


src/module/actor/creature/document.ts
```diff
@@ -374,30 +374,20 @@ abstract class CreaturePF2e<
...
-        if (game.ready && game.scenes.active) {
-            const tokens = this.getActiveTokens(true, true);
-            const highestGrade = Math.max(...tokens.map((t) => t.difficultTerrain));
-            if (highestGrade > 0) {
-                this.rollOptions.all["self:position:difficult-terrain"] = true;
-                const gradeOption = highestGrade === 2 ? "greater" : "normal";
-                this.rollOptions.all[`self:position:difficult-terrain:${gradeOption}`] = true;
-            }
-        }
-
```
CarlosFdez, line 387: Its not implemented, but I suspect this might still be useful for stuff for stuff like Nature's Edge, even if it doesn't impact movement. Unless its already handled elsewhere and no longer necessary?

PR #20185, opened 2025-09-13
Implement equipment grade item alteration

CarlosFdez:
`addOrUpgradeTrait` isn't really the right name for the function anymore, but I don't really have any thoughts right now. `mergeTrait` and `applyTrait` are misleading, `addTrait` lacks information.


stwlam:
What is happening here if not an addition or upgrade of the tracking/resilient trait?

PR #20189, opened 2025-09-15
Add vidya-game trade window

stwlam:

### DIFF
--- a/src/module/apps/sidebar/actor-directory.ts
+++ b/src/module/apps/sidebar/actor-directory.ts
@@ -9,6 +9,7 @@ import type { DropCanvasData } from "@client/helpers/hooks.d.mts";
 import type { ActorUUID } from "@common/documents/_module.d.mts";
 import { htmlClosest, htmlQueryAll } from "@util";
 import * as R from "remeda";
+import { TradeDialog, TradeRequestData } from "../trade-dialog/app.ts";
 
 /** Extend ActorDirectory to show more information */
 class ActorDirectoryPF2e extends fa.sidebar.tabs.ActorDirectory<ActorPF2e<null>> {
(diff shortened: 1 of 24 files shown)


src/module/apps/trade-dialog/app.svelte
```diff
@@ -0,0 +1,448 @@
...
+    // Item lists and summaries
+    const listFormatter = game.i18n.getListFormatter({ style: "narrow" });
+    let selfItems = $derived(
+        self.items
+            .filter((i) => i.visible && i.matchScore > 0)
+            .sort((a, b) => b.matchScore - a.matchScore || a.name.localeCompare(b.name, game.i18n.lang)),
+    );
+    const withQuantity = (i: { quantity: number; marked: number; name: string }) =>
+        i.quantity > 1 ? `${i.marked}x ${i.name}` : i.name;
+    let itemsToSend = $derived.by(() => {
```
CarlosFdez: Minor thing, feel free to resolve, but I think you can const the deriveds. You only need let if its a state or you plan to reassign it if I remember right.

src/module/apps/sidebar/actor-directory.ts
```diff
@@ -244,20 +245,67 @@ class ActorDirectoryPF2e extends fa.sidebar.tabs.ActorDirectory<ActorPF2e<null>>
...
+                        game.actors.find(
+                            (a) => a.isOwner && a !== actor && a.isOfType("character", "npc") && a.isAllyOf(actor),
+                        );
+
+                    return !!createTradeArgs(actor, selfActor);
+                },
+                callback: (li) => {
+                    const token = canvas.tokens.controlled.length === 1 ? canvas.tokens.controlled[0] : null;
+                    const assigned = game.user.character?.isOfType("character", "npc") ? game.user.character : null;
+                    const selfActor = token?.actor ?? assigned;
```
CarlosFdez:
```suggestion
                    const selfActor = token?.actor ?? game.user.character;
```
Would this work? You already check if its a character or npc in the if statement after.

static/templates/popups/item-transfer-dialog.hbs
```diff
@@ -1,39 +1,20 @@
...
-                <span class="price"></span>
-            </button>
-        {{else}}
-            <button type="submit">
-                <i class="fa-regular fa-floppy-disk"></i>{{localize "PF2E.loot.MoveLoot"}}
-            </button>
-        {{/if}}
+{{#if lockStack}}
+    <input type="hidden" name="newStack" value="true" data-dtype="Boolean" />
+{{else if (ne action give)}}
```
CarlosFdez: Should `give` be `"gift"`? Quotes and all?

src/module/apps/sidebar/actor-directory.ts
stwlam: It works because it's checked after, and the controlled token should always be preferred over assigned character.

src/styles/actor/_trade-dialog.scss
```diff
@@ -0,0 +1,9 @@
+#trade-dialog {
+    max-height: 67vh;
+    &:not(.maximizing, .minimizing, .minimized) {
+        min-height: 400px;
+    }
+    .window-content {
+        flex-direction: row;
+    }
+}
```
CarlosFdez: This is an idea I had recently, but what's the viability of doing `global(#trade-dialog)` from the svelte application itself? Would that keeps things tidier?

src/module/item/base/sheet/sheet.ts
```diff
@@ -612,10 +612,6 @@ class ItemSheetPF2e<TItem extends ItemPF2e> extends fav1.sheets.ItemSheet<TItem,
         return buttons;
     }
 
-    protected override _canDragDrop(_selector: string): boolean {
-        return this.item.isOwner;
-    }
-
```
CarlosFdez, line 618: Does dropping stuff in class or deity item sheets work without this?

stwlam: You can't edit a class or deity item unless you own it.

CarlosFdez: I mean yeah, the function was testing for that. But I double checked and it still worked, and it looks like its implemented upstream by checking for isEditable, so it seems this was unnecessary.

PR #20273, opened 2025-09-21
Fix the weapon damage dice upgrade from the item clay sphere

### LINKED ISSUE #20105: Clay Sphere Dice Upgrade does not work
(EDIT: Forgot to label this as a bug)

Recently, while testing, I found that Clay Sphere's dice upgrade does not work (even with no module installed to check for module conflicts). However, I have personally fixed it by adjusting the RE. 

As seen in a fresh world with lv1 character, applying Clay Sphere to a Longbow item: 
Image
As seen with roll inspector, the dice size upgrade effect is applied but damage dice did not change. 
Image

Using the DamageDice RE instead ({"key":"DamageDice","override":{"upgrade":true},"selector":["{actor|flags.pf2e.claySphere.weapon}-damage"],"slug":"base"}) produces the desired result: 

Image

CotillionTheRope:
Fixes https://github.com/foundryvtt/pf2e/issues/20105

This is my first foray into rule elements, and I used the one proposed in the issue noted above. Happy for feedback on this.

### DIFF
--- a/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
+++ b/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
@@ -22,14 +22,10 @@
         },
         "rules": [
             {
-                "key": "DamageDice",
-                "override": {
-                    "upgrade": true
-                },
-                "selectors": [
-                    "{actor|flags.pf2e.claySphere.weapon}-damage"
-                ],
-                "slug": "base"
+                "key": "ItemAlteration",
+                "itemId": "{actor|flags.pf2e.claySphere.weapon}",
+                "mode": "upgrade",
+                "property": "damage-dice-faces"
             }
         ],
         "start": {
--- a/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
+++ b/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
@@ -22,13 +22,14 @@
         },
         "rules": [
             {
-                "key": "DamageAlteration",
-                "mode": "upgrade",
-                "property": "dice-faces",
-                "selectors": [
+                "key":"DamageDice",
+                "override": {
+                    "upgrade":true
+                },
+                "selector": [
                     "{actor|flags.pf2e.claySphere.weapon}-damage"
                 ],
-                "slug": "base"
+                "slug":"base"
             }
         ],
         "start": {
--- a/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
+++ b/packs/equipment-effects/effect-clay-sphere-weapon-damage-increase.json
@@ -22,14 +22,14 @@
         },
         "rules": [
             {
-                "key":"DamageDice",
+                "key": "DamageDice",
                 "override": {
-                    "upgrade":true
+                    "upgrade": true
                 },
-                "selector": [
+                "selectors": [
                     "{actor|flags.pf2e.claySphere.weapon}-damage"
                 ],
-                "slug":"base"
+                "slug": "base"
             }
         ],
         "start": {


pedrogrullada:
We are purposefully moving DamageDice override REs to DamageAlteration REs. There is actually nothing wrong with the DamageAlteration in that effect, so if it is in fact not working, that would be a bug.

You could try having `dice:slug:base` in the predicate and deleting the `"slug"` property, but even if that works, it would only be a bandaid.

CotillionTheRope:
> We are purposefully moving DamageDice override REs to DamageAlteration REs. There is actually nothing wrong with the DamageAlteration in that effect, so if it is in fact not working, that would be a bug.
> 
> You could try having `dice:slug:base` in the predicate and deleting the `"slug"` property, but even if that works, it would only be a bandaid.

I can confirm that the current RE does not work. I am struggling to understand everything happening here to attempt to fix it. I have a couple questions that I don't see answers to in the documentation:

- Is `"slug": "base"` just shorthand for the predicate `dice:slug:base`? If so, why is the predicate just a bandaid.
- Most other examples of using DamageAlteration with mode `upgrade` and property `dice-faces` have a `value` attribute. Does this need a `value` of 1 to indicate it should be updated one step? Or would 1 be interpreted as `d1`?
- Similarly, most other examples of using DamageAlteration with mode `upgrade` and property `dice-faces` have a predicate along with a `slug`, many of which are `base`. If the predicate `dice:slug:base` is not correct here, might some other predicate be missing?
- When doing as suggested above and removing the slug and adding the predicate, the base damage die of any weapon I choose to use with this effect active is upgraded by one step. However, the flavor of the item which this effect comes from implies it is only active for a specific weapon, not all weapons. And when using the DamageDice override RE as committed I get the option to enable or disable the effect on a damage roll. This seems more accurate to the items effect, so would this be a case where we want the DamageDice instead of DamageAlteration?

I'm sorry if this is not the correct place for these questions, but I thought the context provided in the PR and linked issue would help with clarification. If it would be preferred, I can move this to the Discord and link to the PR and issue.

In the end, I'm just trying to understand this better so I can help with more complex issues in the future and hopefully with the SF2e module work.

Thanks!

pedrogrullada:
> Is `"slug": "base"` just shorthand for the predicate `dice:slug:base`? If so, why is the predicate just a bandaid.

It'd be a bandaid because the `slug` key not working would be an underlying bug that would remain unaddressed.

> Most other examples of using DamageAlteration with mode `upgrade` and property `dice-faces` have a `value` attribute. Does this need a `value` of 1 to indicate it should be updated one step? Or would 1 be interpreted as `d1`?

The value is the number of faces (hence the `dice-faces` property) the die is being upgraded to. `8` for a d8, `10` for a d10, etc. (which should upgrade one step if the value omitted)
 
> Similarly, most other examples of using DamageAlteration with mode `upgrade` and property `dice-faces` have a predicate along with a `slug`, many of which are `base`. If the predicate `dice:slug:base` is not correct here, might some other predicate be missing?

There is no predicate missing, and `dice:slug:base` is a correct way of doing this. However, `"slug": "base"` should be working, which is the actual issue that needs addressing. To be clear, it would not be a data issue but rather a bug, but ideally we should have clear reproduction steps so that the devs can take a proper look at some point.

> When doing as suggested above and removing the slug and adding the predicate, the base damage die of any weapon I choose to use with this effect active is upgraded by one step. However, the flavor of the item which this effect comes from implies it is only active for a specific weapon, not all weapons. And when using the DamageDice override RE as committed I get the option to enable or disable the effect on a damage roll. This seems more accurate to the items effect, so would this be a case where we want the DamageDice instead of DamageAlteration?

The `{actor|flags.pf2e.claySphere.weapon}-damage` should restrict the rule element to only the chosen weapon. If that's not working, there may be a different issue with the rule element setting that flag.

pedrogrullada:
All that being said, since this is a weapon base damage dice upgrade, it should be an Item Alteration anyway. So if nothing else we should change the rule element to this. The same should be done eventually for other DamageAlterations that are upgrading base damage dice of weapons.
```json
{
  "key": "ItemAlteration",
  "itemId": "{actor|flags.pf2e.claySphere.weapon}",
  "mode": "upgrade",
}
```

CotillionTheRope:
> > Is `"slug": "base"` just shorthand for the predicate `dice:slug:base`? If so, why is the predicate just a bandaid.
> 
> It'd be a bandaid because the `slug` key not working would be an underlying bug that would remain unaddressed.

Understood.

> > Most other examples of using DamageAlteration with mode `upgrade` and property `dice-faces` have a `value` attribute. Does this need a `value` of 1 to indicate it should be updated one step? Or would 1 be interpreted as `d1`?
> 
> The value is the number of faces (hence the `dice-faces` property) the die is being upgraded to. `8` for a d8, `10` for a d10, etc.

Understood and as I thought. So not specifying a value here just means "go up one step"?
 
> > When doing as suggested above and removing the slug and adding the predicate, the base damage die of any weapon I choose to use with this effect active is upgraded by one step. However, the flavor of the item which this effect comes from implies it is only active for a specific weapon, not all weapons. And when using the DamageDice override RE as committed I get the option to enable or disable the effect on a damage roll. This seems more accurate to the items effect, so would this be a case where we want the DamageDice instead of DamageAlteration?
> 
> The `{actor|flags.pf2e.claySphere.weapon}-damage` should restrict the rule element to only the chosen weapon. If that's not working, there may be a different issue with the rule element setting that flag.

I see. I was not getting a selection dialog because I only had one weapon so it defaulting.

CotillionTheRope:
> All that being said, since this is a weapon base damage dice upgrade, it should be an Item Alteration anyway. So if nothing else we should change the rule element to this. The same should be done eventually for other DamageAlterations that are upgrading base damage dice of weapons.
> 
> ```json
> {
>   "key": "ItemAlteration",
>   "itemId": "{actor|flags.pf2e.claySphere.weapon}",
>   "mode": "upgrade",
>   "property": "damage-dice-faces"
> }
> ```

Can confirm this works in this case. I am confused about the `itemId` property here though. I have proved to myself that is required, but I'm not sure why or why this is the correct Id. The Quick Start guide doesn't show this property anywhere, and I'm confused about why it's looking for the flag on the actor and not `"{actor|flags.pf2e.claySphere.weapon}-damage"` on the item which does show up as a selector on the damage roll.

pedrogrullada (approved):

PR #20274, opened 2025-09-21
Include distance roll options in roll-context actor clones

stwlam:


src/module/actor/roll-context/base.ts
```diff
@@ -256,12 +240,12 @@ abstract class RollContext<
             target: unresolved.target?.actor ?? null,
             item,
             domains: this.domains,
-            options: [...this.rollOptions, ...itemOptions, ...markOptions],
+            options: [...this.rollOptions, itemOptions, markOptions, distanceOption ?? []].flat(),
```
CarlosFdez, line 243:
```suggestion
            options: [this.rollOptions, itemOptions, markOptions, distanceOption ?? []].flat(),
```
You'd do this if you're gonna flatten right?

stwlam: I am flattening right there

PR #20282, opened 2025-09-21
Replace `CompendiumBrowser` checkbox filters with a tri-state filter component

### LINKED ISSUE #15606: Compendium browser doesn't disable some particular sources
As the GM, in the compendium browser window, on the Settings tab, in the Sources sub-tab, I have only selected: Pathfinder Player Core

Logged in as a player, across the Abilities, Equipment, Feats/Features and Spells tabs, the Source filter section has some combination of the following selections:

```
Pathfinder Extinction Curse Player's Guide
Pathfinder Player Core
Pathfinder Stolen Fate Player's Guide
Pathfinder Wake the Dead #1
Pathfinder Wake the Dead #2
Pathfinder Wake the Dead #3
Pathfinder Wake the Dead #4
Pathfinder Wake the Dead #5
```

These entries appear in the GMs window in the Settings->Sources tab, but are definitely NOT selected.

These extra sources only appear if the GM has TICKED the option "Do not filter entries with unknown sources".

In3luki:
This one has grown a bit bigger than I anticipated so I'm pushing this up for review before adding any more to it. The browser should overall work the same as before but there might be some filter edge cases that are not covered.

- Adds a tri-state (neutral, include, exclude) chip filter to make it possible to exclude certain options from searches.
  - The first click includes the option, the second click excludes the options, the third click clears the option.
  - A right clicks clears the option immediately.
  - Shift left click excludes the option immediately.
- Replaces the custom filter implementation with a filter based on a `Predicate` which should make it even easier to implement new filter options.
- Updates all filter options to use arrays of labeled values instead of objects, eliminating the need for conversion before iteration.
- Adds a `Defense` filter to the spell tab that allows filtering by save, AC, basic save, etc.
- Introduces the concept of a custom trait filter that is not based on an actual trait to solve the problem of showing ancestry items that have no ancestry trait:
image

- Moves the `FilterContainer` wrapper to the indiviudal filters to remove the need for the `getClearFunction` function.
  
  
image

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -1289,11 +1289,10 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
         }
 
         // Assign categories
-        const category = filter.checkboxes.category;
+        const category = filter.chips.category;
         for (const value of featFilters.categories ?? []) {
             category.isExpanded = true;
-            category.options[value].selected = true;
-            category.selected.push(value);
+            category.selected.push({ value });
         }
 
         // Assign traits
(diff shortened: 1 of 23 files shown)


stwlam:
Any way to get a minimal fix extracted from this?

PR #20296, opened 2025-09-22
Fix closing choice set without selection

CarlosFdez:


src/module/rules/rule-element/choice-set/rule-element.ts
```diff
@@ -251,6 +251,11 @@ class ChoiceSetRuleElement extends RuleElement<ChoiceSetSchema> {
                 rule.ignored = false;
             }
         } else {
+            if (this.choices.length > 0 && !this.allowNoSelection) {
+                ui.notifications.warn(
+                    game.i18n.format("PF2E.UI.RuleElements.Prompt.NoSelectionMade", { item: this.item.name }),
+                );
```
stwlam:
```suggestion
                ui.notifications.warn("PF2E.UI.RuleElements.Prompt.NoSelectionMade", {
                    format: { item: this.item.name },
                });
```

PR #20323, opened 2025-09-24
Add darkvision with Signifer's Sight while wearing Signifer's Mask

pedrogrullada:
Also correct typo in the item's name

### DIFF
--- a/packs/feats/archetype/hellknight-signifer/signifers-sight.json
+++ b/packs/feats/archetype/hellknight-signifer/signifers-sight.json
@@ -29,15 +29,7 @@
             "remaster": false,
             "title": "Pathfinder Lost Omens Character Guide"
         },
-        "rules": [
-            {
-                "key": "Sense",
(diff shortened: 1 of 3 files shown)


packs/feats/archetype/hellknight-signifer/signifers-sight.json
```diff
@@ -29,7 +29,15 @@
...
-        "rules": [],
+        "rules": [
+            {
+                "key": "Sense",
+                "predicate": [
+                    "signifers-mask"
+                ],
+                "selector": "darkvision"
+            }
+        ],
```
TikaelSol: Why not put the rule on the mask predicate on having this feat? That seems a bit cleaner unless multiple things rely on the mask being worn like this.

pedrogrullada: Just trying to keep REs contained in the items they're automating, no other reason

PR #20326, opened 2025-09-24
Correct Critical Specialization Rule Element on Iruxi Armaments feat.

ammalagonc:


packs/feats/ancestry/lizardfolk/iruxi-armaments.json
```diff
@@ -143,7 +143,9 @@
                     {
                         "or": [
                             "item:id:{item|id}",
-                            "item:slug:claw"
```
ammalagonc, line 146: I think `item:id:{item|id}` is a nothing-burger - should we delete?

pedrogrullada: It's for the Strikes granted by the feat

ammalagonc: OK, so keep it.

Thanks for checking!

packs/feats/ancestry/lizardfolk/iruxi-armaments.json
```diff
@@ -143,7 +143,9 @@
...
+                            "item:slug:claw",
+                            "item:slug:fangs",
+                            "item:slug:tail"
```
pedrogrullada: These are already covered by `item:id:{item|id}`

pedrogrullada: The thing you actually need to do with this rule element is make sure that the specialization applies only if the `iruxi-armaments:slug` roll option exists alongside the other conditions to identify the Strike.

Additionally, I'd suggest going to the Lizardfolk claws feature and tagging the Claw strike given by it, and using that to identify it in this rule element. See how Hungerseed tags its horn Strike, for an example.

packs/feats/ancestry/lizardfolk/iruxi-armaments.json
```diff
@@ -131,6 +131,17 @@
...
+                "itemType": "weapon",
+                "key": "ItemAlteration",
+                "mode": "add",
+                "predicate": [
+                    "item:slug:claw",
+                    "iruxi-armaments:claw"
+                ],
+                "property": "other-tags",
+                "value": "iruxi-armaments-claws"
+            },
```
pedrogrullada: Whoops. What I meant is that the Claws ancestry feature should tag its Strike, just like how Hungerseed does. Then your Crit Spec rule element here can look for that tag!

So you'd add something like this to the lizardfolk Claws feature
```json
{
    "itemType": "weapon",
    "key": "ItemAlteration",
    "mode": "add",
    "predicate": [
        "item:id:{item|id}"
    ],
    "property": "other-tags",
    "value": "lizardfolk-claws"
}
```
Then instead of `item:slug:claws` in the Critical Specialization RE, you can use `item:tag:lizardfolk-claws`.

packs/ancestryfeatures/claws.json
```diff
@@ -45,6 +45,16 @@
...
+            },
+            {
+                "itemType": "weapon",
+                "key": "ItemAlteration",
+                "mode": "add",
+                "predicate": [
+                    "item:slug:claw"
+                ],
+                "property": "other-tags",
+                "value": "lizardfolk-claws"
```
pedrogrullada, line 58:
```suggestion
            },
            {
                "itemType": "weapon",
                "key": "ItemAlteration",
                "mode": "add",
                "predicate": [
                    "item:id:{item|id}"
                ],
                "priority": 119,
                "property": "other-tags",
                "value": "lizardfolk-claws"
```
Making sure we're targeting the right claw attack, also lowered the priority a bit to make sure it fires off before the Item Alterations in Iruxi Armaments

packs/feats/ancestry/lizardfolk/iruxi-armaments.json
```diff
@@ -142,8 +143,13 @@
                     },
                     {
                         "or": [
-                            "item:id:{item|id}",
-                            "item:slug:claw"
+                            {
+                                "and": [
+                                    "item:slug:claw",
```
pedrogrullada:
```suggestion
                                    "item:tag:lizardfolk-claw",
```
Sorry for the double review, overlooked this bit!

pedrogrullada:
Ugh, hate to do this, but if you don't mind, that DamageDice override should be done as an Item Alteration upgrade. Do you mind doing that in this PR as well? If not, I can get it after :)

packs/feats/ancestry/lizardfolk/iruxi-armaments.json
```diff
@@ -63,21 +63,24 @@
...
-                    "iruxi-armaments:claw"
+                    "iruxi-armaments:claw",
+                    "item:tag:lizardfolk-claws"
                 ],
-                "selector": "claw-damage"
+                "property": "dice-faces",
+                "selectors": [
+                    "claw-damage"
+                ],
+                "value": 6
```
pedrogrullada: This Damage Alteration would affect every single damage dice for the Iruxi's claw Strike damage, including bonus ones. For base weapon damage alteration, Item Alteration is the preferred method, like this.
```suggestion
                "itemType": "weapon",
                "key": "ItemAlteration",
                "mode": "upgrade",
                "predicate": [
                    "iruxi-armaments:claw",
                    "item:tag:lizardfolk-claws"
                ],
                "property": "damage-dice-faces"
```

ammalagonc: Check when you get a moment.

Thanks!

pedrogrullada (approved):

Issue #13848, opened 2024-02-19
Localization issues with Proficiency Ranks on PC Shet's Sidebar, and with Area Template's verbiage, and World Clock


mclemente:
For Proficiency Rank:
If Perception's or Saving Throws' prof rank is set to 12+ characters-long string, it overlaps.

![image](https://github.com/foundryvtt/pf2e/assets/5288872/f195e5cf-d3b8-4e75-8bd6-db53e0642840)

This isn't the case for ranks on the Proficiencies tab, where they get ellipsis due to their select's style, which is preferable on this case.

For Area Templates:
Area templates such as a Spell's Area is perfect for `5-Foot Burst`, but it falls apart for things like `Burst of 5 Foot` since the unit used is always the singular for Foot.

In Portuguese, "Foot/Feet" is translated to "Pé/Pés", so a simple fix is to use "Pé(s)", but that would have the undesirable effect that every other place where Foot is used would be displayed like that.
Not a big deal, but worth of note.

Edit:
Also, the World Clock's SyncDarknessScene setting's selector produces an unlocalized string if you toggle the SyncDarkness setting.

mclemente:
Here's the results of my tries at changing the styling:
![image](https://github.com/foundryvtt/pf2e/assets/5288872/a2bba8fb-3a5c-4596-9ace-d7882d891ad3)

Perception and Skills tab can be fixed by increasing their width from 6.5em to 7.5em (as shown above).
8em looks spacier, but the rightmost skills overflow if the name of the skills take up a lot of space.
![image](https://github.com/foundryvtt/pf2e/assets/5288872/2c4466a1-6882-4223-955c-60bfc7dc7734)

Saving Throws might look nicer with a `width: 100%`, but it still overflows if a character has 3 saves with long strings (3 Expert ranks)
![image](https://github.com/foundryvtt/pf2e/assets/5288872/bbaa822d-6ee8-4127-832d-37122855fb9c) ![image](https://github.com/foundryvtt/pf2e/assets/5288872/b580fd28-b495-4a26-980d-056c4fa82bdb)

PR #20337, opened 2025-09-25
Update sidebar's proficiency ranks to fix localization styling regression

mclemente:
There has been a regression of #13848 on the character sheet's sidebar due to `.pf-rank` being changed back to a flex, which makes the element to not make ellipsis if the text is too big.

Before/After:

image
image

The only drawback is Perception's rank being displayed 1px higher, which should be imperceptile unless people have the sheets open side by side.


src/module/scene/token-document/sheets/mixin.ts
```diff
@@ -26,11 +26,13 @@ function TokenConfigMixinPF2e<TBase extends ReturnType<typeof TokenApplicationMi
...
-            parts["appearance"].template = "systems/pf2e/templates/scene/token/appearance.hbs";
-            return parts;
-        })();
+        static override PARTS = {
+            ...super.PARTS,
+            appearance: {
+                scrollable: [""],
+                template: "systems/pf2e/templates/scene/token/appearance.hbs",
+            },
+        };
```
stwlam, line 35: ?

mclemente: That wasn't supposed to be part of this PR, it is from another change elsewhere, I'm sorry.
I'll update my fork so these don't happen.

CarlosFdez: I know you did this locally to get your stuff to work, but I updated world explorer to handle scene config overrides better, since I ran into the same issue for Daggerheart. Since your OG implementation was world explorer's, I figure you should know.

PR #20365, opened 2025-09-30
Add support for configuring npc attacks as area/auto fire

CarlosFdez:
No actual support yet, they're still interpreted as strikes. The actual implementation will be kinda hairy, and it might be best to separate the guaranteed parts first. Setting it is dev only for now.

### DIFF
--- a/src/module/item/melee/data.ts
+++ b/src/module/item/melee/data.ts
@@ -1,3 +1,4 @@
+import type { ModelPropsFromSchema, SourceFromSchema } from "@common/data/fields.mjs";
 import type { MeleePF2e } from "@item";
 import { ItemSystemModel, ItemSystemSchema } from "@item/base/data/model.ts";
 import type {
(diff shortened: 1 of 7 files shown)


src/module/item/melee/data.ts
```diff
@@ -81,6 +110,12 @@ class MeleeSystemData extends ItemSystemModel<MeleePF2e, NPCAttackSystemSchema>
             }),
         };
     }
+
+    override prepareBaseData(): void {
+        if (this.action !== "strike") {
+            this.area ??= { type: "burst", value: 5 };
+        }
+    }
```
stwlam, line 117: super up and reduce gratuitous white space

src/module/item/melee/sheet.ts
```diff
@@ -18,9 +24,15 @@ export class MeleeSheetPF2e extends ItemSheetPF2e<MeleePF2e> {
 
         return {
             ...sheetData,
+            attackActions: R.mapValues(NPC_ATTACK_ACTIONS, (a) => game.i18n.localize(a)),
+            areaShapes: R.mapToObj(EFFECT_AREA_SHAPES, (s) => [s, `PF2E.Area.Shape.${s}`]),
             damageTypes: CONFIG.PF2E.damageTypes,
             damageCategories: damageCategoriesUnique,
             attackEffects: createSheetOptions(this.getAttackEffectOptions(), this.item.system.attackEffects),
```
stwlam:
```suggestion
            attackEffects: createSheetOptions(this.getAttackEffectOptions(), item.system.attackEffects),
```

src/module/item/melee/sheet.ts
```diff
@@ -56,12 +68,24 @@ export class MeleeSheetPF2e extends ItemSheetPF2e<MeleePF2e> {
             formData[key] ||= null;
         }
 
+        // Subtract 10 from the dc to get the modifier if its area/auto fire
+        if (this.item.system.action !== "strike" && typeof formData["system.bonus.value"] === "number") {
+            formData["system.bonus.value"] -= 10;
+        }
```
stwlam, line 74: This seems a bit wonky UX-wise. Can it be a modifier no matter what and show a DC alongside if applicable?

CarlosFdez: I'd argue its cleaner UX wise, it ends up looking somewhat like this (under the hood, the modifier is 5).

image

That way the data they're inputting is the save dc the book gave them, rather than having to see DC22 and enter 11 as the modifier.

EDIT: Though perhaps I should rename to reflex save dc?

src/module/item/melee/data.ts
CarlosFdez: I think in this case the readability is hampered a bit by removing the braces, its not one of those super simple early return type statements, but its minor enough that I'll just do it.

PR #20371, opened 2025-09-30
Add area trait to npc auto and area fire and substitute label during rendering

CarlosFdez:


CarlosFdez:
I'm just gonna push the whole thing combined

PR #20380, opened 2025-10-02
Remove `isObject` helper

stwlam:


src/module/actor/hazard/document.ts
```diff
@@ -55,7 +55,7 @@ class HazardPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e |
         // Hazards without hit points are unaffected by damage
         const damageType = objectHasKey(CONFIG.PF2E.damageTypes, effect)
             ? effect
-            : isObject(effect)
+            : R.isPlainObject(effect)
```
CarlosFdez: I don't think ConditionPF2e is a plain object

PR #20382, opened 2025-10-02
Remove area trait from actions and convert effect area to tag

CarlosFdez:


src/module/actor/npc/sheet.ts
```diff
@@ -332,7 +340,7 @@ class NPCSheetPF2e extends AbstractNPCSheet {
                         label: v.label,
                         breakdown: idx === 0 ? breakdown : null,
                     })),
-                    traits,
+                    traits: tags.sort((a, b) => a.label.localeCompare(b.label, game.i18n.lang)),
```
stwlam: Let's call this `traitsAndTags` or the like and add a comment with some explanation of Paizo's internal tagging system for NPC attacks

CarlosFdez: Done!

PR #20385, opened 2025-10-03
Retire and migrate away RE brackets

stwlam:


packs/iconics/samo-level-3.json
```diff
@@ -3663,24 +3663,7 @@
...
-                                    "value": "@actor.level -1"
-                                },
-                                {
-                                    "start": 7,
-                                    "value": "@actor.level"
-                                }
-                            ],
-                            "field": "actor|level"
-                        }
+                        "value": "match(when(lte(@actor.level, 4), \"@actor.level -2\"), when(btwn(@actor.level, 5, 6), \"@actor.level -1\"), when(gte(@actor.level, 7), \"@actor.level\"))"
```
CarlosFdez: I would imagine that `@actor.level -2` without quotes or escaped quotes is the more correct one? Does match support a double resolve?

PR #20422, opened 2025-10-08
Implement weapon reloading and create ammo type

CarlosFdez:
Converts certain consumables into the ammo item type. Unlike the original implementation, ammo types are flattened and use the actual item slug, which means that there is some chaos with the magazines in the en.json. Oh well. Also includes reloading.

### RELOADING

Direct editing exists because auto fire is cursed and we're going to need it.

https://github.com/user-attachments/assets/7fb5ec13-e8af-44bc-9d90-69dbcefdf83f

### Capacity 

Weapons that don't use magazines have a capacity stat. This checks the config to determine that. This is the capacity/double barreled stats in pf2e, or the magazine stat in pf2e (which have nothing to do with pf2e magazines, separately clarified).

image

Note that air repeaters *do not* have a capacity stat, and we omit it. Having one would be confusing

### Dragging special ammo to an actor shows all possible things you can make with it.

This is handled by the "parent" property. For example, if a valid type is rounds, "rounds" and anything with "rounds" as a parent become valid choices. If you want, we can hardcode omit rounds (and magazines?) from the possible selections.

image

### The only craftable items are top level non-magazine items

image

### Only magazines type items have uses

This prevents loading a magazine of projectile ammo into a stellar cannon, mucking up the UI. I don't want to handle having to display both quantity *and* uses in the UI, just either or.


src/module/migration/migrations/950-ammo-consumable-to-ammo-ammo.ts
```diff
@@ -0,0 +1,405 @@
...
+            const fromSlug = slug.includes("round") ? "rounds" : slug.includes("bolt") ? "bolts" : "arrows";
+            system.baseItem = fromSlug;
+        }
+
+        source.img = source.img.replace(/\bconsumable\.svg$/, "ammo.svg") as ImageFilePath;
+        source.type = "ammo";
+        if ("game" in globalThis) {
+            // todo: handle == in the migration framework. Right now its fine, but will eventually cause problems
+            // If we don't use ==, changing the item type will be blocked by foundry
+            source["==system"] = system;
```
CarlosFdez, line 152: I'd appreciate some help here so I can focus on sf2e matters. What I had to do here was quite hacky.

Its not possible to run later migrations on ammo in the same batch as this one if we don't handle == somehow.

PR #20436, opened 2025-10-11
Use `basePath` in autogenerated RE forms

stwlam:


static/templates/items/rules/default.hbs
```diff
@@ -3,19 +3,19 @@
...
+                {{formGroup field name=(concat basePath "." key) value=(lookup ../rule key)
                             rootId=../rootId
                             elementType=(../form.elementType field)
                             placeholder=(ifThen (eq key "label") ../item.name "")}}
             {{/if}}
         {{/unless}}
     {{/each}}
     {{#unless (includes omittedFields "predicate")}}
-        {{formGroup fields.predicate name=(concat "system.rules." index ".predicate")
+        {{formGroup fields.predicate name=(basePath ".predicate")
```
kanongil, line 18: You are calling `basePath` as a helper! Since it resolves to a string it will throw an exception while rendering trying to use `.call()` on it. This can be seen when trying to show the sheet for "Shadow Signet".

I expect the fix is to use the `concat` helper again.

PR #20457, opened 2025-10-12
Fix transferring physical items into a container

CarlosFdez:
Dragging to a container has been broken for a while it seems, but the stacking in a container is new functionality.

### DIFF
--- a/src/module/actor/sheet/base.ts
+++ b/src/module/actor/sheet/base.ts
@@ -1275,8 +1275,8 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
             throw ErrorPF2e("Unexpected missing actor(s)");
         }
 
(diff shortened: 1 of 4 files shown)


src/module/actor/sheet/base.ts
```diff
@@ -1275,8 +1275,8 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
             throw ErrorPF2e("Unexpected missing actor(s)");
         }
 
-        const containerId = htmlClosest(event.target, "[data-is-container]")?.dataset.containerId?.trim();
-        const stackable = !!recipient.inventory.findStackableItem(item._source);
+        const containerId = htmlClosest(event.target, "[data-is-container]")?.dataset.itemId?.trim();
```
CarlosFdez, line 1278: This part is the bugfix

PR #20466, opened 2025-10-14
Fix copy to clipboard icons in the rules tab

CarlosFdez:


stwlam:
Not sure what this is fixing. Is there an issue #?

PR #20469, opened 2025-10-14
Fix certain invalid predicates from potency and striking migration

### LINKED ISSUE #20464: Runic Body spell effect does not apply to unarmed strikes
The implementation of the spell effect for _Runic Body_ seems to only apply to weapons, and not to any player unarmed strikes.
### Reproduction steps
1. Apply the "Spell Effect: Runic Body" effect at any level to a player actor with an unarmed strike and no Handwraps of Mighty Blows.
2. Perform an Unarmed Attack or other unarmed Strike.

Expected results: +1 Item bonus to attack roll and 2 damage dice.
Actual results: No Item bonus or increase in damage dice.

Image
Image

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/20464

### DIFF
--- a/src/module/migration/migrations/946-retire-potency-striking-res.ts
+++ b/src/module/migration/migrations/946-retire-potency-striking-res.ts
@@ -26,7 +26,7 @@ export class Migration946RetirePotencyStrikingREs extends MigrationBase {
             typeof rule.selector === "string" &&
             !predicate.some((p) => typeof p === "string" && p.startsWith("item:"))
         ) {
-            predicate.push(`item:slug:${rule.selector.replace(/-(?:attack|damage)$/, "")}`);
+            predicate.push(`item:slug:${rule.selector.replace(/-(?:attack|damage|attack-roll)$/, "")}`);
         }
         const unarmedIndex = predicate.findIndex((p) => p === "unarmed");
         if (unarmedIndex !== undefined && unarmedIndex >= 0) {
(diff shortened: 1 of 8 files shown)


src/module/migration/migrations/946-retire-potency-striking-res.ts
```diff
@@ -26,7 +26,10 @@ export class Migration946RetirePotencyStrikingREs extends MigrationBase {
             typeof rule.selector === "string" &&
             !predicate.some((p) => typeof p === "string" && p.startsWith("item:"))
         ) {
-            predicate.push(`item:slug:${rule.selector.replace(/-(?:attack|damage)$/, "")}`);
+            const slug = rule.selector.replace(/-(?:attack|damage|attack-roll)$/, "");
+            if (!["strike", "unarmed", "ranged", "melee"].includes(slug)) {
+                predicate.push(`item:slug:${slug}`);
+            }
```
stwlam: This is theoretically untested, right? Why not leave 946 untouched since 947 will fix it?

stwlam:
Issue #?

src/module/migration/migrations/946-retire-potency-striking-res.ts
CarlosFdez: We could omit the predicate exclusion, though we would still need to leave in including attack-roll if we hope to get someone's homebrew REs before its too late.

PR #20476, opened 2025-10-15
Fix handling of copy-to-clipboard buttons on item sheet RE panel

### LINKED ISSUE #20471: "Copy UUID" button not working
Image

The button to copy UUID is grayed out on compendium items, but works for imported items. Additionally, the button to copy Source ID does nothing when clicked on on either compendium or imported items.

stwlam:
- Closes #20471


CarlosFdez:
Does this also work on compendium items or nah?

PR #20477, opened 2025-10-15
Fix retrieval of in-memory unarmed attacks for use in damage application

### LINKED ISSUE #20121: Add the `origin` flag to applied damage messages generated by character unarmed attacks.
Right now, the applied damage messages generated by a character unarmed attack (i.e. monk `Fist`) does not contains the `origin` flag while message generated by weapon strikes or npc strikes will have it.

This makes it impossible to retrieve the actor that dealt the damage nor the item associated with it.

Image
In the screenshot above, the first message is from a monk `Fist` strike while the second one was from a weapon strike.

stwlam:
- Closes #20121

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -1,5 +1,5 @@
 import type { ActorPF2e } from "@actor";
-import { StrikeData } from "@actor/data/base.ts";
+import type { StrikeData } from "@actor/data/base.ts";
 import type { Rolled } from "@client/dice/roll.d.mts";
 import type { DataModelConstructionContext } from "@common/abstract/_module.d.mts";
 import type {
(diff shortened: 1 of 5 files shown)


src/module/chat-message/document.ts
```diff
@@ -114,25 +114,27 @@ class ChatMessagePF2e extends ChatMessage {
...
-            const embeddedSpell = this.flags.pf2e.casting?.embeddedSpell;
-            if (actor && embeddedSpell) return new ItemProxyPF2e(embeddedSpell, { parent: actor });
-
-            const origin = this.flags.pf2e?.origin ?? null;
-            const item = origin?.uuid && !origin.uuid.startsWith("Compendium.") ? fromUuidSync(origin.uuid) : null;
-            return item instanceof ItemPF2e ? item : null;
-        })();
-        if (!item) return null;
+        const embeddedSpell = this.flags.pf2e.casting?.embeddedSpell;
+        if (actor && embeddedSpell) return new ItemProxyPF2e(embeddedSpell, { parent: actor });
```
CarlosFdez: This may need to be heightened or set with a variant. That's what the if is spell block later is for.

CarlosFdez:
lgtm

PR #20492, opened 2025-10-17
Refactor Compendium Browser to use a `Predicate` for filtering entries

In3luki:
I finally had some more time to work on this. Pulled out from #20282

This PR only includes the switch to predication and no other additions or UI changes. If this is approved I can start to pull out the other changes from the old PR.

Current filter:
Current

Predicate filter:
CachedPredicate

### DIFF
--- a/src/module/apps/compendium-browser/tabs/action.ts
+++ b/src/module/apps/compendium-browser/tabs/action.ts
@@ -1,5 +1,4 @@
 import { getActionIcon } from "@module/sheet/helpers.ts";
-import { sluggify } from "@util";
 import * as R from "remeda";
 import { CompendiumBrowser } from "../browser.ts";
 import { ContentTabName } from "../data.ts";
@@ -51,23 +50,18 @@ export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
                     );
                     continue;
                 }
-                const options: string[] = [];
+
                 const system = actionData.system;
                 // update icons for any passive actions
                 if (system.actionType.value === "passive") actionData.img = getActionIcon("passive");
-                options.push(`action-type:${system.actionType.value}`);
-                options.push(...system.traits.value.map((t: string) => `trait:${t.replace(/^hb_/, "")}`));
-
-                // Prepare publication source
                 const pubSource = String(system.publication?.title ?? system.source?.value ?? "").trim();
-                const sourceSlug = sluggify(pubSource);
-                if (pubSource) {
-                    publications.add(pubSource);
-                    options.push(`source:${sourceSlug}`);
-                }
-
-                options.push(`type:${actionData.type}`);
-                options.push(`category:${system.category}`);
+                const options: string[] = [
+                    `action-type:${system.actionType.value}`,
+                    ...system.traits.value.map((t: string) => `trait:${t.replace(/^hb_/, "")}`),
+                    `type:${actionData.type}`,
+                    `category:${system.category}`,
+                    this.preparePublicationSource(pubSource, publications),
+                ];
 
                 actions.push({
                     name: actionData.name,
--- a/src/module/apps/compendium-browser/tabs/action.ts
+++ b/src/module/apps/compendium-browser/tabs/action.ts
@@ -51,33 +51,30 @@ export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
                     );
                     continue;
                 }
-                const options = new Set<string>();
+                const options: string[] = [];
                 const system = actionData.system;
                 // update icons for any passive actions
                 if (system.actionType.value === "passive") actionData.img = getActionIcon("passive");
-                options.add(`action-type:${system.actionType.value}`);
-
-                for (const trait of system.traits.value) {
-                    options.add(`trait:${trait.replace(/^hb_/, "")}`);
-                }
+                options.push(`action-type:${system.actionType.value}`);
+                options.push(...system.traits.value.map((t: string) => `trait:${t.replace(/^hb_/, "")}`));
 
                 // Prepare publication source
                 const pubSource = String(system.publication?.title ?? system.source?.value ?? "").trim();
                 const sourceSlug = sluggify(pubSource);
                 if (pubSource) {
                     publications.add(pubSource);
-                    options.add(`source:${sourceSlug}`);
+                    options.push(`source:${sourceSlug}`);
                 }
 
-                options.add(`type:${actionData.type}`);
-                options.add(`category:${system.category}`);
+                options.push(`type:${actionData.type}`);
+                options.push(`category:${system.category}`);
 
                 actions.push({
                     name: actionData.name,
                     originalName: actionData.originalName, // Added by Babele
                     img: actionData.img,
                     uuid: actionData.uuid,
-                    options,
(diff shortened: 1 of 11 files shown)


src/module/system/predication.ts
```diff
@@ -53,15 +53,15 @@ class Predicate extends Array<PredicateStatement> {
     }
 
     /** Is the provided statement true? */
-    #isTrue(statement: PredicateStatement, domain: Set<string>): boolean {
+    protected isTrue(statement: PredicateStatement, domain: Set<string>): boolean {
```
stwlam: Need an extremely good justification to open up any of this class's private methods, and I'm not seeing it here. Why can't the result be cached by the compendium browser?

In3luki: I was afraid you'd say that. Let me see if I can get it to work another way.

CarlosFdez: TBH, I"m looking at your example, and for something that occurs on user input when searching, 23ms is nothing. Are you sure it needs optimization to that degree?

In3luki: Welp, I rechecked how much slower the normal `Predicate` was and it's not much of a difference after all. It's been a few weeks so I can't remember what the exact results were the last time I tested it before doing the `CachedPredicate`.

image

I tend to get lost in micro optimizations 😅

src/module/apps/compendium-browser/tabs/base.svelte.ts
```diff
@@ -141,30 +140,75 @@ export abstract class CompendiumBrowserTab {
...
-    protected filterTraits(
-        traits: string[],
-        selected: TraitData["selected"],
-        condition: TraitData["conjunction"],
-    ): boolean {
-        const selectedTraits = selected.filter((s) => !s.not).map((s) => s.value);
-        const notTraits = selected.filter((t) => t.not).map((s) => s.value);
-        if (notTraits.some((t) => traits.includes(t))) {
-            return false;
+    /** Build a `CachedPredicate` from the applied filters */
```
CarlosFdez: Out of date comment

src/module/apps/compendium-browser/tabs/bestiary.ts
```diff
@@ -50,24 +50,34 @@ export class CompendiumBrowserBestiaryTab extends CompendiumBrowserTab {
                     );
                     continue;
                 }
-
+                const domains = new Set<string>();
```
CarlosFdez: Domains is a very misleading name in this context

In3luki: That was taken from the argument name in `Predicate##isTrue` (but should have been singular not plurar) but I can see why that could be misleading. I'll rename them to `options` unless I can think of something better. Maybe `tags`?

CarlosFdez: We call them options everywhere else, or rollOptions. Why the name change to tags? It is better than domains at least.

In3luki: Ok, they  are `options` now. I thought `tags` was also used to describe this kind of data but that might have been a DnD/Core thing. I shouldn't make these changes after just waking up.

CarlosFdez:
Fix it one more time and poke me on discord.

Is this the smallest you can get it? Are there no refactors here you can section off?

In3luki:
I may be able to do this without refactoring the options from objects to arrays. That should make the diff a bit smaller. I'll take another look over the weekend.

src/module/apps/compendium-browser/tabs/feat.ts
```diff
@@ -47,66 +47,85 @@ export class CompendiumBrowserFeatTab extends CompendiumBrowserTab {
...
+                // Prepare source
+                const pubSource = system.publication?.title ?? system.source?.value ?? "";
+                const sourceSlug = sluggify(pubSource);
+                if (pubSource) {
+                    publications.add(pubSource);
+                    options.add(`source:${sourceSlug}`);
+                }
+
+                options.add(`level:${system.level.value}`);
+                options.add(`rarity:${system.traits.rarity}`);
```
CarlosFdez: Most everything here is fine so far but this gives me pause. What is the purpose of using predicates if you end up having to add the roll options anyways?

In3luki: Using a predicate should make it easier to implement new filters without the need for custom logic. It already covers pretty much anything that a filter could need and especially exclusion is much easier.
And for adding roll options, I'm pretty sure a predicate without options wouldn't do anything? I might be misunderstanding what you mean.

CarlosFdez: It looks like you're creating options, rather than using the item's existing options. So there's potential for desync or forgetting that an option has to be added for the predication to even work.

CarlosFdez:
Maybe make the options an array? You might be able to cut down on a lot of lines of setup code if you use `options.push(...traditions.map(stuff))`.

In3luki:
Updated with the suggestion. Options are now converted to a Set as the last step. They could technically stay arrays in the filter data but the Predicate would convert them anyways on each filter operation.

The basic shape of the roll (or filter) options can't really be changed as the Predicate expects a colon as a separator for some things. The only thing I could think of would be to add a `filter:` prefix to all options to make it clear they are not normal roll options.

src/module/apps/compendium-browser/tabs/action.ts
```diff
@@ -44,34 +44,38 @@ export class CompendiumBrowserActionTab extends CompendiumBrowserTab {
...
+                // Prepare publication source
+                const pubSource = String(system.publication?.title ?? system.source?.value ?? "").trim();
+                const sourceSlug = sluggify(pubSource);
+                if (pubSource) {
+                    publications.add(pubSource);
+                    options.push(`source:${sourceSlug}`);
                 }
+
+                options.push(`type:${actionData.type}`);
+                options.push(`category:${system.category}`);
```
CarlosFdez: Can you keep the simple one liner additions together? In other words move this up to the other options.push? It saves only a bit of whitespace if you do it throughout, but right now the main wart of the new style is the manual addition of options, so if the code can be kept tight, I think it'll be fine.

You could do this as well I suppose to save on whitespace

```
const options: string[] = [
    `action-type:${system.actionType.value}`,
    ...system.traits.value.map((t: string) => `trait:${t.replace(/^hb_/, "")}`),
    `type:${actionData.type}`,
    `category:${system.category}`
];
```

Try to keep option initialization LOC tight for each object type if possible.

CarlosFdez:
> The basic shape of the roll (or filter) options can't really be changed as the Predicate expects a colon as a separator for some things. The only thing I could think of would be to add a `filter:` prefix to all options to make it clear they are not normal roll options.

Don't worry about it. While it being confused for genuine roll options is a problem, I don't think making them different is going to help anyways, nor is it enough reason on its own to deviate for the sake of deviating.

CarlosFdez (approved):
Code looks fine to me, but I plan to check for bugs personally.

src/module/apps/compendium-browser/tabs/action.ts
In3luki: Should be better now.

PR #20514, opened 2025-10-20
Convert TreasureSystemData to `TypeDataModel`

stwlam:


src/module/item/treasure/data.ts
```diff
@@ -1,35 +1,234 @@
...
+            }),
+            material: new fields.SchemaField({
+                type: new fields.StringField({
+                    required: true,
+                    nullable: true,
+                    choices: PRECIOUS_MATERIAL_TYPES.values().toArray(),
+                }),
+                grade: new fields.StringField({ required: true, nullable: true, choices: ["low", "standard", "high"] }),
+            }),
+            price: new fields.SchemaField({
```
CarlosFdez: We have a PriceField lying somewhere.

static/system.json
```diff
@@ -69,6 +69,15 @@
                 ]
             }
         },
+        "Item": {
+            "treasure": {
+                "htmlFields": [
+                    "description.value",
+                    "description.gm",
+                    "unidentified.data.description.value"
```
CarlosFdez: Should this be `identification.unidentified.data.description.value`?

CarlosFdez (approved):

CarlosFdez:
Looks good to me. I would like to one day migrate the price value object to a number when we're able to migrate data models, but we can cross that bridge when we get there.

It looks like the oldest treasure related migration is 750. While we can manually run migrations and `applySpecialKeys()` in migrateData() to convert some into data models earlier, this one doesn't look like it needs it.

PR #20515, opened 2025-10-20
Defer DC value computation for inline check

kanongil:
The main purpose of this patch is to allow `EphemeralEffect` REs to apply before the final DC is computed for inline checks using `against / defence`, eg. `@Check[athletics|against:fortitude]`.

This allows an action with a check in its description to (ephemerally) modify or create the statistic on the target before evaluation. Specifically I want to use it for the Esoteric Lore check on [Exploit Vulnerability](https://2e.aonprd.com/Actions.aspx?ID=1227). This way I can create a "level-dc" `SpecialStatistic` on it* to roll against. It is likely useful elsewhere, and this patch could even be considered a bug fix.

**Not yet, as the `baseModifier.dc` value option only takes a fixed number.*

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -1488,7 +1488,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                     domains: attackDomains,
                     origin: { actor: this, statistic: action, item: weapon },
                     target: { token: targetToken },
-                    against: "armor",
+                    against: { slug: "armor" },
                     options: new Set([...baseOptions, ...params.options]),
                     viewOnly: params.getFormula,
                     traits: actionTraits,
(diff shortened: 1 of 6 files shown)


stwlam:
For level-based DCs, I think it'd be better to simply have a math function.

kanongil:
> For level-based DCs, I think it'd be better to simply have a math function.

A math function would certainly be nice, but it doesn't materially help implement the check. I can already do level => dc using `match(when(lte(@actor.level, 20), 14 + floor(max(@actor.level, 0) * 1.345)), 40 + (min(@actor.level, 25) - 20) * 2)`.

stwlam:
The motivating reason you give isn't really suitable for the system: a level-based DC isn't a statistic, and trying to carve a way to being able to turn one into the other is a hack more suitable for a macro or module. If you want a level-based DC and aren't able to acquire one, make a PR for that.

CarlosFdez:
I agree with stwlam and I prefer the math function. That formula works but its really cumbersome. against is meant to contest against an actor, and while level based dc of the actors dc could suffice, it would prohibit level based dcs with other values or in other contexts (for example, what if its your own ability like overdrive? is it against yourself?)

EDIT: and having a math function would allow us to support proficiency without level for them.

PR #20521, opened 2025-10-22
Fix dropping actors to parties from compendium

CarlosFdez:
The changes in actor-directory.ts add the dropped actor to the party. The changes in the partial prevent the event from firing twice. So far I don't see any visual changes.

### DIFF
--- a/src/module/apps/sidebar/actor-directory.ts
+++ b/src/module/apps/sidebar/actor-directory.ts
@@ -212,21 +212,35 @@ class ActorDirectoryPF2e extends fa.sidebar.tabs.ActorDirectory<ActorPF2e<null>>
     }
 
     protected override async _handleDroppedEntry(target: HTMLElement, data: ActorSidebarDropData): Promise<void> {
-        await super._handleDroppedEntry(target, data);
+        if (!data.uuid) return super._handleDroppedEntry(target, data);
 
-        // Handle dragging members to and from parties (if relevant)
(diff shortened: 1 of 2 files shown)


src/module/apps/sidebar/actor-directory.ts
```diff
@@ -212,21 +212,35 @@ class ActorDirectoryPF2e extends fa.sidebar.tabs.ActorDirectory<ActorPF2e<null>>
...
+        const toParty = game.actors.get(toPartyId ?? "");
+        const droppedEntry = await this._getDroppedEntryFromData(data);
+
+        // If this is a entry we have to create, and we have a party, then create it here without going upstream
+        if (toParty?.isOfType("party") && !this._entryAlreadyExists(droppedEntry)) {
+            const entry = await ActorPF2e.create(droppedEntry.toObject(), { render: false });
+            if (entry?.isOfType("creature")) {
+                await toParty.addMembers(entry);
+            } else {
+                this.render();
```
stwlam: ?

CarlosFdez: To avoid issues due to the render: false above. I can check if droppedEntry is a creature before entering this code instead, but I'll have to cast the newly created actor in the addMembers() function.

CarlosFdez: Gave the change a shot. I casted since we already checked that droppedEntry was a creature.

PR #20541, opened 2025-10-24
Fix Hydra Spear base item and properties

ZzinnerGamer:
Base item, traits, price and hands were incorrect based on AoNethys page.
https://2e.aonprd.com/Equipment.aspx?ID=4004

### DIFF
--- a/packs/equipment/hydra-spear.json
+++ b/packs/equipment/hydra-spear.json
@@ -76,7 +76,8 @@
         "traits": {
             "rarity": "unique",
             "value": [
-                "magical"
+                "magical",
+                "reach"
             ]
         },
         "usage": {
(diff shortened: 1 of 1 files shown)


ammalagonc (approved):
As a rule of thumb, we don't use online resources like Archives of Nethys or Demiplane as data sources.

I checked the book and the changes you are making match, so we should be good to merge this.

packs/equipment/hydra-spear.json
```diff
@@ -76,14 +76,12 @@
         "traits": {
             "rarity": "unique",
             "value": [
-                "magical",
-                "monk",
-                "thrown-20"
+                "magical"
```
pedrogrullada:
```suggestion
                "magical",
                "reach"
```

ZzinnerGamer:
> As a rule of thumb, we don't use online resources like Archives of Nethys or Demiplane as data sources.

Oh ok, didnt knew this, pointed from now on

PR #20549, opened 2025-10-25
Always include action roll options in basic actions

### LINKED ISSUE #20516: Administer First Aid and it's sub actions don't have roll options
I inspected the rolls and look into the action list on the consoles and found out that Administer first aid and Administer First Aid: Stabilize and Administer First Aid: Stop Bleeding did not have roll options to reference.

Image

Image

Image

xyzzy42:
Action checks should have the roll option `action:${actionSlug}`, and if it is a named variant, also `action:${actionSlug}:${variantSlug}`.

Every single action with a check roll that exists now has these options, except the non-legacy administer first aid action (the legacy action has it).

Since all actions with check rolls have or should have these options, have them added automatically in the SingleCheckAction and SingleCheckActionVariant constructors.

The rollOptions field in the action data is then documented to be the additional options added beyond the standard ones.

### DIFF
--- a/src/module/actor/actions/single-check.ts
+++ b/src/module/actor/actions/single-check.ts
@@ -38,6 +38,8 @@ interface SingleCheckActionVariantData extends BaseActionVariantData {
     difficultyClass?: CheckDC | DCSlug;
     modifiers?: RawModifier[];
     notes?: SingleCheckActionRollNoteData[];
+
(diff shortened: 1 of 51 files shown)


xyzzy42:
Fixes #20516

CarlosFdez:
This looks fine to me, but I'm uncertain, so I'll wait for an extra approval (from @nikolaj-a or @stwlam) before merging it.

PR #20558, opened 2025-10-28
Transfer contents along with their container

CarlosFdez:
For purchases, it blocks purchasing containers with items in them.

### DIFF
--- a/src/module/actor/helpers.ts
+++ b/src/module/actor/helpers.ts
@@ -1018,28 +1018,37 @@ function createActorGroupUpdate(data: Partial<ActorGroupUpdate> = {}): ActorGrou
 async function applyActorGroupUpdate(
     actor: ActorPF2e,
     data: Partial<ActorGroupUpdate>,
-    { render = true }: { render?: boolean } = {},
+    { render = true, keepId }: { render?: boolean; keepId?: boolean } = {},
 ): Promise<void> {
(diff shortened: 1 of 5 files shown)


CarlosFdez:
Changed behavior for purchases. They will now block purchasing bags with items in them.

src/module/actor/helpers.ts
```diff
@@ -1018,28 +1018,37 @@ function createActorGroupUpdate(data: Partial<ActorGroupUpdate> = {}): ActorGrou
...
+        await actor.updateEmbeddedDocuments("Item", itemUpdates, { render: lastRender === "update" });
     }
     if (itemDeletes.length > 0) {
-        await actor.deleteEmbeddedDocuments("Item", itemDeletes, { render: false });
+        await actor.deleteEmbeddedDocuments("Item", itemDeletes, { render: lastRender === "delete" });
     }
-
-    const changed = !!actorUpdates || itemCreates.length || itemUpdates.length || itemDeletes.length;
-    if (changed && render) actor.render();
 }
```
stwlam, line 1052: Does this change the behavior in any way?

CarlosFdez: if there is only one person there is no change. However actor.render() is local only, so any other user looking at the actor being updated won't get the re-render. For example, if a GM is transferring items on behalf of a player.

This was likely causing a bug somewhere else as well, but I don't know where, so I'm good to avoid mentioning it.

Issue #15149, opened 2024-06-19
Vampires are unable to accept bleed damage


KaruiKage:
In the Remastered Monster Core, bleed immunity was omitted from vampires. On confirming with Paizo whether or not this was an error, I received this response from their team:
Q: " The guys and I had a quick rules question for the devs - we noticed in Monster Core, the majority of undead now have "immunity to bleed" explicitly stated in their statblock, which is great! However, we noticed that the Vampire was missing this immunity. It makes some sense, vampires need blood to "live", but we just wanted to confirm - is it intentional that vampires can take bleed damage now?"
A: "Per Logan Bonner, that is intentional!"

As it is intentional that vampires be able to take bleed damage now, they should be able to accept that condition in Foundry. The vampire trait exists, which can be used as a predicate for this exception to the undead rules.

farling42:
Issue #12472 tried to address why bleed was always included with the Undead trait.

But the Monster Core book explicitly lists "bleed" as a separate immunity for certain types of undead (e.g. skeleton guard) rather than all undead (such as Vampire).

The old Bestiary books did not explicitly list an immunity to bleed on creatures.

iDantar:
Looks like this was resolved in Spring Errata 2026:

> Page 409: The bleed damage text has been updated to be less absolute. **Since bleed immunity appears in creatures' stat blocks more consistently in the remastered rules, refer to a individual creature’s Immunities entry.**
> 
> “Another special type of physical damage is bleed damage. This is typically persistent damage, and represents loss of blood. Typically nonliving creatures and living ones that don’t need blood to live are immune to bleed. Weaknesses and resistances to physical damage apply. Bleed damage ends automatically if you’re healed to your full Hit Points.”

Could the attached PR be re-opened or maybe can I take a crack at it?

TikaelSol:
Closed by #21142

PR #20569, opened 2025-10-30
Remove bleed from being implicitly ignored by non-living actors

FalconOscuro:
Currently bleed damage is implicitly ignored by any non-living actors, post remaster this behavior no longer appears to be the case, with bleed immunity now being explicitly stated on stat blocks as noted in #15149.

Of note here is that the Skeleton ancestry and Zombie/Ghoul dedications are now no longer immune to bleed damage, due to their legacy content status, which may be undesirable when compared to similar monster stat-blocks.


stwlam:
Not accepting PRs for this, sorry.

Issue #20571, opened 2025-10-30
Cannot use strings in Battleforms Speeds


MrVauxs:
Attempting to create a Battleform Rule Element with variable speeds causes the element to silently fail, which also disallows resolving any values inside that string.

```
{"key":"BattleForm","overrides":{"speeds":{"land":"40"}}}
```

vs.

```
{"key":"BattleForm","overrides":{"speeds":{"land":40}}}
```

The example at the top does not work, without any warnings given, while the bottom example does.

https://github.com/user-attachments/assets/21b37d79-d9fc-401b-88c3-24cc6fbbad6c

CarlosFdez:
The linked PR was merged, let me know if something is wrong.

PR #20572, opened 2025-10-30
Resolve Battleform speeds

MrVauxs:
Implements #20571, resolving the speed value to its intended type.


CarlosFdez:
An issue was revealed and I had to draft on his behalf since he's not a contributor. I'll undraft once he feels its ready.

PR #20591, opened 2025-11-02
Perform forced replacement in armor-to-shield migration

stwlam:


CarlosFdez:
This fixes 899 but I'm looking ahead and 905 and 919 will not work. Those will need to see if ==system exists and perform the migration changes there instead of in system. If the repercussions of those failing isn't a big deal, perhaps its fine, let me know if that's the case.

PR #20601, opened 2025-11-03
Move NPC range migration to data model

CarlosFdez:


src/module/item/melee/data.ts
```diff
@@ -140,6 +141,21 @@ class MeleeSystemData extends ItemSystemModel<MeleePF2e, NPCAttackSystemSchema>
...
+    static override migrateData<T extends foundry.abstract.DataModel>(
+        this: ConstructorOf<T>,
+        source: Record<string, unknown>,
+    ): MeleeSystemSource {
+        const migrated = super.migrateData<MeleeSystemData>(source);
+        const rangeData = getLegacyRangeData(migrated.traits.value);
+        if (rangeData) {
+            if (!migrated.range) {
+                migrated.range = { increment: rangeData.increment, max: rangeData.max };
+            }
```
stwlam:
```suggestion
            migrated.range ??= { increment: rangeData.increment, max: rangeData.max };
```

src/module/item/melee/data.ts
```diff
@@ -1,5 +1,5 @@
 import type { ModelPropsFromSchema, SourceFromSchema } from "@common/data/fields.mjs";
-import type { MeleePF2e } from "@item";
+import { MeleePF2e } from "@item";
```
stwlam: ?

stwlam (dismissed):

src/module/item/melee/data.ts
CarlosFdez: You saw nothing

PR #20652, opened 2025-11-09
Fix invalid item `Documents` preventing all physical item updates

In3luki:
The `handleHPChange` function creates an actor clone on every physical item update even when the HP did not change. The observed error happens when the actor clone is created with an invalid item source as `Actor#toObject` still includes the source of invalid embedded documents. The function is called from `_preUpdate` so a thrown error prevents the update from going through.

### DIFF
--- a/src/module/item/physical/helpers.ts
+++ b/src/module/item/physical/helpers.ts
@@ -270,7 +270,13 @@ function handleHPChange(item: PhysicalItemPF2e, changed: DeepPartial<PhysicalIte
     const itemIndex = actorSource?.items.findIndex((i) => i._id === item._id);
     if (itemIndex === -1) return;
     actorSource?.items.splice(itemIndex ?? 0, 1, changedSource);
-    const actorClone = actorSource ? new ActorProxyPF2e(actorSource) : null;
+    const actorClone = ((): ActorPF2e | null => {
+        // Catch invalid items in the actor clone preventing other items from updating
+        try {
+            return actorSource ? new ActorProxyPF2e(actorSource) : null;
+        } catch {}
+        return null;
+    })();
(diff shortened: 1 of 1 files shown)


src/module/item/physical/helpers.ts
```diff
@@ -264,28 +264,31 @@ function handleHPChange(item: PhysicalItemPF2e, changed: DeepPartial<PhysicalIte
...
-    if (maxHPDifference !== 0) {
-        changed.system = fu.mergeObject(changed.system ?? {}, {
-            hp: { value: Math.max(item.system.hp.value + maxHPDifference, 0) },
-        });
-    }
+    // Catch invalid items in the actor clone preventing other items from updating
+    try {
+        // Get a clone of the item, through an actor clone if owned
+        const actorSource = item.actor?.toObject();
+        const changedSource = item.clone(fu.deepClone(changed), { keepId: true }).toObject();
```
stwlam: Just this part

In3luki: The error happens on line 275 when the new actor is created from the `actorSource` that may contain invalid item data.
`itemClone` on 276 is then referencing `actorClone` and `itemClone` itself is referenced on line 287.
I could make `actorClone` an `IIFE` to make the block smaller:
```ts
    const actorClone = ((): ActorPF2e | null => {
        try {
            return actorSource ? new ActorProxyPF2e(actorSource) : null;
        } catch {}
        return null;
    })();
```

Edit: Done that.

CarlosFdez (approved):

CarlosFdez:
After having credsticks block unrelated stuff when working on other branches, imma just merge this. We do need a way to trash erroring items from the inventory though, but given this is an extra thing and not the main document flow, I think avoiding explosions here will be handy.

PR #20695, opened 2025-11-11
Check if the party could roll checks they speak before short-circuiting actor selection

### LINKED ISSUE #20568: Inline rolls fail with missing statistic when spoken by a Party actor
To reproduce:
* Select the Party actor
* In chat, post `@Check[dc:10|type:ac]` (or any other `@Check` that uses a PC statistic; that AC worked amused me)
* Select a PC
* Click on the link
* Check fails with a "missing statistic" error as the button assumes the Party must be the rolling actor.

---
I keep encountering this issue when trying to post clickable checks from adventure journals on map screens

alexbrault:
Fixes #20568

### DIFF
--- a/src/scripts/ui/inline-roll-links.ts
+++ b/src/scripts/ui/inline-roll-links.ts
@@ -154,8 +154,8 @@ export class InlineRollLinks {
                 return [sheetActor];
             }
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Thanks for the PR!

PR #20723, opened 2025-11-14
Migrate several incorrect uses of strike-damage selectors

CarlosFdez:
And fixes bloodrager scaling


pedrogrullada (approved):
Data looks good to me! Thanks for catching my goof on weapon specialization.

CarlosFdez:
Not sure what goof you're referring to, if you mean using strike-damage, git blame says I was the goof.

PR #20728, opened 2025-11-14
Resolve templates and assets relative to system root path

CarlosFdez:
This is an attempt using a SYSTEM_ROOT global and a handlebars helper. While vite does support %GLOBAL_HERE% replacement in HTML files (at least for those in `import.meta.env`), it does not seem like those replacements work for static assets.

To prevent explosion during the build, an action icons helper was moved to sheet helpers.


src/global.ts
```diff
@@ -361,6 +361,7 @@ declare global {
     }
 
     const SYSTEM_ID: "pf2e" | "sf2e";
+    const SYSTEM_ROOT: `systems/${"pf2e" | "sf2e"}`;
```
stwlam:
```suggestion
    const SYSTEM_ROOT: `systems/${typeof SYSTEM_ID}`;
```

PR #20750, opened 2025-11-17
Convert add coins dialog to appv2

CarlosFdez:
image

The gap between form elements bothers me a bit but compact also removes overall padding.


static/templates/actors/add-currency-dialog.hbs
```diff
@@ -0,0 +1,20 @@
+<div class="form-group">
+    <label for="{{actor.uuid}}-add-coins-pp">{{localize "PF2E.CurrencyPP"}}:</label>
+    <input id="{{actor.uuid}}-add-coins-pp" type="number" name="pp" value="0" min="0" step="1" />
```
stwlam: The input needs to be wrapped in a `form-fields`-classed element, and make sure the DialogV2 class is applying the `standard-form` class..

static/templates/actors/add-currency-dialog.hbs
```diff
@@ -0,0 +1,28 @@
+<div class="form-group">
+    <label for="{{actor.uuid}}-add-coins-pp">{{localize "PF2E.CurrencyPP"}}</label>
+    <div class="form-fields">
+        <input id="{{actor.uuid}}-add-coins-pp" type="number" name="pp" value="0" min="0" step="1" />
```
stwlam: Use the application ID, not the actor UUID, since this form can be open multiple times.

CarlosFdez: Turns out it actually can't cause of the id in default options, but I set it to the application's id.

static/templates/actors/add-currency-dialog.hbs
CarlosFdez: It was, but now its a regular application with a form tag with it added explicitly.

src/module/actor/sheet/popups/add-currency-dialog.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    protected override async _prepareContext(options: fa.ApplicationRenderOptions): Promise<AddCurrencyDialogContext> {
+        const denominations = DENOMINATIONS.map((d) => ({
+            name: d,
+            field: new fields.NumberField({
+                min: 0,
+                integer: true,
+                label: game.i18n.localize(CONFIG.PF2E.currencies[d]),
+            }),
+        }));
```
stwlam, line 43: This already exists

CarlosFdez: Its better to do it like this because we're going to need to support credits and upb in this dialog, which don't have treasure fields. The treasure price fields days are numbered as well, once its easier to migrate such things I want to convert it to a number.

PR #20757, opened 2025-11-18
Convert distribute coins to appv2

CarlosFdez:
Need a good window title icon

image


static/templates/actors/distribute-coins.hbs
```diff
@@ -1,26 +1,19 @@
...
+        <label>
+            <input type="checkbox" name="actorIds" value="{{info.id}}" {{checked info.checked}} />
+            <span>{{info.name}}</span>
+        </label>
+    {{/each}}
+</fieldset>
+<div class="form-group">
+    <label for="{{rootId}}-breakCoins">{{localize "PF2E.loot.AllowConversion"}}</label>
+    <input type="checkbox" name="breakCoins" id="{{rootId}}-breakCoins"/>
+</div>
```
stwlam: use `formGroup` helper

CarlosFdez: Something doesn't really sit right about creating fields for non-data form props for the purposes of the formGroup helper, especially given that it pushes visual aspects away from HTML into typescript. Is it standard practice in foundry nowadays to do that?

I'll do it since I need this in anyways to continue, but its very strange.

PR #20792, opened 2025-11-24
Add support for credits in currency summary

CarlosFdez:
Primarily, this allows us to show credits in the currency summary's total coins and total wealth. This also changes the GP in coinage to lowercase, which is the more appropriate way to show it anyways.

Internally, this also attempts to remove the coins toString()'s normalize property to replace with a new "raw" pseudo-unit. I'm willing for alternatives in how to handle that, but ideally the less properties the better. 

In summary, these are all the ways we have to render coins:
- System dependent GP with 2 decimals OR credits, such 52.50 gp (used for currency summaries)
- A specific denomination usually without decimals (used for displaying currencies in the sheet)
- A system normalized value. 15 sp is 1 gp, 5 sp. 5 cp stays 5 cp. 5 credits is 5 credits (used for price displays)
- A variant system normalized value that culls "credits", currently this is "short" (used for inventories....this is what sf2e player core does when showing items, they just show a number for the credits)
- as is with no convert. 0 cp is 0 cp. 5 gp is 5 gp. (used in compendium browser filters)

The params to reproduce are the following, in order:
- `decimals: true` (default unit "primary")
- `unit: "sp"` or `unit: "pp"` or whatever
- no parameters, this is the default (default unit "primary")
- `short: true`
- `unit: "raw"`

### DIFF
--- a/src/module/actor/inventory/index.ts
+++ b/src/module/actor/inventory/index.ts
@@ -6,7 +6,7 @@ import { ItemSourcePF2e, KitSource, PhysicalItemSource } from "@item/base/data/i
 import { itemIsOfType } from "@item/helpers.ts";
 import { RawCoins } from "@item/physical/data.ts";
 import { Coins } from "@item/physical/helpers.ts";
-import { DENOMINATIONS } from "@item/physical/values.ts";
+import { DENOMINATION_RATES, DENOMINATIONS } from "@item/physical/values.ts";
 import { DelegatedCollection, ErrorPF2e, groupBy } from "@util";
 import * as R from "remeda";
 import { InventoryBulk } from "./bulk.ts";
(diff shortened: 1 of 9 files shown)


CarlosFdez:
If item prices were rendered with a different means rather than everything going through coins.toString(), that may simplify the arguments. The problem then becomes parsing.

These are specific to coin summaries only:
* System dependent GP with 2 decimals OR credits, such 52.50 gp (used for currency summaries)

These are the cases that show up related to physical items, including in the compendium browser
* A specific denomination usually without decimals (used for displaying currencies in the sheet)
* A system normalized value. 15 sp is 1 gp, 5 sp. 5 cp stays 5 cp. 5 credits is 5 credits (used for price displays)
* A variant system normalized value that culls "credits", currently this is "short" (used for inventories....this is what sf2e player core does when showing items, they just show a number for the credits)
    
These are the cases that show up in the compendium browser filters
* As is with no convert. 0 cp is 0 cp. 5 gp is 5 gp. (used in compendium browser filters)

PR #20857, opened 2025-12-01
Add the ability to use target's initiative for effect expiration

### LINKED ISSUE #18990: Feature request: New effect duration "Expire at start of next turn of affected creature"
Currently, when setting an effect that expires at the start of a creature’s turn, we must define a duration in rounds. However, there are situations where we need the effect to expire at the start of the creature’s *next* turn, regardless of how much time has passed.

### **Use Cases:**

1. **Defend Exploration Action:** A player applies the "Raise Shield" effect at the start of combat. It should expire at the start of the creature's next turn, even if no full round has passed.
2. **Custom Reminder for Reactions:** When adding a custom effect for tracking whether a creature has used their Reaction, the effect should expire at the start of their next turn, regardless of whether a full round has passed.

### **Current Limitations:**

- **Option: "Duration: 1 Round; Expire on: Start of Turn"**  
  This option does not work correctly in certain cases. If the effect is applied during a higher initiative count than when the creature will act (e.g., after a Reaction by a creature higher in initiative; or at the start of combat with the Defend action), the effect doesn’t expire because not enough time (moved to next round) has passed.
  
- **Option: "Duration: 0 Round; Expire on: Start of Turn"**  
  This causes the effect to expire at the end of the round in which it was applied, which is not useful for situations where the effect needs to persist until the start of the creature's next turn.

### **Proposal:**

Add a checkbox or toggle that, when enabled, allows the effect to expire at the start of the creature’s next turn. This would mean that the effect expires when the creature starts their next turn, whether it’s during the current round or the next round, depending on when the effect was applied. To make it applicable to more situations than only single-round duration, it could be made so it expires at the start of the creature's turn if the duration has been reduced to 1 round or less, but I'd already be very happy with a single-round duration solution.

failwiz:
изображение

This is somewhat a big change, so if it's not accepted, hopefully, someone will get inspired and makes a better attempt.
Adds a checkbox to the effect properties, that, when set, will use the target's initiative for the expiration of that effect.

Hopefully, closes #18990

Some considerations:
 - Due to the way durations are handled, every combination of initative order of target and origin except for one (target being earlier, end of the turn) requires the "next turn" to be a duration of 0; this is handled
 - This currently doesn't handle changing of target's initiative after the effect is applied. I think it's possible to handle, either per-effect before the expiry is checked _or_ on the initiative change; might not be worth it though
 - Having the ability to set the initiative of the effect's start manually would be nice, and probably would help with the previous point

### DIFF
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -62,6 +62,11 @@ class EffectSystemData extends ItemSystemModel<EffectPF2e, EffectSystemSchema> {
                     nullable: true,
                     initial: null,
                 }),
+                target: new fields.BooleanField({
+                    required: false,
+                    nullable: false,
+                    initial: false,
+                }),
                 sustained: new fields.BooleanField({ required: true, nullable: false, initial: false }),
             }),
             tokenIcon: new fields.SchemaField({
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -62,7 +62,7 @@ class EffectSystemData extends ItemSystemModel<EffectPF2e, EffectSystemSchema> {
                     nullable: true,
                     initial: null,
                 }),
-                target: new fields.BooleanField({
+                useTarget: new fields.BooleanField({
                     required: false,
                     nullable: false,
                     initial: false,
--- a/src/module/item/effect/document.ts
+++ b/src/module/item/effect/document.ts
@@ -123,8 +123,18 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
         user: fd.BaseUser,
     ): Promise<boolean | void> {
         if (this.isOwned) {
-            const initiative = this.origin?.combatant?.initiative ?? game.combat?.combatant?.initiative ?? null;
-            this._source.system.start = { value: game.time.worldTime, initiative };
+            const use_target = data.system?.duration?.target ?? false;
+            const originInitiative = this.origin?.combatant?.initiative ?? game.combat?.combatant?.initiative ?? null;
+            const targetInitiative = this.actor?.combatant?.initiative ?? null;
+            const initiative = use_target ? targetInitiative : originInitiative;
+            const isTargetBefore = (targetInitiative ?? 0) > (originInitiative ?? 0);
+            const expiry = this._source.system.duration.expiry;
+            if (use_target && expiry) {
+                if (!(expiry === "turn-end" && isTargetBefore) && this._source.system.duration.value > 0) {
+                    this._source.system.duration.value -= 1;
+                }
+            }
+            this._source.system.start = { value: game.time.worldTime, initiative: initiative };
         }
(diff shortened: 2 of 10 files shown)


src/module/item/effect/document.ts
```diff
@@ -123,8 +123,18 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
         user: fd.BaseUser,
     ): Promise<boolean | void> {
         if (this.isOwned) {
-            const initiative = this.origin?.combatant?.initiative ?? game.combat?.combatant?.initiative ?? null;
-            this._source.system.start = { value: game.time.worldTime, initiative };
+            const use_target = data.system?.duration?.target ?? false;
+            const originInitiative = this.origin?.combatant?.initiative ?? game.combat?.combatant?.initiative ?? null;
+            const targetInitiative = this.actor?.combatant?.initiative ?? null;
```
failwiz, line 128: So this is the main thing that I don't like, but afaik there's no other way to access target actor's initiative. There is supposed to be target data in data.system.context.target, but it's null. Apparently, it's not set (and I assume there's a reason for that).

src/module/item/effect/document.ts
```diff
@@ -123,8 +123,18 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
         user: fd.BaseUser,
     ): Promise<boolean | void> {
         if (this.isOwned) {
-            const initiative = this.origin?.combatant?.initiative ?? game.combat?.combatant?.initiative ?? null;
-            this._source.system.start = { value: game.time.worldTime, initiative };
+            const use_target = data.system?.duration?.target ?? false;
```
failwiz: Probably should camelize it, too.

src/module/item/effect/data.ts
```diff
@@ -62,6 +62,11 @@ class EffectSystemData extends ItemSystemModel<EffectPF2e, EffectSystemSchema> {
                     nullable: true,
                     initial: null,
                 }),
+                target: new fields.BooleanField({
```
failwiz: The name of the field should be changed to something else (`useTarget` seems like a good option), as there is another one named like this. It's one level higher, so there's no conflict, but for clarity's sake field names should be different.

JamzTheMan:
Dumb question but why do we even care about "time" for these sort of effects?

From a UX perspective, we already have a "Until encounter ends" option, could we just add a "Until next turn"option? 

On the start of an actor's combat turn, just expire/remove effects flagged as such? (not sure if there are any "Until the end of your next turn" so worst case, you add "Until end of next turn" as well or leave the "start/end" of turn dropdown.

image

Just seems that way you don't have to worry about any time math or when it was applied, if the token moved in initiative, etc. Have a separate "start of turn" check to expire these specific effects?

stwlam:
I don't think we're going to address this until V14 when there'll be some core tooling available.

PR #20873, opened 2025-12-03
Fix ancestry reach and speed resetting 0 to 5 when updating

CarlosFdez:
The range sliders already do the clamping in most practical cases.

### DIFF
--- a/src/module/item/ancestry/document.ts
+++ b/src/module/item/ancestry/document.ts
@@ -173,12 +173,13 @@ class AncestryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends 
             changed.system.hands = Math.floor(Math.clamp(changed.system.hands || 0, 0, 12) / 2) * 2;
         }
 
-        for (const fieldName of ["hp", "speed", "reach"] as const) {
-            if (changed.system[fieldName] !== undefined) {
-                const minimum = fieldName === "speed" ? 5 : 1;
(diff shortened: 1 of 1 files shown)


src/module/item/ancestry/document.ts
```diff
@@ -175,8 +175,7 @@ class AncestryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends
 
         for (const fieldName of ["hp", "speed", "reach"] as const) {
             if (changed.system[fieldName] !== undefined) {
-                const minimum = fieldName === "speed" ? 5 : 1;
-                const value = Math.floor(Math.clamp(Number(changed.system[fieldName]) || 0, minimum, 99));
+                const value = Math.floor(Math.clamp(Number(changed.system[fieldName]) || 0, 0, 99));
```
stwlam: This probably ought to be broken out more:
hp: increments of 2 between 0 and 12
speed: increments of 5 between 0 and 100
reach: same as speed

CarlosFdez: Perhaps, but I'm ok with leaving it to the sheet inputs range sliders. These clamps are really only for people calling update manually. Honestly we might as well remove the clamps or just coerce to a number, we don't really fine tune convert other properties in other documents.

But fine sure.

stwlam: Eventually a `NumberField` will take care of it for us!

PR #20894, opened 2025-12-05
Condense Living Fortification effect and make it a self-applied effect

pedrogrullada:


static/lang/re-en.json
```diff
@@ -4671,6 +4671,9 @@
                 "FleshWound": {
                     "ToggleLabel": "Flesh Wound — Reduce Strike damage"
                 },
+                "LivingFortification": {
+                    "Prompt": "Does your drawn weapon have the parry trait?"
```
TikaelSol: We already have a general prompt for this. `PF2E.SpecificRule.Parry.Prompt`

pedrogrullada: Yeah, I know. Didn't use that prompt because it refers to a weapon you're wielding, whereas this one is about the specific weapon you drew as part of it.

But I'll go with the generic one then.

TikaelSol: We could change it to `Does your weapon have the parry trait?` instead and make it work for both.

PR #20899, opened 2025-12-05
Update dragon instinct with Draconic Codex dragons

### LINKED ISSUE #20898: Dragon Instinct Missing Draconic Codex options
Dragon / Draconic Instinct has not been updated with Draconic Codex options from page 207.

pedrogrullada:
- Closes #20898

Worth noting that the draconic benefactor options are not explicitly called out as being available for the dragon instinct. This is all making some assumptions based on these two paragraphs.

From Draconic Codex (page 206)
> The table below lists all of the dragons included in Draconic Codex and can be used to determine the benefits of your draconic connection or the benefits you gain when using dragon form **and similar effects**. If you choose a dragon from a different source (such as Monster Core 2), work with your GM to determine if you have access to it and what options that dragon grants you.

From Player Core 2, page 74
> You summon the fury of a mighty dragon and manifest incredible abilities. Perhaps your culture reveres draconic majesty, or you gained insights by drinking or bathing in dragon’s blood or watching a marauding wyrm burn your village. Select a type of dragon from the Dragon Instincts table as your instinct’s dragon type. These are the dragons from Monster Core, **but your GM might allow you to choose dragons from other sources and determine their tradition and dragon breath type.**

emphasis mine in both cases. Might be worth waiting on PFS clarifications before going through with this.


MrVauxs:
> Worth noting that the draconic benefactor options are not explicitly called out as being available for the dragon instinct. This is all making some assumptions based on these two paragraphs.

Thought so as well initially but it is mentioned in the first sentence of Draconic Connections paragraph.

> Some abilities reference your draconic benefactor, draconic exemplar, or a **draconic instinct**, which is a specific type of dragon from which you draw your draconic powers—typically the type of dragon you serve or whose blood runs in your veins.

And so goes the "similar effects" paragraph.

pedrogrullada:
Ah, they used the wrong name for it

MrVauxs:
> Ah, they used the wrong name for it

Yeah, that was the exact thing that caught me off guard lol, searching for Dragon Instinct and it bringing nothing.

PR #20939, opened 2025-12-09
Convert Craft and Repair Action Macros to Appv2 and Svelte

### LINKED ISSUE #20915: Craft and Repair macros - Button labels disappear in Dark theming for Application, X Close button opens roll dialog
Noticed these couple of issues. Happens with no modules, in both Chrome and Firefox. Weirdly, the Craft button is visible in Dark theme up until an item is added to the window

Foundry Virtual Tabletop: Version 13 Stable, 13.351
Game System: pf2e, 7.8.0

Video was too big so some screenshots instead. 

Image

Image

Image

Image

jfn4th:
Resolves #20915

Tried my best to follow the patterns I've seen in other AppV2 commits.
Screenshot 2025-12-09 at 2 45 50 PM
Screenshot 2025-12-09 at 2 46 04 PM
Screenshot 2025-12-09 at 2 46 15 PM
Screenshot 2025-12-09 at 2 46 24 PM


src/module/system/action-macros/crafting/select-item.svelte
```diff
@@ -0,0 +1,110 @@
...
+    async function handleDrop(event: DragEvent) {
+        const dataString = event.dataTransfer?.getData("text/plain");
+        const dropData: DropCanvasItemData | undefined = JSON.parse(dataString ?? "");
+
+        if (dropData?.type !== "Item") {
+            ui.notifications.error(game.i18n.localize(`PF2E.Actions.${actionKey}.Error.ItemReferenceMismatch`));
+            return;
+        }
+
+        const droppedItem = await ItemPF2e.fromDropData(dropData);
```
CarlosFdez: If you do `const droppedItem = dropData?.type === "Item" ? await ItemPF2e.fromDropData(dropData) : null;` and then check droppedItem, you might be able remove the duplicate error message.

src/module/system/action-macros/crafting/select-item.svelte
```diff
@@ -0,0 +1,110 @@
...
+        updateSelection(droppedItem);
+    }
+
+    function handleCancel() {
+        updateSelection(null);
+        resolve(null);
+    }
+
+    function handleSelect() {
+        resolve(data.item);
```
CarlosFdez: I think you don't need these methods, and can maybe do `onclick={() => resolve(data.item)}` directly. resolve can also change the selection immediately.

CarlosFdez:
Thank you for this. Can you post some screenshots of what it looks like into the main post?

jfn4th:
@CarlosFdez Updated with screenshots.

src/module/system/action-macros/crafting/select-item.svelte
jfn4th: Nope, no need to handle null!

CarlosFdez:
Thanks! It looks perfectly fine. That said I think if you close the dialog with X, it shouldn't resolve with the current selection. In a pick a thing it does so intentionally, but this seems different. This means that selection shouldn't be set until resolve, and the actual selection in the prompt could be a local state variable.

jfn4th:
Correct, that's actually the current behavior, hence not updating `this.selection` in `updateSelection` and only updating the local state. `selection` defaults to null in the app and isn't updated until `resolve` is called in svelte. Pressing Craft or Repair will pass the item to `resolve`, Cancel will pass `null`, and X will trigger `_onClose`, resolving with the default of `null` (the macro handles things from there).

CarlosFdez:
Gotcha. Well, I think you can get rid of updateSelection then and replace it for a `let selection: PhysicalItemPF2e | null = $state(null);` in the svelte. Keeps the application handling a bit thinner. Updating the application state directly is also a bit risky since re-renders overwrite the state, though that wouldn't be a problem for this particular one.

CarlosFdez (dismissed):
Looks fine to me. We may want to do a minor release soon (we'll see) but I'll approve it and merge it in and a later date unless someone finds an issue with it.

Thanks for the PR.

CarlosFdez:
Actually wait found one more thing. select-item-dialog content class?

jfn4th:
Oops, should it be `standard-form` or omitted?

CarlosFdez:
Omitted if you don't need it. content classes is only required if there's some top level css you need to set that can't be set in svelte.

CarlosFdez (approved):

PR #20940, opened 2025-12-09
Add heightened effect for See the Unseen

DocSchlock:
See the Unseen was missing its 8 hour version like Tailwind has.


pedrogrullada:
There's conflicts because of the recent compendium split. You need to rebase.

PR #20941, opened 2025-12-09
Cleanup Holy Host and Unholy Army and add effects

DocSchlock:
This:
Adds in-lines to Unholy Army like Holy Host has.
Adds effects for the bonuses certain Holy Host choices provide.


pedrogrullada:
There's conflicts because of the recent compendium split. You need to rebase.

PR #20942, opened 2025-12-10
Delete and redirect additional actors to their Monster Core 2 reprints

ammalagonc:


TikaelSol:
Can't blame me for these conflicts

PR #20945, opened 2025-12-10
Flip multiplier of held ctrl/shift keys in Trade Dialog to match actor sheet inventory

pedrogrullada:
Inventory multiplies by 5 with shift and 10 with ctrl, figured this should match


CarlosFdez:
👍

PR #20955, opened 2025-12-12
Correct slugs in Dragon Spit Choice Set to include Ignition

### LINKED ISSUE #20951: Dragon Spit Selections Missing
As title says, the "Dragon Spit" Ancestry Feat is missing an option. Specifically Ignition (previously named produce flame prior to remaster). 
I have attached an image as evidence. This is in System Version 7.8.0 Pathfinder Second Edition and was present in previous versions too.
Image

ammalagonc:
- Closes #20951

### DIFF
--- a/packs/pf2e/feats/ancestry/human/dragon-spit.json
+++ b/packs/pf2e/feats/ancestry/human/dragon-spit.json
@@ -37,7 +37,7 @@
                             "or": [
                                 "item:slug:acid-splash",
                                 "item:slug:electric-arc",
-                                "item:slug:produce-flame",
+                                "item:slug:ignition",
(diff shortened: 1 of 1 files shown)


packs/pf2e/feats/ancestry/human/dragon-spit.json
```diff
@@ -37,7 +37,7 @@
                             "or": [
                                 "item:slug:acid-splash",
                                 "item:slug:electric-arc",
-                                "item:slug:produce-flame",
+                                "item:slug:ignition",
```
simonward, line 40: Should we include `"item:slug:produce-flame"` as wel? Might be desired if using the pre-remaster content.

PR #20965, opened 2025-12-12
Link Share Hunter's Edge effects and add edge-sharing feat description addenda in Hunt Prey action

pedrogrullada:
It gets quite long at high levels, so I've localized shortened, fluff-free versions of the descriptions to make it more bearable. Worth noting that, as of now, these alterations don't show up on self-applied effects. But it should help with getting easier access to these effects normally linked within their own feats.

image


TikaelSol:
Sorry, should have gone on a merging spree before moving all the feats

PR #20969, opened 2025-12-14
Add effect for Flensing Slice

pedrogrullada:


static/lang/re-en.json
```diff
@@ -5623,6 +5623,7 @@
                 "Color": "Select a Color.",
                 "ConsciousMind": "Select a conscious mind.",
                 "CreatureSize": "Select a creature size.",
+                "DamageDice": "Select a number of damage dice.",
```
```suggestion
                "DamageDice": "Select a number of damage dice.",
```
Maybe WeaponDamageDice and "number of weapon damage dice" would be better? Since it's more clear what is meant.

pedrogrullada: Yeah, that makes sense. Will get that done when I rebase.

PR #20970, opened 2025-12-14
Convert Generate Check Prompt Macro to AppV2/Svelte

failwiz:
old app
new app 
results

 - Converted app to v2 using Svelte
 - Small changes to make the tag select fields work with Svelecte
 - Also add the checkbox to copy the check to clipboard instead of posting it to chat
 - Some reshuffling of the fields; check traits now also apply to saves
 - Besides that, old functionality is preserved

Thanks everyone who helped me over on discord!
Old appv1 app is preserved - I can either leave it as is, or fully replace it with the new one. Of course, any suggestions for changes are welcome.


stwlam:
You can go ahead and drop the V1 app as part of the PR. We don't keep those around without good reason.

stwlam:
Mind posting side-by-side screenshots?

failwiz:
right, forgot about linting too

failwiz:
I was made aware the resulting string is wrapped in `<p></p>` tags, which probably would be annoying to remove manually every time if you use the copy to clipboard feature.
Also, the original app is a bit old, and  predates the implementation of the inline action links syntax(I think). I actually had a PR for that (#15463), but it's not really good, and I would like to re-do it on this converted app. It probably would be a bit easier to handle, too.
Other things that could be improved in a later PR are:
 - adding ability to set stuff like `against:` options to traits
 - same to DCs - while not very useful for journals and chat links, could be helpful for generating inlines for items
 - having a preview of the resulting string - either updating on changing the data, or shown manually on generating.

Most of those things probably are better left for future PRs, but I could just keep adding to this one.

failwiz:
изображение

Refined the copy/post interaction - now there's preview for every inline check (if you have several skills/saves selected). Each one can be copied to clipboard.
Posting uses the old logic and sends all the checks in one message.

I can either add a commit here, or hold it until/if this PR is accepted.

failwiz:
Yeah, I think I'm remaking that after the dust settles, sorry about the mess.

PR #20973, opened 2025-12-14
Remove improper `select-item` id override

jfn4th:
Fixes an issue from #20939. While peaking at some other things that could be converted to AppV2, I learned the id getter should not be overridden, instead should be configured in `_initializeApplicationOptions`. In this case there isn't really a need to ensure there's only one instance of the window though.

### DIFF
--- a/src/module/system/action-macros/crafting/select-item.ts
+++ b/src/module/system/action-macros/crafting/select-item.ts
@@ -21,6 +21,7 @@ class SelectItemDialog extends SvelteApplicationMixin(fa.api.ApplicationV2) {
     }
 
     static override DEFAULT_OPTIONS: DeepPartial<SelectItemConfiguration> = {
+        id: "select-item-dialog",
         position: { width: 270 },
         window: {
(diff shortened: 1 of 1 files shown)


stwlam:
That ID doesn't ensure a single instance of it but one per action. We should probably have one total, which means setting a static `id` in `DEFAULT_OPTIONS`.

jfn4th:
Sure, I can push that up in a minute. Why only one and not one per action though? I'm curious if there's a rule of thumb behind when to restrict things to only one instance, vs one per actor, etc. since it isn't always consistent across apps.

CarlosFdez:
There's no rule of thumb, just one out of practicality I think. One ever is simpler and less code, so we only do something more complicated if there is a perceived value. One per action doesn't seem to add value, you're only going to be one crafting thing at a time. Whereas a once per actor or once per item might allow comparing things (such as if preparing spellcasting entries).

More of an art than a science, with a preference towards simpler.

PR #20982, opened 2025-12-16
Add an Archetypes and Dedications page and link it to the GM and Player Screen journals

ammalagonc:
image


Cuingamehtar:
Is there a need for it? There is already a searchable TOC on the left that is alphabetically ordered.

ammalagonc:
It creates one page that can be linked to both the GM and Player Screen journal entries. 

Initially, I was going to link to just the Rules page and call it day, but then that didn't  look right.

PR #20983, opened 2025-12-16
Fix sf2e image paths

CarlosFdez:


packs/sf2e/equipment/zero-pistol.json
pedrogrullada, line 1: Innocent bystander here. Did you mean to delete this?

CarlosFdez: yup! We had two, and the one tikael added wins out (there's also the extra ownership stuff in there)

PR #20991, opened 2025-12-17
Fix NPCs and characters with 0 base speed increasing to 5.

CarlosFdez:
Fixes an issue reported in the bugs channel. Creatures with 0 land speed are buffed to 5 land speed. Also, the clamping did not apply to rule elements, allow penalties to reduce speed below 5 regardless.


stwlam:
Penalties can reduce speed below 5; just not armor speed penalty. Is the idea to address that after?

CarlosFdez:
Actually, I assumed it was an overall min, not just to armor. I reverted the clamp on the final computed value.

PR #20995, opened 2025-12-17
Correct Action Cost of Nizca's Channel Smite

ammalagonc:


packs/pf2e/shades-of-blood-bestiary/book-3-to-blot-out-the-sun/nizca-iricol.json
simonward:
```suggestion
                    "value": "<p>Nizca siphons death through a melee attack and into their foe. Nizca expends a @UUID[Compendium.pf2e.spells-srd.Item.Harm] spell and makes a melee Strike. On a hit, they cast the 1-action version of the expended spell to damage the target, in addition to the normal damage from the Strike. The target automatically gets a failure on its save (or a critical failure if Nizca's Strike was a critical hit). The spell doesn't have the manipulate trait when cast this way.</p>"
```

PR #21018, opened 2025-12-20
Restructure Actions compendium to add folders

ammalagonc:
Also removed a bunch of Animal Companion actions and one unneeded that's already an Ancestry Feat


TikaelSol:
What's with the deletions?

ammalagonc:
They're Animal Companion actions that already exist in the Animal Compendia module, and the other one is a duplicate of an existing Ancestry feat. 

For some odd reason, all of the AC actions from Book of the Dead got added into the system.

PR #21036, opened 2025-12-21
Convert Attribute Builder to AppV2 and Svelte

jfn4th:
On top of conversion, this fixes a few things from the AppV1 version

- Builder updates when ABP setting is changed (AppV1 version would require a page refresh if setting was changed after the first time the builder was opened for a character)
- `text-overflow: ellipses` styles work now
- Adds `aria` attributes throughout
- Scrollable
- Fully responsive to changes in core font size setting

The `AttributeButton` component is also made to be usable with the Kingdom attribute builder when that gets converted to Svelte.

v2dark
v2light


CarlosFdez:
As weird as it is, I'm wondering if it should force light mode because the character sheet itself is light mode, and then do a swap once the character sheet itself is light mode. I'd like a second opinion on this thought though, I'm hesitant about it.

jfn4th:
I'd argue you'll often have just come from using the ABC pickers, which already have dark mode support, so why force this one? Dark mode users are used to support being a WIP and will take anything they can get.

CarlosFdez:
that is fair, though those are smaller dialogs and this is almost a sheet. Even once we have dark mode character sheets, I think this should match whatever mode the character sheet is set to on account of how big it is. Its possible to override dark/light mode on a sheet by sheet basis.

I guess beyond that, I don't think that dark mode variant looks nice. There's something jarring about it. Maybe its the area on the bottom left (and the contrast with the checkbox), maybe its the lack of transparency on the background, maybe its the way the faded out buttons do. Something is off but I can't put my finger on what.

EDIT: Light mode too actually, the hr looks different.

jfn4th:
There is transparency, I just didn't have a scene loaded or anything behind it.

The `<hr>` change actually comes from core appv2 styles

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+
+/** Number of ability scores in the game (str, dex, con, int, wis, cha) */
+const ALL_ATTRIBUTES_COUNT = 6;
+/** Maximum alternate ancestry boosts allowed */
+const MAX_ALTERNATE_ANCESTRY_BOOSTS = 2;
+/** Maximum voluntary flaws in legacy mode */
+const MAX_VOLUNTARY_FLAWS_LEGACY = 2;
+/** Threshold at which boosts become "partial" (require 2 boosts per +1) */
+const PARTIAL_BOOST_THRESHOLD = 5;
+/** Minimum allowed manual ability modifier */
```
stwlam:
```suggestion
/** Minimum allowed manual attribute modifier */
```

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+const MODIFIER_MIN = -5;
+/** Maximum allowed manual ability modifier */
```
stwlam:
```suggestion
/** Maximum allowed manual attribute modifier */
```

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+const MODIFIER_MAX = 10;
```
stwlam: Let's make these static properties since there's no use of them as project-wide importable constants

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+        options: Partial<AttributeBuilderConfiguration>,
+    ): AttributeBuilderConfiguration {
+        const initialized = super._initializeApplicationOptions(options) as AttributeBuilderConfiguration;
+        initialized.uniqueId = `attribute-builder-${initialized.actor.uuid}`;
+        return initialized;
+    }
+
+    override get title(): string {
+        return game.i18n.localize("PF2E.Actor.Character.Attribute.Boosts");
+    }
```
stwlam: Set this in DEFAULT_OPTIONS without localizing

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+        return game.pf2e.variantRules.AutomaticBonusProgression.isEnabled(this.#actor);
+    }
+
+    protected override _initializeApplicationOptions(
+        options: Partial<AttributeBuilderConfiguration>,
+    ): AttributeBuilderConfiguration {
+        const initialized = super._initializeApplicationOptions(options) as AttributeBuilderConfiguration;
+        initialized.uniqueId = `attribute-builder-${initialized.actor.uuid}`;
+        return initialized;
+    }
```
stwlam: Go with a constant id instead

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -0,0 +1,286 @@
...
+        options: fa.ApplicationRenderOptions,
+    ): Promise<void> {
+        await super._onFirstRender(context, options);
+        this.options.actor.apps[this.id] = this;
+    }
+
+    protected override _onClose(options: fa.ApplicationClosingOptions): void {
+        delete this.options.actor.apps[this.id];
+        super._onClose(options);
+    }
```
stwlam: Move to `_tearDown`

src/module/apps/attribute-builder/helpers.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/** Creates a record of button states keyed by attribute */
+function createButtonRecord<TAttr extends BuilderAttribute, TState extends BoostFlawState>(
+    attributes: readonly TAttr[],
+    builder: (attr: TAttr) => TState,
+): Record<TAttr, TState> {
+    return Object.fromEntries(attributes.map((attr) => [attr, builder(attr)])) as Record<TAttr, TState>;
+}
+
+export { createButtonRecord, type BoostFlawState, type BuilderAttribute, type BuilderButton };
```
stwlam:
```suggestion
export { createButtonRecord };
export type { BoostFlawState, BuilderAttribute, BuilderButton };
```

src/module/actor/character/apps/attribute-builder/app.svelte
```diff
@@ -0,0 +1,943 @@
...
+        };
+    });
+
+    const legacyFlaws = $derived(data.ancestry?.system.voluntary?.boost !== undefined);
+
+    const voluntaryFlaws = $derived.by(() => {
+        const ancestry = data.ancestry;
+        if (!ancestry) {
+            return null;
+        }
```
stwlam:
```suggestion
        if (!ancestry) return null;
```
tariffs on whitespace have sent prices through the roof :(

src/module/actor/character/apps/attribute-builder/app.svelte
```diff
@@ -0,0 +1,943 @@
...
+            voluntaryBoostsRemaining: 0,
+            labels: getBoostFlawLabels(ancestry.system.flaws),
+        };
+    });
+
+    const backgroundBoosts = $derived.by(() => {
+        const background = data.background;
+        if (!background) {
+            return null;
+        }
```
stwlam:
```suggestion
        if (!background) return null;
```

stwlam (changes requested):

jfn4th:
I tweaked the contrast on the manual entry checkbox a little bit, and tweaked the dark mode red a bit.
Screenshot 2025-12-27 at 12 51 58 PM

stwlam:
Try the notification colors for red/green. For dark mode in general, don't bother making it look unique to satisfy idiosyncratic tastes: just do the best you can with bog-standard core styling.

jfn4th:
With the following, things look mostly fine.
```css
--color-boost: var(--color-level-success-border);
--color-boost-dark: color-mix(in srgb, var(--color-level-success-border) 70%, black);
--color-flaw: var(--color-level-error);
--color-flaw-dark: color-mix(in srgb, var(--color-level-error) 70%, black);
--button-color: var(--color-boost);
--button-locked-color: var(--color-boost-dark)
```

I ended up going with the border color for green since the standard success was pretty bright to my eyes in light mode, but can switch it if you'd like. Also tried using `--color-level-success-bg` and `--color-level-error-bg` as darker shades, but those have transparency and would require defining border styles.

Light mode using `color-level-success-border`
light-builder-border

Light mode using `color-level-success`
light-builder

Dark mode using `color-level-success-border`
dark-builder-border

Dark mode using `color-level-success`
dark-builder

stwlam:
Did the button color change for light mode as well?

jfn4th:
Are you referring to the screenshots? There is a difference between them in light mode. Here's a more focused pic
`color-level-success-border`
Screenshot 2026-01-20 at 6 01 21 PM

`color-level-sucess`
Screenshot 2026-01-20 at 6 00 58 PM

stwlam (approved):

PR #21038, opened 2025-12-22
Add sf2e pack list and support for copying items from pf2e

CarlosFdez:
This will cause bad pack reference errors in the build, but that's because the folders don't exist yet. They will be resolved once we have actual data in here.


build/conditions.ts
```diff
@@ -16,13 +16,17 @@ const args = argv
...
-    CompendiumPack.loadJSON("conditions", options).finalizeAll(),
-    CompendiumPack.loadJSON("campaign-effects", options)
-        .finalizeAll()
-        .filter((e) => "type" in e && e.type === "condition"),
-].flat();
-console.log(JSON.stringify(conditions));
+const conditions = [CompendiumPack.loadJSON("conditions", options).finalizeAll()];
+if (options.systemId === "pf2e") {
+    conditions.push(
+        CompendiumPack.loadJSON("campaign-effects", options)
```
stwlam:
```suggestion
        ...CompendiumPack.loadJSON("campaign-effects", options)
```

static/icons/default-icons/feats-sf2e.webp
stwlam, line 1: Can this be the core standard of 256x256?

stwlam (dismissed):

build/conditions.ts
CarlosFdez: It flattens at the end, but we can ... in both sure.

PR #21043, opened 2025-12-24
Standardize script name and fix hot with no params

CarlosFdez:
Fixes running sf2e. I chose main since we already use that for the main scss file, and "system" as a filename has additional meanings in a foundry context.

It also fixes `npm run hot` defaulting to sf2e instead of pf2e.


stwlam:
What additional meaning in a Foundry context?

PR #21053, opened 2025-12-26
Add automation to Underhanded Assault  stealth penalty via a rolloption

failwiz:


packs/pf2e/feats/class/rogue/level-2/underhanded-assault.json
```diff
@@ -12,7 +12,7 @@
         },
         "category": "class",
         "description": {
-            "value": "<p>You capitalize on the distraction provided by an ally. You @UUID[Compendium.pf2e.actionspf2e.Item.Sneak] up to a foe adjacent to one of your allies. You can roll against the foe you're Sneaking up on, even if it's currently observing you, as though you were @UUID[Compendium.pf2e.conditionitems.Item.Hidden]. You take a –2 penalty on your Stealth check. If your Stealth check against the chosen foe succeeds, you can make a melee Strike against that foe at the end of your Sneak.</p>"
+            "value": "<p>You capitalize on the distraction provided by an ally. You [[/act sneak]] up to a foe adjacent to one of your allies. You can roll against the foe you're Sneaking up on, even if it's currently observing you, as though you were @UUID[Compendium.pf2e.conditionitems.Item.Hidden]. You take a –2 penalty on your Stealth check. If your Stealth check against the chosen foe succeeds, you can make a melee Strike against that foe at the end of your Sneak.</p>"
```
simonward: Consider adding the RollOption to the `act` rather than a toggle?

failwiz: Is there an established guideline on the inlines vs checkbox? I went through some actions and they mostly use the toggle for similar automation.
I personally would prefer using the inline, but I feel like having the inline and the toggle is a good compromise to facilitate both usage styles - player can turn the option on in the pre-roll modifier window, without actually using the checkbox.

simonward: In general, we’re moving away from toggles where it makes sense in order to reduce character sheet clutter. Players can always add a toggle to their copy if they want one.

ammalagonc: Not an official guideline, but I tend to follow:
- Is the ability/feat something that can be done in a variety of situations? The RollOption toggle.
- Is the ability/feat used for a specific situation/scenario? Then inline RollOption. 

In this case, the ability/feat has a specific condition that needs to be met, so I would be in favor of doing an inline RollOption.

packs/pf2e/feats/class/rogue/level-2/underhanded-assault.json
```diff
@@ -29,7 +29,18 @@
             "remaster": true,
             "title": "Pathfinder Player Core"
         },
-        "rules": [],
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "underhanded-assault"
```
simonward: Could include “action:sneak” in the predicate.

packs/pf2e/feats/class/rogue/level-2/underhanded-assault.json
```diff
@@ -29,7 +29,18 @@
...
+                ],
+                "selector": [
+                    "stealth-check"
```
simonward: Consider a selector of “skill-check”, since the description doesn’t require using stealth to sneak?

failwiz: As far as I can tell, the only specifically noted instance of using other skills to Sneak is an AP-specific dedication (Verduran Shadow). Other than that, RAW Sneak is used with stealth - and Underhanded Assault doesn't specify otherwise.
Don't see the harm in changing that, though.

PR #21094, opened 2025-12-29
Rename SF2E traits

CarlosFdez:
Since we're including SF2E traits in the PF2e build.


CarlosFdez:
I removed the "installed in X" trait descriptions since that's usage data, but is expend-x in use somewhere? If so, by what?

PR #21098, opened 2025-12-29
Use core color vars in `svelecte-pf2e' and abc picker

### LINKED ISSUE #21095: Svelecte dropdown on light theme is slightly hard to read
Image

Affects every app that uses it (Pick a Thing, trait filter in Compendium browser, my in-progress conversion of check prompt macro).

jfn4th:
Closes #21095 

Also adds functionality to sync the detached dropdown theme with changes to foundry theme without reopening apps.

Before:
Screenshot 2025-12-29 at 1 19 16 PM
Screenshot 2025-12-29 at 1 19 44 PM

After:
Screenshot 2025-12-29 at 1 19 26 PM
Screenshot 2025-12-29 at 1 20 03 PM

### DIFF
--- a/src/module/sheet/components/svelecte-pf2e.svelte
+++ b/src/module/sheet/components/svelecte-pf2e.svelte
@@ -3,15 +3,46 @@
     type SvelecteParams = Parameters<Required<Svelecte>["$set"]>[0];
     let { value = $bindable(), ...props }: SvelecteParams = $props();
 
-    // Svelecte position resolver that puts it on the body
-    function popoutPositionResolver(node: HTMLElement) {
-        // Add classes to element, including theme
-        const themed = node.closest(".themed, body");
-        const theme = [...(themed?.classList ?? [])].find((c) => c.startsWith("theme-"));
-        node.classList.add("detached", "pf2e");
+    // Track the detached dropdown node and its theme source for syncing
+    let dropdownNode: HTMLElement | null = $state(null);
+    let themeSource: HTMLElement | null = $state(null);
+
+    // Sync theme classes from the theme source to the dropdown
+    function syncTheme(node: HTMLElement, source: HTMLElement) {
+        const theme = [...source.classList].find((c) => c.startsWith("theme-"));
+        // Remove any existing theme classes
+        for (const cls of [...node.classList]) {
+            if (cls.startsWith("theme-")) {
+                node.classList.remove(cls);
+            }
+        }
+        // Add current theme
         if (theme) {
             node.classList.add("themed", theme);
+        } else {
+            node.classList.remove("themed");
         }
+    }
+
+    // Watch for theme changes on the theme source and sync to dropdown
+    $effect(() => {
+        if (!dropdownNode || !themeSource) return;
+
+        syncTheme(dropdownNode, themeSource);
+
+        const observer = new MutationObserver(() => syncTheme(dropdownNode!, themeSource!));
+        observer.observe(themeSource, { attributes: true, attributeFilter: ["class"] });
+
+        return () => observer.disconnect();
+    });
+
+    // Svelecte position resolver that puts it on the body
+    function popoutPositionResolver(node: HTMLElement) {
+        // Find the closest themed ancestor before detaching, or fall back to body
+        themeSource = node.closest<HTMLElement>(".themed") ?? document.body;
+
+        node.classList.add("detached", "pf2e");
+        dropdownNode = node;
 
         let destroyed = false;
         const selectElement = node.parentElement;
@@ -35,13 +66,15 @@
         return {
             destroy: () => {
                 destroyed = false;
+                dropdownNode = null;
+                themeSource = null;
                 selectElement?.appendChild(node);
             },
         };
     }
 
     const positionResolver = $derived(props.positionResolver ?? popoutPositionResolver);
-    const className = $derived((props.class ?? "" + " pf2e").trim());
+    const className = $derived(((props.class ?? "") + " pf2e").trim());
 </script>
 
 <Svelecte {...props} bind:value class={className} {positionResolver} />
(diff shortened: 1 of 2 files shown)


stwlam:
I saw this issue and was surprised to see all the CSS defined in `:global`. What makes it such that it can't be scoped to the component?

CarlosFdez:
I can answer this one. Because of the portal (the positionResolver), the dropdown is re-attached to the global scope. Svelecte no longer can access it with its own css solution when that happens.

src/module/sheet/components/svelecte-pf2e.svelte
```diff
@@ -104,6 +137,7 @@
         body > .sv_dropdown.pf2e.detached {
             min-width: 0;
             z-index: 5000 !important;
+            color: var(--sv-color);
```
stwlam: I think these theme variables are mostly unnecessary: there are core ones that already switch by theme: see, e.g., `--color-text-primary`.

jfn4th:
Yeah, the dropdown is appended to `document.body`, outside of the component's style tree.

CarlosFdez:
How well does this work though?

```css
.svelecte.pf2e,
:global(.sv_dropdown.pf2e) {
    :global {
        blah
    }
}
```

That might end up being clearer in intent, if messy.

src/module/sheet/components/svelecte-pf2e.svelte
CarlosFdez: If I remember right the theme variables are used internally (which is why I set them to the core ones, though they look like they need updates), but something has gone wrong if you need to set something to one of the svelecte internal theme variables.

src/module/sheet/components/svelecte-pf2e.svelte
```diff
@@ -94,7 +128,6 @@
         body.theme-dark .application:not(.themed) .svelecte.pf2e,
         .themed.theme-dark .svelecte.pf2e,
         .themed.theme-dark.sv_dropdown.pf2e {
-            --sv-color: var(--color-light-3);
             --sv-control-bg: var(--color-cool-4);
```
CarlosFdez: For these dark mode variants, if you can't find a good theme sensitive css variable in core, try swapping it to using `light-dark`. This is something that's sensitive to light/dark mode settings, and seems to be sensitive to something the theme vars do (color-scheme css var likely).

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/light-dark

jfn4th:
> How well does this work though?
> 
> ```css
> .svelecte.pf2e,
> :global(.sv_dropdown.pf2e) {
>     :global {
>         blah
>     }
> }
> ```
> 
> That might end up being clearer in intent, if messy.

You can't nest a :global { ... } block inside a selector with declarations. Instead, you'd have to use :global() on individual nested selectors, like

```css
.svelecte.pf2e,
:global(.sv_dropdown.pf2e) {
    blash;
    
    :global(.sv-input--text) {
        width: auto;
    }
    
    :global(button) {
        height: unset;
    }
}
```

That won't actually work either though, as  `.svelecte.pf2e` (without `:global`) will get scoped by Svelte, so really we end up with

```css
:global(.svelecte.pf2e),
:global(.sv_dropdown.pf2e) {
    blah;
    
    :global(.sv-input--text) {
        width: auto;
    }
}
```

which isn't really better

src/module/sheet/components/svelecte-pf2e.svelte
jfn4th: I'm familiar with `light-dark` and have become familiar with foundry's vars while working on attribute builder. Will take a look at cleaning this up in a bit.

CarlosFdez:
No luck with 

```
.svelcte. :global,
:global(.sv_dropdown.pf2e) :global {
    blah
}
```

either?

Then I guess leave it as is. The thing I was going for is that .svelcte. doesn't need the global, its just the nested stuff (which are internal and cannot be scoped) and sv_dropdown that does.

CarlosFdez:
Ah I see why the other needs a global as well. `.svelcte` is assigned internally, and `.pf2e` is dynamic. What if you do `class:svelecte` and have the top level be .svelecte in the prior example without global? Or is that not doable with custom components?

src/module/sheet/components/svelecte-pf2e.svelte
CarlosFdez: I was not familiar with light-dark when I wrote this, and I think it was also during v12 or early v13, so it def could use some cleaning up.

src/module/sheet/components/svelecte-pf2e.svelte
jfn4th: Normally svelecte will apply the `color: var(--sv-color)` to the `.svelecte` element, but being detached it's not inheriting anymore.

jfn4th:
Unfortunately the `class:xxx` directive is not valid on components.

jfn4th:
Updated with `light-dark` and foundry css vars
Screenshot 2025-12-29 at 3 03 01 PM
Screenshot 2025-12-29 at 3 03 14 PM

src/module/sheet/components/svelecte-pf2e.svelte
```diff
@@ -50,18 +83,29 @@
...
             --sv-item-btn-color-hover: var(--color-text-trait);
-            --sv-control-bg: rgba(0, 0, 0, 0.1);
-            --sv-icon-color: var(--color-dark-6);
+            --sv-control-bg: var(--input-background-color);
+            --sv-icon-color: var(--button-text-color);
             --sv-item-selected-bg: var(--color-bg-trait);
             --sv-item-btn-bg: var(--color-bg-trait);
+            --sv-placeholder-color: var(--color-form-hint);
+            --sv-dropdown-bg: light-dark(white, var(--color-cool-4));
+            --sv-dropdown-active-bg: light-dark(rgba(0, 0, 0, 0.1), var(--color-text-selection-bg));
```
CarlosFdez, line 101: is color-text-selection-bg not already good for light mode and dark mode?

Ideally we should be using css variables from core foundry if one exists that works, light-dark is a last resort.

jfn4th: `color-text-selection-bg`'s light mode doesn't suit this case very well
Screenshot 2025-12-29 at 3 26 09 PM

jfn4th:
In the (hopefully) not too far future, we'll be able to ditch the position resolver/theme sync via [CSS Anchor Positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning). Firefox support still in beta...

Even then, that won't get rid of the `:global` since everything being styled is svelecte internals, but still cleans things up a fair bit.

```svelte
<style>
    :global {
        .svelecte.pf2e {
            anchor-name: --svelecte-anchor;

            /* ... */

            .sv_dropdown {
                position: fixed;
                position-anchor: --svelecte-anchor;
                min-width: anchor-size(width);
            }
        }
    }
</style>

File: src/module/sheet/components/svelecte-pf2e.svelte (hunk 3 above)
Comment (stwlam) [inline src/module/sheet/components/svelecte-pf2e.svelte:101]:
```suggestion
            --sv-dropdown-bg: var(--table-header-bg-color);
            --sv-dropdown-active-bg: var(--table-row-color-highlight);
```
How about these?

jfn4th: Dark mode looks identical, but still no good with light mode. `--table-header-bg-color` is too transparent and `--table-row-color-highlight` is even darker than `--color-text-selection-bg` as supe suggested.

Screenshot 2026-01-06 at 11 32 52 PM

PR #21103, opened 2025-12-30
Add inline checks for Influence Rumor

ammalagonc:


packs/pf2e/actions/archetype/dandy/influence-rumor.json
```diff
@@ -12,7 +12,7 @@
...
         "description": {
-            "value": "<p>You spend at least 1 day of downtime attempting to manipulate the course, tone, or content of a rumor to your benefit. You need to succeed at a Diplomacy check to shift the rumor as you intend. The difficulty is determined by the GM based on the size of the community, the relative perceptiveness of the inhabitants, and the agency of other rumormongers, but it typically is no lower than DC 15 for a small village, DC 20 for a town, DC 30 for a city, or DC 40 for a metropolis.</p>"
+            "value": "<p>You spend at least 1 day of downtime attempting to manipulate the course, tone, or content of a rumor to your benefit. You need to succeed at a Diplomacy check to shift the rumor as you intend. The difficulty is determined by the GM based on the size of the community, the relative perceptiveness of the inhabitants, and the agency of other rumormongers, but it typically is no lower than @Check[diplomacy|dc:15|traits:secret]{ } for a small village, @Check[diplomacy|dc:20|traits:secret]{ } for a town, @Check[diplomacy|dc:30|traits:secret]{ } for a city, or @Check[diplomacy|dc:40|traits:secret]{ } for a metropolis.</p>"
```
pedrogrullada: I don't really like this personally, seems like dancing around a problem. If it weren't a downtime activity I'd suggest using suboptions to change the inline DC that way, and making the DC immutable. (And thinking about it, those DCs are merely upper bounds anyway!)

But it being an downtime activity makes it awkward since toggles don't show in that tab. So probably best to just make it a Diplomacy check with no DC attached. It's also not supposed to be secret, right?

```suggestion
            "value": "<p>You spend at least 1 day of downtime attempting to manipulate the course, tone, or content of a rumor to your benefit. You need to succeed at a @Check[diplomacy] check to shift the rumor as you intend. The difficulty is determined by the GM based on the size of the community, the relative perceptiveness of the inhabitants, and the agency of other rumormongers, but it typically is no lower than DC 15 for a small village, DC 20 for a town, DC 30 for a city, or DC 40 for a metropolis.</p>"
```

PR #21122, opened 2026-01-02
Add strike rule to solar weapon to change damage dice with two hand d10 trait

Zendelll:
My players found out that you can't change grip on strikes that were created with a strike rule element. My guess it's because it doesn't create a real item, so we can't change grip. But since when you are creating a solar weapon with two hand d10 trait, you certainly want to use it in 2h, I guess it might be okay to just add a new strike rule with d10, that will be used only with the solar weapon with this trait. Not the best approach, but it least it'll work

### DIFF
--- a/packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
+++ b/packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
@@ -351,6 +351,29 @@
                 ]
             },
             {
+              "domain": "damage",
+              "key": "RollOption",
+              "option": "solar-weapon-two-hand",
+              "toggleable": true,
+              "value": false,
+              "predicate": [
+                "solar-weapon-trait-one:two-hand-d10"
+              ],
+              "label":"Solar Weapon 2H"
+            },
+            {
+              "key": "DamageAlteration",
+              "property": "dice-faces",
+              "value": 10,
+              "predicate": [
+                "solar-weapon-two-hand"
+              ],
+              "mode": "override",
+              "selectors": [
+                "{item|id}-damage"
+              ]
+            },
+            {
                 "itemType": "weapon",
                 "key": "ItemAlteration",
                 "mode": "add",
--- a/packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
+++ b/packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
@@ -353,20 +353,20 @@
             {
               "domain": "damage",
               "key": "RollOption",
-              "option": "solar-weapon-two-hand",
+              "option": "twin-solar-weapon-two-hand",
               "toggleable": true,
               "value": false,
               "predicate": [
-                "solar-weapon-trait-one:two-hand-d10"
+                "twin-solar-weapon-trait-one:two-hand-d10"
               ],
-              "label":"Solar Weapon 2H"
+              "label":"SF2E.SpecificRule.Solarian.SolarWeapon.TwoHandToggle.TwinLabel"
             },
             {
               "key": "DamageAlteration",
               "property": "dice-faces",
               "value": 10,
               "predicate": [
-                "solar-weapon-two-hand"
+                "twin-solar-weapon-two-hand"
               ],
               "mode": "override",
               "selectors": [
--- a/packs/sf2e/feat-effects/effect-solar-weapon.json
+++ b/packs/sf2e/feat-effects/effect-solar-weapon.json
@@ -402,9 +402,6 @@
                 "predicate": [
                     {
                         "not": "stellar-attunement:unattuned"
-                    },
-                    {
-                        "not": "solar-weapon-trait-one:two-hand-d10"
                     }
                 ],
                 "range": null,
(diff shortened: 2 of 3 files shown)


TikaelSol:
Rather than adding a new strike, which would potentially complicate some other automation, we should maybe consider just adding a toggle for it along with a DamageDice override. It's kludgy but less kludgy than making an entire second strike.

Ideally At least since this is on an effect we can always remove this stuff later when the system properly supports two hand from Strike REs

Zendelll:
Yep, I think it'll be better. I'll change it to a toggle!

pedrogrullada:
When you do, use Damage Alteration rather than Damage Dice, as the latter's override mode is being phased out.

Zendelll:
Used DamageAlteration with RollOption toggle, it works much better now

packs/sf2e/feat-effects/effect-solar-weapon.json
```diff
@@ -411,6 +411,30 @@
...
+            },
+            {
+              "key": "DamageAlteration",
+              "property": "dice-faces",
+              "value": 10,
+              "predicate": [
+                "solar-weapon-two-hand"
+              ],
+              "label": "Solar Weapon Two Hand",
+              "mode": "override",
```
pedrogrullada, line 432:
```suggestion
              ],
              "mode": "override",
```
This label is not needed

packs/sf2e/feat-effects/effect-solar-weapon.json
```diff
@@ -411,6 +411,30 @@
...
+            {
+              "domain": "damage",
+              "key": "RollOption",
+              "option": "solar-weapon-two-hand",
+              "toggleable": true,
+              "value": false,
+              "predicate": [
+                "solar-weapon-trait-one:two-hand-d10"
+              ],
+              "label":"Solar Weapon 2H"
```
pedrogrullada:
```suggestion
              "label":"Solar Weapon 2H"
```
This label needs to be localized. When you do, a label like "Solar Weapon — Use two hands" would work.

pedrogrullada (commented):
Technically you need to do add this for a twin solar weapon as well, unlikely as it may be that someone actually wields one two-handed. I suppose there's multi-armed ancestries in this game, after all!

packs/sf2e/feat-effects/effect-solar-weapon.json
Zendelll: Do we have any examples of localization in rule elements? I'm not sure how to properly implement this

packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
```diff
@@ -350,6 +350,29 @@
...
                 ]
             },
+            {
+              "domain": "damage",
+              "key": "RollOption",
+              "option": "solar-weapon-two-hand",
+              "toggleable": true,
+              "value": false,
+              "predicate": [
+                "solar-weapon-trait-one:two-hand-d10"
```
pedrogrullada:
```suggestion
                "twin-solar-weapon-trait-one:two-hand-d10"
```

packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
```diff
@@ -350,6 +350,29 @@
                     "solarian"
                 ]
             },
+            {
+              "domain": "damage",
+              "key": "RollOption",
+              "option": "solar-weapon-two-hand",
```
pedrogrullada:
```suggestion
              "option": "twin-solar-weapon-two-hand",
```

packs/sf2e/feat-effects/effect-solar-weapon-twin-weapons.json
```diff
@@ -350,6 +350,29 @@
...
+                "solar-weapon-trait-one:two-hand-d10"
+              ],
+              "label":"Solar Weapon 2H"
+            },
+            {
+              "key": "DamageAlteration",
+              "property": "dice-faces",
+              "value": 10,
+              "predicate": [
+                "solar-weapon-two-hand"
```
pedrogrullada:
```suggestion
                "twin-solar-weapon-two-hand"
```

packs/sf2e/feat-effects/effect-solar-weapon.json
pedrogrullada: `Spell Effect: Quantum Analysis (Fortune)` has a localized label [here](https://github.com/foundryvtt/pf2e/blob/a4f649d1d6ca707569d54b1588f902af8317415a/packs/sf2e/spell-effects/spell-effect-quantum-analysis-fortune.json#L27), and the key that is referencing is [here](https://github.com/foundryvtt/pf2e/blob/a4f649d1d6ca707569d54b1588f902af8317415a/static/lang/re-en.json#L8071).

So nearly every time there would be raw text in a rule element, we add it to that language file (`re-en.json`) and reference it in the label like you see in the linked effect. For this particular case, you'd likely add
```json
                    "TwoHandToggle": {
                        "Label": "Solar Weapon — Use two hands",
                        "TwinLabel": "Twin Solar Weapon — Use two hands"
                    }
```
under `SF2E.SpecificRule.Solarian.SolarWeapon`, such that the labels for the toggles on Solar Weapon and Twin Solar Weapon's effects would be: `SF2E.SpecificRule.Solarian.SolarWeapon.TwoHandToggle.Label` and `SF2E.SpecificRule.Solarian.SolarWeapon.TwoHandToggle.TwinLabel` respectively.

Zendelll:
Done! Thank you so much for your help!

PR #21134, opened 2026-01-04
Fix Stone Lion notes formatting

Cuingamehtar:


simonward:
Consider making the same fix in MC2 Stone Lion.

Cuingamehtar:
For some reason VS Code search bar only showed me 3 files (2 cubs and 1 lion), so I thought one was already being redirected. Fixed it.

PR #21136, opened 2026-01-04
Tidy `WeaponReloader` application

stwlam:


CarlosFdez:
How does this handle the case where you reload an sf2e gun with an empty row, load one bullet into it, and then it has two rows where the first is the loaded bullet and the second is the new empty row? I was using the interval so that the popup would move with the shifting elements as you continue to load the gun. Did you find an alternative I didn't notice?

PR #21139, opened 2026-01-04
Automate Traditional Resistances Dragonblood feat

TikaelSol:
Also handle/remove some unneeded notes on feats


ammalagonc:
Should this be for Dragonblood feat?

PR #21140, opened 2026-01-05
Remove the Remaster Journal popup on new worlds

TikaelSol:


simonward:
Does "seenRemasterJournalEntry" need to be removed from ClientSettingsPF2e and registerTrackingSettings()?

PR #21151, opened 2026-01-06
Added missing traditions to Share Pain

Arrrthritis:
Noticed that the spell Share Pain didn't have any traditions listed. This PR should address that.


simonward:
It also looks like it's incorrectly heightening at 4 rather than 6.

PR #21163, opened 2026-01-07
Move NPC glossary and custom IWR to PF2E namespace

CarlosFdez:
Glossaries and custom IWR are part of a more general set, where the pf2e ones show up in sf2e as well and an sf2e entry could eventually show up in pf2e. So they need to be with everything else for long term compatibility.


static/lang/re-en.json
```diff
@@ -7509,13 +7497,6 @@
             }
         },
         "SpecificRule": {
-            "AreaFire": {
-                "One": "1",
-                "Three": "3",
-                "TrackingLabel": "Tracking Value",
-                "Two": "2",
-                "Zero": "0"
-            },
```
TikaelSol, line 7518: Were these unused?

CarlosFdez: Seems that way.

PR #21176, opened 2026-01-08
Fix reload on attack popup

CarlosFdez:
This also prevents holding onto a reference to a stale html element, and makes the re-render reposition more resilient in cases where the actor updates due to something a different user (such as the GM) did. Manual sheet re-renders are not broadcasted.

image

### DIFF
--- a/src/module/item/weapon/apps/weapon-reloader/app.ts
+++ b/src/module/item/weapon/apps/weapon-reloader/app.ts
@@ -1,23 +1,31 @@
+import { CreaturePF2e } from "@actor";
 import type { CharacterPF2e } from "@actor/character/document.ts";
+import { CreatureSheetPF2e } from "@actor/creature/sheet.ts";
 import { type PhysicalItemPF2e, WeaponPF2e } from "@item";
 import { getLoadedAmmo } from "@item/weapon/helpers.ts";
 import { ChatMessagePF2e } from "@module/chat-message/document.ts";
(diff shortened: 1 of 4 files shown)


src/module/item/weapon/apps/weapon-reloader/app.ts
```diff
@@ -1,23 +1,31 @@
...
-    anchor: HTMLElement | null;
 }
 
 class WeaponReloader extends SvelteApplicationMixin<
     AbstractConstructorOf<fa.api.ApplicationV2> & { DEFAULT_OPTIONS: DeepPartial<WeaponReloaderConfiguration> }
 >(fa.api.ApplicationV2) {
+    constructor(options: DeepPartial<WeaponReloaderConfiguration> & { anchor: HTMLElement | null }) {
+        super(R.omit(options, ["anchor"]));
+        this.#anchorId = options.anchor?.dataset.anchorId ?? null;
+        this.#appId = htmlClosest(options.anchor, "[data-appid]")?.dataset.appid ?? null;
```
stwlam: Maybe call this `sheetId` since the WeaponReloader is also an app with an id

CarlosFdez: you rite

CarlosFdez: actually, #anchorAppId is the least ambiguous.  I went with that one.

PR #21179, opened 2026-01-09
Override sickened description in Starfinder and include serums in its description addendum

### LINKED ISSUE #21175: SF2e: Sickened is not a full double
Sickened condition has a bit different description in SF2e from PF2e:
_You feel ill. Sickened always includes a value. You take a status penalty equal to this value on all your checks and DCs. You can't willingly ingest anything—**including serums**—while sickened._

PF2e:
_You feel ill. Sickened always includes a value. You take a status penalty equal to this value on all your checks and DCs. You can't willingly ingest anything—**including elixirs and potions**—while sickened._

pedrogrullada:
- Closes #21175

This is bound to hit a conflict with #21178 but I can deal with it.

Also moves serum from an equipment trait to a consumable trait.
image


CarlosFdez:
Would it not be possible to make the pf2e one inclusive of sf2e traits and vice versa? Anachronism is a thing after all.

It looks like the description being different means that can't be duplicates though.

pedrogrullada:
We can always do `@Localize` and use a localization override. We already store the descriptions as localization keys.

CarlosFdez:
That would work. I think I'm ok with either that, or updating the rule elements in both to be inclusive of both systems. Its not that the behavior of sickened is any different (you just can't ingest), its just that serums aren't a thing in pf2e normally.

pedrogrullada:
I've gone the `@Localize` route

PR #21195, opened 2026-01-11
Add automation to Versatile Mutation feat

ammalagonc:


TikaelSol:
This may be doable in a cleaner way rather than dropping a bunch of different roll options we should be able to just check if they have the feat already and are level 16, then use all the same roll option, no?

ammalagonc:
> This may be doable in a cleaner way rather than dropping a bunch of different roll options we should be able to just check if they have the feat already and are level 16, then use all the same roll option, no?

I tried that approach (checking if `feat:versatile-mutation` was present via predicates), and for some reason, the drop-down box was disappearing whenever the actor reached level 16. Hence why I went with using the RollOptions as predicates and that seemed to work.

pedrogrullada:
Before going deep into the specifics, I think I disagree with using character level to predicate a Choice Set here. Mostly because you could be retraining a lower level feat slot, or just building a high level character and filling in their feats, and the feat will think you're using a high level feat slot. Basically the same deal as why we don't automate "advanced" archetype feats

ammalagonc:
> Before going deep into the specifics, I think I disagree with using character level to predicate a Choice Set here. Mostly because you could be retraining a lower level feat slot, or just building a high level character and filling in their feats, and the feat will think you're using a high level feat slot. Basically the same deal as why we don't automate "advanced" archetype feats

Suggestions are welcome

pedrogrullada:
~~Unfortunately my suggestion is the same as with advanced feats, can't be done well enough yet, unless I'm missing something.~~ I was missing something! Since this can only be taken a second time after level 16, that's the only case you need to account for, so there's no need to really predicate on it. 

I think you can use a Roll Option RE with `count`, and then use Choice Sets predicated on the number of times the counter has (1 or 2)

ammalagonc:
> ~Unfortunately my suggestion is the same as with advanced feats, can't be done well enough yet, unless I'm missing something.~ I was missing something! Since this can only be taken a second time after level 16, that's the only case you need to account for, so there's no need to really predicate on it.
> 
> I think you can use a Roll Option RE with `count`, and then use Choice Sets predicated on the number of times the counter has (1 or 2)

And this works!

PR #21204, opened 2026-01-12
Add missing SF2e vehicle-related actions and correct their source

Cuingamehtar:


CarlosFdez:
These need to be entries in system/action-macros/index.ts (horrible place for them giving they are not macros) and be in SF2eOnlySystemActions. Then the item needs an [[/act]] for it.

Cuingamehtar:
I'll look into adding them there, but 
a) they are effectively the same between PF and SF, so no need to have them in SF-only actions,
b) Run Over and Stop don't require checks and
c) Drive and Board are not straightforward about their checks, with Drive's result being different for different number of actions spent, and Board requiring either Acrobatics or Athletics and only of the vehicle is in motion. I can reference how Escape handles it, but it might not be worth the effort.

CarlosFdez:
I suppose the implementations of the drive actions need to differ between systems. What are you trying to fix here specifically then? Is there something erroring, or just a description thing? If its the latter, this is fine until we address the vehicle actions.

Cuingamehtar:
Just the descriptions.
I've found out that some of the links we had in the translation are invalid and decided to add the missing action Items.

PR #21208, opened 2026-01-12
Fix grant items exploding in sf2e build

CarlosFdez:


stwlam:
No explosions for me?

CarlosFdez:
The building itself works but it causes an issue when you try to add soldier to a character (or any class feature that invokes grant item). It was reported in the alpha test thread.

The reason is that folder is a string id for most of them but folder is an object for the ysoki folder remap, so schema validation fails.

CarlosFdez:
The error was unhelpful but this is what happened.

image

PR #21219, opened 2026-01-13
Update currency to use Credits instead of Silver on Iconic actors and corrects Chk-Chk's and Zemir's spellcasting attributes

ammalagonc:


TikaelSol:
Apparently some of Zemir's DCs are also incorrectly based on cha

PR #21220, opened 2026-01-13
Fix fly speed calculation in Soaring Wings effect

Reiaka:


simonward:
These were changed to total in:
https://github.com/foundryvtt/pf2e/pull/20374 and https://github.com/foundryvtt/pf2e/pull/20403
@TikaelSol?

TikaelSol:
Shark changed the location of the data on us so I was still correcting things to the old (now non-working) data path. This one is good to merge whenever the pipeline finishes

PR #21225, opened 2026-01-15
Fix references to glossary items on NPCs

TikaelSol:


CarlosFdez:
Merging, but this will cause issues with anachronism referring to sf2e only bestiary glossary entries. We should put a pin on this for a proper solution somehow.

PR #21229, opened 2026-01-16
Support resilient item alterations on non-investment armor

### LINKED ISSUE #21217: Adding Resilient Runes via ItemAlteration RE does not add the invested trait to the modified armor.
As said in title, it is required to either manually add the trait or to add it with the ItemAlteration RE. Otherwise the rune gets sucessfully applied but cannot be used as you are unable to invest into the armor.

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/21217

### DIFF
--- a/src/module/item/physical/document.ts
+++ b/src/module/item/physical/document.ts
@@ -146,6 +146,7 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         return this.system.traits.value.some((t) => t === "magical" || setHasElement(MAGIC_TRADITIONS, t));
     }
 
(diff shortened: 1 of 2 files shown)


src/module/item/physical/document.ts
```diff
@@ -146,6 +146,7 @@ abstract class PhysicalItemPF2e<TParent extends ActorPF2e | null = ActorPF2e | n
         return this.system.traits.value.some((t) => t === "magical" || setHasElement(MAGIC_TRADITIONS, t));
     }
 
+    /** Returns true if invested, false if not invested and requires investment, and null otherwise */
```
```suggestion
    /** Returns true if invested, false if not invested and requires investment, and otherwise null */
```

PR #21247, opened 2026-01-17
Add action inline for Demoralize in Battle Cry skill feat

ammalagonc:


packs/pf2e/feats/skill/level-7/battle-cry.json
simonward:
```suggestion
```

Issue #21091, opened 2025-12-28
Powerful Alchemy applying to incorrect modifiers


20maxm211-max:
One of my players is running a lvl 7 bomber Alchemist, we noticed towards the end of a session that his infused items DCs were too high at a 32 and 34 in different moments. 

After some testing, discovered the source of the issue is Powerful Alchemy, as it appears to be applying to the frightened and sickened conditions resulting in heightened DCs beyond capability. 

This is how the DC looks with frightened or sickened 1 conditions applied to the character:

<img width="485" height="254" alt="Image" src="https://github.com/user-attachments/assets/474ef0d0-7e11-48cb-bafc-8b81c34ad87a" />

This is the DC with the condition increased to 2:

<img width="490" height="259" alt="Image" src="https://github.com/user-attachments/assets/e3f2dea0-d062-42ef-876d-e17a09138b50" />

This is how the DC looks without the frightened or sickened conditions applied to the character:

<img width="491" height="260" alt="Image" src="https://github.com/user-attachments/assets/b96bcbe5-b3b3-469c-a1ba-102e2ff916d8" />

For ease of access, here is the JSON for the character:

[Seeril.json](https://github.com/user-attachments/files/24362326/Seeril.json)

manuel-hegner:
#21251 did not fix this completely.

While now the class DC (or more specifically class DC mod) is applied only once to the base of inline DCs, effects are still counted twice.

This happens because for example frightened affects the class DC and the inline DCs.

PR #21251, opened 2026-01-18
Add base slug to Powerful Alchemy

rkukuchka:
Powerful Alchemy was lacking a base slug which resulted in the target DC going wacky if any conditions on the actor impacted DCs. As detailed in https://github.com/foundryvtt/pf2e/issues/21091

### DIFF
--- a/packs/pf2e/class-features/powerful-alchemy.json
+++ b/packs/pf2e/class-features/powerful-alchemy.json
@@ -32,6 +32,7 @@
                     "item:trait:infused"
                 ],
                 "selector": "inline-dc",
+                "slug": "base",
                 "value": "@actor.classDCs.alchemist.mod"
             }
         ],


packs/pf2e/class-features/powerful-alchemy.json
```diff
@@ -32,6 +32,7 @@
                     "item:trait:infused"
                 ],
                 "selector": "inline-dc",
+                "slug": "base",
```
ammalagonc:
```suggestion
                "slug": "powerful-alchemy-base",
```

ammalagonc, line 35: Always a good practice to have slug names match the modifier for ease of reference.

pedrogrullada: In this case, they *want* to reference the base modifier only, so `base` is correct

PR #21254, opened 2026-01-18
Update Elemental Magic to account for Wizard and Sorcerer adjustments

ammalagonc:


packs/pf2e/class-features/elemental-magic.json
```diff
@@ -57,8 +57,164 @@
...
+                "predicate": [
+                    "elemental-focus-adjustment:yes",
+                    {
+                        "or": [
+                            "item:slug:elemental-toss",
+                            "item:slug:elemental-motion"
+                        ]
+                    }
+                ],
+                "priority": 150,
```
pedrogrullada: This priority is too late. Item Alterations relying on this tag usually fire off at 121, so your tag needs to be gone by then for this to do anything. Perhaps changing the priority on those features to be something like 125 will give you enough wiggle room to squeeze this removal at some number in-between. You can do a [search](https://github.com/search?q=repo%3Afoundryvtt%2Fpf2e+%22item%3Atag%3Ablood-magic-spell%22&type=code) for `item:tag:blood-magic-spell` to see which automations rely on this tag. It's not as many as it seems considering all the bloodline ones are practically identical except for their specific spells.

PR #21261, opened 2026-01-19
Add Battle for Nova Rush actors

ammalagonc:


packs/sf2e/standalone-adventure-bestiary/_folders.json
```diff
@@ -1,4 +1,26 @@
 [
+    {
+        "_id": "1jA4waT9aciKvt6k",
+        "color": null,
+        "description": "",
+        "flags": {},
+        "folder": null,
+        "name": "Free RPG Day",
```
TikaelSol: I don't think we need this top level folder

PR #21268, opened 2026-01-20
Add aura to 3rd rank protection spell effect

Drental:
I hope I did this right, and I hope this doesn't conflict with the text of the spell being:

Heightened (3rd) You **can** choose to have the benefits also affect all your allies in a 10-foot emanation around the target.

I am also not sure if this is intended to act like an aura, or to apply to everyone at the time of casting for the duration of the spell

But @InfamousSky made me make this PR, so he's at fault

### DIFF
--- a/packs/pf2e/spell-effects/spell-effect-protection-aura.json
+++ b/packs/pf2e/spell-effects/spell-effect-protection-aura.json
@@ -1,7 +1,7 @@
 {
-    "_id": "RawLEPwyT5CtCZ4D",
+    "_id": "Ydjj9h3EwhVLmv11",
     "img": "systems/pf2e/icons/spells/protection.webp",
-    "name": "Spell Effect: Protection",
+    "name": "Spell Effect: Protection (Aura)",
     "system": {
         "description": {
             "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Protection].</p>\n<p>You gain a +1 status bonus to Armor Class and saving throws.</p>"
@@ -12,6 +12,7 @@
             "unit": "minutes",
             "value": 1
         },
+        "fromSpell": false,
         "level": {
             "value": 1
         },
@@ -30,23 +31,6 @@
                 ],
                 "type": "status",
                 "value": 1
-            },
-            {
-                "effects": [
-                    {
-                        "uuid": "Compendium.pf2e.spell-effects.Item.Ydjj9h3EwhVLmv11"
-                    }
-                ],
-                "key": "Aura",
-                "predicate": [
-                    {
-                        "gte": [
-                            "parent:level",
-                            3
-                        ]
-                    }
-                ],
-                "radius": 10
             }
         ],
         "start": {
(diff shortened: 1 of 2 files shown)


Drental:
I don't really understand why the pipeline would fail on this 😅

packs/pf2e/spell-effects/spell-effect-protection.json
```diff
@@ -30,6 +30,23 @@
                 ],
                 "type": "status",
                 "value": 1
+            },
+            {
+                "effects": [
+                    {
+                        "uuid": "Compendium.pf2e.spell-effects.Item.Ydjj9h3EwhVLmv11"
```
pedrogrullada:
```suggestion
                        "uuid": "Compendium.pf2e.spell-effects.Item.Spell Effect: Protection (Aura)"
```
I think this might be why the pipeline is failing. That said, you can do this without creating a new effect, see how we do heightened silence.

But I don't think this is an aura at all since the spell doesn't say it. Seems closer to courageous anthem et. al.

ammalagonc:
Extract the effect first then the spell.

Drental:
they are both effects^^

packs/pf2e/spell-effects/spell-effect-protection.json
```diff
@@ -30,6 +30,23 @@
                 ],
                 "type": "status",
                 "value": 1
+            },
+            {
+                "effects": [
```
simonward, line 35: Should probably add `"affects": "allies"`.

packs/pf2e/spell-effects/spell-effect-protection-aura.json
```diff
@@ -0,0 +1,48 @@
+{
+    "_id": "Ydjj9h3EwhVLmv11",
+    "img": "systems/pf2e/icons/spells/protection.webp",
+    "name": "Spell Effect: Protection (Aura)",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Protection].</p>\n<p>You gain a +1 status bonus to Armor Class and saving throws.</p>"
```
simonward, line 7:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Protection]</p>\n<p>You gain a +1 status bonus to Armor Class and saving throws.</p>"
```

simonward:
Might be better to add both effects to the spell itself rather than use a predicate? That way the caster can decide which to use.

simonward:
Actually perhaps a ChoiceSet would be better?

packs/pf2e/spell-effects/spell-effect-protection-aura.json
```diff
@@ -0,0 +1,48 @@
...
+        },
+        "publication": {
+            "license": "ORC",
+            "remaster": true,
+            "title": "Pathfinder Player Core"
+        },
+        "rules": [
+            {
+                "key": "FlatModifier",
+                "label": "Protection",
```
simonward, line 27: Consider removing or localizing the label.

TikaelSol:
Superseded by #21307 since I don't want to wonder why the pipeline is actually failing here :D

PR #21274, opened 2026-01-21
Update sf2e modular weapons to new modular system

mechamaya:
Data entry to test #20576. This should include every modular weapon in sf2e so far. There's going to be a lot more that need updating in pf2e but those are generally much simpler.


packs/sf2e/equipment/weapons/cryopike.json
```diff
@@ -26,7 +26,7 @@
             "value": "<p><strong>Upgrades:</strong> 1</p><hr /><p>This polearm's shaft houses rapidly circulating supercooled gas that condenses and channels endothermic energy into a cryogenic blade.</p>"
         },
         "expend": null,
-        "grade": "commercial",
```
CarlosFdez, line 29: Looks like this is a bug. I will handle this later.

EDIT: basically modular weapons where the tech/analog is in the modular config get interpreted as pf2e weapons.

mechamaya: Ahh, that makes sense now, I was wondering why the export changed the grade to null. Is there a workaround for this like mark them all as tech or analog by default, and then have the modular override it? Or does it just have to be a code fix to recognize tech or analog inside of the modular traits?

CarlosFdez: A code fix is required. https://github.com/foundryvtt/pf2e/pull/21284 should resolve it.

CarlosFdez:
should be good now to continue, make sure the grades aren't set back to null in the final result! Thanks for working on this.

mechamaya:
> should be good now to continue, make sure the grades aren't set back to null in the final result! Thanks for working on this.

Thanks, I'll get the updated ones up in a little bit.

CarlosFdez (approved):

PR #21292, opened 2026-01-22
Update pf2e switchscythe to new modular system

mechamaya:
For some reason I thought there was a lot more...

### DIFF
--- a/packs/pf2e/equipment/big-boom-gun.json
+++ b/packs/pf2e/equipment/big-boom-gun.json
@@ -69,22 +69,6 @@
             "value": 0
         },
         "traits": {
-            "config": {
-                "modular": [
-                    {
-                        "damageType": "bludgeoning",
-                        "traits": []
-                    },
-                    {
-                        "damageType": "piercing",
-                        "traits": []
-                    },
-                    {
-                        "damageType": "slashing",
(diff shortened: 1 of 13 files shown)


CarlosFdez:
So modular on its own should be presuming BPS. Is that not functioning?

CarlosFdez:
It is, but its not obvious in the UI. Let me add something very quickly.

mechamaya:
Ah ok, if BPS is default then this PR is probably moot then. Cause they're all just BPS here in PF2e.

CarlosFdez:
I think the switchscythe might be the only expection?

mechamaya:
Huh somehow I missed the obvious one.

mechamaya:
Yeah BPS seems to work by default, but I definitely didn't realize it was working from looking at the item description. Only looking at the sheet actions.

CarlosFdez:
Can you keep only the switchscythe?

mechamaya:
> Can you keep only the switchscythe?

Yeah should have only switchscythe now. The rest should be fine by default the UI just needs to indicate that it's working?

CarlosFdez:
Yeah basically, there's a PR up for that. Ideally its better to have less data than more. Thank you!

PR #21294, opened 2026-01-22
Change Release Github action to include system-specific language files

Cuingamehtar:


CarlosFdez (approved):
Looks good to me but I want a second opinion from @stwlam

Cuingamehtar:
There should really be a better way of removing all the merged branches than going over them manually

PR #21297, opened 2026-01-22
Correct internal currency values for select items

### LINKED ISSUE #21276: [SF2e] Some data error and/or inconsistency with equipment pricing
Some equipment in the compendium have their price listed as `gp` instead of `sp`. For some it is simply a bad entry like `Diva's Microphone (Advanced)` which has its price set as 190400 `gp` while it cost 190400 `credits`. Some others have their price divided by `10` set as `gp` which makes them technically correct i suppose but inconsistent with all the other entries in the compendium.

Here is the list of items that use `gp` instead of `sp`: 

    "Diva's Microphone (Advanced)",
    "Machine Mind Serum (Advanced)",
    "Vesk Scales",
    "Petrified Bark",
    "Flash Grenade (Paragon)",
    "Shuriken Drone",
    "Diva's Microphone (Tactical)",
    "Invisibility Potion",
    "Diva's Microphone (Commercial)",
    "Machine Mind Serum (Tactical)",
    "Dizzying Anesthetic"

ammalagonc:
- Closes #21276

Note that Invisibility Potion is not a native item from SF2e, but was added because of typo shenanigans/links in the Detect Magic description. Those links have been removed, so the item in question could be removed.

### DIFF
--- a/packs/sf2e/equipment/grenades/flash-grenade-paragon.json
+++ b/packs/sf2e/equipment/grenades/flash-grenade-paragon.json
@@ -42,7 +42,7 @@
         },
         "price": {
             "value": {
-                "gp": 5000
+                "sp": 5000
             }
         },
         "publication": {
--- a/packs/sf2e/equipment/magic-items/divas-microphone-advanced.json
+++ b/packs/sf2e/equipment/magic-items/divas-microphone-advanced.json
@@ -27,9 +27,9 @@
         "price": {
             "value": {
                 "cp": 0,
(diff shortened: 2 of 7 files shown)


reonZ:
- `Flash Grenade (Paragon)` should be 50,000 `sp` not 5,000
- `Machine Mind Serum (Advanced)` should be 2,000 `sp`  not 200
- `Machine Mind Serum (Tactical)` should be 100 `sp`  not 10

PR #21308, opened 2026-01-22
Fix attribute builder in sf2e

jfn4th:


CarlosFdez:
Thank you for the followup

jfn4th:
I break, I fix!

PR #21311, opened 2026-01-23
Fix level and traits of Big Mouth

Shazburg:
Corrected big-mouth feat level from 1 to 9.
Corrected big-mouth feat trait value from empty to ratfolk so it stops matching all ancestries.


Shazburg:
Found a couple errors in the ysoki level 9 ancestry feat Big Mouth. Corrected wrong character level and trait.

PR #21315, opened 2026-01-23
Rename and merge Double Draw

Cuingamehtar:
I've left the shobhad feat as a separate item, as it has different publication and description.

Beasts of Bo Bandit I can't rename, since I don't know the numbers that are supposed to be used in the name.


TikaelSol:
These have two different descriptions, so we should have both and use an alteration to set the right one on actors

Cuingamehtar:
Should the difference in publication source be ignored?

TikaelSol:
I think we can keep player core as the source

Cuingamehtar:
I've reproduced the REs from Perception Mastery and Magical Fortitude. Even the one that adds a trait, but that one does nothing, since the trait is already present in the source item.

packs/sf2e/feats/ancestry/shared/level-5/double-draw.json
```diff
@@ -0,0 +1,63 @@
...
+        },
+        "rules": [
+            {
+                "itemId": "{item|id}",
+                "key": "ItemAlteration",
+                "mode": "add",
+                "property": "traits",
+                "value": "{actor|ancestry.slug}"
+            },
+            {
```
TikaelSol, line 29:
```suggestion
        "rules": [
            {
```

PR #21318, opened 2026-01-24
Add Android Level 5 feats and select duplicate backgrounds to the duplicates.json

ammalagonc:
Added only feats with the same text (checked using Diff Checker); anything else is going in as a separate PR.


TikaelSol:
Conflicts!

PR #21328, opened 2026-01-25
Add support for weapon capacity item alterations

AlexClark:
Added a capacity item alteration property that allows modifying the capacity for ranged weapons. Allows users to make things like an "Extended Magazine" mod.

Doesn't work for weapons with ammo types that track usage separately, and prevents attempting to set capacity when it is not available or not a positive value.


src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -226,6 +226,41 @@ const ITEM_ALTERATION_HANDLERS = {
...
+            value: new fields.NumberField({
+                required: true,
+                nullable: false,
+                integer: true,
+                positive: true,
+                initial: undefined,
+            } as const),
+        },
+        validateForItem(item): validation.DataModelValidationFailure | void {
+            const weapon = item as WeaponPF2e;
```
AlexClark: I tried a variety of ways to get the editor to shut up about accessing the ammo property, and casting the object worked. I referenced the "group" handler here in the same file, which does the same thing.

CarlosFdez: itemIsOfType should work. That should narrow it down to WeaponPF2e | WeaponSource.

That said I don't know if the extra validation is necessary. We don't usually error outright on incompatible properties, we usually early return instead. For example, grade doesn't error on non-tech weapons.

AlexClark: > itemIsOfType should work. That should narrow it down to WeaponPF2e | WeaponSource.
> 
> That said I don't know if the extra validation is necessary. We don't usually error outright on incompatible properties, we usually early return instead. For example, grade doesn't error on non-tech weapons.

I have switched to itemIsOfType, which also resolved the issue. Sorry for the confusion, I was trying item.isOfType without success.

AlexClark:
I ran the linter, it added a space to address the linting formatting.

CarlosFdez:
TY for the PR. I hope it ends up working out for you.

PR #21329, opened 2026-01-25
Fix scaling item price by size in SF2e

TikaelSol:
PF2e
image

SF2e
image


CarlosFdez:
While this weirdness will still exist and analog is also a thing, I think we can better solve this in most cases by changing physical/document.ts line 352 to also include tech items. That will at least make it slightly more compatible at least.

> Higher-level magic and tech items that cost significantly more than 8 times the cost of a mundane item use their listed Price regardless of size.

Though its still possible to do both, but this one will need a comment and some page references, its too weird to go without one imo.

PR #21342, opened 2026-01-26
Add missing monster strike traits from Alien Core

mechamaya:
Many of the monsters in Alien Core are missing traits from SF2e like boost, expend, powered, and others. Even if some of these aren't automated on NPCs yet, I figured it would be good to get this data on the creatures now so it's at least listed. ~~(Many of these creatures also had Mag traits but the system doesn't accept the mag trait yet)~~

### DIFF
--- a/packs/sf2e/alien-core-bestiary/aeon-guard-commander.json
+++ b/packs/sf2e/alien-core-bestiary/aeon-guard-commander.json
@@ -510,7 +510,10 @@
                 "slug": "boom-pistol",
                 "traits": {
                     "value": [
+                        "boost-1d6",
+                        "expend-2",
                         "razing",
+                        "reload-1",
                         "tech",
                         "tracking-1"
                     ]
--- a/packs/sf2e/alien-core-bestiary/aeon-guard-trooper.json
+++ b/packs/sf2e/alien-core-bestiary/aeon-guard-trooper.json
@@ -515,6 +515,10 @@
                 "slug": "aeon-rifle",
                 "traits": {
                     "value": [
+                        "aeon",
+                        "caster",
+                        "expend-1",
+                        "reload-1",
                         "tech"
                     ]
                 }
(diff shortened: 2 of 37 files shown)


CarlosFdez:
The system does support mag but only when its tied to a linked weapon. Do these creatures with mag not list a integrated weapon in their inventory in the actual book or on foundry.

mechamaya:
I believe they do have linked weapons yeah, but the system wasn't allowing me to enter Mag traits in the strike field. I'll double check if maybe those links got broken when we moved from Anachronism to system.

mechamaya:
Ok the links still appear to be there but the issue persists. Like take the Aeon Guard Commander for instance, their Advanced Boom Pistol has a linked Boom Pistol strike, and that's supposed to have Mag 20, but mag traits aren't recognized.

CarlosFdez:
Even on the NPC sheet? I'll double check later then.

CarlosFdez:
> Ok the links still appear to be there but the issue persists. Like take the Aeon Guard Commander for instance, their Advanced Boom Pistol has a linked Boom Pistol strike, and that's supposed to have Mag 20, but mag traits aren't recognized.

It works for me?
image

mechamaya:
Alright then. Looks like mag is working, though I'm still unsure how it's actually generating it from the linked item. Mag 20 is different from the boom pistol's usual 10 charge battery. Does it see that the actor has 2 tactical batteries which have charge 20 each? I guess it's smarter than I thought.

Then disregard the mag trait stuff, but the PR still adds in these missing weapon traits to Alien Core creatures:
Boost, expend, reload, powered, arc, aeon, caster, unwieldy and a few others that we may have just missed during data entry.

These creatures already had these traits in Starfinder Anachronism, but they must have been lost during the transition to the system data because nearly every instance of them are gone now.

CarlosFdez:
sorry yeah. This was a data PR, but I was concerned that something might have broken when you mentioned mag, or that it didn't handle a specific use case.  Yeah it checks the batteries in the inventory.

Issue #20599, opened 2025-11-03
[PF2e] Auras with [visual] trait get “blind spots” at specific squares near map edges


Rutz179:
### What happened?

Hi all,

I am running PF2e 7.6.4 on Foundry VTT v13-3.4.8 and I am getting a very specific issue with auras that have the visual trait. No additional modules or basemaps were used.

What happens

![Image](https://github.com/user-attachments/assets/31574c5a-d668-40f0-91d3-c8625c9941b4)

Figure 1. Specific squares that produce the blind spot, excluding the center square.

    I place a token with a 30 ft aura that is set to be visible.

    If that aura sits close to both edges of the map (for example, token is near bottom and right edges), some squares inside the aura do not receive the effect.

![Image](https://github.com/user-attachments/assets/a9230309-24a4-444f-a4cb-7d3850fb274a)

Figure 2. These six tokens experience blind spots

    Only tokens that stand on those exact squares fail to give the aura effect.

    If I move the token one square in any direction, the aura works again.

Things I tested

    Different map sizes (4000x3000 e.g.)

    With and without light

    With and without walls

![Image](https://github.com/user-attachments/assets/467f55d8-0b70-475f-bae2-aa71ea01a049)

Figure 3. When the aura was increased to 40 ft, the blind spot shifted 2 squares to the right and 3 squares downward.

![Image](https://github.com/user-attachments/assets/ce539105-20fb-46af-8b18-aa58da4ed87a)

Figure 4. After the token was moved just one square south, the aura behaved normally again.

    Different aura sizes (20 ft, 30 ft, 40 ft). The blind spot shifts with the radius.

    If I remove the visual trait, the issue goes away.

    If I just avoid those exact squares, it also goes away.

    So it looks like it is related to how visible auras are drawn or checked near map boundaries.

What it looks like

    Token is clearly inside 30 ft.

    Aura UI shows the token as if it should be affected.

    But the character does not actually get the aura benefit.

    Only at some very specific coordinates near the edge.

My guess

This feels like some kind of rounding or clipping near the canvas edge when the aura is both visible and touching the map boundary. Maybe PF2e’s aura detection uses the rendered shape when visible is on. Maybe Foundry is cutting part of the shape off when it hits the edge.

Question

    Has anyone seen this?

    Is this a known issue with PF2e auras on v13?

    Is there a module setting or system setting that forces aura detection to use distance instead of the rendered shape?

[...]

Issue #20478, opened 2025-10-15
doors sometimes block auras in unexpected ways.


Drental:
This effect seems to be difficult to recreate between different instances, I am happy to provide a login to the instance I used to recreate a version of this issue shown in the screenshots below:

I have seen or heard of this issue on three other instances, usually the whole aura is completely "supressed" when the door is open. I should be able to get access to one other instance beyond mine if needed for testing.

<img width="846" height="819" alt="Image" src="https://github.com/user-attachments/assets/171fa18a-9314-46da-b7e7-784cb0df2f5b" />

<img width="820" height="750" alt="Image" src="https://github.com/user-attachments/assets/5128345e-fb02-48ca-94fd-044b04792495" />

And this is the used actor and scene:

[fvtt-Scene-old-fishery-lTqy7zj584cjTnVD.json](https://github.com/user-attachments/files/22932873/fvtt-Scene-old-fishery-lTqy7zj584cjTnVD.json)
[fvtt-Actor-renzi-ToZmCIpN4BxOWEkA.json](https://github.com/user-attachments/files/22932872/fvtt-Actor-renzi-ToZmCIpN4BxOWEkA.json)

PR #21356, opened 2026-01-27
Fix aura-blind areas appearing when aura touches a wall

Cuingamehtar:
Make `getAreaSquares` use radius in pixels instead of bounding box for polygons.
I still think it treats a symptom, and not a cause, but that would require delving deep into how core Foundry handles polygon intersection beyond what I already did, and I don't want to.

Handles https://github.com/foundryvtt/pf2e/issues/20478 and https://github.com/foundryvtt/pf2e/issues/20599

Before:

https://github.com/user-attachments/assets/8507cdb5-e44f-48de-a5fd-6fd7bf31bd78

After:

https://github.com/user-attachments/assets/346a2bcd-b035-49d8-b2b8-6a844badc859

### DIFF
--- a/src/module/canvas/token/aura/util.ts
+++ b/src/module/canvas/token/aura/util.ts
@@ -48,7 +48,7 @@ export function getAreaSquares(data: GetAreaSquaresParams): EffectAreaSquare[] {
         }),
     );
 
-    const squareGrid = Array.from({ length: rowCount * rowCount }, (_, i) => {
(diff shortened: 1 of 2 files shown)


Cuingamehtar:
My understanding is that Foundry has issues when calculating polygon intersections that have parallel edges with small offset. Polygons also pad their bounding boxes by 1 px, which might be the cause of the issue with a parallel wall.

PR #21360, opened 2026-01-28
Add support for hardlight handwraps

CarlosFdez:
This uses the suggestion of including `{handwraps?.id}-attack` in the domains so that the nested upgrade subitems also apply those unarmed strikes, without having to clone the upgrades themselves. If this doesn't work out we can change to cloning them, but so far this seems to work pretty well.

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -48,7 +48,8 @@ import { getPropertyRuneDegreeAdjustments, getPropertyRuneStrikeAdjustments } fr
 import type { EffectAreaShape, ItemType } from "@item/types.ts";
 import type { WeaponSource } from "@item/weapon/data.ts";
 import { processTwoHandTrait } from "@item/weapon/helpers.ts";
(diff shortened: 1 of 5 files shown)


src/module/actor/character/document.ts
```diff
@@ -1026,46 +1027,58 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         const unarmedRunes = fu.deepClone(handwraps?._source.system.runes) ?? { potency: 0, striking: 0, property: [] };
```
Cuingamehtar, line 1027: Can you also change so that `unarmedRunes` don't use source data? They currently don't work with REs that improve runes (the upgrades will show on the handwraps but won't transfer to Strikes)

CarlosFdez: That can be done and I saw the issue but I'd rather not bundle that under a different thing.

Cuingamehtar: I understand

PR #21364, opened 2026-01-28
Update Specialty Crafting feat (SF2e) and delete incorrect condition links

Longstrider:
Also added the missing Translate spell link to the Chattermander feat.


static/lang/re-en.json
```diff
@@ -8015,6 +8015,19 @@
...
+                "Chemicals": "Chemicals",
+                "Computer": "Computer",
+                "Engineering": "Engineering",
+                "Gunsmithing": "Gunsmithing",
+                "LifeScience": "Life Science",
+                "Prompt": "Select a specialty.",
+                "Robotics": "Robotics",
+                "Textiles": "Textiles"
+            },   
             "Spellshape": {
```
TikaelSol: There's already a Specialty Crafting list elsewhere in the file, we should just expand that one

PR #21368, opened 2026-01-29
Distinguish null from 0 saves in hazards

CarlosFdez:
Migration will be run as a followup.

### DIFF
--- a/src/module/actor/hazard/document.ts
+++ b/src/module/actor/hazard/document.ts
@@ -182,10 +182,7 @@ class HazardPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | 
             const save = system.saves[saveType];
             const label = game.i18n.localize(CONFIG.PF2E.saves[saveType]);
             const base = save.value;
-
-            // Saving Throws with a value of 0 are not usable by the hazard
-            // Later on we'll need to explicitly check for null, since 0 is supposed to be valid
-            if (!base) return saves;
+            if (base === null) return saves;
 
             const statistic = new Statistic(this, {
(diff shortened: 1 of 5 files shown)


src/module/actor/hazard/document.ts
```diff
@@ -182,10 +182,7 @@ class HazardPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e |
             const save = system.saves[saveType];
             const label = game.i18n.localize(CONFIG.PF2E.saves[saveType]);
             const base = save.value;
-
-            // Saving Throws with a value of 0 are not usable by the hazard
-            // Later on we'll need to explicitly check for null, since 0 is supposed to be valid
-            if (!base) return saves;
+            if (typeof base !== "number") return saves;
```
stwlam: Isn't this either a number or null?

src/module/migration/migrations/955-hazard-null-saves.ts
```diff
@@ -0,0 +1,25 @@
...
+        if (source.type !== "hazard") return;
+
+        // This migration is not idempotent and cannot be run multiple times
+        const version = source.system._migration?.version;
+        if (version && version >= this.version) return;
+
+        for (const save of Object.values(source.system.saves)) {
+            if (save.value === 0) {
+                save.value = null;
+            }
```
stwlam:
```suggestion
            save.value ||= null;
```

src/module/actor/hazard/document.ts
CarlosFdez: Just some caution that was unnecessary due to data models.

PR #21373, opened 2026-01-29
Fix Armor Expertise resistances on SF2e

surged20:
SF2e's Armor Expertise rule element references
@armor.system.traits.config.resilient in its formula, but this property is only set when armor has a resilient-X trait. This caused formula resolution failures on SF2e actors with Armor Expertise when wearing armor without the resilient trait.

Initialize traits.config.resilient to 0 during prepareBaseData to ensure the property is always defined, matching the pattern used by PF2e's system.runes.potency field (which defaults to 0 via schema).

### DIFF
--- a/src/module/item/base/data/model.ts
+++ b/src/module/item/base/data/model.ts
@@ -105,6 +105,7 @@ class TraitConfigField extends PrunedSchemaField<TraitConfigSchema> {
                 }),
                 { required: false, nullable: false },
             ),
+            resilient: new fields.NumberField({ required: false, nullable: false, initial: 0, min: 0 }),
         });
     }
 }
@@ -117,6 +118,7 @@ type TraitConfigSchema = {
         false,
         false
     >;
+    resilient: fields.NumberField<number, number, false, false, true>;
 };
 
 type ModularConfigSchema = {
--- a/src/module/item/base/data/system.ts
+++ b/src/module/item/base/data/system.ts
@@ -28,7 +28,7 @@ interface TraitConfig {
     deadly?: string;
     fatal?: string;
     modular: ModularConfig[] | undefined;
-    resilient?: number;
+    resilient: number;
     thrown?: number;
     tracking?: number;
     versatile?: DamageType[];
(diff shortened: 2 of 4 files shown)


stwlam:
Can we update the data field definition? It should cast any null value to 0

surged20:
> Can we update the data field definition? It should cast any null value to 0

Yes, I see for melee items, it's using TraitConfigField in MeleeSystemData so this works. But for armor items, it's just using an interface with TraitConfig and so no model initialization logic (so trying this I see I still need an explicit janky `{resilient:0}` initializer for TraitConfig where that interface is used in document/helpers). It seems it would need a refactor to add something like ArmorSystemData that uses TraitConfigField properly. What's the recommended path here or did I miss something?

CarlosFdez:
Late reply but I think running the weapon through the trait config field at certain points is fine as a temporary workaround until weapons are data models.

surged20:
> Late reply but I think running the weapon through the trait config field at certain points is fine as a temporary workaround until weapons are data models.

Updated with this approach. I initially tried to add a clean traits helper that's used in addOrUpgradeTrait, but saw the build tooling doesn't like that since it needs to resolve runtime interfaces. Fell back to using  the hardcoded init in that helper like with modular.

FWIW, Obodaya (Level 7) is an existing actor that triggers this.

CarlosFdez:
I'm looking at this and thinking ahead to future trait config work. I'm not sure about this solution, but I don't have a a good alternative in mind.

So the trait config feature would pop up an input in the details when a trait becomes present. Said field should validate the allowed values, for example, resilient would have a minimum of 1. But this solution requires resilient to have a value of 0, which is an invalid value when the resilient trait is present. For this to work, we'd need to make it so that resilient is always present in sf2e armor, and hidden when 0 or something, which I'm not too big a fan of.

For a solution where fetching resilient returns 0, we would need some way to define a "value when missing". So no tracking/resilient would have 0, and when the trait is present the min is 1. But I'm not sure of a clean way to do that yet.

PR #21379, opened 2026-01-30
Add Starfinder Pact Standard (AG) to world clock options

mechamaya:
Adds Pact Standard time measured in After Gap (AG), with an offset of -1700 years as the current year should be 326 AG. Includes localization for the Starfinder days of the week which are just Firstday, Seconday, etc.


CarlosFdez:
You should probably make the default be based on the SYSTEM_ID

mechamaya:
Yeah good point, I was wondering if we wanted to do that or not. So I just set the initial choice based on system.

CarlosFdez (approved):

CarlosFdez:
Thank you for the PR!

PR #21389, opened 2026-01-31
Correct usage on Name Pendant

### LINKED ISSUE #21388: Name pendant usage and predicate is incorrect.
The Name Pendant (https://2e.aonprd.com/Equipment.aspx?ID=3974) has "held in 1 hand" listed for usage, when it should be "worn". Its bonus also only applied against effects that explicitly have the "mental" trait, even though the "emotion" trait states that all effects with said trait have the "mental" trait implicitly as well.

ammalagonc:
- Closes #21388

Predicates look correct.
image

### DIFF
--- a/packs/pf2e/equipment/name-pendant.json
+++ b/packs/pf2e/equipment/name-pendant.json
@@ -55,7 +55,7 @@
             ]
         },
         "usage": {
-            "value": "held-in-one-hand"
+            "value": "worn"
         }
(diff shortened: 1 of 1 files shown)


TikaelSol:
I think with the traits they are saying we should have an `or: mental, emotion`, but we should instead just come up with a cleaner way to handle these kind of trait riders.

PR #21395, opened 2026-01-31
Fix Status Effect Duplication

### LINKED ISSUE #21391: Condition display prints multiple times at start of turn with multiple GMs
See #20858 

Item transfers from the stash now work without a hitch, but it seems that there's still the same issue with combat status effects. It even looks like the cause is the same: https://github.com/foundryvtt/pf2e/blob/99a8f708724f3a7dff253dd432ade6ff4ef9c52f/src/module/canvas/status-effects.ts#L184 contains the same `isGM()` check which should be `isActiveGM()` for it to not run multiple times.

Ashgar225:
This changes the `isGM()` check in status-effects.ts to `isActiveGM()`, preventing it from running multiple times.

This should resolve #21391

### DIFF
--- a/src/module/canvas/status-effects.ts
+++ b/src/module/canvas/status-effects.ts
@@ -181,7 +181,7 @@ export class StatusEffects {
 
     /** Called by `EncounterPF2e#_onUpdate` */
     static onUpdateEncounter(encounter: EncounterPF2e): void {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Thank you for the PR

PR #21398, opened 2026-01-31
Change SF2e Summon Celestial spell to a unique entry

Longstrider:
Technically not a full duplicate, but the change is really minor and doesn't contradict the SF lore. Decided to make it a separate entry just in case.

Also added the Stitch Flesh feat from GotGW Player's Guide.

### DIFF
--- a/packs/sf2e/feats/skill/level-1/stitch-flesh.json
+++ b/packs/sf2e/feats/skill/level-1/stitch-flesh.json
@@ -12,7 +12,7 @@
         },
         "category": "skill",
         "description": {
-            "value": "<p>You can use @UUID[Compendium.pf2e.actionspf2e.Item.Treat Wounds] to restore Hit Points to undead creatures, not just living ones. The techniques you use to do so vary, but all require sutures, bandages, and other tools included in a medkit. Some conditions that might raise the DC of treating undead's wounds differ from that of living creatures. For instance, the GM might increase the DC if the undead is being treated in a Pharasmin medbay or the wounds were caused by powerful vital energy.</p>"
+            "value": "<p>You can use @UUID[Compendium.sf2e.actions.Item.Treat Wounds] to restore Hit Points to undead creatures, not just living ones. The techniques you use to do so vary, but all require sutures, bandages, and other tools included in a medkit. Some conditions that might raise the DC of treating undead's wounds differ from that of living creatures. For instance, the GM might increase the DC if the undead is being treated in a Pharasmin medbay or the wounds were caused by powerful vital energy.</p>"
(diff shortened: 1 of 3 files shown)


packs/sf2e/feats/skill/level-1/stitch-flesh.json
```diff
@@ -0,0 +1,42 @@
...
+    "system": {
+        "actionType": {
+            "value": "passive"
+        },
+        "actions": {
+            "value": null
+        },
+        "category": "skill",
+        "description": {
+            "value": "<p>You can use @UUID[Compendium.pf2e.actionspf2e.Item.Treat Wounds] to restore Hit Points to undead creatures, not just living ones. The techniques you use to do so vary, but all require sutures, bandages, and other tools included in a medkit. Some conditions that might raise the DC of treating undead's wounds differ from that of living creatures. For instance, the GM might increase the DC if the undead is being treated in a Pharasmin medbay or the wounds were caused by powerful vital energy.</p>"
```
pedrogrullada: Probably should be `@UUID[Compendium.sf2e.actions.Item.Treat Wounds]`? Not gonna make a suggestion since the pipeline might not like it if it's fixed that way :D

TikaelSol:
```suggestion
            "value": "<p>You can use @UUID[Compendium.sf2e.actions.Item.Treat Wounds] to restore Hit Points to undead creatures, not just living ones. The techniques you use to do so vary, but all require sutures, bandages, and other tools included in a medkit. Some conditions that might raise the DC of treating undead's wounds differ from that of living creatures. For instance, the GM might increase the DC if the undead is being treated in a Pharasmin medbay or the wounds were caused by powerful vital energy.</p>"
```

TikaelSol: Works fine when you do it that way

packs/sf2e/feats/skill/level-1/stitch-flesh.json
```diff
@@ -0,0 +1,42 @@
...
+            "value": [
+                {
+                    "value": "trained in Medicine"
+                }
+            ]
+        },
+        "publication": {
+            "license": "ORC",
+            "remaster": true,
+            "title": " Starfinder Guilt of the Grave World Player's Guide"
```
TikaelSol:
```suggestion
            "title": "Starfinder Guilt of the Grave World Player's Guide"
```

Longstrider:
Thank you! I was in a bit of a rush when commiting this one and missed those details.

PR #21405, opened 2026-02-01
Fix logic in DamageAlteration dice face upgrades/downgrades

TikaelSol:
Noticed that the way it currently is wouldn't work quite as expected from other uses of upgrade/downgrade. If you set an upgrade to 6 then it would turn a d4 to a d6, but also upgrade any dice d6 or above to the next step up.


CarlosFdez:
This looks like a bugfix actually? It seems like what it does is prevent upgrading to a d6 upgrading past a d6 in certain circumstances.

TikaelSol:
I wasn't sure if it was intended to be that way, but yeah it is probably just a bug.

PR #21413, opened 2026-02-01
Add miscellaneous missing effects to SF2e

ammalagonc:
Aquatic Combat
Mounted
Remaining Air
Engulf & Swallow

### DIFF
--- a/build/duplicates.json
+++ b/build/duplicates.json
@@ -56,7 +56,10 @@
                 "Tremorsense",
                 "Void Healing",
                 "Wavesense"
-            ]
+            ],
+        "bestiary-effects": [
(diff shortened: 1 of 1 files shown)


pedrogrullada:
Aquatic Combat links Remaining Air in its description. The rules are the same between both games, so you can duplicate it as well.

While you're at it, you can add Effect: Engulf & Swallow Whole too. They probably have broken links to the effect.

ammalagonc:
> Aquatic Combat links Remaining Air in its description. The rules are the same between both games, so you can duplicate it as well.
> 
> While you're at it, you can add Effect: Engulf & Swallow Whole too. They probably have broken links to the effect.

Thanks Rigo!

PR #21421, opened 2026-02-02
Fix Recall Knowledge links

Cuingamehtar:


simonward:
This is because the links are missing `.Item`. If that’s repaired, the extract will strip the labels.

Cuingamehtar:
Ah, that explains why I couldn't parse them for translation.
I must be subconsciously blanking out the entry types already.

PR #21440, opened 2026-02-04
Add missing Murder in Metal City Item

mechamaya:
An item that was missed in data entry which was printed only on a card handout in Murder in Metal City, was also missed by AoN, but it's on Demiplane https://app.demiplane.com/nexus/starfinder2e/items/consecrated-keycaps

It has a usage "Installed in a Datapad"


mechamaya:
Nevermind Tikael beat me to it :D

PR #21446, opened 2026-02-04
Update tech trait description to include all portions

### LINKED ISSUE #21357: SF2e Tech Trait Description Incorrect
The current tech trait description is out of date, still using the text from the playtest which references the Machine Magic feat that doesn't exist anymore. Additionally, we should display different text on the Tech trait for Armor, Weapons, Shields, Augmentations, as they all have different properties granted to them by the trait.

> Equipment with the tech trait rely on electricity and may gain the glitching condition. Some creatures and hazards also have the tech trait, such as artificial intelligence, robots, and turrets.
> 
> Armor with the tech trait incorporate electronics, computer systems, integrated power sources, and a comm unit.
> 
> Shields with the tech trait incorporate electronics, computer systems, and integrated power sources.
> 
> Weapons with the tech trait incorporate electronics, computer systems, and power sources. Usually the weapons rely on integrated power sources (such as melee weapons that don't have the powered trait), while others drain batteries with each attack.
> 
> Cybernetic augmentations are technological, not magical, though they aren't subject to any effect or attack that targets technology unless it specifies that it affects cybernetics.

steve148:
Fixes https://github.com/foundryvtt/pf2e/issues/21357

Simple version of a fix to make sure that the tech trait in Foundry matches the official wording on Archives of Nethys.

https://2e.aonsrd.com/traits/179-tech

Issue mentioned showing a different description based on the item type. I'm new to contributing to Foundry so I went with the simple approach to start. Would be happy to implement this if desired.

Screenshot 2026-02-04 at 18 37 32


mechamaya:
Not sure if we want to just dump the entire description in the trait like this, or wait until there's a way to localize them differently depending on whether it's armor, or shields, or weapons, etc?

TikaelSol:
Thanks for the PR. As a note, we use imperative case for PR titles, basically completing the sentence "If merged, this PR will..." (This makes it easy to just use the PR title in the changelog).

For the actual PR, just move this to the en.json file instead of the SF2e overrides. Also, make sure you check wording against the book, not AoN.

steve148:
@TikaelSol Thanks for the feedback and for fixing the PR title.

Looking at the text in the book, there were two item types with differences. I think the AoN version reads better, but I'm happy to make the change to use the book's phrasing.
```
# AON
Armor with the tech trait incorporate electronics, computer systems, integrated power sources, and a comm unit.
# Book
This armor incorporates electronics, computer systems, integrated power sources, and a comm unit.

# AoN
Shields with the tech trait incorporate electronics, computer systems, and integrated power sources.
# Book
This shield incorporates electronics, computer systems, and integrated power sources.
```

I put the change in the overrides file so that the trait in pathfinder wouldn't change. Is it ok using the starfinder language for the tech trait in pathfinder worlds?

TikaelSol:
The tech trait in PF is the SF tech trait, we're ignoring the two Kingmaker items with that trait (Or, rather forcibly swapping them to use SF rules). Also we can stick to the cleaner description

steve148:
Appreciate the context! Moved the description into en.json ✅

PR #21455, opened 2026-02-06
Add folders to the Iconics compendium

ammalagonc:


TikaelSol:
The migration test (`tests/module/migration.test.ts`) references Amiri, apparently. Just have to update that path.

PR #21467, opened 2026-02-07
Add build script for pf2e and sf2e anachronism

CarlosFdez:


build/lib/compendium-pack.ts
```diff
@@ -56,9 +56,11 @@ function isItemSource(docSource: PackEntry): docSource is ItemSourcePF2e {
 }
 
 interface ConstructorParams {
+    id?: string;
     dirName: string;
     data: unknown[];
     folders: unknown[];
+    documentType?: CompendiumDocumentType;
```
stwlam: Prefer `documentName` (core is going in this direction as well)

build/lib/compendium-pack.ts
```diff
@@ -338,7 +347,7 @@ class CompendiumPack {
         }
 
         // Stamp actors and items with partial stats data so the server won't attempt to migrate
-        const systemJSON = JSON.parse(fs.readFileSync(`system.${this.systemId}.json`, { encoding: "utf-8" }));
+        const systemJSON = JSON.parse(fs.readFileSync(`system.${targetSystem}.json`, { encoding: "utf-8" }));
```
stwlam: This is always going to be `this.systemId`, no?

build/lib/level-database.ts
```diff
@@ -183,7 +184,8 @@ interface DBFolder {
 }
 
 interface LevelDatabaseOptions {
-    systemId: SystemId;
+    systemId?: SystemId;
```
stwlam: Why optional now? These are much less obvious as to their purpose and could use some jsdoc

build/lib/compendium-pack.ts
CarlosFdez: I also changed the internal variable name as well to match.

build/lib/compendium-pack.ts
CarlosFdez: It is, it used to not, but I got it to and forgot to adjust this.

build/lib/level-database.ts
CarlosFdez: The property is used to load a system manifest file to then use to validate the packs, however in the case of anachronism we want to load a module manifest instead. So I added manifest as an alternative way to initialize.

We could either make the core system pass in a manifest file always instead of system (since manifest file supports all use cases) or include a packageType: "system" | "module" param. What do you prefer?

CarlosFdez: It now always accepts `manifest` instead of `manifest` or `systemId`. Let me know if this is preferable.

PR #21483, opened 2026-02-08
Bump Thaumaturge martial proficiency rank to expert in Weapon Expertise

stwlam:
Plus delete and redirect legacy Thaumaturge Weapon Expertise class feature


simonward:
`ABYmUcLdxDFXEtzu` is also referenced in classes.json.

PR #21495, opened 2026-02-10
Allow modification of options.keep for pf2e.preReroll and pf2e.reroll

### LINKED ISSUE #21494: Pass keep as an object to the hooks pf2e.preReroll and pf2e.reroll so it can be modified
At the moment both `unevaluatedNewRoll` and `newRoll` are passed as an object to the hooks allowing modification of them. However, the same is not true for `options.keep` which is passed as a primitive and can't be modified. It would be nice if it was wrapped up into an object so that it can be modified and use that object.

7H3LaughingMan:
* Closes #21494

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -522,12 +522,13 @@ class Check {
             );
         })();
         unevaluatedNewRoll.options.isReroll = true;
-        Hooks.callAll("pf2e.preReroll", Roll.fromJSON(oldRollJSON), unevaluatedNewRoll, resource, options.keep);
+        const optionsKeep = { keep: options.keep };
+        Hooks.callAll("pf2e.preReroll", Roll.fromJSON(oldRollJSON), unevaluatedNewRoll, resource, optionsKeep);
 
         // Evaluate the new roll and call a second hook allowing the roll to be altered
         const allowInteractive = context.rollMode !== "blindroll";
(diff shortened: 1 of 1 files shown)


stwlam:
Isn't this in the roll?

7H3LaughingMan:
> Isn't this in the roll?

The system uses `options.keep` from the parameters of  `rerollFromMessage` to determine which roll to keep, the problem is that since it just passes it as `options.keep` it's a primitive type so Javascript is just passing around a copy of the data instead of a reference which means it can't be modified.

stwlam:
`roll.options.keep`

7H3LaughingMan:
The system doesn't use that to determine which roll to keep, it uses `options.keep` from the parameters.

https://github.com/foundryvtt/pf2e/blob/dddda4519e1d7ca4163ca64276de596b1278513b/src/module/system/check/check.ts#L536-L544

stwlam:
I don't think we can do this until system version 8. To the extent the system has an API at all, this is part of it.

7H3LaughingMan:
I mean, this works the same way as how it currently works with `unevaluatedNewRoll` and `newRoll` which can be modified at the moment since the object itself is being passed. Modules are already using these hooks to modify the new roll.

CarlosFdez:
Any module listening to the hook would be receiving an object instead of a string now, so it would be a breaking change. I think this is a good idea but I agree that it should wait until system version 8 (which is when everything breaks anyways).

PR #21512, opened 2026-02-12
Update predicate for commanders banner effect

MrVauxs:
Makes it consistent with Courageous Anthem, making it work for effects such as Demoralize that are not `items` and thus do not have `item:trait:` prefixes.


TikaelSol:
We need to fix the action macros rather than break data

MrVauxs:
Does that mean Courageous Anthem is also broken?

MrVauxs:
I'll close this stale issue.

PR #21513, opened 2026-02-12
Use token assets for all SF2e iconic prototype tokens

surged20:
SF2e iconic prototype tokens currently point at the portrait assets.
Update them to use the token assets.

### DIFF
--- a/packs/sf2e/iconics/chk-chk/chk-chk-level-1.json
+++ b/packs/sf2e/iconics/chk-chk/chk-chk-level-1.json
@@ -2730,7 +2730,7 @@
         "texture": {
             "scaleX": 1.2,
             "scaleY": 1.2,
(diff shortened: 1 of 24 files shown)


static/icons/iconics/tokens/obozaya.webp
pedrogrullada, line 1: Unless we migrate, we might need to stick with the typo for now as to not break images on imported copies of this actor.

As an aside, we use imperative case for PR titles. It needs to complete the sentence "If merged, this PR will...", which makes it easy for you to be credited when the changelog is assembled.

surged20: I dropped the asset rename completely for now (doesn't seem worth a migration right now) and left the functional token asset fix. Updated PR description.

PR #21514, opened 2026-02-12
Add support for troop segment tokens

CarlosFdez:
https://github.com/user-attachments/assets/bf31fd35-6ff6-488a-8764-61f7a12b2014

https://github.com/user-attachments/assets/19dd438b-8dc7-4bdb-904d-43c3dabae43a

The following features are missing, but should be fine to split off since they may require new discussions:
* Manually adding segments in the case of accidental deletes
* Some way for users to know which tokens are interlinked
* Flanking using the targeted token (mirror thaumaturge needs the same handling, so we can maybe do it in a separate PR)
* Handling for actor linked troops (or non-handling?)

I don't know if we should try to also propagate floaty text. 

As for the code itself, I'm uncertain about the parent/child naming. Its more of a real/illusion sort of illusion, but I don't know what to call that. `hasChildTokens` is a performance consideration so we're not checking every single token on non-troops for everything. It is also unclear if troops should be one actor or four (conditions affecting only one is odd, since you can always attack with the best one, and I have no idea how I'd handle draining a single segment). If troops should be an actor per segment with linked HP, it will need a different implementation.

### DIFF
--- a/src/module/actor/npc/document.ts
+++ b/src/module/actor/npc/document.ts
@@ -205,7 +205,7 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
             system.attributes.hp = stat;
             setHitPointsRollOptions(this);
 
-            // Troop Thresholds
+            // Troop Thresholds and special segment token size
             const hpMax = stat.max;
             system.attributes.hp.thresholds = null;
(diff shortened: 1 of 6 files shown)


CarlosFdez:
Closed in favor of https://github.com/foundryvtt/pf2e/pull/21564

PR #21519, opened 2026-02-13
Brush up Nodocite Incarcerator's Catchapon

Longstrider:


simonward:
Would it be better to use action skill overrides like `[[/act disable-device skill=crafting dc=15]]`?

Longstrider:
> Would it be better to use action skill overrides like `[[/act disable-device skill=crafting dc=15]]`?

For most cases - definitely. I initially considered it in this case as well, but eventually chose `@Checks` for two reasons:
1. Labels fit the description better. Like "... or a [DC 15 Computers], [DC 15 Crafting], or [DC 15 Thievery] check to Disable Device" instead of "... or a [Disable Device (DC 15 Computers)], [Disable Device (DC 15 Crafting)], or [Disable Device (DC 15 Thievery)] check to Disable Device".
2. The original actions' degrees of success are irrelevant for this ability, so there's no real reason to use `[[/act]]` for those checks in my opinion.

SpartanCPA:
One benefit of [[/act]] in lieu of @Check is that the display the action glyphs, for GMs who may not realize that Disable a Device requires 2 actions in combat

Longstrider:
Alright, changed it all to [[/act]]  with fitting labels. Can get rid of those, if needed.

PR #21525, opened 2026-02-13
Add a parameter to hide /act skill label

Cuingamehtar:
image

image

The goal is to use those in descriptions and hazard disables when the button label is the same as the used skill (to avoid `Athletics (Athletics)`).
I wanted to have an option to display the default statistic too, but the current type of actions collection doesn't allow getting the default statistic of the action, so I skipped it for now. The parameter name was chosen similar to "show-dc", but as I don't think there is a need to hide it from non-GM, the choice would be simpler.
I'm not sure what would be the better name for the parameter value. Currently it's "false" or "never" to not show the skill label (potentially expanding to "true" or "always" in case there is ever a need to show the default skill label). Those are just my suggestions.

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -464,7 +464,13 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
         // statistic
         const statistic = (params["statistic"] || params["stat"] || params["skill"])?.trim();
 
-        if ((dc && showDC) || statistic) {
+        // "false" or "never" to never show
+        const showStatisticOverride = (params["show-statistic"] || params["show-stat"] || params["show-skill"])
+            ?.trim()
+            .toLowerCase();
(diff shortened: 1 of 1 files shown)


iDantar:
Hi, trying my hand at reviews - hope this helps :)
Noticed small thing on the loop that copies macro options onto `data-pf2-…`: `show-statistic` (and aliases) still get written as `data-pf2-show-statistic`, but nothing reads that at click time, only the display block uses it. Was that intentional? If it’s enrichment-only, might be worth skipping it in the loop so the element doesn’t carry a flag that looks like roll config. Non-blocker, of course

p.s. looks like there's a merge conflict

Cuingamehtar:
@iDantar 
Thanks, It's not something that I considered at the time.
Since the goal of this PR was just to create more natural-looking action buttons in descriptions for cases where contextual information is sufficient (like statistic substitution in feats, or a list of disable a device buttons in hazards), there is no need to further the introduced parameter to the actual check.
I've added an ignore-list for this parameter.

PR #21526, opened 2026-02-13
Add defense filter to Compendium Browser spell tab

In3luki:
Since nobody has pinged me yet for breaking the browser filter, I’ll continue with smaller PRs pulled out of the big one.

image


stwlam:
Can we have a checkbox for "basic" that disables non-applicable options?

PR #21538, opened 2026-02-15
Include party tokens in door reach checks

surged20:
When party members are collected into a party token, check if the party token is in reach of doors so players can operate them.

### DIFF
--- a/src/module/canvas/door-control.ts
+++ b/src/module/canvas/door-control.ts
@@ -18,8 +18,13 @@ export class DoorControlPF2e extends fc.containers.DoorControl {
             { x: wallDoc.c[2], y: wallDoc.c[3] },
             this.center,
         ];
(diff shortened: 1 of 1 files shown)


CarlosFdez:
It looks like it already works with parties if the party actor is controlled. Is the purpose to also work for uncontrolled party tokens?

stwlam:
This is going to allow door opening in situations that aren't intended. As Supe noted, party tokens are already included but must be controlled.

surged20:
Ok, rereading the activeTokens support, I see now that this should always be the case and I reached the wrong conclusion. When I first saw this happen and was testing it, the user had ownership permission and thus controlled. In any case, sorry for the noise, went back to my production world with same user and can't even reproduce the reach failure. PEBCAK.

PR #21547, opened 2026-02-16
Add roll options for area-fire and auto-fire saves

### LINKED ISSUE #21481: Rolled area and auto fire saving throws don't include their action slug
The message itself contains it, but the rolled saving throw by the target does not have `origin:action:slug:area-fire`, making it not possible to have conditional modifiers against area- or auto-fire saves. Using `item:trait:area` and `item:trait:auto` are close approximations, but not quite correct, especially if a Strike with such a weapon has a Note with a saving throw.

Image

surged20:
Fixes #21481

### DIFF
--- a/src/module/chat-message/listeners/cards.ts
+++ b/src/module/chat-message/listeners/cards.ts
@@ -220,8 +220,18 @@ class ChatCards {
                     return;
                 }
                 case "roll-area-save": {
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I'm checking actual examples and it does not seem like these options are present in other actions like saving throws to shoving shot nor any spell. I don't know if this is what should happen. I'm going to poke the issue for more details.

PR #21549, opened 2026-02-17
Fix Rule Element contention on Sextant of the Night

AshenShade:
added a slug to one of colliding rule elements on sextant of the night so both FlatModifiers on the item function as intended

### DIFF
--- a/packs/pf2e/equipment/sextant-of-the-night.json
+++ b/packs/pf2e/equipment/sextant-of-the-night.json
@@ -38,6 +38,7 @@
             {
                 "key": "FlatModifier",
                 "selector": "survival",
+                "slug": "sextant-base-bonus",
                 "type": "item",
                 "value": 1
             },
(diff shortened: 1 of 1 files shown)


packs/pf2e/equipment/sextant-of-the-night.json
```diff
@@ -47,6 +47,7 @@
                     "action:sense-direction"
                 ],
                 "selector": "survival",
+                "slug": "sextant-of-the-night-action",
```
ammalagonc:
```suggestion
                "slug": "sextant-sense-direction-bonus",
```

ammalagonc:
Consider adding slug to the base flat modifier as well to ensure they don't collide.

AshenShade:
> Consider adding slug to the base flat modifier as well to ensure they don't collide.

I used the existing Bracers of Strength as a template for the change, so did not add it to the other because the existing example did not do so (though my original testing on Foundry did)
https://github.com/foundryvtt/pf2e/blob/v13-dev/packs/pf2e/equipment/bracers-of-strength.json

pedrogrullada:
Yeah it doesn't really need it necessarily, it will inherit the item's slug if left alone, which will be different from your explicitly stated slug. But it doesn't hurt either.

ammalagonc (approved):

PR #21554, opened 2026-02-17
Update SF2e PFS tab for SFS2

surged20:
Factions, reputation, and level bump aren't used in SFS2 play.


CarlosFdez:
If possible, I'd prefer passing in the system id in the template data and using the eq helper instead.

surged20:
> If possible, I'd prefer passing in the system id in the template data and using the eq helper instead.

just updated with this change.

CarlosFdez:
Thanks for the PR

PR #21557, opened 2026-02-18
Fix Walking Armory penalties

### LINKED ISSUE #21531: [SF2E - Data Issue] Walking Armory - speed penalty issue
For walking armory:

- On a 10ft speed penalty armor, if you have both enough STR and CON, it reduces the penalty twice.

Image

I believe adding `"lt":["attribute:str:mod","armor:strength"]}` to the predicate should fix it.

> Your endurance training empowers you to carry a load of heavy equipment. When determining your Strength for using armor, you can instead choose to use your Constitution modifier. If you already meet the Strength requirement for the armor you're wearing, reduce that armor's Bulk by 1 (to a minimum of 1).
> 
> In addition, increase the amount of Bulk you can carry before becoming encumbered by half your Constitution modifier rounded up. Increase your maximum carried Bulk by your Constitution modifier.

steve148:
Fixes https://github.com/foundryvtt/pf2e/issues/21531

Initially tried putting `lt` predicate in the same object as the 'gte' clause.

```
{
    "gte": [
    "attribute:con:mod",
    "armor:strength"
    ],
    "lt": [
    "attribute:str:mod",
    "armor:strength"
    ]
}
```

This didn't work, giving an error message saying "predicate.1: must be a recognized predication statement". Next thing tried was using an `and` predicate to combine the two clauses together.

```
{
    "and": [
        {
            "gte": [
                "attribute:con:mod",
                "armor:strength"
            ]
        },
        {
            "lt": [
                "attribute:str:mod",
                "armor:strength"
            ]
        }
    ]
}
```

That did work! Confirmed all of this on a test actor with a strength of 3 and a constitution of 4 with Commercial Aegis Series armor. Where before there would be no speed reduction, this change makes sure that there is a speed penalty of -5 (reduced from -10).

The issue didn't mention this, but the same problem existed for the check penalty on armor. I applied the same fix to that as for the the speed penalty.

## Testing

Recording which shows the fix to speed and check penalty against this change.

https://github.com/user-attachments/assets/1276a6e0-f0b8-4161-835d-6cb45d1226d5

### DIFF
--- a/packs/sf2e/class-features/soldier/walking-armory.json
+++ b/packs/sf2e/class-features/soldier/walking-armory.json
@@ -35,6 +35,12 @@
                             "attribute:con:mod",
                             "armor:strength"
                         ]
+                    },
+                    {
+                        "lt": [
+                            "attribute:str:mod",
+                            "armor:strength"
+                        ]
                     }
                 ],
                 "selector": "skill-check",
@@ -46,10 +52,20 @@
                 "predicate": [
                     "penalty:slug:armor-speed-penalty",
                     {
-                        "gte": [
-                            "attribute:con:mod",
-                            "armor:strength"
-                        ]
+                        "and": [
+                            {
+                                "gte": [
+                                    "attribute:con:mod",
+                                    "armor:strength"
+                                ]
+                            },
+                            {
+                                "lt": [
+                                    "attribute:str:mod",
+                                    "armor:strength"
+                                ]
+                            }
+                        ]   
                     }
                 ],
                 "selector": "speed",


TikaelSol:
We put PR titles in imperative case. Basically Pr titles should finish the sentence "If merged, this PR will...". This makes it easy to use the PR title for the changelog.

steve148:
@TikaelSol sorry for not fixing that, you told me this in a previous PR.

packs/sf2e/class-features/soldier/walking-armory.json
```diff
@@ -31,10 +31,20 @@
...
+                            },
+                            {
+                                "lt": [
+                                    "attribute:str:mod",
+                                    "armor:strength"
+                                ]
+                            }
+                        ]   
                     }
                 ],
```
TikaelSol, line 45: The and isn't necessary, so everything in it can come up a level. The top level of a predicate is implicitly an `and`.

steve148: Done! Amended the commit with the change. I had originally tried putting multiple conditions in a single object as I thought that would be equivalent to an "and", but it didn't work, leading me to use the verbose `and` predicate to get it to work.

```
{
    "gte": [
    "attribute:con:mod",
    "armor:strength"
    ],
    "lt": [
    "attribute:str:mod",
    "armor:strength"
    ]
}
```

PR #21559, opened 2026-02-18
Brush up Variable Gravity spell and effect

simonward:


packs/pf2e/spell-effects/spell-effect-variable-gravity-high-gravity.json
```diff
@@ -33,7 +38,10 @@
                         ]
                     }
                 ],
-                "selector": "saving-throw",
+                "selector": [
+                    "saving-throw",
+                    "saving-throw-dc"
```
pedrogrullada: Are you sure this works? The saving-throw-dc thing, that is. I only see it mentioned/used in one other effect, which raises some suspicions.

simonward: Hm. Yeah, maybe it doesn't have any effect that the saving-throw selector won't handle.

Issue #20169, opened 2025-09-12
Troops need data to express individual constituent-creature size


Cuingamehtar:
Although it moves as if its Gargantuan, troops usually consist of Medium size creatures.
Currently, a creature on top of a 20-foot cliff is considered the same distance away as if it was on the ground.

Steps to reproduce:
1. Place a troop token.
2. Place a normal creature 10 feet away from the troop and then give it +20 foot elevation.

Expected distance: 25 feet (2 diagonals + 2 cubes up), actual distance: 10 feet.

<img width="449" height="331" alt="Image" src="https://github.com/user-attachments/assets/841cef2a-129d-4110-aab9-f7609b16f16f" />

PR #21564, opened 2026-02-19
Support creation of troop segment actors

CarlosFdez:
There is no handling for HP syncing yet, nor death handling. I want to be sure this looks ok before continuing on.

### DIFF
--- a/src/module/actor/npc/document.ts
+++ b/src/module/actor/npc/document.ts
@@ -214,6 +214,8 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
                     { hp: Math.floor((hpMax * 2) / 3), segments: 3 },
                     { hp: Math.floor(hpMax / 3), segments: 2 },
                 ];
+                this.system.traits.size.wide = 10;
+                this.system.traits.size.long = 10;
             }
         }
 
--- a/src/module/encounter/document.ts
+++ b/src/module/encounter/document.ts
@@ -8,6 +8,7 @@ import type {
     DatabaseCreateCallbackOptions,
     DatabaseCreateOperation,
     DatabaseDeleteCallbackOptions,
+    DatabaseDeleteOperation,
     DatabaseUpdateCallbackOptions,
 } from "@common/abstract/_types.d.mts";
 import type EmbeddedCollection from "@common/abstract/embedded-collection.d.mts";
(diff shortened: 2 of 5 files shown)


src/module/actor/npc/document.ts
```diff
@@ -214,6 +214,11 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
                     { hp: Math.floor((hpMax * 2) / 3), segments: 3 },
                     { hp: Math.floor(hpMax / 3), segments: 2 },
                 ];
+                this.system.traits.size = new ActorSizePF2e({
+                    value: this.system.traits.size.value,
+                    wide: 10,
+                    long: 10,
+                });
```
stwlam: Is this not already constructed for NPCs?

CarlosFdez: No, only in the base actor. It is standard practice to replace it though, familiars and characters both do that.

In this case we need an actor size that continues to report gargantuan but creates a 2x2 actor.

stwlam: you can adjust the wide/long values after it's constructed

Cuingamehtar: Would it be reasonable to request height here? So that a segment is not a cube for measuring vertical distance.

stwlam: height of what?

Cuingamehtar: Of an actor/segment. Troop is made of individual creatures, which are not necessary Gangantuan/Large size each.
Current distance calculation considers tokens to be cubes stretching from the ground up, which results in distance calculation that is suitable for individual creatures, but doesn't necessary make sense for large groups of smaller creatures (https://github.com/foundryvtt/pf2e/issues/20169).

CarlosFdez: I'm poking into it and while I agree, that doesn't seem to be part of the actor size object.

The overwrites worked. Dunno why I though they'd be read only. Must have skimmed.

CarlosFdez: It looks like the height is the lesser of long/wide, and is part of the actor dimensions property. Its already handled.

stwlam: > Of an actor/segment.

I meant height of what value? We can't just assign some arbitrary height. The way it's setup in this PR will result in a token (z) height of 10 feet, making a cube.

Cuingamehtar: That would require a new field on an Actor for the size of the creatures that form the troop.

But if it is beyond the scope of this PR, there is no rush to consider that now.

PR #21581, opened 2026-02-21
Fix missing DC value in some check context flags

### LINKED ISSUE #21205: Rerolling an escape check tries to resolve over an undefined DC
When you make an Escape roll (either with the base macro, with an [[/act escape]] check or other, then reroll it, in resulting chat card, the DC is shown as `undefined` and the result shows Success or Failure by +NaN

Image

Tested on a world with no module with PF2e version 7.8.0

In3luki:
When  `context.dc` is a `StatisticDifficultyClass`, the DC value is a getter and is therefore non-enumerable, causing it to be dropped by the `omit` helper.

Closes #21205

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -304,7 +304,12 @@ class Check {
             title: context.title ?? "PF2E.Check.Label",
             traits: context.traits ?? [],
             substitutions,
-            dc: context.dc ? R.omit(context.dc, ["statistic"]) : null,
+            dc: context.dc
+                ? {
+                      ...R.omit(context.dc, ["statistic"]),
+                      value: context.dc.value,
+                  }
+                : null,
             skipDialog: context.skipDialog,
             isReroll: context.isReroll ?? false,
             outcome: context.outcome ?? null,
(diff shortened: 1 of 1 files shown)


CarlosFdez:
How long is the list of properties we're seeking to pull? Perhaps a pick would be cleaner?

In3luki:
Looks like picking would work and is indeed cleaner.
```ts
{
      dc: context.dc ? R.pick(context.dc, ["label", "scope", "slug", "value", "visible"]) : null,
}
```
I've looked at a check roll and an attack roll before & after and the flags look the same. The only downside is, that the list of picked properties may need to be updated if the `CheckDC` properties change.

CarlosFdez:
Hopefully typescript would take care of that. I just prefer to avoid non-multiple of 4 indentation that prettier likes to add in these situations where possible. Its either that or adding a getTraceData() equivalent that uses that and falls back to the dc.

If I remember right value is a getter because we want to be able to defer its computation, so we can't simply convert that to a variable.

In3luki:
Yeah, it's probably fine. The `dc` flag is only used for rerolls in the system and modules can request other hypothetical new properties if they need them.

PR #21598, opened 2026-02-23
Add Suspended Respiration sarcesian feature and fix journal embed

Cuingamehtar:
Only the Item with description, no automation.
I've chosen a reasonably ethereal icon for it, but haven't searched thoroughly for other options.
And replaced the embed in it's journal, as, apparently, "Add manually" is not a valid uuid.


Cuingamehtar:
Didn't notice the duplicate

PR #21602, opened 2026-02-23
Generate area fire and auto fire NPC attacks

### LINKED ISSUE #21546: Implement generating area fire and and auto fire npc attacks
Was messing around on the Starfinder2e foundry module and noticed2 things. First, attacks using either Area-Fire or Auto-Fire do not expend ammo. And secondly, generating an attack on an NPC using weapons that have either Area-Fire or Auto-Fire does not generate attacks of those types, only normal Strike actions.

surged20:
Also improve PC and NPC area fire chat cards to show item traits, range increment, and weapon name in the title.

Fixes #21546

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -45,9 +45,9 @@ import type { AbilityTrait } from "@item/ability/types.ts";
 import { ARMOR_CATEGORIES } from "@item/armor/values.ts";
 import { ActionCost } from "@item/base/data/index.ts";
 import { getPropertyRuneDegreeAdjustments, getPropertyRuneStrikeAdjustments } from "@item/physical/runes.ts";
-import type { EffectAreaShape, ItemType } from "@item/types.ts";
+import type { ItemType } from "@item/types.ts";
 import type { WeaponSource } from "@item/weapon/data.ts";
-import { processTwoHandTrait } from "@item/weapon/helpers.ts";
+import { computeWeaponArea, processTwoHandTrait } from "@item/weapon/helpers.ts";
 import { PROFICIENCY_RANKS, ZeroToFour, ZeroToTwo } from "@module/data.ts";
 import {
     extractDegreeOfSuccessAdjustments,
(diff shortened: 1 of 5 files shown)


surged20:
Couple notes here:

- Refactored some of the PC area/auto-fire support to helpers to avoid duplicating code
- Item traits and range increments previously weren't included on chat cards so added those and with the newly shared helpers so that adds display of those traits/properties for PC area/auto-fire as well
- For area-fire weapons, it just generates the general rule area-fire only attack. There could be a case where an NPC has a Soldier-like primary target ability in which case that Strike could be added manually since it's an exception case.

src/module/actor/helpers.ts
```diff
@@ -858,17 +859,30 @@ async function createAreaAttackMessage({
     const dc = statistic.dc;
 
     const key = sluggify(action, { camel: "bactrian" });
-    const title = game.i18n.localize(`PF2E.Actions.${key}.Title`);
+    const title = game.i18n.format(`PF2E.Action.${key}.Label`, { weapon: item.name });
```
CarlosFdez: This seems a bit out of scope. I'd imagine we'd get some discussion with regards to adding the weapon name into the title.

CarlosFdez:
Thanks for working on this. Could split this into separate concerns? It would be easier both for me reviewing when I get time as well as the patch notes.

Perhaps show a context menu when clicking the generation button if its an area/auto weapon? That way people can prompt if they want to generate a strike or area/auto fire.

src/module/actor/helpers.ts
surged20: I've dropped this out to just focus on the base area/auto-fire attack generation.

surged20:
I removed the title and trait improvements so they can be proposed separately.

Added a dialog that verifies which type of attack to generate (or both) while grenades skip this dialog just generate an area fire attack. This works properly with the regenerate prompt as well, whether generating only a strike, area/auto-fire, or both attacks the previous attack(s) will be replaced.

image

PR #21616, opened 2026-02-25
Sync HP between troop segments

CarlosFdez:


src/module/actor/npc/document.ts
```diff
@@ -600,6 +614,18 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
...
         this.#updatePrototypeToken(changed);
 
+        // Propagate HP updates to all sibling segments as well
+        const hpUpdates = changed.system.attributes?.hp;
+        if (hpUpdates && !options.fromTroop) {
+            const segments = this.getActiveTokens()?.[0].document.segments ?? [];
+            for (const segment of segments) {
+                const actor = segment.actor;
+                if (actor?.isOfType("npc")) {
+                    actor.update({ "system.attributes.hp": hpUpdates }, { fromTroop: true });
```
stwlam, line 622: Isn't this going to include itself?

src/module/scene/document.ts
```diff
@@ -67,9 +67,7 @@ class ScenePF2e extends Scene {
 
     override prepareData(): void {
         super.prepareData();
-        Promise.resolve().then(() => {
-            this.checkAuras();
-        });
+        this.checkAuras();
```
stwlam: ??

CarlosFdez: Gonna undo. I presumed it was just some sillinessy, but I did some testing and it looks like this is the equivalent of setTimeout 0. I'm not sure why its needed, but I had assumed this was from the time when peeps thought you had to wrap promises for them to run.

setTimeout 0 would have been more obvious if that was the intent but ohwell.

src/module/actor/npc/document.ts
CarlosFdez: 1) segments only includes sibling segments
2) fromTroop avoids the recursion

CarlosFdez: Though in retrospect I think I can reuse the class property

CarlosFdez: Should be swapped over now

src/module/scene/document.ts
stwlam: `setTimeout` 0 is an indirect way to send something to the end of the task queue, predating `Promise`s. `Promise.resolve` directly appends to the end of the queue.

PR #21620, opened 2026-02-25
Fix Calculate XP Macro to use the correct XP budget and Award methods

ChasarooniZ:
**Issue**
Currently the system uses the following method to calculate XP Award and Budget `baseXP * 4 / partyCount`, however this isn't RAW.

**Intended Results**

RAW XP award is as follows [1](https://2e.aonprd.com/Rules.aspx?ID=2651&Redirected=1)
> Encounters with adversaries and hazards grant a set amount of XP. When the group overcomes an encounter with creatures or hazards, each character gains XP equal to the total XP of the creatures and hazards in the encounter (**this excludes XP adjustments for different party sizes**; see Party Size for details)

And then the party size page gives a table of encounter budget and adjustments [2](https://2e.aonprd.com/Rules.aspx?ID=2717)

Thus i believe it should probably instead of using the above formula do the following (which I've attempted to do in this PR)

**Encounter Budget** = `BASE XP + (adjustment * extra characters)`

### DIFF
--- a/src/scripts/macros/xp/index.ts
+++ b/src/scripts/macros/xp/index.ts
@@ -50,6 +50,14 @@ const xpSimpleHazardDifferences = new Map([
     [4, 32],
 ]);
 
+const xpCharacterAdjustment = new Map([
+    ['trivial', 10],
+    ['low', 20],
+    ['moderate', 20],
+    ['severe', 30],
+    ['extreme', 40],
+])
+
 function getXPFromMap(partyLevel: number, entityLevel: number, values: Map<number, number>): number {
     // add +1 to all levels to account for -1 levels
     const difference = entityLevel + 1 - (partyLevel + 1);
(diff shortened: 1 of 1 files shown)


CarlosFdez:
As a general rule, please don't make major system changes that impact all users without first getting approval in an issue. Especially for something as obvious as XP. These sorts of changes impact everyone.

For the implementation itself, you're conveniently ignoring context. That entire section is derived under the assumption that your combat is intended to be at a specific difficulty level, and you're adjusting it for party size to maintain that difficulty level. In those situations, the XP you award is unaffected by the adjustments made to rebalance the encounter. 

The goal underneath all this is very clearly that you should try to maintain difficulty, but the party should continue to level at the same rate.

CarlosFdez:
Its easier if I throw some images. 

Lets say we have a moderate encounter. 80XP. 4 PL-2 creatures vs 4 PCs
image

After scaling a fight (making an adjustment) we should give the same XP without those adjustments (pre scaling XP). So after scaling this fight, we should still award 80XP regardless of any adjustments we do for party size. So lets add a party member, which increases the XP budget by 20. We add another Xoarian to maintain the same difficulty.

image

The encounter budget increased, and had an adjustment of an extra 20XP for the same difficulty. But the awarded XP is without the adjustment for different party size. And players don't level any faster for the same difficulty fight.

ChasarooniZ:
Ah, sorry for the trouble! going to close this PR

PR #21638, opened 2026-02-27
Add Tetanus passive to Rusty Spike Launcher hazard

ammalagonc:


TikaelSol:
I meant adding the tetanus description as a passive to the Rusty Spike Launcher hazard

Issue #21579, opened 2026-02-21
Could you please add a link to the Adept Benefit (Chalice) effect in the Drink from the Chalice action?


Bomberman1988:
Currently you gotta look at the feature to get the effet. I thought the effect just wasn't working.

<img width="387" height="679" alt="Image" src="https://github.com/user-attachments/assets/a377fe0d-0018-4a66-859e-315e2cfb5c49" />

<img width="386" height="396" alt="Image" src="https://github.com/user-attachments/assets/c72fcc38-e4fd-4dea-871f-9d71c07285a9" />

pedrogrullada:
- Closed by #21672

PR #21672, opened 2026-03-01
Display Adept Benefit on Drink from the Chalice at all times

iDantar:
Based on the issue https://github.com/foundryvtt/pf2e/issues/21579
Changes: 
- add full text of Adept's Benefit to the Drink from the Chalice, including the effect
- show text of Adept's Benefit to the Drink from the Chalice at all times - previously it was only displayed if the Thaumaturge was already under the Effect: Chalice Adept Benefit (Self)
- updated text formatting in the note displayed when Thaumaturge fulfills the Adept Benefit's conditions
- removed DescriptionAlteration that's unused now

P.S. First time contributing to an open source, please correct me if I messed something up :)


pedrogrullada:
Hi, I'm the one who made this implementation. Moving that description alteration to the action itself makes `Effect: Chalice Adept Benefit (Self)` confusing due to the way adding tags via Item Alteration works. That aside, looking at it, the current implementation is wrong in a different way in that it buffs _every_ use of Drink from the Chalice, whereas only one particular creature actually benefits from the increased healing. That's my mistake.

Ideally we want chalice adept thaumaturges to be able to choose each time what kind of healing they do. Would you mind trying a different implementation to allow that? For that, you'd need to:
- Remove `Effect: Chalice Adept Benefit (Self)` from the Chalice Adept Benefit description, but leave the effect itself unchanged to not break current functionality.
- Restore the description alteration localization key, but replacing its text with the following to add the option for greater healing, instead of enforcing it with a Flat Modifier.
`On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature @Damage[(5*@actor.level)[healing,vitality]|traits:healing,vitality]{5 Hit Points for each level you have}`
Use this restored localization key in your description Item Alteration in `Adept Benefit (Chalice)` 
- Add the `chalice-adept-benefit` tag Item Alteration to the `Adept Benefit (Chalice)` feature so we can use it in Drink from the Chalice's effect.
- Add a Choice Set to `Effect: Drink from the Chalice`, asking whether the temporary Hit Points are increased due to Chalice's Adept Benefit. The Choice for "Yes" being predicated on `"parent:origin:item:tag:chalice-adept-benefit"` to ensure it only comes up if the thaumaturge has that benefit, otherwise the Choice Set is auto-resolved. Then update the predicate on each temporary Hit Points rule element in it accordingly.

This makes it so chalice adept thaumaturges have the option to use the adept benefit each time instead of enforcing it, and it keeps current chalice thaumaturges working.

iDantar:
Thanks for the response and a detailed suggestion, appreciate it!
Though small question: I think it makes sense to keep the trigger in the text of description alteration - in the same style as Transcendence actions have for Exemplar, what do you think?

pedrogrullada:
> Thanks for the response and a detailed suggestion, appreciate it! Though small question: I think it makes sense to keep the trigger in the text of description alteration - in the same style as Transcendence actions have for Exemplar, what do you think?

Yeah, you're right. That part should be there as well.

iDantar:
Applied the suggested fixes, please review.
I added a localization key for Choice Set prompt and overall not too sure I used the correct syntax in it - would appreciate comments/advise

static/lang/re-en.json
```diff
@@ -6990,8 +6990,9 @@
...
+                            "DescriptionAlteration": "<p>If you or an ally within 30 feet takes piercing or slashing damage from a foe's critical hit or takes persistent bleed damage, @UUID[Compendium.pf2e.actionspf2e.Item.D91jQs0wleU5ml4K]{Drinking from the Chalice} before the end of your next turn grants that injured creature greater restoration to make up for its lost vitality.</p>\n <p>On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature @Damage[(5 * @actor.level)[healing,vitality]]{5 Hit Points for each level} you have. Same for the creature with @Damage[(5 * @actor.level)[healing,void]]{void healing}.</p>",
```
pedrogrullada:
```suggestion
                            "DescriptionAlteration": "If you or an ally within 30 feet takes piercing or slashing damage from a foe's critical hit or takes persistent bleed damage, Drinking from the Chalice before the end of your next turn grants that injured creature greater restoration to make up for its lost vitality. On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature @Damage[(5*@actor.level)[healing,vitality]|shortLabel] Hit Points (or @Damage[(5*@actor.level)[healing,void]] if they have void healing).",
```
Item Alterations don't support html. Let's go with this formatting, I hadn't thought of the void healing.

packs/pf2e/feat-effects/effect-chalice-adept-benefit-self.json
pedrogrullada, line 1: We can keep this effect unchanged, even with the new description alteration, it still leaves thaumaturges with the old implementation working as normal.

static/lang/re-en.json
```diff
@@ -6990,8 +6990,9 @@
...
+                            "Note": "If you or an ally within 30 feet takes piercing or slashing damage from a foe's critical hit or takes persistent bleed damage, @UUID[Compendium.pf2e.actionspf2e.Item.D91jQs0wleU5ml4K]{Drinking from the Chalice} before the end of your next turn grants that injured creature greater restoration to make up for its lost vitality.<p>On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature 5 Hit Points for each level you have.</p>",
+                            "Prompt": "Does adept benefit apply?"
```
pedrogrullada:
```suggestion
                            "Prompt": "Does the chalice's adept benefit apply?"
```

packs/pf2e/feat-effects/effect-drink-from-the-chalice.json
```diff
@@ -22,18 +22,34 @@
...
-                "key": "TempHP",
-                "predicate": [
+                "choices": [
                     {
-                        "not": "parent:origin:item:tag:chalice-adept-benefit"
+                        "label": "Yes",
+                        "predicate": [
+                            "parent:origin:item:tag:chalice-adept-benefit"
+                        ],
+                        "value": "yes"
```
pedrogrullada:
```suggestion
                        "label": "PF2E.UI.RuleElements.ChoiceSet.YesLabel",
                        "predicate": [
                            "parent:origin:item:tag:chalice-adept-benefit"
                        ],
                        "value": "adept"
```

packs/pf2e/feat-effects/effect-drink-from-the-chalice.json
```diff
@@ -22,18 +22,34 @@
...
+                    },
+                    {
+                        "label": "No",
+                        "value": "no"
```
pedrogrullada:
```suggestion
                        "label": "PF2E.UI.RuleElements.ChoiceSet.NoLabel",
                        "value": "initiate"
```

packs/pf2e/feat-effects/effect-drink-from-the-chalice.json
```diff
@@ -22,18 +22,34 @@
...
                     }
                 ],
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Thaumaturge.Implement.Chalice.Adept.Prompt",
+                "rollOption": "adept-benefit-chalice-applies"
```
pedrogrullada:
```suggestion
                "rollOption": "drink-from-the-chalice-temp-hp"
```

packs/pf2e/feat-effects/effect-drink-from-the-chalice.json
```diff
@@ -22,18 +22,34 @@
...
+            },
+            {
+                "key": "TempHP",
+                "predicate": [
+                    "adept-benefit-chalice-applies:no"
```
pedrogrullada:
```suggestion
                    "drink-from-the-chalice-temp-hp:initiate"
```

packs/pf2e/feat-effects/effect-drink-from-the-chalice.json
```diff
@@ -22,18 +22,34 @@
...
+                ],
                 "value": "max(3,@item.origin.level)"
             },
             {
                 "key": "TempHP",
                 "predicate": [
-                    "parent:origin:item:tag:chalice-adept-benefit"
+                    "adept-benefit-chalice-applies:yes"
```
pedrogrullada:
```suggestion
                    "drink-from-the-chalice-temp-hp:adept"
```

static/lang/re-en.json
iDantar: Sounds good! The only changes I'd leave though are spaces around * - otherwise the text between them becomes _italic_ and inline rolls are not rendered correctly. I spent ~an hour figuring that out :)

pedrogrullada: That seems like a bug. I don't have time to verify but in that case go ahead and keep them.

packs/pf2e/feat-effects/effect-chalice-adept-benefit-self.json
iDantar: Good call, thanks! Some part of me wants to mark it as deprecated so users can switch to a new one, but that's probably out of scope

iDantar:
Updated, tested - looks good to go :)
Thank you so much for a detailed review and all of the suggestions!

static/lang/re-en.json
```diff
@@ -6990,8 +6990,9 @@
...
+                            "DescriptionAlteration": "If you or an ally within 30 feet takes piercing or slashing damage from a foe's critical hit or takes persistent bleed damage, Drinking from the Chalice before the end of your next turn grants that injured creature greater restoration to make up for its lost vitality. On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature @Damage[(5 * @actor.level)[healing,vitality]|shortLabel] Hit Points (or @Damage[(5 * @actor.level)[healing,void]] if they have void healing).",
```
pedrogrullada:
```suggestion
                            "DescriptionAlteration": "If you or an ally within 30 feet takes piercing or slashing damage from a foe's critical hit or takes persistent bleed damage, Drinking from the Chalice before the end of your next turn grants that injured creature greater restoration to make up for its lost vitality. On a sip, the temporary Hit Points granted to the creature increase to twice your level. When drained, the chalice heals the creature @Damage[(5 * @actor.level)[healing,vitality]|shortLabel|traits:healing,vitality] Hit Points (or @Damage[(5 * @actor.level)[healing,void]|traits:healing,void] if they have void healing).",
```
Oh yeah these traits need to be set explicitly. Other than that, looks like we're ready.

iDantar: Huh, I was suggested on Discord to remove them - I should have tested it myself, my bad.
Thanks for catching that!

pedrogrullada (approved):

PR #21689, opened 2026-03-04
Fix `@Check` slugs

Cuingamehtar:


simonward:
"The Deft" looks like it needs a fix: `"path": "system.skills.theivery.rank"`

PR #21691, opened 2026-03-05
Add automation for NPC abilities Consecrated Fists, Fiendish Lunge, Infected Wounds, Midnight Depths

simonward:
Plus some other tidying

### DIFF
--- a/packs/pf2e/pathfinder-monster-core-2/nasurgeth.json
+++ b/packs/pf2e/pathfinder-monster-core-2/nasurgeth.json
@@ -751,9 +751,6 @@
                 "rules": [
                     {
                         "key": "Aura",
-                        "predicate": [
-                            "terrain:underwater"
-                        ],
                         "radius": 60,
                         "slug": "midnight-depths",
                         "traits": [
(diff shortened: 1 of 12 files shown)


packs/pf2e/pathfinder-monster-core-2/nasurgeth.json
```diff
@@ -763,6 +763,16 @@
                             "void"
                         ]
                     },
+                    {
+                        "damageType": "cold",
+                        "diceNumber": 2,
+                        "dieSize": "d10",
+                        "key": "DamageDice",
+                        "predicate": [
+                            "terrain:underwater"
```
pedrogrullada: Wouldn't this activate even if only the nasurgeth is underwater and not its target? The aura seems like it can extend beyond the water. We might want a toggle here instead.

simonward: Something more like this?

pedrogrullada: Yeah I think that works

PR #21697, opened 2026-03-05
Use weapon potency for Fox Arson bonus damage calculation

### LINKED ISSUE #21133: Fox Arson flat modifier not applying
9th level kitsune feat Fox Arson doesn't apply the item bonus damage correctly. 
Instead of its current value, it should use `@weapon.system.runes.potency`

iDantar:
Fixes https://github.com/foundryvtt/pf2e/issues/21133
Quote from the feat: 'You gain an item bonus to this persistent damage equal to the foxfire’s item bonus to attack rolls.'
Unfortunately this is not a clean fix, as it doesn't correctly work with other item bonuses to attack (Quicksilver Mutagen for example), but it works most of the time. 
If there's a better reference for this, would appreciate the advice

### DIFF
--- a/packs/pf2e/feats/ancestry/kitsune/level-9/fox-arson.json
+++ b/packs/pf2e/feats/ancestry/kitsune/level-9/fox-arson.json
@@ -43,7 +43,7 @@
                 "damageCategory": "persistent",
                 "key": "FlatModifier",
                 "selector": "foxfire-damage",
-                "value": "@weapon.system.runes.potency"
+                "value": "@weapon.flags.system.attackItemBonus"
             }
         ],
         "traits": {
--- a/packs/pf2e/feats/ancestry/kitsune/level-9/fox-arson.json
+++ b/packs/pf2e/feats/ancestry/kitsune/level-9/fox-arson.json
@@ -43,7 +43,7 @@
                 "damageCategory": "persistent",
                 "key": "FlatModifier",
                 "selector": "foxfire-damage",
-                "value": "@item.flags.system.attackItemBonus"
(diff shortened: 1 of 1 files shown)


packs/pf2e/feats/ancestry/kitsune/level-9/fox-arson.json
```diff
@@ -43,7 +43,7 @@
                 "damageCategory": "persistent",
                 "key": "FlatModifier",
                 "selector": "foxfire-damage",
-                "value": "@item.flags.system.attackItemBonus"
+                "value": "@weapon.system.runes.potency"
```
pedrogrullada:
```suggestion
                "value": "@weapon.flags.system.attackItemBonus"
```
The getter originally used does work, but it needs to point to the weapon. By using `@item` it was trying to get the data from the feat instead.

iDantar: This for sure looks better than relaying on potency, thanks! 
Unfortunately it still doesn't cover the edge case of having a temporary buff:
image
Is there maybe a way to get that bonus, or that's good enough for now?

pedrogrullada: Sounds like a separate bug with that specific flag, so it should be reported and fixed. We should still use it and assume it'll be fixed to work as intended.

iDantar: Sounds good, I'll report it and apply the suggestions then
Thanks again!

pedrogrullada (approved):

PR #21707, opened 2026-03-06
Add inline check to Cover Tracks skill action

ammalagonc:


packs/pf2e/actions/skill/cover-tracks.json
```diff
@@ -12,7 +12,7 @@
...
+            "value": "<p>You cover your tracks, moving up to half your travel Speed, using the @UUID[Compendium.pf2e.journals.JournalEntry.S55aqwWIzpQRFhcq.JournalEntryPage.DUOb8QrlC5t0adkL]{Travel Speed} rules. You don't need to attempt a Survival check to cover your tracks, but anyone tracking you must succeed at a @Check[survival|against:survival|options:track] check against your Survival DC if it is higher than the normal DC to @UUID[Compendium.pf2e.actionspf2e.Item.Track].</p>\n<p>In some cases, you might Cover Tracks in an encounter. In this case, Cover Tracks is a single action and doesn't have the exploration trait.</p>"
```
simonward:
```suggestion
            "value": "<p>You cover your tracks, moving up to half your travel Speed, using the @UUID[Compendium.pf2e.journals.JournalEntry.S55aqwWIzpQRFhcq.JournalEntryPage.DUOb8QrlC5t0adkL]{Travel Speed} rules. You don't need to attempt a Survival check to cover your tracks, but anyone tracking you must succeed at a @Check[survival|against:survival|options:action:track] check against your Survival DC if it is higher than the normal DC to @UUID[Compendium.pf2e.actionspf2e.Item.Track].</p>\n<p>In some cases, you might Cover Tracks in an encounter. In this case, Cover Tracks is a single action and doesn't have the exploration trait.</p>"
```

PR #21714, opened 2026-03-07
Add late preparation for generated shield attacks in CharacterPF2e

### LINKED ISSUE #19680: Unable to upgrade shield bash damage dice with Item Alteration
Not sure if this is a bug report or a feature request. Using this broad Item Alteration rule element, the Strike created by holding any shield is not upgraded by it.
```json
{
  "key": "ItemAlteration",
  "itemType": "weapon",
  "mode": "upgrade",
}
```
It would also be useful to be able to target shield bash attacks in general with such an Item Alteration (or Damage Alteration), maybe by way of a roll option that identified the Strike as such.

iDantar:
Fixes https://github.com/foundryvtt/pf2e/issues/19680

Item alterations for weapons run over the set of items returned by #getItemsOfType() (i.e. only embedded weapons). Shield Bash is created on the fly, so ItemAlteration didn't work on it.
I found performLatePreparation used in StrikeRuleElement and applied it to strikes generated from shields. 

Testing this using a rule below, it appears to work now. But since this is my first real change to the code, would appreciate advice and comments

```{"key":"ItemAlteration","itemType":"weapon","mode":"upgrade","property":"damage-dice-faces","predicate":["item:equipped","item:group:shield","item:melee"]}```

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -44,6 +44,7 @@ import { WeaponPF2e } from "@item";
 import type { AbilityTrait } from "@item/ability/types.ts";
 import { ARMOR_CATEGORIES } from "@item/armor/values.ts";
 import { ActionCost } from "@item/base/data/index.ts";
+import { performLatePreparation } from "@item/helpers.ts";
 import { getPropertyRuneDegreeAdjustments, getPropertyRuneStrikeAdjustments } from "@item/physical/runes.ts";
 import type { EffectAreaShape, ItemType } from "@item/types.ts";
 import type { WeaponSource } from "@item/weapon/data.ts";
@@ -1075,7 +1076,11 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
             // Generate a shield attacks from the character's shields
             this.itemTypes.shield
                 .filter((s) => !s.isStowed && !s.isBroken && !s.isDestroyed)
-                .map((s) => s.generateWeapon()),
+                .map((s) => {
+                    const w = s.generateWeapon();
+                    if (w) performLatePreparation(w);
+                    return w;
+                }),
             this.inventory.flatMap((i) =>
                 i.isEquipped
                     ? i.subitems.filter(
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -44,7 +44,6 @@ import { WeaponPF2e } from "@item";
 import type { AbilityTrait } from "@item/ability/types.ts";
 import { ARMOR_CATEGORIES } from "@item/armor/values.ts";
 import { ActionCost } from "@item/base/data/index.ts";
-import { performLatePreparation } from "@item/helpers.ts";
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/module/actor/character/document.ts [hunk 1]
```diff
@@ -1075,7 +1076,11 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
             // Generate a shield attacks from the character's shields
             this.itemTypes.shield
                 .filter((s) => !s.isStowed && !s.isBroken && !s.isDestroyed)
-                .map((s) => s.generateWeapon()),
+                .map((s) => {
+                    const w = s.generateWeapon();
+                    if (w) performLatePreparation(w);
+                    return w;
+                }),
```
stwlam: This should probably be done from `ShieldPF2e#generateWeapon` so that it happens regardless of which calls it.

iDantar: Good call, moved it and reduced duplication a bit. Thanks!

src/module/item/shield/document.ts
```diff
@@ -238,7 +239,7 @@ class ShieldPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
         const shieldThrowTrait = this.system.traits.value.find((t) => t.startsWith("shield-throw-"));
         if (shieldThrowTrait) weaponTraits.push(`thrown-${shieldThrowTrait.slice(-2)}` as WeaponTrait);
 
-        const baseData: BaseWeaponData = fu.deepClone({
+        let weaponData: BaseWeaponData = fu.deepClone({
```
stwlam: Continue using `baseData` and assign the merged data to `weaponData`. We prefer to avoid `let` except in case of necessity.

iDantar: Updated, will keep that in mind for the future, thanks.

src/module/item/shield/document.ts
```diff
@@ -257,6 +258,7 @@ class ShieldPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ph
             },
         });
 
+        let weaponData: BaseWeaponData | undefined;
```
stwlam: no `let` :)

src/module/item/shield/document.ts
stwlam: Does appear to be resolved?

src/module/item/shield/document.ts
iDantar: That one felt necessary :)
Switched to iife, but it feels a bit harder to comprehend this way

src/module/item/shield/document.ts
iDantar: Yep, just wasn't sure how you folks usually do this: does reviewer or the author resolve the conversations?

src/module/item/shield/document.ts
stwlam: A separate #private method would be appropriate if you don't prefer an IIFE

iDantar: Good call! I think it looks good now - could you take another look please?
Appreciate you for getting back to me so fast and all the advise :)

PR #21725, opened 2026-03-09
Add build-time check for name duplicates

Cuingamehtar:
Will prevent build from finishing as duplicate entries make source uuid links ambiguous.

Will currently fail SF build due to gambler duplicate, but will work correctly after it's resolved.

Will allow extracting the pack normally, but won't let packing it back.

### DIFF
--- a/build/lib/compendium-pack.ts
+++ b/build/lib/compendium-pack.ts
@@ -112,6 +112,10 @@ class CompendiumPack {
         };
 
         for (const docSource of this.data) {
+            // Assert there are no entries with the same name
+            if (packMap.has(docSource.name))
+                throw PackError(`Compendium ${this.id} (${dirName}) has multiple entries named ${docSource.name}.`);
+
             // Populate CompendiumPack.namesToIds for later conversion of compendium links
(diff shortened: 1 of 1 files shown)


build/lib/compendium-pack.ts
```diff
@@ -112,6 +112,10 @@ class CompendiumPack {
         };
 
         for (const docSource of this.data) {
+            // Assert there are no entries with the same name
+            if (packMap.has(docSource.name))
+                throw PackError(`Compendium ${this.id} (${dirName}) has multiple entries named ${docSource.name}.`);
+
```
CarlosFdez:
```suggestion
            if (packMap.has(docSource.name)) {
                throw PackError(`Compendium ${this.id} (${dirName}) has multiple entries named ${docSource.name}.`);\
            }
```

Cuingamehtar: Added braces.
Have you considered [ESLint "curly" rule](https://eslint.org/docs/latest/rules/curly) so that you won't have to keep track of those?

PR #21727, opened 2026-03-09
Also resolve persistent damage for troop segments and familiars

CarlosFdez:
Should we include familiars in `combatant.tokens`. This would also include them in combatant pings and probably lead to better overall handling in the future of shared turn combatants. But I wasn't sure yet.


stwlam:
I don't think we should: maybe a separate representation in the tracker would be good.

PR #21728, opened 2026-03-09
Fix stealth DC for Shaken Loose

RakuJa:
image


packs/pf2e/myth-speaker-bestiary/book-3-titanbane/shaken-loose.json
```diff
@@ -51,7 +51,7 @@
             "resistances": [],
             "stealth": {
                 "details": "",
-                "value": null
+                "value": 10
```
simonward: Should this be `"value": 0` as it's listed as "Stealth DC 10"?

RakuJa: Yes! Sorry, it was a careless mistake

PR #21729, opened 2026-03-09
[wip] parallelize pack load for pf2e

iDantar:
Not sure how helpful it could be in the main branch, but wanted to share and maybe get some opinions on it.

Locally, each build took ~2.5 minutes, so I wanted to dig around and see if I can condense it a bit.
By running the loads in parallel, local build time for me dropped to 30 seconds


iDantar:
Looks like a long local build time was a fluke - re-testing on v14 didn't show any significant difference

iDantar:
Looks like a long local build time was a fluke - re-testing on v14 didn't show any significant difference

PR #21732, opened 2026-03-10
Parse dice values in trait config

CarlosFdez:


stwlam:
Need an issue since parsing something typically has a reason. Also, retarget merge branch.

CarlosFdez:
it was privately discussed and I had received the go ahead, but if my memory serves since it was a while ago, the goal was to convert things like `3d6` in the traits config to structured data for the purposes of implementing boost in Starfinder. Trait Configs are still an in development incomplete thing, and I don't recall there being stringent issue requirements for in development stuff. 

If you would rather keep them as strings, we can. I don't mind if someone types 3d6. Though we may need a way to create damage dice rule elements from raw string formulas, though that is not insurmountable. If so though, close the issue, and we'll work with string formulas for trait configs for the deadly/fatal/boost/etc traits in the foreseeable future.

CarlosFdez:
I'll close it actually. While trait configs are still generally unpersisted (outside of modular) and are still in development, this particular bit has had too much back and forth, when string formulas work just fine and we have sufficient parsers for them. I'll just add boost to the list in a separate PR and we'll move on.

PR #21735, opened 2026-03-10
Add 'fear' trait to A Falling Star's Memories

RakuJa:
Shouldn't A Falling Star's Memories have the fear trait?
image

https://2e.aonprd.com/Hazards.aspx?ID=622


simonward:
I don't believe "fear" is a currently allowed trait for hazards. It would need to be added to `hazardTraits`.

PR #21741, opened 2026-03-11
Add missing localization strings to SF2e

Cuingamehtar:
Will probably need some review and adjustment.

Only SF here. I'll do PF piecemeal later as I've found 767 entries (most of them in bestiaries).

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -7590,7 +7590,7 @@
             },
             "SihedronGuardThunderstriker": {
                 "NavigationalSpellshape": {
-                    "Alteration": "If the next action the aeon guard uses is to Cast a Spell that has a range, increase that spell's range by 30 feet."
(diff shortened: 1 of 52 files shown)


static/lang/re-en.json
```diff
@@ -7565,6 +7564,35 @@
...
+            },
+            "RallyTheTroops": {
+                "Ally": "Next to Ally",
+                "Commander": "Next to Commander"
+            },
+            "SihedronGuardThunderstriker": {
+                "NavigationalSpellshape": {
+                    "Alteration": "If the next action the aeon guard uses is to Cast a Spell that has a range, increase that spell's range by 30 feet."
+                }
+            },
```
TikaelSol, line 7595: Feel like we can just use the existing reach spell text, no?

Cuingamehtar: It's actually more than that - it also reduces bonus from lesser cover.

PR #21746, opened 2026-03-13
Add `item:trait` roll options to action macros

TikaelSol:
Action macros not including them is a pain, and we have had a standard in data entry for years of using `item:trait:x` for added specificity. The action macros are stand ins for the action items themselves, and should pass the proper roll options for the item they are standing in for.

### DIFF
--- a/src/module/system/action-macros/helpers.ts
+++ b/src/module/system/action-macros/helpers.ts
@@ -166,6 +166,7 @@ class ActionMacroHelpers {
                 const actionTraits = (options.traits ?? []).filter(
                     (t): t is AbilityTrait => t in CONFIG.PF2E.actionTraits,
                 );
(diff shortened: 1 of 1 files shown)


src/module/system/action-macros/helpers.ts
```diff
@@ -166,6 +166,7 @@ class ActionMacroHelpers {
                 const actionTraits = (options.traits ?? []).filter(
                     (t): t is AbilityTrait => t in CONFIG.PF2E.actionTraits,
                 );
+                combinedOptions.push(...actionTraits.map(t => `item:trait:` + t) ?? [])
```
simonward:
```suggestion
                combinedOptions.push(...actionTraits.map((t) => `item:trait:${t}`));
```

PR #21748, opened 2026-03-13
Also sync troop tokens on token creation and deletion

CarlosFdez:
This fixes editing HP on troops that were created, or after adding a segment by copy pasting a token. 

In live they only sync properly after a data reset cycle or when reloading the scene.

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -674,9 +674,7 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
             }
             for (const segment of segments ?? []) {
                 const segmentActor = segment.actor;
-                if (segmentActor) {
-                    TokenDocumentPF2e.#resetBatch.reset(segmentActor);
-                }
+                if (segmentActor) TokenDocumentPF2e.#resetBatch.reset(segmentActor);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/module/scene/token-document/document.ts
```diff
@@ -660,6 +663,22 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
...
+        if (actor?.isOfType("npc") && this.flags[SYSTEM_ID].troop) {
+            const segments = this.segments;
+            if (segments?.length && !actor.otherSegments?.length) {
+                TokenDocumentPF2e.#resetBatch.reset(actor);
+            }
+            for (const segment of segments ?? []) {
+                const segmentActor = segment.actor;
+                if (segmentActor) {
+                    TokenDocumentPF2e.#resetBatch.reset(segmentActor);
+                }
```
stwlam:
```suggestion
                if (segmentActor) TokenDocumentPF2e.#resetBatch.reset(segmentActor);
```

src/module/scene/token-document/document.ts
```diff
@@ -717,6 +736,13 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
                 game.pf2e.effectTracker.unregister(effect);
             }
         }
+
+        for (const segment of this.segments ?? []) {
+            const segmentActor = segment.actor;
+            if (segmentActor) {
+                TokenDocumentPF2e.#resetBatch.reset(segmentActor);
+            }
```
stwlam:
```suggestion
            if (segmentActor) TokenDocumentPF2e.#resetBatch.reset(segmentActor);
```

PR #21757, opened 2026-03-14
Add missing Emotionless Android level 1 ancestry feat

### LINKED ISSUE #21747: Android Ancestry feat named Emotionless is missing from Starfinder 2e system
This ancestry feat Emotionless is missing from the system.

ammalagonc:
- Closes #21747


pedrogrullada:
Might as well bring it up here too: since all fear effects are also emotion, this is mechanically identical to the Pathfinder feat. Since the mechanical text is equivalent and the flavor text only has a minor difference, I think we can makes this a duplicate.

TikaelSol:
I think it's fine as a separate item, since the text is different.

PR #21768, opened 2026-03-15
Add all formula variants when adding one

### LINKED ISSUE #17796: Formulas for any version of an item should automatically grant future and past versions as you level.
It has been clarified in multiple occasions since the remaster that having the formula for a lesser Elixir of life, for example, gives you the formula to craft all the other versions of the Elixir of Life. It would be useful then that if you added a formula for an item with multiple versions, you would get all the versions in your formula book.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/17796

Creates an index of items that have versions for different levels once per session.
When adding such an item, prompts user to add one or all versions, with an option to save the choice for future.


iDantar:
Added an option to only add items up to character level

PR #21775, opened 2026-03-17
Add export log for anachronism builds

CarlosFdez:
Adds export logs of all items (not actors) to the commit log. One fear of anachronism builds is that duplicates may lead to something no longer being exported, and we may need to set up redirects in some of those cases. These logs should hopefully catch that when new anachronism builds are created by checking the diffs in the git log.

A suggestion was placed to make it a duplicates log instead, but a new entry in a duplicates log may be from a new content release and not require a redirect. An export log is better at catching potential broken links.

### DIFF
--- a/build/anachronism.ts
+++ b/build/anachronism.ts
@@ -1,4 +1,6 @@
-import { recursiveReplaceString } from "@util/misc.ts";
+import { ItemSourcePF2e } from "@item/base/data/index.ts";
+import { itemIsOfType } from "@item/helpers.ts";
+import { objectHasKey, recursiveReplaceString, sluggify } from "@util/misc.ts";
 import fs from "fs";
 import path from "path";
 import * as R from "remeda";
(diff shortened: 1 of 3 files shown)


stwlam:
Why?

CarlosFdez:
After a discussion in the discord and a thought today while getting coffee I was actually gonna update the comment description, so this is an interesting coincedence. But in short its to debug problems with the build between releases by comparing to the results of a previous build.

In long, we need to handle these situations.
1) We need to detect when an item is suddenly flagged as a duplicate and removed from the build. In some cases we may need to add manual redirects.
2) We need to detect when items aren't being properly exported.

@Cuingamehtar wanted it to be a log that shows duplicates instead, but while that's better at dealing with the second situation, its poor at handling when an item is no longer being exported, and my priority is on avoiding errors.

Cuingamehtar:
No, I wanted the log that shows the excluded items, not necessary duplicates.

I see no particular reason to log items from the duplicates.json file individually. They are duplicates by definition. With those the check should be straightforward - if it is in the list, don't export it to anachronism.

If one list isn't good enough for now, why don't have both?

But ultimately, the question will be, when will this log be created? Are you going to run the anachronism build manually every time just before making a release?
Or is it going to be automatically updated by github action as a new commit?
Because otherwise the log might be different from what actually goes into the release.

CarlosFdez:
sorry, yeah when I said duplicates I meant in the context of the export, so overlaps as well.

For when it gets committed, I imagine at the same time we increment the version numbers in the manifest. We would run a build at the same time we increment to test very quickly, and in the PR people can see if any new omitted content would be a problem.

CarlosFdez:
So while this would help catch failed links before release, that might not happen enough to warrant the addition to the repo.

PR #21777, opened 2026-03-18
Update data to match Player Core 2 Spring Errata

### LINKED ISSUE #20141: Alchemist Feat Sticky Bomb doing too much damage, inappropriately boosted by other feats / abilities
[Issue as discussed back and forth on the paizo forums](https://paizo.com/threads/rzs6tsza?Remastered-Alchemist#14)

**Issue:** The Alchemist Feat Sticky Bomb presently gains increased damage from other Alch feats and abilities that boost splash damage of bomb throws. Notably, the Bomber L5 field discovery and Expanded Splash feat.

**Expected behavior:** The damage from Sticky Bomb should be unaffected by improved throws.

[Sticky Bomb](https://2e.aonprd.com/Feats.aspx?ID=5786) text:   
> You can mix in a sticky additive to an [alchemical bomb](https://2e.aonprd.com/Weapons.aspx?ID=431) to make its contents adhere and continue to deal damage. A creature hit by a sticky bomb also takes [persistent damage](https://2e.aonprd.com/Conditions.aspx?ID=29) **equal to and of the same type as the bomb's splash damage.** If the bomb already deals persistent damage, combine the two amounts.

This extra persistent damage should only key into the splash value inherent to the bomb item, not any "final splash damage" of a potential throw. Feats and effects that specifically wish to key into such "final" values will use terms like "the ___ damage of **your attack**." 
This feat's text, meanwhile, is written in a manner that is similar to the "number of damage dice" style effects; a static bonus that is unaffected by other boosts. (Bombs are an abnormality that lack damage dice. )

The present behavior can result in a Sticky dealing flat 10 persistent damage per application once Expanded is added, roughly 3x of what it should be. 
This boost also means that at L10, an Alchemist throwing an Acid Flask, the standard persistent damage option, will deal 3d6 +10 persistent. The passive boost of Sticky is about doubling the strike damage, which as far as I know would be completely without equal for a passive damage boost in the pf2 system.

pedrogrullada:
- Closes #20141
(coincidentally!)


static/lang/re-en.json
```diff
@@ -3353,8 +3353,8 @@
...
-                        "PartTwo": "You add to this spell repertoire as you increase in level. Each time you get a spell slot (see the Bard Spells per Day table), you add a spell to your spell repertoire of the same rank. At 2nd level, you select another 1st-rank spell; at 3rd level, you select two 2nd-rank spells, and so on. When you add spells, you might add a higher-rank version of a spell you already have, so you can cast a heightened version of that spell.",
+                        "PartOne": "The collection of spells you can cast is called your spell repertoire. At 1st level, you learn three 1st-rank divine spells of your choice and five divine cantrips of your choice. You choose these from the common spells from the occult list or from other occult spells to which you have access. You can cast any spell in your spell repertoire by using a spell slot of an appropriate spell rank.",
```
Cuingamehtar: I think you meant to edit the oracle's entry, not the bard's one.

pedrogrullada: whoopsie

PR #21780, opened 2026-03-18
Lightly brush up Swashbuckler

Longstrider:


ammalagonc:
Curious, why are we removing the links to features and effects on these items?

Longstrider:
As far as I can see, in most cases panache is referenced not as a class feature or effect, but as a state without any links - so I removed the links for the sake of formatting consistency. I can roll it back if that's not needed, or put links in other feats as well - it just makes sense to figure out if there should be links to the feature, to the effect or none. As for the separate effect links, the ones I removed from the description I put in self-applied effects.

packs/pf2e/feats/class/swashbuckler/level-4/guardians-deflection-swashbuckler.json
pedrogrullada, line 1: Making this a self effect would point users towards the use button, which then hides the Guardian's Deflection effect. I think it's better left as is.

pedrogrullada (commented):
I would leave all the references to the Panache class feature when it's not meant to be applied but rather as a condition for another effect. It's useful to have a link to the feature as a rules reference.

PR #21784, opened 2026-03-19
Change Recognize Spell (PF) and Face in the Crowd (SF) localization keys

Cuingamehtar:
The keys were changed because they were the only entries in their subcategory.

Primary reason to sort keys that was so that I could just add RE localizations without bothering with order and have them sorted automatically later.
Unfortunately, this is not the complete sort, as there are cases where I felt preserving sequential order is more valuable than having keys in alphabetical.

### DIFF
--- a/packs/pf2e/feats/skill/level-1/recognize-spell.json
+++ b/packs/pf2e/feats/skill/level-1/recognize-spell.json
@@ -32,7 +32,7 @@
         "rules": [
             {
                 "key": "RollOption",
-                "label": "PF2E.SpecificRule.SkillFeat.RecognizeSpell.RollOptionLabel",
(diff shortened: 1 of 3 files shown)


TikaelSol:
Can we hold off on the sort until after the books drop on April 1st? The localization work has already started on that pre-release branch, and this would be a pretty hard one to avoid conflicts with

Cuingamehtar:
Sure, no problem. I've kept only the key changes.

ammalagonc:
Conflicts

Cuingamehtar:
Actually, at this point I'll drop it.
You've already started populating PF2E SkillFeats prefix with ScareToDeath, and according to Tikael, all SpecificRules go to PF2E prefix, so I'm not even sure this shuffling of keys even matters.

PR #21792, opened 2026-03-19
Apply resistances once across split damage

### LINKED ISSUE #17428: Oath of the Defender resistance
Oath of defenders reads: "Allies in your champion's aura, not including you, get resistance 2 to damage dealt by creatures with the chosen trait. If such a creature deals more than one damage type at once, apply this resistance only to the highest amount of damage."

But it seems to behave like a regular resistance in foundry, getting applied to all damage types.
image

Eliott45:
## Summary
- update overall IWR resistance handling so resistances apply once per effect across split damage
- keep the highest applicable resistance for each damage instance while preventing the same resistance from applying multiple times to one effect

It fixes [#17428](https://github.com/foundryvtt/pf2e/issues/17428)


CarlosFdez:
Thanks for this. However the errata kinda states that all resistances are apply once. I think this would be handled by a update to the IWR handling overall.

Eliott45:
> Thanks for this. However the errata kinda states that all resistances are apply once. I think this would be handled by a update to the IWR handling overall.

Thanks, that makes sense. I've reworked the change so it updates the overall IWR resistance handling instead of adding Oath-specific resistance data handling.

I've also tested the updated behavior manually in Foundry with Oath of the Defender and a few other split-damage resistance cases. I considered adding automated tests as well, but I wasn't fully sure yet what the best coverage would look like for these IWR edge cases, so I wanted to avoid adding a weak or misleading test.

Issue #21771, opened 2026-03-16
No option to create a spell chip


Longstrider:
While dragging a spell to the character's inventory, there is no option to create a spell chip - only a spell gem.

<img width="340" height="280" alt="Image" src="https://github.com/user-attachments/assets/e0ca325d-269f-4ee0-8bb6-3342ce24e332" />

PR #21798, opened 2026-03-20
Implement SF2e spell chips

Cuingamehtar:
I've based the implementation on spell wands, but since the spell chip has to be installed into a comm unit, the changes went significantly beyond just configuration.

1) Added a new consumable subtype "spell-chip" and changed spell chips to consumable type
2) Added a new install type and **additional logic for physical items that allow them to be attachable if their usage includes installed or attached**.
3) Made Comm Unit (by slug) accept items that are installable in it.
4) Because subitems inherit their usage from parent items, when the Comm Unit is held in hand, the spell can be used.
5) Made interface tweaks - show SC charges. Made "Use" button appear on subitems. When the chip is drawn from the activations tab, it will draw the parent item instead.
6) Made chips recharge on rest.

image

A helper function to iterate over subitems would be useful. Using flatmaps feel really ugly.
And another one to obtain subitems by id.

Relevant issue https://github.com/foundryvtt/pf2e/issues/21771

Note for the future - enable these items with anachronism.

### DIFF
--- a/packs/sf2e/equipment/magic-items/spell-chip-1st-rank-spell.json
+++ b/packs/sf2e/equipment/magic-items/spell-chip-1st-rank-spell.json
@@ -8,6 +8,7 @@
         "bulk": {
             "value": 0.1
         },
+        "category": "spell-chip",
         "containerId": null,
         "description": {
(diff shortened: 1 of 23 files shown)


stwlam:
I don't think we're going to continue expanding the legacy usage of consumable items for wand-like things.

CarlosFdez:
Yeah there was a reason I didn't do all this beyond a wand relabel.

There's ideas and plans for changing how spells are embedded as well as making proper spell embeddables, though little good ideas yet on how to handle spell activations in general.

PR #21801, opened 2026-03-21
Fix healing calculation in Administer Ambient Magic

MarpoOcknew:
During gameplay, I noticed the previous calculation was listed as 2*0d4. 

This moves the `2*` into the first brackets to make it match the values listed under [AON](https://2e.aonprd.com/Feats.aspx?ID=5452)


TikaelSol:
As a note, we don't use AoN as a reference, since we always want to double check against the PDF.

PR #21805, opened 2026-03-21
Add option to NPC Attack items to exempt them from MAP

stwlam:

### DIFF
--- a/static/templates/items/melee-details.hbs
+++ b/static/templates/items/melee-details.hbs
@@ -13,7 +13,7 @@
                 value="{{data.area.value}}"
             />
             <select id="{{fieldIdPrefix}}area-type" name="system.area.type">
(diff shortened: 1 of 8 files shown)


static/templates/items/melee-details.hbs
```diff
@@ -13,7 +13,7 @@
                 value="{{data.area.value}}"
             />
             <select id="{{fieldIdPrefix}}area-type" name="system.area.type">
-                {{selectOptions areaShapes selected=data.area.type blank="" localize=true}}
+                {{selectOptions areaShapes selected=source.area.type blank="" localize=true}}
```
CarlosFdez: From reading source, this should be source.system.area right? data is system data

CarlosFdez:
Looks good to me!

PR #21809, opened 2026-03-21
Add deployment toggle to Ultralight Wings

### LINKED ISSUE #21778: Ultralight Wings implant item doesn't apply Fly speed
A player noticed they have the **Ultralight Wings (Commercial)** implant equipped at level 3 and it does not add the Fly speed of 20 feet to the character sheet as expected. It is implanted, the character is not encumbered, and is only wearing light armor (commercial microcord armor). That said, the +1 item bonus does appear as a toggle on the "Manuever in Flight" check.
https://2e.aonsrd.com/treasure/125-ultralight-wings

**Workaround** (until there's a fix in a future update)
In the item's Rules I applied a Roll Option that was suggested on Discord and got it to work. The Rule creates a toggle on the Character Sheet that when checked deploys the wings and adds the Fly speed.
{"key":"RollOption","label":"Wings Deployed","option":"wings-deployed","toggleable":true,"value":false}

Eliott45:
- add the missing `wings-deployed` roll option toggle to all Ultralight Wings variants
- allow the existing fly Speed and Maneuver in Flight bonus rules to activate correctly when the wings are deployed
- match the item behavior described in Starfinder Player Core

Fixes [#21778](https://github.com/foundryvtt/pf2e/issues/21778)

### DIFF
--- a/packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
+++ b/packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
@@ -41,7 +41,7 @@
         "rules": [
             {
                 "key": "RollOption",
-                "label": "Wings Deployed",
+                "label": "PF2E.SpecificRule.Equipment.UltralightWings.WingsDeployed",
                 "option": "wings-deployed",
                 "toggleable": true,
                 "value": false
--- a/packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
+++ b/packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
@@ -40,14 +40,18 @@
         "quantity": 1,
         "rules": [
             {
+                "key": "RollOption",
+                "label": "Wings Deployed",
+                "option": "wings-deployed",
+                "toggleable": true,
+                "value": false
+            },
+            {
                 "key": "BaseSpeed",
                 "predicate": [
                     "wings-deployed",
                     {
-                        "nor": [
-                            "armor:category:heavy",
-                            "armor:category:medium"
-                        ]
+                        "nor": ["armor:category:heavy", "armor:category:medium"]
                     }
                 ],
                 "selector": "fly",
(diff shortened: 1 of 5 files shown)


packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
```diff
@@ -39,39 +39,36 @@
...
             {
                 "key": "BaseSpeed",
                 "predicate": [
                     "wings-deployed",
                     {
-                        "nor": [
-                            "armor:category:heavy",
-                            "armor:category:medium"
-                        ]
+                        "nor": ["armor:category:heavy", "armor:category:medium"]
```
TikaelSol: Did these get extracted? They should lint like they were before.

packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
```diff
@@ -39,39 +39,36 @@
         },
         "quantity": 1,
         "rules": [
+            {
+                "key": "RollOption",
+                "label": "Wings Deployed",
```
TikaelSol: We don't want non-localized text in rule element labels. So these would need to be localized. Basically add a new entry to re-en.json under `PF2E.SpecificRule.Equipment.UltralightWings.WingsDeployed`

packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
Eliott45: My bad. I restored the pack formatting so it matches the extracted style again.

packs/sf2e/equipment/augmentations/biotech/ultralight-wings-tactical.json
Eliott45: Thanks, fixed

static/lang/re-en.json
```diff
@@ -7776,6 +7757,9 @@
                 "Magboots": {
                     "ClimbMetalSurfaces": "Magboots — Climb metal surfaces"
                 },
+                "UltralightWings": {
+                    "WingsDeployed": "Wings Deployed"
```
```suggestion
                    "WingsDeployed": "Ultralight Wings — Wings Deployed"
```

PR #21811, opened 2026-03-21
Update shield warfare to work with shield bash

### LINKED ISSUE #19991: Guardian Warfare not increasing damage die of shield bash
Guardian Warfare is only increasing the damage dice for shield spikes and shield boss attacks, but is not doing anything for shield bash itself. 

Image

Image

iDantar:
Closes https://github.com/foundryvtt/pf2e/issues/19991

### DIFF
--- a/packs/pf2e/feats/class/guardian/level-1/shield-warfare.json
+++ b/packs/pf2e/feats/class/guardian/level-1/shield-warfare.json
@@ -27,18 +27,15 @@
         },
         "rules": [
             {
-                "itemType": "weapon",
                 "key": "ItemAlteration",
+                "itemType": "weapon",
                 "mode": "upgrade",
+                "property": "damage-dice-faces",
                 "predicate": [
-                    {
-                        "or": [
-                            "item:base:shield-boss",
-                            "item:base:shield-spikes"
-                        ]
-                    }
-                ],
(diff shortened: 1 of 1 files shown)


packs/pf2e/feats/class/guardian/level-1/shield-warfare.json
```diff
@@ -27,18 +27,15 @@
...
-                            "item:base:shield-boss",
-                            "item:base:shield-spikes"
-                        ]
-                    }
-                ],
-                "property": "damage-dice-faces"
+                    "item:equipped",
+                    "item:group:shield",
+                    "item:melee"
+                ]
```
pedrogrullada: I don't think it's that likely to come up, but this is broader than what the rules intend if new shield group weapons are introduced by Paizo or third party products.

Just worth noting, as we don't have a way to identify a Strike as a shield bash other than exhaustively listing shield base types, which runs into the opposite problem, not taking into account future or third party shields. This might still be good enough for our purposes.

Also, you'll want to sort your keys alphabetically or they'll throw out noise when extracting packs

iDantar: Heh, that was my thought process, give or take :)

Fixed the key order - thanks, will keep it in mind

PR #21815, opened 2026-03-22
Add some automation for Unshadowed Okoro

simonward:
Also add some missing rolloptions to Unshadowed inlines
Remove clumsy link from "Effect: Swarming Bites"


packs/pf2e/strength-of-thousands-bestiary/book-6-shadows-of-the-ancients/unshadowed-okoro.json
```diff
@@ -3154,7 +3172,23 @@
...
+                        "toggleable": true
+                    },
+                    {
+                        "keep": "lower",
+                        "key": "RollTwice",
+                        "predicate": [
+                            "push-my-luck-misfortune"
+                        ],
+                        "selector": "all"
+                    }
```
TikaelSol: Kind of feel like this should be an effect with a badge of 3 to allow counting down (and no localization needed)

Issue #20775, opened 2025-11-21
Roll options in `@Check` rolls do not satisfy predicates in DC-targeting Flat Modifiers


pedrogrullada:
If you have a Flat Modifier like this, as an example
```json
{
  "key": "FlatModifier",
  "selector": [
    "fortitude-dc"
  ],
  "value": 1,
  "type": "status",
  "predicate": [
    "forced-movement"
  ]
}
```
It will increase the parent actor's fortitude DC against Shove attempts (which include `forced-movement`, but this applies to any roll option present).

If you instead target the parent actor with a roll from an inline check, such as `@Check[athletics|against:fortitude|options:forced-movement]`, the modifier will not be taken into account in the roll, despite the required roll option being present.

PR #21820, opened 2026-03-23
Fix inline check options for against DCs

Eliott45:
Pass inline check `options` to `against` defense statistic resolution

**Testing**
- tested against a target with `Root Boots` and `Grip Ground` enabled
- confirmed `Flowing Hair` uses the same increased Fortitude DC as `Shove`

this fixes contextual inline checks #20775, but not bare chat-authored checks without actor or item context


stwlam:
Need to retarget branch

stwlam:
What's a circumstance that isn't addressed? I'd prefer it if we can close the issue.

PR #21823, opened 2026-03-23
Limit Knave's Standard damage to off-guard targets only

### LINKED ISSUE #21822: Knave's standard effect applies on every critical hit
Knave's standard is only supposed to be applied on a critical hit on an off-guard opponent, but is being applied to every critical hig.

pedrogrullada:
- Closes #21822

### DIFF
--- a/packs/pf2e/equipment-effects/effect-knaves-standard.json
+++ b/packs/pf2e/equipment-effects/effect-knaves-standard.json
@@ -29,6 +29,9 @@
                 "dieSize": "d4",
                 "hideIfDisabled": true,
                 "key": "DamageDice",
+                "predicate": [
+                    "target:condition:off-guard"
+                ],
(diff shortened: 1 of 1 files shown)


packs/pf2e/equipment-effects/effect-knaves-standard.json
simonward:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.equipment-srd.Item.Knave's Standard], @UUID[Compendium.pf2e.equipment-srd.Item.Knave's Standard (Greater)], @UUID[Compendium.pf2e.equipment-srd.Item.Knave's Standard (Major)]</p>\n<p>Whenever you critically succeed with a Strike against an off-guard target, the Strike deals additional precision damage.</p>"
```

PR #21835, opened 2026-03-25
Create logs of exclusions from anachronism and refine logic

CarlosFdez:
Supercedes https://github.com/foundryvtt/pf2e/pull/21775

While this can't really be used to detect broken links, it does make it easier to refine the logic. Entries must now match name, type, and feat category (if feat) to be excluded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 icons/\!dox/
 
 coverage
+build/logs
 dist/
 docs/
 foundryconfig.json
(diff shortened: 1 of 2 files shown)


CarlosFdez:
From checking the log, some bugs were found, so once those are resolved I will undraft.

CarlosFdez:
this is better than what we had before, so I'll merge it now, but it will need further refinement. Better for dev builds right now though.

PR #21852, opened 2026-03-28
Add Damage Note to Tolokand's Toxic Blood ability

ammalagonc:


simonward:
Consider setting the visibility to owner.

PR #21859, opened 2026-03-29
Fix chthonian in shade (dead vault)

rectulo:
chthonian fhtagn ph'nglui mglw'nafh cthulhu r'lyeh wgah'nagl fhtagn


simonward:
~~Consider making the same fix in Shade (Outer Rifts)'s Planar Incarnation - Abyss (The Larvae).~~

Never mind, I see that's in a separate PR.

PR #21874, opened 2026-03-31
Fix action type of Warden's Reload

TikaelSol:


simonward:
Consider setting the frequency.

PR #21877, opened 2026-03-31
Avoid eager construction of all synthetic actors during world load

stwlam:
Tested with a fresh Kingmaker world

Before:

After:


CarlosFdez (approved):
Looks good to me. I'm gonna assume this is some precursor v14 thing

PR #21887, opened 2026-04-01
Add description and built-in upgrades to Shobhad Longrifle

Cuingamehtar:
Since they are built-in and upgradable, it felt reasonable to include them in the item itself.
But I note that I didn't subtract their price from the item's.

~~Ultralight wings got extracted with everything else automatically.~~


stwlam:
Please exclude unrelated revisions

PR #21893, opened 2026-04-01
Add Hellfire Dispatches dedication details to Archetypes journal

ammalagonc:
Also removes Hellfire Armiger and Hellfire Signifer pages, as those got gobbled up ᗧ···ᗣ···ᗣ··


TikaelSol:
Do we need to redirect the links to the journals?

ammalagonc:
> Do we need to redirect the links to the journals?

UUID redirects don't work with Journal links. They break on build/don't pass checks

PR #21902, opened 2026-04-02
Add missing Crysmal NPC

ammalagonc:


simonward:
From Rage of Elements Errata (Summer 2024, 1st Printing):
"Page 103: The crysmal's name wasn't updated during the remaster process. These creatures are now called cullitox."

PR #21906, opened 2026-04-02
Add missing Galactic Ancestries and Hellfire Dispatches localization strings

Cuingamehtar:


packs/sf2e/ancestry-features/moyishuu/soulfeel.json
```diff
@@ -32,19 +32,19 @@
...
                         "value": "joy"
                     },
                     {
-                        "label": "PF2E.SpecificRule.Moyishuu.Soulfeel.Sorrow",
+                        "label": "SF2E.SpecificRule.Moyishuu.Soulfeel.Sorrow",
                         "value": "sorrow"
                     },
                     {
-                        "label": "PF2E.SpecificRule.Moyishuu.Soulfeel.Wonder",
+                        "label": "SF2E.SpecificRule.Moyishuu.Soulfeel.Wonder",
```
TikaelSol: We're only using the SF2E prefix for overrides

Cuingamehtar: Overrides use PF2E prefix. They wouldn't work as overrides otherwise without additional data manipulation in memory.

This is a SF ancestry feature.
Ultralight Wings' RE was also recently accepted into SF2e prefix.

TikaelSol: Oversight on my part then, should be PF2E

PR #21912, opened 2026-04-03
Deadly Simplicity properly upgardes two-hand trait

hollyvalenta:
As per 2026 Player Core errata


pedrogrullada:
This should be handled by the code rather than a bespoke data update, in my opinion.

PR #21919, opened 2026-04-04
Add strike buttons for quick alchemy bombs

CarlosFdez:
Handy if the alchemist has other actions or decided to macro the quick alchemy button

https://github.com/user-attachments/assets/ae1dfe69-f65c-42f4-97a7-ad0729eced63

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -1504,14 +1504,17 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         ];
 
         action.variants = ([0, 1, 2] as const).map((mapIncreases) => ({
-            get label(): string {
+            get penalty(): number {
                 const penalty = createMAPenalty(initialMAPs, mapIncreases);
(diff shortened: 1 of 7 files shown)


stwlam:
Since the quick alchemy bomb ends up in your hand if one is free, isn't it already macroable?

CarlosFdez:
With a custom script macro, maybe, but it'd be a tricky macro where you have to read the message, parse it to get the item, then attack with it. Or pull the last action in the list (I think they're unsorted internally?), in case they're already holding another weapon. If we don't already we can add some flags to the message I guess.

The macro depicted in the video is just drag dropping quick alchemy to the hot bar.

CarlosFdez:
I considered making the [Activate] button work for weapons but I couldn't think of a decent workflow for that. Sending a bomb to chat doesn't put strike buttons there or anything. In general bombs are a bit clunky in terms of using them, given they often have additional effects which aren't reported in a strike, but I have no ideas for that problem.

stwlam:
retarget merge branch

CarlosFdez:
Rebased

PR #21924, opened 2026-04-04
Update weapon damage die calculation for errata

### LINKED ISSUE #17741: Deadly Simplicity not working with Two-Hand trait
As title - Two-Hand changes a weapon's damage die to the one listed in the trait, which Deadly Simplicity does not buff to the next die size up.

iDantar:
Allows stacking of different ways to change damage die size, following the second spring errata 2026
- Fixes https://github.com/foundryvtt/pf2e/issues/17741

Changes:
1. Apply damage die alteration to two-hand trait directly
2. Allow overriding damage die on weapons without one - to enable implementation of things like [Guerrilla Weaponry](https://2e.aonprd.com/Feats.aspx?ID=7932)


src/module/item/weapon/helpers.ts
```diff
@@ -18,6 +21,50 @@ function upgradeWeaponTrait(trait: string): string {
...
+        return trait.replace(new RegExp(String.raw`d${value}$`), downgraded);
+    }
+    return trait;
+}
+
+/**
+ * When item alterations change weapon damage faces, keep an existing `two-hand-d*` trait in sync so
+ * `processTwoHandTrait` still applies the correct two-hand die after the base die changes.
+ */
+function adjustTwoHandTraitForDamageFacesChange(weapon: WeaponPF2e, mode: "downgrade" | "override" | "upgrade"): void {
```
stwlam: Move this to inside the handler; no need to have a system-wide export if it's used in only one place.

iDantar: Moved, thanks. Ready for review

src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -341,17 +342,20 @@ const ITEM_ALTERATION_HANDLERS = {
...
+                item.system.damage.die = nextDamageDieSize({ downgrade: die });
+                adjustTwoHandTraitForDamageFacesChange(item, mode);
             } else if (mode === "override" && typeof data.alteration.value === "number") {
-                if (data.alteration.value > Number(item.system.damage.die.replace("d", ""))) {
+                if (die && data.alteration.value > Number(die.replace("d", ""))) {
                     item.flags[SYSTEM_ID].damageFacesUpgraded = true;
                 }
                 item.system.damage.die = `d${data.alteration.value}`;
+                adjustTwoHandTraitForDamageFacesChange(item, mode);
             }
```
stwlam: Better as a switch on mode: there's no point in checking down the conditional chain if `die` isn't set, for example.

iDantar: It does look cleaner that way, thanks. 
P.S. One exception in override, allowing to set the die if it wasn't, to enable [Guerrilla Weaponry](https://2e.aonprd.com/Feats.aspx?ID=7932)

PR #21932, opened 2026-04-05
Fix precision immunity crit damage miscalculation

### LINKED ISSUE #19318: Precision immunity, critical damage and fatal/deadly
Version: 6.12.4
With the Title combination, the damage is not correctly calculated.
Only half of the precision damage is deducted.
I think the issue is while calculating the instance.componentTotal("precision") it returns only half, probably related to the deadly/fatal extra die.
Issue happens with or without critical immunity.
Example without critical immunity
Image
Example with critical immunity:
Image

iDantar:
- Closes #19318

2 changes needed for the fix:
1. Setting `.options.crit` in both Grouping and ArithmeticExpression. Previously .options.crit was only set in Grouping's child node - having formula with any damage that doesn't multiply on crit caused that flag to not be set. 
2. Check for .options.crit in deepFindTerms instead of componentTotal - before `.options.crit` was only checked on the head node, so had to account for nested ones.

Tested on:
- normal/critical damage
- with/without fatal
- with/without deadly
- against a target with crit immunity, precision immunity, both present and both absent

Please let me know if additional testing is required

### DIFF
--- a/src/module/system/damage/roll.ts
+++ b/src/module/system/damage/roll.ts
@@ -599,15 +599,9 @@ class DamageInstance extends AbstractDamageRoll {
     }
 
     componentTotal(component: "precision" | "splash"): number {
-        if (!this._evaluated) {
-            throw ErrorPF2e("Component totals may only be accessed from an evaluated damage instance");
-        }
-
-        const terms = deepFindTerms(this.head, { flavor: component });
-        const rawTotal = terms.reduce((total, t) => total + (Number(t.total) || 0), 0);
-        const critMultiplier = Number(this.head.options.crit) || 1;
-
-        return rawTotal * critMultiplier;
+        if (!this._evaluated) throw ErrorPF2e("Component totals may only be accessed from evaluated damage instances");
+        const matches = deepFindTerms(this.head, { flavor: component });
+        return matches.reduce((total, [term, multiplier]) => total + (Number(term.total) || 0) * multiplier, 0);
     }
 
     /**
--- a/src/module/system/damage/terms.ts
+++ b/src/module/system/damage/terms.ts
@@ -13,14 +13,16 @@ class ArithmeticExpression extends terms.RollTerm<ArithmeticExpressionData> {
         super(termData);
 
         this.operator = termData.operator;
-        this.operands = termData.operands.slice(0, 2).map((datum) => {
-            if (datum instanceof terms.RollTerm) return datum;
+        this.operands = termData.operands.slice(0, 2).map((operand) => {
+            if (operand instanceof terms.RollTerm) return operand;
             const TermCls =
-                CONFIG.Dice.termTypes[datum.class ?? ""] ??
-                Object.values(CONFIG.Dice.terms).find((t) => t.name === datum.class) ??
+                CONFIG.Dice.termTypes[operand.class ?? ""] ??
+                Object.values(CONFIG.Dice.terms).find((t) => t.name === operand.class) ??
                 terms.Die;
-            return simplifyTerm(TermCls.fromData(datum));
+            return simplifyTerm(TermCls.fromData(operand));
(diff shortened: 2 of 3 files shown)


src/module/system/damage/roll.ts
```diff
@@ -598,16 +598,14 @@ class DamageInstance extends AbstractDamageRoll {
...
         if (!this._evaluated) {
             throw ErrorPF2e("Component totals may only be accessed from an evaluated damage instance");
         }
 
-        const terms = deepFindTerms(this.head, { flavor: component });
-        const rawTotal = terms.reduce((total, t) => total + (Number(t.total) || 0), 0);
-        const critMultiplier = Number(this.head.options.crit) || 1;
-
-        return rawTotal * critMultiplier;
+        return componentDamageTotal(this.head, component, {
```
stwlam: Looks like the single-use helper is the whole method?

iDantar: Basically yes. I tried to stay close to the old structure, but this indeed looks silly - would it be better to drop the helper and just move the functionality here?

iDantar: Made a commit moving it all into the main method, just in case the answer is positive

src/module/system/damage/roll.ts
```diff
@@ -598,16 +598,39 @@ class DamageInstance extends AbstractDamageRoll {
...
 
-        const terms = deepFindTerms(this.head, { flavor: component });
-        const rawTotal = terms.reduce((total, t) => total + (Number(t.total) || 0), 0);
-        const critMultiplier = Number(this.head.options.crit) || 1;
+        const stripCritDoubling = options?.stripCritDoubling ?? false;
+        const flavorMatches = (t: RollTerm): boolean => t.flavor.split(",").includes(component);
 
-        return rawTotal * critMultiplier;
+        const visit = (t: RollTerm, mult: number): number => {
+            let sum = flavorMatches(t) ? (Number(t.total) || 0) * mult : 0;
```
stwlam: no `let`

src/module/system/damage/roll.ts
```diff
@@ -598,16 +598,39 @@ class DamageInstance extends AbstractDamageRoll {
...
+                const innerMult = stripCritDoubling ? mult : mult * (Number(t.options.crit) || 1);
+                sum += visit(t.term, innerMult);
+                return sum;
+            }
+
+            if (t instanceof ArithmeticExpression) {
+                const [left, right] = t.operands;
+                if (t.operator === "*" && left instanceof terms.NumericTerm && tupleHasValue([2, 3], left.number)) {
+                    const critOpt = Number(t.options.crit);
+                    const rightMult = stripCritDoubling ? mult : critOpt === left.number ? mult * critOpt : mult;
```
stwlam: This is kind of nonsensical:
- What's the point of using `tupleHasValue` here?
- Why is `left.number` then checked for truthiness a second time?

iDantar:
While debugging found that the root cause is not in applyIWR, but in how the damage is constructed - still figuring out what exactly it is.

iDantar:
Sorry Shark, I re-wrote the whole fix from scratch - could you please take another look?

src/module/system/damage/terms.ts
```diff
@@ -221,7 +238,7 @@ class Grouping extends terms.RollTerm<GroupingData> {
         } else {
             super(termData);
             this.term = childTerm;
-            if (this.#dataIsCriticalDoubling(termData.term)) {
+            if (dataIsCriticalDoubling(termData.term)) {
                 this.options.crit = 2;
             }
```
stwlam: Isn't this already determined now by `ArithmeticExpression`?

iDantar: Good call! I wasn't sure that removing it from Grouping wouldn't break anything else, but after testing looks like it's fine. Thanks :)

PR #21933, opened 2026-04-05
Fix Ammunition RE applying to Combination Weapons

### LINKED ISSUE #20445: Ammunition RE applying to Combination Weapons
I've discovered a handful of ammunition types' (Elemental Ammunition, Exsanguinating Ammunition, and Magnetic Shot for example. Not a comprehensive list) REs `selector` to be `{item|id}-damage` instead of `{item|id}-ranged-damage`. This causes them to apply to the melee half of Combination Weapons.

iDantar:
Fixes https://github.com/foundryvtt/pf2e/issues/20445

### DIFF
--- a/packs/pf2e/equipment/antler-arrow.json
+++ b/packs/pf2e/equipment/antler-arrow.json
@@ -43,7 +43,7 @@
                 "outcome": [
                     "success"
                 ],
-                "selector": "{item|_id}-damage",
+                "selector": "{item|id}-ranged-damage",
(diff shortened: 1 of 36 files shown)


simonward:
Some ammunition also uses the older `"{item|_id}-damage"` selector - for example "Eroding Bullet".

iDantar:
Thanks! Updated those as well - looks like all ammunition now uses the new selector. (aside from things like Roc-Shaft Arrow - which is a consumable, for some reason)

PR #21945, opened 2026-04-07
Align the implementation of Bless spell effect with Benediction

iDantar:
Currently the two spells have different implementations, which may confuse new players - this makes Benediction work as Bless by separating the aura into it's own spell-effect, that auto-applies the buff to allies


pedrogrullada:
Benediction was made this way on purpose so it's all a single-effect. If anything, bless would need to be condensed instead.

iDantar:
As long as they can be done in the same fashion - sounds good to me :)

iDantar:
Updated the code: implemented Bless to have a single effect being responsible for aura and the bonus.
Used the Aura: Bless item to avoid updating all places that reference it.

Slight problem: Bless and Benediction now have the same implementation, but different naming conventions - would it be better to rename 'Spell Effect: Benediction' to 'Aura: Benediction' to keep them in line, or to rename the Bless and replace it's references?

pedrogrullada:
I'd follow the `Spell Effect` convention

iDantar:
Renamed and updated the references, please let me know if anything else needs fixing
And thanks for quick responses, appreciate you!

pedrogrullada:
We do need to handle links on existing copies of the spell in people's worlds such that nothing changes from their perspective. UUID-redirecting the deleted effect to the condensed effect should be enough. If you haven't done redirects before, searching "redirect" in the closed PRs should give plenty of examples. (Otherwise feel free to ask in the Discord's  #development-help channel!)

iDantar:
Thanks for the details, found and added a redirect!
Just in case: does that mean I could have avoided updating all those npc/monster sheets with the new effect name? Or is it still good to keep things clean?

pedrogrullada:
Not doing so wouldn't break anything, but we would have typically wanted to update the NPCs to match anyway. It's better for consistency.

(Also you probably want to rename the PR if you are able to, otherwise it's up to whoever merges this 🙃)

iDantar:
Sounds good!
Oh, right, totally forgot about that, thanks!

pedrogrullada (approved):
Looks good to me!

PR #21948, opened 2026-04-08
Address roll inspector svelte warnings and fix modifier scrolling

CarlosFdez:
This was an issue in v13 as well, so I made it a PR instead.

### DIFF
--- a/src/module/apps/roll-inspector/app.svelte
+++ b/src/module/apps/roll-inspector/app.svelte
@@ -89,7 +89,7 @@
             {/if}
         </header>
 
-        <ul class="modifier-list">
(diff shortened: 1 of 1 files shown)


src/module/apps/roll-inspector/app.svelte
```diff
@@ -236,8 +234,6 @@
     }
 
     .modifiers {
-        grid-area: "modifiers";
```
CarlosFdez: Grid areas weren't named, and quotes means this didn't work anyways. It actually breaks the layout if you remove the quotes so that css actually resolves it.

PR #21956, opened 2026-04-10
Fix  detection of inline damage rolls

In3luki:
Looks like the core anchor element no longer has a `mode`. I'm not 100% sure if removing the check without replacement is safe.

Before:
image

After:
image

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -64,7 +64,7 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
     ): Promise<HTMLAnchorElement | null> {
         const anchor = await super._createInlineRoll(match, rollData, options);
         const formula = anchor?.dataset.formula;
-        if (!formula || anchor.dataset.flavor) return anchor;
+        if (!formula || !("flavor" in anchor.dataset)) return anchor;
         const roll = ((): DamageRoll | null => {
             try {
                 return new DamageRoll(formula);
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -64,7 +64,7 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
     ): Promise<HTMLAnchorElement | null> {
         const anchor = await super._createInlineRoll(match, rollData, options);
         const formula = anchor?.dataset.formula;
-        if (!formula || !("flavor" in anchor.dataset)) return anchor;
+        if (!formula || anchor.dataset.flavor) return anchor;
         const roll = ((): DamageRoll | null => {
             try {
                 return new DamageRoll(formula);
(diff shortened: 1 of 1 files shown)


stwlam:
Not perfect, but look for `dataset.flavor`, too, I guess.

In3luki:
Done

src/module/system/text-editor.ts
```diff
@@ -64,7 +64,7 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
     ): Promise<HTMLAnchorElement | null> {
         const anchor = await super._createInlineRoll(match, rollData, options);
         const formula = anchor?.dataset.formula;
-        if (!formula || !objectHasKey(CONFIG.ChatMessage.modes, anchor.dataset.mode)) return anchor;
+        if (!formula || anchor.dataset.flavor) return anchor;
```
stwlam:
```suggestion
        if (!formula || !("flavor" in anchor.dataset)) return anchor;
```

In3luki:
Oh wait. I commited the changes here without testing but isn't an empty `flavor` always present in the dataset? Can't test right now.

In3luki:
`!("flavor" in anchor.dataset)` is always `false`. From `TextEditor#_createInlineRoll`:
```js
    a.dataset.flavor = matchedFlavor ?? label ?? "";
```

CarlosFdez:
Looks like !anchor.dataset.flavor then?

stwlam:
Why were we looking for `mode` in the first place?

src/module/system/text-editor.ts
```diff
@@ -64,7 +64,7 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
     ): Promise<HTMLAnchorElement | null> {
         const anchor = await super._createInlineRoll(match, rollData, options);
         const formula = anchor?.dataset.formula;
-        if (!formula || !objectHasKey(CONFIG.ChatMessage.modes, anchor.dataset.mode)) return anchor;
+        if (!formula || !anchor.dataset.flavor) return anchor;
```
stwlam:
```suggestion
        const messageMode = anchor.dataset.mode ?? "public";
        if (!formula || !(messageMode in CONFIG.ChatMessage.modes)) return anchor;
```

In3luki: No idea what the check is for but that should work.

src/module/system/text-editor.ts
```diff
@@ -64,7 +64,8 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
     ): Promise<HTMLAnchorElement | null> {
         const anchor = await super._createInlineRoll(match, rollData, options);
         const formula = anchor?.dataset.formula;
-        if (!formula || !objectHasKey(CONFIG.ChatMessage.modes, anchor.dataset.mode)) return anchor;
+        const messageMode = anchor.dataset.mode ?? "public";
```
In3luki:
```suggestion
        const messageMode = anchor?.dataset.mode ?? "public";
```

PR #21964, opened 2026-04-10
Add module tab registration API to PartySheetPF2e

scooper4711:
Hello,
 I'm the developer for https://github.com/scooper4711/pfs-chronicle-generator - a module that adds an additional tab to the Party sheet to allow PFS GMs to generate chronicles for their players right from inside Foundry. It works together with a browser extension I've also written to make PFS Session reporting easier. Think "rpg chronicles lite, but from inside of Foundry".
 
 Currently, I have to do a several hacks to add a tab to the party sheet, and because of this, when players make changes, the party sheet is re-drawn and my new Society tab is de-selected and the GM is put back into the Overview tab. It's proven to be quite disruptive for the workflow. 
 
 I'm humbly submitting this patch to allow any module to register a new tab with associate callback hooks. This would work for any module, not just mine. 

**Problem statement**
Currently modules have to hook renderPartySheetPF2e, use setTimeout hacks to wait for DOM readiness, and manually inject tab elements. This API gives modules a first-class way to add tabs with proper lifecycle management.

**Solution**

Introduce a static registerModuleTab() method on PartySheetPF2e that allows modules to add custom tabs to the party sheet. Modules can optionally provide renderCallback and activateListeners functions to integrate with the sheet's region-based rendering lifecycle.

When a renderCallback is provided, the tab element receives a data-region attribute so that #renderRegions invokes the callback during both initial render and partial re-renders (e.g. when party membership changes). The sheet app instance is passed as the first argument to both callbacks, giving modules access to the actor and sheet state without relying on hook timing or DOM traversal.

Tabs registered without callbacks get an empty content div that modules can populate via the existing renderPartySheetPF2e hook.

### DIFF
--- a/src/module/actor/party/sheet.ts
+++ b/src/module/actor/party/sheet.ts
@@ -29,7 +29,45 @@ interface PartySheetRenderOptions extends ActorSheetRenderOptionsPF2e {
     actors?: boolean;
 }
 
+interface ModuleTab {
+    /** A unique tab identifier used as the data-tab attribute value */
+    tabId: string;
+    /** The display label rendered in the navigation link */
+    label: string;
+    /** Optional async callback that returns an HTML string for the tab content region */
+    renderCallback?: (app: PartySheetPF2e, element: HTMLElement, data: object) => Promise<string>;
+    /** Optional callback to attach event listeners after renderCallback populates innerHTML. Ignored if renderCallback is absent. */
+    activateListeners?: (app: PartySheetPF2e, element: HTMLElement) => void;
+}
+
+/** Options for registering a module tab with render callback integration */
+interface ModuleTabOptions {
+    /** Async callback that returns an HTML string for the tab content region */
+    renderCallback: (app: PartySheetPF2e, element: HTMLElement, data: object) => Promise<string>;
+    /** Optional callback to attach event listeners after renderCallback populates innerHTML */
+    activateListeners?: (app: PartySheetPF2e, element: HTMLElement) => void;
+}
+
 class PartySheetPF2e extends ActorSheetPF2e<PartyPF2e> {
+    static readonly moduleTabs: ModuleTab[] = [];
+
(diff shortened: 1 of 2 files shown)


CarlosFdez:
DOM readiness should be assured already in the render hook and I've written code that assumed such, the fact that its not is a bit odd. Furthermore, party sheet is appv1, and its very likely we're gonna update it to appv2 sometime within the v14 cycle. And finally, any sort of tab hijacking thing would be best as a more general thing.

To avoid tabs swapping back I remember you have to add it to the tab groups somehow, but that might only be available once its AppV2.

scooper4711:
I’m happy to update it to Appv2. 

The reason the tabs swap back and forth is because when focus is decided, my code hasn’t had a chance to manipulate the dom and re-insert my tab. 

I had a simpler patch which didn’t do anything regarding callbacks and that resolved the issue for me, but I wanted to make something generic that would work for any module. So I added in the callback support. 

My users are reporting it as very annoying to have to keep switching back to my tab when their players are logged in. This patch resolves that. I worked at making the patch both small and explanatory. 

Are there any changes I can make so that this patch would be accepted? Any guidance you can provide?

CarlosFdez:
Looking at foundry source code, you can resolve the tab refocus issue in appv1 by adding your custom tab in `_tabs` but I don't have sample code for it.

Party sheet will need to be rewritten in svelte specifically. If you happen to be very familiar with that framework then sure but maybe best to let one of us handle it otherwise. I have a half finished conversion that was on ice due to other things.

If Svelte makes it absolutely impossible for module devs to do tabs we could consider it but until then its definitely possible to add a tab and have it refocus, its just a bit clunky to do so.

scooper4711:
It got pretty hacky to keep the tab selected, both scroll areas in the right locations and to not lose focus on the input field you're tying in.
As you work on the AppV2 conversion, I request that you add support for dyamic tab registration.

PR #21967, opened 2026-04-11
Fix migrating old spell consumables

CarlosFdez:
There is an error that manifests running abomination vaults since some of the wand items are old. It works fine at first, but using Troubleshooting -> Remigrate errors because embedded spells do not automatically receive a _stats if missing like real documents do. However the type is a bit awkward and made me null out a bunch of things, so if there's a cleaner way let me know.

This error is not new to v14 and exists in v13 as well.


CarlosFdez:
Updated to push to v13

CarlosFdez:
I don't see how to draft from mobile, but I realized that I may need to change and test something so it's not ready for merge yet.

PR #21969, opened 2026-04-11
Fix Region shape resetting after release of shape control handles

### LINKED ISSUE #21965: v14 Regions Snap Back To Original Position When Dragged By Handles
PF2e 8.0.0 (and SF2e 1.0.0), Foundry v14.360 Windows Portable, Windows 10

When dragging regions by handles in a fresh world with no modules installed, the regions snap back to their original sizes once the handle is released. Observed in both PF2e and SF2e systems, but in DnD5e 5.3.0 this did not occur, so I believe it must be something the XF2e systems are doing.

The overall region can be moved fine by clicking and dragging anywhere within a region, but the handles for resizing plus the circular handle for moving an individual shape in the region will snap back. 

Additionally there are no errors or messages of any kind in the console when attempting to drag by the handle

Image

In3luki:
- Closes #21965

Should work the same as the base `ShapeObjectMixin#_onDragLeftDrop` now:
```js
    /** @inheritDoc */
    _onDragLeftDrop(event) {
      const handle = event.interactionData.handle;
      if ( handle ) return handle.controls._onDragDrop(event);
      return super._onDragLeftDrop(event);
    }
```

### DIFF
--- a/src/module/canvas/region.ts
+++ b/src/module/canvas/region.ts
@@ -33,7 +33,6 @@ class RegionPF2e<TDocument extends RegionDocumentPF2e = RegionDocumentPF2e> exte
     protected override async _onDragLeftDrop(event: PlaceablesLayerPointerEvent<this>): Promise<RegionDocument[]> {
         const handle = event.interactionData.handle;
         if (handle) return handle.controls._onDragDrop(event) ?? [];
-
         const clones = event.interactionData.clones ?? [];
         const updates = clones.map((clone) => {
             const shapes = clone.document.shapes.map((s) => s.toObject(false));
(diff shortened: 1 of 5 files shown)


src/module/canvas/region.ts
```diff
@@ -31,6 +31,9 @@ class RegionPF2e<TDocument extends RegionDocumentPF2e = RegionDocumentPF2e> exte
     /** Save the coordinates of the new drop location(s). */
     protected override async _onDragLeftDrop(event: PlaceablesLayerPointerEvent<this>): Promise<TDocument[]>;
     protected override async _onDragLeftDrop(event: PlaceablesLayerPointerEvent<this>): Promise<RegionDocument[]> {
+        const handle = event.interactionData.handle;
+        if (handle) return handle.controls._onDragDrop(event) ?? [];
+
```
stwlam:
```suggestion
```

PR #21972, opened 2026-04-11
Fix two type errors in TS 6.0 and TS Go

In3luki:
These are the only errors I found while using the ts-go preview in VSCode. Typescript 6.0 has the same two errors.

I've added the pf2e repo as a git submodule to one of my module projects as the source of foundry and system types. Merging this would unblock me from updating the module to TS6.

### DIFF
--- a/src/module/system/action-macros/crafting/repair.ts
+++ b/src/module/system/action-macros/crafting/repair.ts
@@ -8,14 +8,16 @@ import { SelectItemDialog } from "./select-item.ts";
 
 async function repair(options: RepairActionOptions): Promise<void> {
     // resolve item
-    const item = options.item ?? (await (options.uuid ? fromUuid(options.uuid) : SelectItemDialog.getItem("repair")));
+    const item =
+        options.item ?? (await (options.uuid ? fromUuid<ItemPF2e>(options.uuid) : SelectItemDialog.getItem("repair")));
 
     // ensure specified item is a valid repair target
     if (!(item instanceof ItemPF2e)) {
         console.warn("PF2e System | No item selected to repair: aborting");
         return;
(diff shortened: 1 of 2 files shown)


src/module/system/action-macros/crafting/repair.ts
```diff
@@ -8,14 +8,15 @@ import { SelectItemDialog } from "./select-item.ts";
...
     // resolve item
-    const item = options.item ?? (await (options.uuid ? fromUuid(options.uuid) : SelectItemDialog.getItem("repair")));
+    const item =
+        options.item ?? (await (options.uuid ? fromUuid<ItemPF2e>(options.uuid) : SelectItemDialog.getItem("repair")));
 
     // ensure specified item is a valid repair target
     if (!(item instanceof ItemPF2e)) {
         console.warn("PF2e System | No item selected to repair: aborting");
         return;
     } else if (!(item instanceof PhysicalItemPF2e)) {
```
stwlam:
```suggestion
    }
    if (!(item instanceof PhysicalItemPF2e)) {
```

stwlam:
Main problem I noticed with ts 6 were "Type instantiation is excessively deep" errors

In3luki:
I haven't seen any of those yet but I'm mostly using the ts-go preview right now.

In3luki:
The main blocker for the systems is probably that `svelte-preprocess` doesn't support ts 6 yet. I've just `npm ci --force`d my module around that which seems to work for now.

src/module/system/action-macros/crafting/repair.ts
stwlam: Actually, can't this just be an `isOfType` check?

In3luki: yep

PR #21988, opened 2026-04-12
Fix incorrect assets paths in Vite hmr mode

In3luki:
Some files (e.g. `systems/pf2e/assets/system-banners/03.webp`) were being replaced with a "This file is for a running vite dev server and is not copied to a build." stub.
After some digging I found out that for a normal build these assets are copied from `static/assets/system-banners/pf2e/` to `dist/pf2e/assets/system-banners/` but that doesn't happen when the files are served via HMR. Only the original paths in Vite's `publicDir` are available so the banner is served at `system/pf2e/assets/system-banners/pf2e/03.webp` instead.

Adding an explicit proxy rule for the assets ensures that they are serverd from the foundry server with the correct path.

Before:
image

After:
image

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -285,7 +285,7 @@ const config = Vite.defineConfig(({ command, mode }): Vite.UserConfig => {
             open: "/game",
             proxy: {
                 [`^(?!/systems/${SYSTEM_ID}/)`]: `http://localhost:${foundryPort}/`,
-                [`^/systems/${SYSTEM_ID}/lang`]: `http://localhost:${foundryPort}/`,
+                [`^/systems/${SYSTEM_ID}/(?:assets|lang)`]: `http://localhost:${foundryPort}/`,
                 "/socket.io": {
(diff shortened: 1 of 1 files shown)


vite.config.ts
```diff
@@ -285,6 +285,7 @@ const config = Vite.defineConfig(({ command, mode }): Vite.UserConfig => {
             open: "/game",
             proxy: {
                 [`^(?!/systems/${SYSTEM_ID}/)`]: `http://localhost:${foundryPort}/`,
+                [`^/systems/${SYSTEM_ID}/assets`]: `http://localhost:${foundryPort}/`,
                 [`^/systems/${SYSTEM_ID}/lang`]: `http://localhost:${foundryPort}/`,
```
stwlam:
```suggestion
                [`^/systems/${SYSTEM_ID}/(?:assets|lang)`]: `http://localhost:${foundryPort}/`,
```

In3luki: Right, that's cleaner.

PR #21999, opened 2026-04-13
Handle some svelte warnings

CarlosFdez:
This handles everything I noticed that wasn't in the trade dialog.

While $derived() would work I didn't really see a need for that in these cases.

### DIFF
--- a/src/module/apps/compendium-browser/components/result-item.svelte
+++ b/src/module/apps/compendium-browser/components/result-item.svelte
@@ -12,7 +12,6 @@
         activeTabName: ContentTabName | "";
         entry: CompendiumBrowserIndexData;
     }
-
     const { entry, activeTabName }: ResultItemProps = $props();
 
     async function onClickButton(uuid: string, action: "buy-item" | "open-sheet" | "take-item"): Promise<void> {
--- a/src/module/apps/compendium-browser/components/result-item.svelte
+++ b/src/module/apps/compendium-browser/components/result-item.svelte
@@ -8,9 +8,12 @@
     import type { ContentTabName } from "../data.ts";
     import type { ActorPF2e } from "@actor";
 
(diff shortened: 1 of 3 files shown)


src/module/sheet/components/item-traits.svelte
```diff
@@ -9,12 +9,11 @@
...
 
     const { rarity, traits = [], properties = [] }: ItemTraitsProps = $props();
-    const rarityLabel = rarity ? CONFIG.PF2E.rarityTraits[rarity] : null;
 </script>
 
 <div class="tags paizo-style">
-    {#if rarity && rarity !== "common" && rarityLabel}
-        <span class={["tag", "rarity", rarity].join(" ")}>{_loc(rarityLabel)}</span>
+    {#if rarity && rarity !== "common"}
+        <span class={["tag", "rarity", rarity].join(" ")}>{_loc(CONFIG.PF2E.rarityTraits[rarity])}</span>
```
stwlam: can we just interpolate `rarity` into a string?

src/module/apps/compendium-browser/components/result-item.svelte
```diff
@@ -8,9 +8,12 @@
...
     import type { ActorPF2e } from "@actor";
 
-    const props: { activeTabName: ContentTabName | ""; entry: CompendiumBrowserIndexData } = $props();
-    const entry = props.entry;
-    const activeTabName = props.activeTabName;
+    interface ResultItemProps {
+        activeTabName: ContentTabName | "";
+        entry: CompendiumBrowserIndexData;
+    }
+
```
stwlam:
```suggestion
```

src/module/sheet/components/item-traits.svelte
CarlosFdez: Yeah, was just habit from projects with optional class names.

PR #22008, opened 2026-04-13
Reduce width increase of scene config

CarlosFdez:
Base width is 600, v14 support made it 700 to fit the tab. This attempts to do something a bit more middleground, since I think the wide scene config looks strange.

image

### DIFF
--- a/src/module/scene/sheet.ts
+++ b/src/module/scene/sheet.ts
@@ -10,7 +10,7 @@ import fields = foundry.data.fields;
 export class SceneConfigPF2e<TDocument extends ScenePF2e> extends fa.sheets.SceneConfig<TDocument> {
     static override DEFAULT_OPTIONS: DeepPartial<fa.api.DocumentSheetConfiguration> = {
         sheetConfig: false,
-        position: { width: 650 },
+        position: { width: 660 },
         actions: {
             openAutomationSettings: SceneConfigPF2e.#onClickOpenAutomationSettings,
             openWorldClockSettings: SceneConfigPF2e.#onClickOpenWorldClockSettings,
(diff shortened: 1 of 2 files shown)


src/module/scene/sheet.ts
```diff
@@ -10,7 +10,7 @@ import fields = foundry.data.fields;
 export class SceneConfigPF2e<TDocument extends ScenePF2e> extends fa.sheets.SceneConfig<TDocument> {
     static override DEFAULT_OPTIONS: DeepPartial<fa.api.DocumentSheetConfiguration> = {
         sheetConfig: false,
-        position: { width: 700 },
+        position: { width: 650 },
```
stwlam:
```suggestion
        position: { width: 660 },
```

660 is a standard with for core applications if we want to keep some kind of uniformity

src/styles/scene/_sheet.scss
```diff
@@ -1,4 +1,7 @@
 .application.scene-config {
+    .top-tabs {
+        gap: var(--spacer-8);
```
stwlam, line 3: The variable name is wrong here, so I guess it's not needed?

src/module/scene/sheet.ts
CarlosFdez: Making it wider feels weird but that may be a case of not being accustomed. So sure.

src/styles/scene/_sheet.scss
CarlosFdez: Isn't that the core one now? It was spacer 16 B4 and I saw it get smaller

src/module/scene/sheet.ts
stwlam: You still knocked it down by 40px!

PR #22011, opened 2026-04-14
Update TokenConfig for v14

### LINKED ISSUE #22006: [v14] Token Depth
With the v14 update it looks like they added `depth` as a new field alongside of `width` / `height`, while it doesn't appear to be used for anything it would be nice if the system allowed this field to be edited and set it when the token size is linked with actor size. Based on what I am reading the depth would be set to the same value as width/height. So a large creature would have a width/height/depth of 2 while a huge creature would have a width/height/depth of 3 and so forth.

Image

Edit: Here is a screenshot showing that depth can't be edited, only width/height.

Image

In3luki:


stwlam:
I got the template in a separate PR

PR #22015, opened 2026-04-14
Allow modification of options.keep for pf2e.preReroll and pf2e.reroll

### LINKED ISSUE #21494: Pass keep as an object to the hooks pf2e.preReroll and pf2e.reroll so it can be modified

7H3LaughingMan:
Closes #21494

This is pretty much the same as #21495 but had to remake the PR to target the v14-dev branch

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -510,7 +510,6 @@ class Check {
         })();
         unevaluatedNewRoll.options.isReroll = true;
         const hookOptions = { keep: options.keep };
-
         Hooks.callAll("pf2e.preReroll", Roll.fromJSON(oldRollJSON), unevaluatedNewRoll, resource, hookOptions);
 
         // Evaluate the new roll and call a second hook allowing the roll to be altered
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -498,7 +498,7 @@ class Check {
         Hooks.callAll("pf2e.reroll", Roll.fromJSON(oldRollJSON), newRoll, resource, hookOptions);
 
         // Generate Flavor Text
-        let rerollFlavor = _loc(`PF2E.RerollMenu.MessageKeep.${options.keep}`);
+        let rerollFlavor = _loc(`PF2E.RerollMenu.MessageKeep.${hookOptions.keep}`);
         if (resource) {
             // If the reroll costs a hero or mythic point, first check if the actor has one to spare and spend it
             if (rerollingActor?.isOfType("character")) {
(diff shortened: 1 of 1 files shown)


src/module/system/check/check.ts
```diff
@@ -509,21 +509,23 @@ class Check {
             );
         })();
         unevaluatedNewRoll.options.isReroll = true;
-        Hooks.callAll("pf2e.preReroll", Roll.fromJSON(oldRollJSON), unevaluatedNewRoll, resource, options.keep);
+        const hookOptions = { keep: options.keep };
+
```
CarlosFdez:
```suggestion
```

CarlosFdez:
I was about to merge since I saw we agreed on doing it in v8, but I'm looking at it more carefully, and it looks like options.keep is already used earlier to create the reroll flavor, and its not affected by this option. Are you sure the label works?

7H3LaughingMan:
> I was about to merge since I saw we agreed on doing it in v8, but I'm looking at it more carefully, and it looks like options.keep is already used earlier to create the reroll flavor, and its not affected by this option. Are you sure the label works?

I totally missed that portion, yeah the flavor text would be wrong if a hook changes it so I moved it to after the hook calls.

src/module/system/check/check.ts
```diff
@@ -509,21 +489,43 @@ class Check {
...
+        Hooks.callAll("pf2e.preReroll", Roll.fromJSON(oldRollJSON), unevaluatedNewRoll, resource, hookOptions);
 
         // Evaluate the new roll and call a second hook allowing the roll to be altered
         const allowInteractive = context.messageMode !== "blind";
         const newRoll = await unevaluatedNewRoll.evaluate({ allowInteractive });
-        Hooks.callAll("pf2e.reroll", Roll.fromJSON(oldRollJSON), newRoll, resource, options.keep);
+        Hooks.callAll("pf2e.reroll", Roll.fromJSON(oldRollJSON), newRoll, resource, hookOptions);
+
+        // Generate Flavor Text
+        let rerollFlavor = _loc(`PF2E.RerollMenu.MessageKeep.${options.keep}`);
```
In3luki:
```suggestion
        let rerollFlavor = _loc(`PF2E.RerollMenu.MessageKeep.${hookOptions.keep}`);
```

7H3LaughingMan: Oops, fixed

CarlosFdez:
Alrighty, this might break some modules that somehow survived to v14, but thanks for the PR.

PR #22035, opened 2026-04-15
Convert `AncestrySystemData` to `DataModel`

In3luki:
Getting in some practice before going for a bigger one. No errors in my test world.

### DIFF
--- a/src/module/item/ancestry/data.ts
+++ b/src/module/item/ancestry/data.ts
@@ -1,8 +1,14 @@
-import { CreatureTrait, Language } from "@actor/creature/index.ts";
-import { AttributeString } from "@actor/types.ts";
-import { ABCSystemData, ABCSystemSource } from "@item/abc/index.ts";
-import { BaseItemSourcePF2e, ItemTraits } from "@item/base/data/system.ts";
-import { Size, TraitsWithRarity, ValuesList } from "@module/data.ts";
+import type { CreatureTrait, Language } from "@actor/creature/index.ts";
+import type { AttributeString } from "@actor/types.ts";
+import { ATTRIBUTE_ABBREVIATIONS } from "@actor/values.ts";
+import { ABCFeatureEntryField, type ABCSystemSource } from "@item/abc/index.ts";
+import { ItemSystemModel, type ItemSystemSchema } from "@item/base/data/model.ts";
+import type { BaseItemSourcePF2e, ItemTraits } from "@item/base/data/system.ts";
+import { SIZES, type Size, type TraitsWithRarity, type ValuesList } from "@module/data.ts";
+import { RarityField } from "@module/model.ts";
+import { RecordField, SlugField } from "@system/schema-data-fields.ts";
+import type { AncestryPF2e } from "./document.ts";
+import fields = foundry.data.fields;
 
 type AncestrySource = BaseItemSourcePF2e<"ancestry", AncestrySystemSource>;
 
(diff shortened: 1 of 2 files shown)


stwlam:
We need to make sure that the model cleaning isn't going to hit before a system migration.

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +47,159 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                rarity: new RarityField(),
+                value: new fields.ArrayField(new fields.StringField()),
+            }),
+            items: new RecordField(
+                new fields.StringField({ required: true, nullable: false }),
+                new ABCFeatureEntryField(),
+            ),
+            additionalLanguages: new fields.SchemaField({
+                count: new fields.NumberField({ nullable: false, initial: 1 }),
+                value: new fields.ArrayField(new fields.StringField()),
```
stwlam: the element field needs to be required in array fields

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +47,159 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                custom: new fields.StringField(),
+            }),
+            alternateAncestryBoosts: new fields.ArrayField(
+                new fields.StringField({ required: true, nullable: false, choices: [...ATTRIBUTE_ABBREVIATIONS] }),
+            ),
+            boosts: new RecordField(
+                new fields.StringField({ required: true, nullable: false }),
+                new fields.SchemaField({
+                    value: new fields.ArrayField(new fields.StringField({ choices: [...ATTRIBUTE_ABBREVIATIONS] })),
```
stwlam: same

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +47,159 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                    selected: new fields.StringField({ nullable: true }),
+                }),
+            ),
+            flaws: new RecordField(
+                new fields.StringField({ required: true, nullable: false }),
+                new fields.SchemaField({
+                    value: new fields.ArrayField(new fields.StringField({ choices: [...ATTRIBUTE_ABBREVIATIONS] })),
```
stwlam: ditto

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +47,159 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+            ),
+            voluntary: new fields.SchemaField(
+                {
+                    boost: new fields.StringField({
+                        nullable: true,
+                        required: false,
+                        initial: undefined,
+                        choices: [...ATTRIBUTE_ABBREVIATIONS],
+                    }),
+                    flaws: new fields.ArrayField(new fields.StringField({ choices: [...ATTRIBUTE_ABBREVIATIONS] })),
```
stwlam: etc.

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +47,159 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                value: new fields.ArrayField(
+                    new fields.StringField({
+                        choices: [
+                            "common",
+                            ...LANGUAGES_BY_RARITY.common,
+                            ...LANGUAGES_BY_RARITY.uncommon,
+                            ...LANGUAGES_BY_RARITY.rare,
+                            ...LANGUAGES_BY_RARITY.secret,
+                        ],
+                    }),
```
stwlam: We can't restrict; languages can be added via homebrew settings

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                rarity: new RarityField(),
+                value: new fields.ArrayField(new fields.StringField({ required: true })),
+            }),
+            items: new RecordField(
+                new fields.StringField({ required: true, nullable: false }),
+                new ABCFeatureEntryField(),
+            ),
+            additionalLanguages: new fields.SchemaField({
+                count: new fields.NumberField({ nullable: false, initial: 1 }),
+                value: new fields.ArrayField(new fields.StringField({ required: true })),
```
stwlam:
```suggestion
                value: new fields.ArrayField(new fields.StringField({ required: true, blank: false })),
```

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+class AncestrySystemData extends ItemSystemModel<AncestryPF2e, AncestrySystemSchema> {
+    static override defineSchema(): AncestrySystemSchema {
+        return {
+            ...super.defineSchema(),
+            traits: new fields.SchemaField({
+                otherTags: new fields.ArrayField(
+                    new SlugField({ required: true, nullable: false, initial: undefined }),
+                ),
+                rarity: new RarityField(),
+                value: new fields.ArrayField(new fields.StringField({ required: true })),
```
stwlam:
```suggestion
                value: new fields.ArrayField(new fields.StringField({ required: true, blank: false })),
```

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                        initial: undefined,
+                        choices: [...ATTRIBUTE_ABBREVIATIONS],
+                    }),
+                    flaws: new fields.ArrayField(
+                        new fields.StringField({ choices: [...ATTRIBUTE_ABBREVIATIONS], required: true }),
+                    ),
+                },
+                { required: false, initial: undefined },
+            ),
+            hp: new fields.NumberField({ required: true, initial: 6, nullable: false }),
```
stwlam: Validate along the same lines as the sheet (only odd integers between 0 and 12)

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+            languages: new fields.SchemaField({
+                value: new fields.ArrayField(new fields.StringField({ required: true })),
```
stwlam:
```suggestion
                value: new fields.ArrayField(new fields.StringField({ required: true, blank: false })),
```

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+            }),
+            speed: new fields.NumberField({ required: true, initial: 25, nullable: false }),
+            size: new fields.StringField({ required: true, nullable: false, initial: "med", choices: [...SIZES] }),
+            hands: new fields.NumberField({ required: true, initial: 2, nullable: false }),
+            reach: new fields.NumberField({ required: true, initial: 5, nullable: false }),
```
stwlam: Same as hp

src/module/item/ancestry/data.ts
```diff
@@ -40,7 +46,155 @@ interface AncestrySystemSource extends ABCSystemSource {
...
+                        new fields.StringField({ choices: [...ATTRIBUTE_ABBREVIATIONS], required: true }),
+                    ),
+                },
+                { required: false, initial: undefined },
+            ),
+            hp: new fields.NumberField({ required: true, initial: 6, nullable: false }),
+            languages: new fields.SchemaField({
+                value: new fields.ArrayField(new fields.StringField({ required: true })),
+            }),
+            speed: new fields.NumberField({ required: true, initial: 25, nullable: false }),
```
stwlam: same as hp (maybe increase the max to 60 to give a little more room for homebrew)

In3luki:
I've been experimenting with data model migrations again but I can't get the unpruned source values to stick around long enough to be available for the system migration. There are some new levers with `DataModel._preCleanData`  and the new `DataModelCleaningOptions` but setting `prune: false` doesn't seem to do anything. When the source hits my ancestry test migration (system.speed to system.speed2) the original source value is gone and only values defined in the data schema remain..
I've tried overriding in the `ItemSystemModel` and in `ActorPF2e` but that didn't do anything either.

PR #22042, opened 2026-04-16
Fix multiclass dedications proficiencies

Longstrider:
Also fixed SF2e Hellknight Dedication Intimidation proficiency and empty choices, plus added skill proficiency automation to some dedications.

### DIFF
--- a/packs/pf2e/feats/archetype/fighter/fighter-dedication.json
+++ b/packs/pf2e/feats/archetype/fighter/fighter-dedication.json
@@ -60,6 +60,39 @@
                 "mode": "override",
                 "path": "system.proficiencies.classDCs.fighter.attribute",
                 "value": "{item|flags.system.rulesSelections.attribute}"
+            },
+            {
+                "choices": [
+                    {
+                        "label": "PF2E.Skill.Acrobatics",
+                        "predicate": [
+                            "skill:acrobatics:rank:0"
+                        ],
+                        "value": "acrobatics"
+                    },
+                    {
+                        "label": "PF2E.Skill.Athletics",
+                        "predicate": [
+                            "skill:athletics:rank:0"
+                        ],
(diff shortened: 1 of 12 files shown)


packs/pf2e/feats/archetype/fighter/fighter-dedication.json
```diff
@@ -60,6 +60,39 @@
...
+                        "label": "PF2E.Skill.Athletics",
+                        "predicate": [
+                            "skill:athletics:rank:0"
+                        ],
+                        "value": "athletics"
+                    }
+                ],
+                "flag": "skill",
+                "key": "ChoiceSet",
+                "prompt": "PF2E.SpecificRule.Prompt.Skill"
```
stwlam, line 78: This needs to not error out if the character already has trained proficiency in both skills.

packs/pf2e/feats/archetype/fighter/fighter-dedication.json
```diff
@@ -60,6 +60,48 @@
...
+                            "skill:athletics:rank:0"
+                        ]
+                    }
+                ],
+                "prompt": "PF2E.SpecificRule.Prompt.Skill"
+            },
+            {
+                "key": "ActiveEffectLike",
+                "mode": "upgrade",
+                "path": "system.skills.{item|flags.system.rulesSelections.skill}.rank",
```
TikaelSol, line 83: This will throw warnings if the above rule is skipped. If trained in both we should have another choice set with the same flag that fires off to ask for another skill. See Skill Training for the choice set and add being trained in both skills as a predicate to it

Longstrider: I was thinking about it, just haven't seen such implementation anywhere else. I'll look into it.

PR #22057, opened 2026-04-17
Fix items and actions sent by players defaulting to blind

### LINKED ISSUE #22046: Items players post to chat from their sheets are sent Blind GM
v8.0.2, as title. Send to chat sends blind gm. Holding ctrl makes it send publicly.

Image

Image

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/22046

### DIFF
--- a/src/module/sheet/helpers.ts
+++ b/src/module/sheet/helpers.ts
@@ -148,8 +148,7 @@ function eventToRollParams(event: Maybe<Event>, rollType: { type: "check" | "dam
 
 /** Set roll mode from a user's input: used for messages that are not actually rolls. */
 function eventToMessageMode(event: Maybe<Event>): ChatMessageMode | undefined {
-    if (!isRelevantEvent(event)) return undefined;
-    return isControlDown(event) || game.user.isGM ? undefined : "blind";
+    return isRelevantEvent(event) && isControlDown(event) ? "blind" : undefined;
 }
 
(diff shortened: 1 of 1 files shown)


CarlosFdez:
The original behavior in v13 was to first check if control was down, and if so, make it either blind/gm based on if its a gm or player. If control wasn't held down, return public roll.

If we need to make a distinction between blind/gm roll I can bring that back pretty easily, but blind and gm are effectively the same thing from the gm's perspective, and I think stwlam's intent was to remove that distinction but I can't be sure.

stwlam:
yep, bring it back

src/module/sheet/helpers.ts
```diff
@@ -148,8 +148,7 @@ function eventToRollParams(event: Maybe<Event>, rollType: { type: "check" | "dam
 
 /** Set roll mode from a user's input: used for messages that are not actually rolls. */
 function eventToMessageMode(event: Maybe<Event>): ChatMessageMode | undefined {
-    if (!isRelevantEvent(event)) return undefined;
-    return isControlDown(event) || game.user.isGM ? undefined : "blind";
+    return isRelevantEvent(event) && isControlDown(event) ? "blind" : undefined;
```
stwlam:
```suggestion
    if (!isRelevantEvent(event) || !isControlDown(event)) return undefined;
    return game.user.isGM ? "gm" : "blind";
```

PR #22071, opened 2026-04-18
Fix Predicate Input Label on Effects in Aura Rules

websterguy:


CarlosFdez:
I looked into the code, this might address the symptoms well enough, but it looks like the problem is that the effect's own predicate field does not have the predicate label assigned in the data model

websterguy:
It's true that the `PredicateField` doesn't have an explicit label set, but neither does any other `new PredicateField` that I can see. If you would rather, I can set it for this instance instead of the posted fix. I confirmed it does fix it as well.

CarlosFdez:
Most of them are set by `LOCALIZATION_PREFIXES`. But the stuff in the aura effects explicitly set a label property. I think PredicateField could maybe have a label set by default.

websterguy:
It looks like setting a label in the constructor of `PredicateField` causes some other forms to fail to localize it.

image

Could be an autogenerated forms issue, as some of them still seem to work. Should I just set the label explicitly in the `new PredicateField` in `effectSchemaField` for the predicate in the effects tab of the aura?

PR #22090, opened 2026-04-20
Move `svelecte-pf2e` to CSS anchor positioning and Popover API

jfn4th:
Full support for `position-anchor` rolled out in January with Chrome/Edge 144 and Firefox 147. Some people needing to update their browsers might pop out of the woodwork.

At first I thought this would be doable with just anchor positioning, but popover proved necessary due to the top level usage of `backdrop-filter` on dark mode windows. Removing the portal and adding a wrapper div also allowed for rewriting the global selectors such that they're scoped to `svelecte-pf2e` instances.

### DIFF
--- a/src/module/sheet/components/svelecte-pf2e.svelte
+++ b/src/module/sheet/components/svelecte-pf2e.svelte
@@ -1,95 +1,48 @@
 <script lang="ts">
     import Svelecte from "svelecte";
     type SvelecteParams = Parameters<Required<Svelecte>["$set"]>[0];
+
     let { value = $bindable(), ...props }: SvelecteParams = $props();
+    const className = $derived(((props.class ?? "") + " pf2e").trim());
+    const positionResolver = $derived(props.positionResolver ?? popoverPositionResolver);
 
-    // Track the detached dropdown node and its theme source for syncing
-    let dropdownNode: HTMLElement | null = $state(null);
-    let themeSource: HTMLElement | null = $state(null);
+    // Promotes dropdown to top layer to escape clipping from ancestor backdrop-filter.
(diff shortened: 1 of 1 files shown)


src/module/sheet/components/svelecte-pf2e.svelte
```diff
@@ -100,14 +53,11 @@
...
             --sv-border: 1px solid var(--input-border-color);
-
             --sv-selection-multi-wrap-padding: 0.15em;
             --sv-selection-gap: 0.2em;
             --sv-min-height: var(--input-height);
 
-            color: var(--color-text-primary);
-
-            .sv-input--text {
+            :global(.sv-input--text) {
```
stwlam, line 60: What's the necessity of a `:global` in a `:global`?

jfn4th: Svelte's scoping is per selector, not inherited from a parent selector. `:global(...)` only marks what's inside the parens as global, Svelte goes back to scoping after the closing paren, so nested selectors inside a `:global(...)` rule still get the component hash unless each one is also wrapped in `:global(...)`. To my understanding this is preferred, as wrapping each descendent in its own `:global(...)` keeps the rules namespaced under the local component scoped `.svelecte-pf2e` wrapper and not leaking to any other svelectes. `:global {}` blocks, by contrast, make everything within fully unscoped. 

We even get a nice warning if you try to nest styles within a `:global(...)` without tagging each selector as `:global(...)`

> `:global(...)` can be at the start or end of a selector sequence, but not in the middle
https://svelte.dev/e/css_global_invalid_placement

stwlam: wack

PR #22107, opened 2026-04-22
Propagate elite/weak updates to all troop siblings

CarlosFdez:
To get this to work without having to do a weird double update, I refactored how HP adjustments work as well. These are the technical details:

Before this update, `NPCPF2e#applyAdjustment()` updated the prepared max HP in order to trick the clamp in `CreaturePF2e#_preUpdate()`. Now, the max HP adjustment is checked during the clamp itself. So an update that passes the new adjustment and the new HP value is sufficient for the update.

The HP adjustment for weak is now a negative number, so we don't have to check for addition/subtraction anymore. I needed to access it outside of the npc class, but as it was private I made it a helper instead of opening it up.

### DIFF
--- a/src/module/actor/creature/helpers.ts
+++ b/src/module/actor/creature/helpers.ts
@@ -152,4 +152,22 @@ function imposeEncumberedCondition(actor: CreaturePF2e): void {
     }
 }
 
-export { AttackTraitHelpers, imposeEncumberedCondition, setImmunitiesFromTraits };
+/** Returns the elite/weak adjustment for a given level */
+function getHpAdjustment(level: number, adjustment: "elite" | "weak" | null): number {
+    if (adjustment === "elite") {
+        // Elite adjustment: Increase/decrease the creature's Hit Points based on its starting level (20+ 30HP, 5~19 20HP, 2~4 15HP, 1 or lower 10HP).
+        if (level >= 20) return 30;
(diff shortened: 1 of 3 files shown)


src/module/actor/creature/helpers.ts
```diff
@@ -152,4 +152,32 @@ function imposeEncumberedCondition(actor: CreaturePF2e): void {
...
+        // Elite adjustment: Increase/decrease the creature's Hit Points based on its starting level (20+ 30HP, 5~19 20HP, 2~4 15HP, 1 or lower 10HP).
+        if (level >= 20) {
+            return 30;
+        } else if (level <= 19 && level >= 5) {
+            return 20;
+        } else if (level <= 4 && level >= 2) {
+            return 15;
+        } else if (level <= 1) {
+            return 10;
+        }
```
stwlam:
```suggestion
        if (level >= 20) return 30;
        if (level <= 19 && level >= 5) return 20;
        if (level <= 4 && level >= 2) return 15;
        if (level <= 1) return 10;
```

stwlam:
Why not make it protected at the creature level?

CarlosFdez:
> Why not make it protected at the creature level?

elite/weak adjustments are NPC only concerns. Checking that the actor is an npc in the creature level is also awkward but at least its an internal only foible.

PR #22110, opened 2026-04-23
Use modifier type to find proficiency modifier in mythic rerolls

### LINKED ISSUE #22081: Reroll with mythic adds level twice when character has untrained improvisation
Had a player roll an untrained athletics check with untrained improvisation then reroll the check with a mythic point. The system still applied the untrained improvisation bonus to the roll on top of the mythic proficiency bonus

Original roll was 13 +13 for untrained improv
Mythic reroll was reported as 50 with a raw roll of 14 which would account for 26 for mythic at level 13 and 13 for untrained improv.

Image

pedrogrullada:
To account for Untrained Improvisation and other differently-slugged modifiers. Also check that the modifier found is enabled
- Closes #22081

Worth noting that there is still a visual bug where the old modifier's tag is not strikethroughed and replaced by the Mythic tag when the modifier is something other than just a normal proficiency modifier (with the `proficiency` slug). I think this is because the code expects that particular span to have the `proficiency` slug, but that doesn't apply to non-standard proficiency modifiers, such as those from rule elements like Untrained Improvisation. I don't know how to fix that though.

image

### DIFF
--- a/src/module/system/check/check.ts
+++ b/src/module/system/check/check.ts
@@ -470,7 +470,7 @@ class Check {
         const unevaluatedNewRoll = ((): CheckRoll => {
             if (resource?.slug !== "mythic-points" || !actor.isOfType("character")) return oldRoll.clone();
             // Create a new CheckRoll in case of a mythic point reroll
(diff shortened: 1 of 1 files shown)


CarlosFdez:
Thanks for the PR!

PR #22127, opened 2026-04-24
Update Base Document types

7H3LaughingMan:
This does a couple of things to update the base documents.

- Removes `DocumentMetadata` and replaces it with `DocumentClassMetadata` in the correct file.
- `static get metadata()` should now return a Readonly version of the metadata
- Adds `DocumentAuthorField` which is used by `BaseNote`
- Fixes `SceneLevelsSetField` so that the interior `DocumentIdField` isn't modelled as `string | null` it should always just be a `string`
- Base documents that have a "parent" defined as a generic parameter now have a default
- Base documents should now have an updated schema/metadata


types/foundry/common/abstract/_types.d.mts
```diff
@@ -57,6 +59,10 @@ export interface DataModelUpdateOptions {
     restoreDelta?: boolean;
 }
 
+export type DatabaseAction = "get" | "create" | "update" | "delete";
+
+export type DatabaseWriteAction = "create" | "update" | "delete";
```
stwlam:
```suggestion
export type DatabaseWriteAction = Exclude<DatabaseAction, "get">;
```

types/foundry/common/abstract/_types.d.mts
```diff
@@ -168,14 +174,17 @@ export interface DatabaseDeleteCallbackOptions extends Omit<
...
     | DatabaseGetOperation<TParent>
     | DatabaseCreateOperation<TParent>
     | DatabaseUpdateOperation<TParent>
     | DatabaseDeleteOperation<TParent>;
 
+export type DatabaseWriteOperation<TParent extends Document | null> =
+    | DatabaseCreateOperation<TParent>
+    | DatabaseUpdateOperation<TParent>
+    | DatabaseDeleteOperation<TParent>;
+
```
stwlam:
```suggestion
export type DatabaseWriteOperation<TParent extends Document | null> = Exclude<DatabaseOperation<TParent>, DatabaseGetOperation<TParent>;
```
(format as necessary)

types/foundry/common/documents/actor.d.mts
```diff
@@ -4,20 +4,17 @@ import {
...
 
 /**
- * The Document definition for an Actor.
+ * The Actor Document.
  * Defines the DataSchema and common behaviors for an Actor which are shared between both client and server.
- * @memberof documents
- *
- * @param data    Initial data from which to construct the Actor
- * @param context Construction context options
+ * @category Documents
```
stwlam:
```suggestion
```
Omit categories: they're haphazardly set in core

types/foundry/common/documents/actor.d.mts
```diff
@@ -94,31 +93,18 @@ type ActorSchema<
...
-    /** The _id of a Folder which contains this Actor */
     folder: fields.ForeignDocumentField<BaseFolder>;
-    /** The numeric sort value which orders this Actor relative to its siblings */
     sort: fields.IntegerSortField;
-    /** An object which configures ownership of this Actor */
     ownership: fields.DocumentOwnershipField;
-    /** An object of optional key/value flags */
     flags: fields.DocumentFlagsField;
-    /** An object of creation and access information. */
     _stats: fields.DocumentStatsField<ActorUUID>;
```
stwlam, line 107: Put all these back (same for all other document subclasses)

stwlam (changes requested):

types/foundry/common/data/fields.d.mts
```diff
@@ -1627,7 +1640,7 @@ export class JavaScriptField<
 /**
  * A nullable set field used to reference a set of specific Scene Level documents.
  */
-export class SceneLevelsSetField extends SetField<DocumentIdField<string>> {
+export class SceneLevelsSetField extends SetField<DocumentIdField<string, true, false, true>> {
```
stwlam, line 1846: Isn't that the default?

7H3LaughingMan: https://github.com/foundryvtt/pf2e/blob/v14-dev/types%2Ffoundry%2Fcommon%2Fdata%2Ffields.d.mts#L1138

It isn't the default, the default allows nullable values which results in this getting modelled as `Set<string | null>`

types/foundry/common/documents/actor.d.mts
7H3LaughingMan: I think I put this in there because I saw it in one of the newer ones so I wasn't sure if you wanted that in there.

stwlam:
looks like you got a rebase snafu here

7H3LaughingMan:
> looks like you got a rebase snafu here

Yeah, I forgot that I needed to do a force push after a rebase. I am going to finish up the comments tomorrow, been busy with some real life stuff and adding comments to the ones that didn't have one. Once I do that I will see if I can fix this or if need be redo it.

PR #22137, opened 2026-04-25
Condense Adapt Self Spell Effects

ammalagonc:


packs/pf2e/spell-effects/spell-effect-adapt-self.json
```diff
@@ -0,0 +1,68 @@
+{
+    "_id": "kTFIq6FOziSP2D3e",
+    "img": "icons/creatures/magical/spirit-undead-ghost-purple.webp",
+    "name": "Spell Effect: Adapt Self",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Adapt Self].</p>\n<p>You gain one of the following options:</p><ul><li><p>A Swim Speed equal to half your Speed.</p></li><li><p>Darkvision</p></li></ul>"
```
simonward:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.spells-srd.Item.Adapt Self]</p>\n<p>You gain one of the following options:</p><ul><li><p>A Swim Speed equal to half your Speed.</p></li><li><p>Darkvision</p></li></ul>"
```

packs/pf2e/spell-effects/spell-effect-adapt-self.json
```diff
@@ -0,0 +1,68 @@
...
+        },
+        "publication": {
+            "license": "ORC",
+            "remaster": true,
+            "title": "Pathfinder Lost Omens Divine Mysteries"
+        },
+        "rules": [
+            {
+                "choices": [
+                    {
```
simonward, line 27: Should we include the choices that we can't automate? It might be confusing otherwise, since they're no longer listed next to the appropriate choices in the spell.
Also, since there will now be a single effect, it could at least be used for tracking duration?

ammalagonc: Hmm - the one kicker to this would be the Environmental Protection effect, which has a different duration (10 minutes vs 1 minute).

PR #22145, opened 2026-04-26
Fix tag strikethrough when rerolling untrained improv using mythic

CarlosFdez:
Fixes a visual issue mentioned here https://github.com/foundryvtt/pf2e/pull/22110. The strikethrough now applies to any proficiency type modifier, not specifically modifiers with the proficiency slug.

image


stwlam:
Isn't that a proficiency bonus and simply replaces any lower one?

CarlosFdez:
Yes, but the previous detection only worked on proficiency bonuses with the proficiency slug, not other proficiency bonuses (like untrained improvisation).

Issue #22139, opened 2026-04-25
Monster Core 2 - Bikkhasura - Incorrect skill?


TroelsL:
I am playing a bit around with this lovely repository, and I stumbled upon something that puzzles me. I am not 100% certain that it is a mistake, but the skills section for the Monster Core 2 Bikkhasura.

So at the very least a question - what does "+38" as a skill represent?
(Note that foundry loads the Actor fine - I just can't find this value anywhere in the UI, nor can I see what it reflects from the sourcebook).

From https://github.com/foundryvtt/pf2e/blob/v14-dev/packs/pf2e/pathfinder-monster-core-2/bikkhasura.json line 2070:
```
          "skills": {
            "+38": {
                "base": null
            },
            "acrobatics": {
                "base": 33,
                "special": []
            },
            "athletics": {
                "base": 38,
                "special": []
            },
            "diplomacy": {
                "base": 36,
                "special": []
            },
            "intimidation": {
                "base": 34,
                "special": []
            },
            "performance": {
                "base": 34,
                "special": []
            },
            "religion": {
                "base": 35,
                "special": []
            },
            "stealth": {
                "base": 33,
                "special": []
            }
        },
```

I have discovered similar skills in:
https://github.com/foundryvtt/pf2e/blob/v14-dev/packs/pf2e/pathfinder-monster-core-2/garadasura.json

PR #22146, opened 2026-04-26
Remove malformed skill entry from Garadasura

Razytos:


Razytos:
also mentioned in #22139

PR #22147, opened 2026-04-26
 Correct HP change from Drained for Svetocher and Sanguine Tenacity

### LINKED ISSUE #22120: Svetocher not calculating Drained 1 correctly
The HP Calculation for drained 1 with Svetocher should be X (Your level) x 0 = 0.
Currently the formula for drained 1 seems to still be  X (Your Level) x 1

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22120

Also fixes a bug where having both Sanguine Tenacity and Svetocher resulted in Drained 1 increasing max HP and Fort

### DIFF
--- a/src/module/rules/rule-element/lose-hit-points.ts
+++ b/src/module/rules/rule-element/lose-hit-points.ts
@@ -18,29 +18,36 @@ class LoseHitPointsRuleElement extends RuleElement<LoseHitPointsRuleSchema> {
     }
 
     override onCreate(actorUpdates: Record<string, unknown>): void {
-        if (this.ignored) return;
-        const value = this.#resolveInitialHpLoss();
+        if (this.ignored || !this.test()) return;
+        const value = Math.trunc(Math.abs(Number(this.resolveValue(this.value)) || 0));
         const currentHP = this.actor._source.system.attributes.hp.value;
         actorUpdates["system.attributes.hp.value"] = Math.max(currentHP - value, 0);
     }
 
     override beforePrepareData(): void {
-        if (this.ignored) return;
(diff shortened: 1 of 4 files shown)


src/module/rules/rule-element/lose-hit-points.ts
```diff
@@ -51,6 +51,18 @@ class LoseHitPointsRuleElement extends RuleElement<LoseHitPointsRuleSchema> {
...
+    #resolveInitialHpLoss(): number {
+        if (
+            this.item.isOfType("condition") &&
+            this.item.slug === "drained" &&
+            this.actor.getRollOptions().includes("drained-hp-effective-minus-one")
+        ) {
+            return this.actor.level * ((this.item.value ?? 1) - 1);
+        }
+        return Math.trunc(Math.abs(Number(this.resolveValue(this.value)) || 0));
+    }
```
stwlam: We're not going to adjust an RE for a specific item. What problem are you trying to solve?

iDantar: Application of Drained reduces current HP - I'm trying to mitigate that for Svetocher and Sanguine Tenacity, since they treat HP loss from that as if condition was 1 lower. 
I didn't find a way to do that with only data changes, so had to adjust the method that applies the HP loss - if I missed it, could you suggest a better way of approaching this please?

iDantar: @stwlam I found a different approach: implemented predicates for LoseHitPoints and adjusted Drained RE - could you take another look please?

PR #22172, opened 2026-04-28
Make Ageless Patience always apply on nat 1

### LINKED ISSUE #21731: Ageless Patience Adjust Degree of Success RE
Currently, the Adjust Degree of Success RE of Ageless Patience only adjusts the degree of success if the result is a critical failure. However, the text of Ageless Patience says this:
> If you do, you gain a +2 circumstance bonus to the check and don't automatically reduce your degree of success on a natural 1 (you get a critical failure only if your result is 10 lower than the DC).

I believe the correct interpretation is that you should adjust the degree of success to be one better regardless if the result is a critical failure or not - this would mean changing `"criticalFailure"` in the RE to `"all"`. This would make it such that, for example, a rolled success on a natural 1 doesn't automatically become a failure, and a rolled critical success on a natural 1 doesn't automatically become a normal success.

iDantar:
Closes https://github.com/foundryvtt/pf2e/issues/21731
Also implements rule for Perception

### DIFF
--- a/packs/pf2e/feats/ancestry/elf/level-5/ageless-patience.json
+++ b/packs/pf2e/feats/ancestry/elf/level-5/ageless-patience.json
@@ -46,7 +46,7 @@
             },
             {
                 "adjustment": {
-                    "criticalFailure": "one-degree-better"
+                    "all": "one-degree-better"
                 },
                 "key": "AdjustDegreeOfSuccess",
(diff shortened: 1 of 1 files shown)


stwlam:
PR title should be in imperative mood

iDantar:
Updated, sorry about that

PR #22174, opened 2026-04-28
Fix regression with combatant start and end turn events

CarlosFdez:
Fixes fast healing, regeneration, and persistent damage events.


stwlam:
What is the point of `roundOfLastTurnEnd`?

CarlosFdez:
Its something that's set by the existing code already when you end the turn. It looks like roundOfLastTurn is set during the turn start event, and roundOfLastTurnEnd is set at the end of turn event. Looking at the diff of the encounter refactor, it was in use by the previous logic as well.

If we want, we can completely kill the flag, it should functionally be equivalent to `roundOfLastTurn - 1` in most reasonable cases unless there is something I'm missing.

PR #22192, opened 2026-05-02
Migrate sf2e-anachronism images to fix broken images

### LINKED ISSUE #22190: Anachronism updates(?) breaks all image paths
It appears that the other day, I'm guessing when pf2e-anachronism was updated (just a guess though), that all the image paths that were in that module changed as the images seem to have now been moved into the pf2e game system itself.

Previously, images were located at: `modules/sf2e-anachronism/art/icons/...`, but now all these images have been migrated to `systems/pf2e/icons/...`.

This seems like a totally good move to consolidate, and it looks like all the compendiums have been updated to be aware of the new image paths, but it doesn't appear that there's any mechanism to migrate existing world content to the new path structure? 

This results in a ton of broken images in my games worlds on actors/sheets/items/etc for any sf2e-anachronism based content.

Is there a system or script to automate fixing this?

I personally have a lot of sf2e world content that is affected and manually fixing it all would be a pretty significant and time consuming undertaking. I was able to see this behavior in effect across two separate foundry servers after they were both updated.

I'm running Foundry v13.351 and the latest versions of the pf2e game system, as well as the sf2e-anachronism module. Not sure if it's relevant, but I also have the pf2e-anachronism module installed on my server, but obviously not active in these game worlds.

Any insight would be appreciated!

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/22190

### DIFF
--- a/src/module/migration/migrations/957-sf2e-anachronism-images.ts
+++ b/src/module/migration/migrations/957-sf2e-anachronism-images.ts
@@ -1,0 +1,39 @@
+import { ActorSourcePF2e } from "@actor/data/index.ts";
+import { TokenSource } from "@common/documents/token.mjs";
+import type { ItemSourcePF2e } from "@item/base/data/index.ts";
+import { recursiveReplaceString } from "@util";
+import { MigrationBase } from "../base.ts";
+
+/**
+ * As of SF2e Anachronism 2.0, images used by sf2e are now part of the repository.
+ */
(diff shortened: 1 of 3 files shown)


stwlam:
system migration??

CarlosFdez:
Last I checked there isn't really a good to do a module migration without bloating up the anachronism code and adding anachronism version numbers in flags. The anachronism code is by us so it should be fine.

PR #22200, opened 2026-05-02
Touch up style of system roll and template links

stwlam:


CarlosFdez:
Tested it locally and it looks like the line spacing between lines still looks good.

PR #22209, opened 2026-05-03
Fix errors deleting min and max effect badge fields

CarlosFdez:
I ran into this while migrating my world, but it occurs if _del is present in a badge's min/max. The effect _preUpdate attempts to clamp to the _del keys and errors. 

The error is reproducible by trying to delete the min or max badge values in the console on an effect that already has a badge
image

### DIFF
--- a/src/module/item/effect/data.ts
+++ b/src/module/item/effect/data.ts
@@ -144,8 +144,8 @@ class EffectBadgeField extends fields.TypedSchemaField<
                 counter: new fields.SchemaField({
                     type: new fields.StringField({ choices: ["counter"], required: true }),
                     labels: createLabelsField(),
-                    min: new fields.NumberField({ required: false, nullable: true, integer: true }),
-                    max: new fields.NumberField({ required: false, nullable: true, integer: true }),
+                    min: new fields.NumberField({ required: true, nullable: true, integer: true, initial: null }),
+                    max: new fields.NumberField({ required: true, nullable: true, integer: true, initial: null }),
                     value: new fields.NumberField({ required: true, nullable: false }),
                     loop: new fields.BooleanField({ required: false, nullable: false }),
                 }),
--- a/src/module/item/effect/document.ts
+++ b/src/module/item/effect/document.ts
@@ -150,34 +150,31 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
             if (duration.value === -1) duration.value = 1;
         }
 
-        // Run all badge change checks. As of V12, incoming data is not diffed, so we check the merged result
-        if (changed.system?.badge) {
+        // Run all badge change checks. Handle that it may be a partial diff and that there may be operators
+        const ForcedDeletion = foundry.data.operators.ForcedDeletion;
+        if (changed.system?.badge && !(changed.system.badge instanceof ForcedDeletion)) {
             const badgeSource = this._source.system.badge;
-            const badgeChange = fu.mergeObject(changed.system.badge, badgeSource ?? {}, { overwrite: false });
-            if (badgeChange.type === "counter") {
-                // Clamp to the counter value, or delete if decremented to 0
-                badgeChange.labels ??= null;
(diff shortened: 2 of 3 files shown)


src/module/item/effect/document.ts
```diff
@@ -156,10 +156,11 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
             const badgeChange = fu.mergeObject(changed.system.badge, badgeSource ?? {}, { overwrite: false });
             if (badgeChange.type === "counter") {
                 // Clamp to the counter value, or delete if decremented to 0
+                const badgeChangeAdjusted = fu.applyDataOperators(badgeChange);
```
stwlam: What's setting the operator, and where is it in the data?

CarlosFdez: I have no idea, other than it did. My data is now fixed as well, so I'd have to load a backup. But min/max are required: false fields and/or should probably not explode if you try to delete them regardless.

Trying to search the code, I have no idea how it even got set in the first place. My best guess right now is some sort of update diff thing.

CarlosFdez: Discussed privately but for posterity it occurs automatically when -=min or -=max is already present in the effect data, or manually if you call `effect.update({ "system.badge.min": _del })`

src/module/item/effect/document.ts
```diff
@@ -156,10 +156,11 @@ class EffectPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ab
...
                 badgeChange.labels ??= null;
                 const [minValue, maxValue] = badgeChange.labels
-                    ? [1, Math.min(badgeChange.labels.length, badgeChange.max ?? Infinity)]
-                    : [badgeChange.min ?? 1, badgeChange.max ?? Infinity];
+                    ? [1, Math.min(badgeChange.labels.length, badgeChangeAdjusted.max ?? Infinity)]
+                    : [badgeChangeAdjusted.min ?? 1, badgeChangeAdjusted.max ?? Infinity];
```
stwlam:
```suggestion
                const withoutOperators = fu.applyDataOperators(badgeChange);
badgeChange.labels ??= null;
const [minValue, maxValue] = badgeChange.labels
                    ? [1, Math.min(badgeChange.labels.length, withoutOperators.max ?? Infinity)]
                    : [withoutOperators.min ?? 1, withoutOperators.max ?? Infinity];
```
let's be clearer about how it's "adjusted"

src/module/item/effect/document.ts
stwlam: I don't think that's true

stwlam:
Upon testing I notice the same error is thrown with or without these changes.

CarlosFdez:
Had one more refactor thought in mind, and while I'm wondering if I should explicitly set min to 1 if labels are empty rather than letting prepared data continue to take care of it, this is otherwise probably near the best I can get without going out of scope.

Moving the deletion to the sheet and effects panel events may lead to better user behavior, but that's out of scope for this change.

stwlam:
hi supe

src/module/item/effect/data.ts
```diff
@@ -144,8 +144,8 @@ class EffectBadgeField extends fields.TypedSchemaField<
                 counter: new fields.SchemaField({
                     type: new fields.StringField({ choices: ["counter"], required: true }),
                     labels: createLabelsField(),
-                    min: new fields.NumberField({ required: false, nullable: true, integer: true }),
-                    max: new fields.NumberField({ required: false, nullable: true, integer: true }),
+                    min: new fields.NumberField({ required: false, nullable: true, integer: true, initial: null }),
+                    max: new fields.NumberField({ required: false, nullable: true, integer: true, initial: null }),
```
stwlam:
```suggestion
                    max: new fields.NumberField({ required: true, nullable: true, integer: true, initial: null }),
```

CarlosFdez: Updated but untested

PR #22229, opened 2026-05-04
Update Hooks Class

7H3LaughingMan:
- Adds `events`
- Updated `on` and `once` so that the last overload takes `any[]` as the parameters for the callback function. With out this change if you try to create a hook for something that is not explicitly defined you will get an error that no overload matches the call.


stwlam:
We don't trust hooks not explicitly defined (see `dice-so-nice-ready.ts`), so it should stay `unknown`.

PR #22230, opened 2026-05-04
[WIP] Tie default window sizes to selected font size

iDantar:
This is an experiment to improve user experience for people with non-default font size selected. Personally, I use font 7 (5 is the default), so each time I have to open any character or monster sheet, it's content doesn't fit into the screen and I have to re-adjust it.
Suggestions/thoughts/comments are very welcome!

Here's an example of changes in action:
Foundry_Virtual_Tabletop_JjylKBCCYE-ezgif com-optimize

### DIFF
--- a/src/global.ts
+++ b/src/global.ts
@@ -85,7 +85,14 @@ interface ClientSettingsPF2e extends fh.ClientSettings {
     get(scope: "core", key: "fontSize"): number;
     get(scope: "core", key: "noCanvas"): boolean;
     get(scope: "core", key: "messageMode"): ChatMessageMode;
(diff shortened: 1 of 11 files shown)


stwlam:
I don't think we can safely do this. Some people use larger fonts despite being on a tiny monitor. It's really a core issue: UI scaling should apply to applications as well as the baseline UI.

iDantar:
Hmm, that's a fair point, enforcing this on everyone may do more bad than good. Perhaps that could be a setting that's turned-off by default? 
Or is this something that's better off as a separate module?

stwlam:
module :)

CarlosFdez:
Gonna echo the concern stwlam has. This is a great module idea that would benefit more than simply pf2e, and is otherwise a core concern. Good attempt and idea though.

PR #22246, opened 2026-05-06
Refresh token bars on Temp HP updates.

### LINKED ISSUE #22239: Temp HP token bar doesnt update properly
Bug Description:

In version 8.1.1 the Temp HP bar on the token doesn't change unless there is a change made to normal HP (and on reload of the server).

Steps to reproduce:

Take a token, open its actor sheet, add/remove Temp HP. See if anything changes on the token.

jfn4th:
Closes #22239 

Before v14, Foundry redrew a token's bars on any actor update. v14 diffs `{value, max}` before and after the update and only redraws when one of those changed.

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -707,6 +707,11 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         const updates = Array.isArray(update) ? update : [update];
         this.simulateUpdate(updates[0]);
 
+        // Core's bar refresh only diffs {value, max}, so a temp-HP-only change won't trigger a redraw on its own.
+        if (this.scene.isView && updates.some((u) => u && fu.hasProperty(u, "system.attributes.hp.temp"))) {
+            this.object?.renderFlags.set({ refreshBars: true });
+        }
+
         // Follow up any actor (or descendant document thereof) modification with a size synchronization
         const actor = this.actor;
         if (!actor?.isOwner) return;


CarlosFdez:
Would you happen to know how and where `refreshBars` happens to get set in foundry normally? I want to take a look before approving this, but I'm not seeing where `redrawEffects` turns into `redrawEffects` + `refreshBars`.

From debugging, doesn't seem to be Token#_onUpdate().

jfn4th:
Yeah, here's the whole flow. In short, v14 sets `refreshBars` via the animation pipeline rather than explicitly on each actor update.

1. `client/documents/token.mjs:3704` - `_onRelatedUpdate` calls `this.object?.animate({bar1: this.bar1, bar2: this.bar2}, ...)` This is new in v14. In v13, `refreshBars` gets set alongside `refreshEffects` a few lines earlier.
2. `client/canvas/placeables/token.mjs:2147` - `Token##animate` runs `diffObject(from, to)` against `#animationData` (which for bars is just `{value, max}`, see `_getAnimationData` on line 2049). Empty diff → empty animation.
3. `client/canvas/placeables/token.mjs:2385-2393` -  each animation tick fires `_onAnimationUpdate(changed, context)`, which is where `refreshBars: ("bar1" in changed) || ("bar2" in changed)` lives.
4. `client/canvas/placeables/token.mjs:1375` - `if (flags.refreshBars) this.drawBars();` in the render flag handler

CarlosFdez:
Oh I see! I was looking in the right place, but there are multiple animation updates. The first one is blank. Thank you!

PR #22247, opened 2026-05-06
Add effect for Feast on Fear

simonward:
Also add inline healing to Skaveling and add some inline roll options


packs/pf2e/bestiary-effects/effect-feast-on-fear.json
```diff
@@ -0,0 +1,75 @@
...
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "self:signature:{item|origin.signature}"
+                ],
+                "selector": "all-speeds",
+                "slug": "feast-on-fear-bonus",
+                "type": "status",
+                "value": "@item.flags.system.rulesSelections.feastOnFear"
+            }
```
TikaelSol, line 61: We should probably adjust modifier instead

simonward: They have different selectors (Speed vs Speeds).

PR #22249, opened 2026-05-06
Display Heritages in the Features Tab

### LINKED ISSUE #22235: Display Heritages in the Features Tab in the Character Sheet
The goal: add a Heritages tab so that feats granted by heritages display here, not under the Bonus feat sections.

Supersedes #7546

iDantar:
- Closes #22235
Supersedes https://github.com/foundryvtt/pf2e/pull/22233

Inserts the heritage in Ancestry Features list and renders it's nested feats, if any.
Some examples:
image

### DIFF
--- a/src/module/actor/character/feats/collection.ts
+++ b/src/module/actor/character/feats/collection.ts
@@ -247,6 +247,9 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<string, Fe
                 if (granter?.isOfType("feat") && granter.grants.includes(feat) && isNested) {
                     continue;
                 }
+                if (granter?.isOfType("heritage") && isNested) {
+                    continue;
+                }
             }
 
             // Find the group then assign the feat
--- a/src/module/actor/character/feats/collection.ts
+++ b/src/module/actor/character/feats/collection.ts
@@ -1,9 +1,9 @@
 import type { CharacterPF2e } from "@actor";
-import type { FeatPF2e, ItemPF2e } from "@item";
+import type { FeatPF2e, HeritagePF2e, ItemPF2e } from "@item";
 import { sluggify } from "@util";
 import * as R from "remeda";
 import { FeatGroup } from "./group.ts";
-import { FeatGroupData, FeatSlotData } from "./types.ts";
+import type { FeatGroupData, FeatSlot, FeatSlotData } from "./types.ts";
 
 class CharacterFeats<TActor extends CharacterPF2e> extends Collection<string, FeatGroup<TActor>> {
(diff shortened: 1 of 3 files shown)


src/module/actor/character/sheet.ts
```diff
@@ -288,7 +310,23 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
+        if (actor.heritage && ancestryFeatures) {
+            const alreadyPresent = ancestryFeatures.feats.some((f) => f.feat?.id === actor.heritage?.id);
+            if (!alreadyPresent) {
+                ancestryFeatures.feats.unshift({
+                    id: "heritage",
+                    label: null,
+                    level: null,
+                    feat: actor.heritage,
+                    children: this.#getNestedSlots(actor.heritage),
+                } as never);
```
CarlosFdez: avoid casting to never. There's gotta be another way.

src/module/actor/character/feats/collection.ts
```diff
@@ -247,6 +247,9 @@ class CharacterFeats<TActor extends CharacterPF2e> extends Collection<string, Fe
                 if (granter?.isOfType("feat") && granter.grants.includes(feat) && isNested) {
```
CarlosFdez, line 247: What is the practicality of including the heritage in granter.grants? It looks like grants in FeatPF2e is FeatPF2e | HeritagePF2e, so doing `granter?.isOfType("feat", "heritage") && granter.grants.includes(feat) && isNested` should be more terse

iDantar: I tried that, but Heritage doesn't have `.grants`, so the check fails and granted items show in bonus feats as duplicates, unfortunately

src/module/actor/character/sheet.ts
iDantar: Thanks, fixed and will keep in mind for the future

iDantar (commented):
Thanks again for the review, finally got back to PC and addressed it - would you be able to take another look please?

CarlosFdez:
Something feels off about hijacking it at the sheet level rather than handling it in the feat groups data structure. Perhaps this is the right solution but it is impossible for me to know without digging deep, and I don't have the gumption for that right at this time.

iDantar:
I share the feeling, thanks for the review. I'll try to see if it can be moved to feat groups

iDantar:
Moved the logic to feat groups + handled the heritages given by feats. Please take another look when you have the capacity.
Here's how it looks now:
image

PR #22272, opened 2026-05-08
Fix incorrect `_loc` function calls

Cuingamehtar:
These are the ones certainly incorrect.
There are a couple questionable ones I found, 
https://github.com/foundryvtt/pf2e/blob/b77619817d147900eaa4fb861fa5eb03ebf82983/src/module/rules/rule-element/choice-set/rule-element.ts#L578
and
https://github.com/foundryvtt/pf2e/blob/b77619817d147900eaa4fb861fa5eb03ebf82983/src/module/rules/rule-element/base.ts#L93
But I didn't do any testing to see if they can fail.


CarlosFdez:
After basic testing the ones you fixed looked fine. For the rule elements, that should maybe be its own thing and I'm happy to wait and see for actual examples.

PR #22274, opened 2026-05-08
Add "robot" trait to creature-identification skills map

### LINKED ISSUE #22273: [SF2E] The Robot Trait on monsters does not give Computers and Crafting as Recall Knowledge Options
SF2E Version 1.1.1
Foundry Version 14.360-14

The Robot Trait on monster actors does not give Computers and Crafting as Recall Knowledge Options in the actor sheet. GM Core page 54-55 lists the Robot trait as having both Crafting and Computers as options for Recall Knowledge, but the option doesn't appear. 

Reproduction Steps:

Most easily seen with the Boneflayer from Guilt of the Graveworld. Has the Robot trait, but no option for Recall Knowledge beside specific Lore skills. 

Image

Also seen with other monsters with the Robot Trait, for Example the Ordinance-Class Civil Robot from Alien Core. By default, it has the Construct trait, which gives Arcana and Crafting, but Computers is not listed despite having Robot

Image

Remove the Construct Trait and the Arcana and Crafting options disappear, and Computers and Crafting are missing. 

Image

stwlam:
- Closes #22273

### DIFF
--- a/src/module/actor/npc/sheet.ts
+++ b/src/module/actor/npc/sheet.ts
@@ -195,7 +195,9 @@ class NPCSheetPF2e extends AbstractNPCSheet {
         sheetData.identificationDCs = ((): NPCIdentificationSheetData => {
             const data = actor.identificationDCs;
             const skills =
-                data.skills.length > 0 ? localizeList(data.skills.map((s) => _loc(CONFIG.PF2E.skills[s].label))) : null;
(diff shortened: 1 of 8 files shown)


CarlosFdez:
Tests are explody it seems

CarlosFdez:
🤖

PR #22281, opened 2026-05-08
Include spell attack and dc domains for special statistic spellcasting

### LINKED ISSUE #22280: Cannot make a spell attack roll with a ` ItemSpellcasting` entry that uses the `kinetic-activation` statistic
So, if i create a `Wand of Snowball` on a kineticist with the `Kinetic Activation` feat, the attack roll ends up being a default check instead.

Image

---

If i create a regular spellcasting entry using the `kineticist` statistic, it does work fine.

Image

CarlosFdez:
- Closes https://github.com/foundryvtt/pf2e/issues/22280

We can't do it in the item spellcasting class itself because scrolls/wands make new item spellcasting classes to wrap the statistic used instead of reusing the existing entries, and better not slow down prep further with more statistic clones.

### DIFF
--- a/src/module/rules/rule-element/special-statistic.ts
+++ b/src/module/rules/rule-element/special-statistic.ts
@@ -37,7 +37,7 @@ class SpecialStatisticRuleElement extends RuleElement<SpecialStatisticSchema> {
             type: new fields.StringField({
                 required: true,
                 choices: ["simple", "check", "attack-roll"],
-                initial: "check",
+                initial: (d) => (d.itemCasting ? "attack-roll" : "check"),
             }),
             extends: new fields.StringField({ required: true, nullable: true, initial: null }),
             attribute: new fields.StringField({
@@ -93,7 +93,7 @@ class SpecialStatisticRuleElement extends RuleElement<SpecialStatisticSchema> {
                   })
                 : { mod: [], check: [], dc: [] };
 
-        const data: Required<Omit<StatisticData, "filter" | "lore" | "proficient" | "rank" | "rollOptions">> = {
+        const data = {
             slug: this.slug,
             label,
             attribute: this.attribute ?? extendedFrom?.attribute ?? null,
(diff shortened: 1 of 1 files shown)


CarlosFdez:
I'm going to move the addition of the domains to earlier in the function and avoid the double extend once I'm less busy. EDIT: had a moment and enough bad sense to update it, should be good now.

src/module/rules/rule-element/special-statistic.ts
```diff
@@ -93,15 +93,25 @@ class SpecialStatisticRuleElement extends RuleElement<SpecialStatisticSchema> {
...
+            data.dc.domains.push("spell-dc");
+            const tradition = this.itemCasting.tradition;
+            if (tradition) {
+                data.check.domains.push(`${tradition}-spell-attack`);
+                data.dc.domains.push(`${tradition}-spell-dc`);
+            }
+        }
 
         const statistic = extendedFrom?.extend(data) ?? new Statistic(actor, data);
         if (statistic) {
```
stwlam: Not sure how this conditional block survived after the nullish coalescence was added

src/module/rules/rule-element/special-statistic.ts
```diff
@@ -93,15 +93,25 @@ class SpecialStatisticRuleElement extends RuleElement<SpecialStatisticSchema> {
...
+        const data = {
             slug: this.slug,
             label,
             attribute: this.attribute ?? extendedFrom?.attribute ?? null,
             domains: [this.slug],
             modifiers: modCheckDC.mod,
             check: { type: checkType, domains: checkDomains, modifiers: modCheckDC.check },
             dc: { domains: [`${this.slug}-dc`], modifiers: modCheckDC.dc },
-        };
+        } satisfies Required<Omit<StatisticData, "filter" | "lore" | "proficient" | "rank" | "rollOptions">>;
```
stwlam: ??

src/module/rules/rule-element/special-statistic.ts
CarlosFdez: I'm a little embarrassed I didn't notice the wart. Removed.

PR #22285, opened 2026-05-09
Fix invalid language keys

Cuingamehtar:


CarlosFdez:
Was this because of an issue with the migration? Or unrelated?

Cuingamehtar:
Change shape ones are related to the migration, but I don't know if they just weren't covered by it, or otherwise slipped by.

PF->SF is caused by miscommunication between me, Tikael and somebody who went and changed the keys after (https://github.com/foundryvtt/pf2e/pull/21906#discussion_r3028495997) - I think there was some misunderstanding here, which has to be talked through.

CarlosFdez:
I'd like it at least if the change shape ones were included in the migration, even if its a bit too late for those who already updated. Remigrates are a thing.

Maybe I'm misreading, but it looks like tikael is saying they should have been using PF2e? If I recall, the reason why we try to use PF2e system wide is that Its important to consider the case of what happens if something like Bantrid is added to PF2e. Would we use the SF2e prefix in pf2e? Or do we migrate to PF2e prefixes? We avoid migrations if we treat it like the prefix should be `SYSTEM` for everything but we got grandfathered into `PF2E`.

Tikael overrides me here of course, I don't know his intent in the message to you, but that was my motivation.

Cuingamehtar:
I've brought the namespace question to the discussion. It's something that we'd better clarify or else one person will be overriding another one's changes back-and-forth (and they will slip by on accident). It's also something that's better to have pinned so it can be referenced.

Are you suggesting updating the existing migration to include the missing keys, or making a new one? Either way, I'm not going to take that up right now (maybe in a day or two when I get free time for it).

CarlosFdez:
I said it mostly as a reminder but also by this point I've also kinda stopped caring too much, While we shouldn't be leaning into "sf2e is an expansion of pf2e", I've kinda realized everyone is gonna mix them no matter what we do even when trying otherwise, and people more familiar with PF2e will try to use SF2E for SF2e only things while they continue to use PF2E for "PF2e or SF2e" stuff. We will likely run into a problem in the future that we'll need to fix, but I don't think this can be avoided unless we somehow find a way to migrate the prefix to "SYSTEM" (and break everything!) so that people stop trying to separate. So I'm happy to leave it to Tikael and see how lucky he's feeling on any particular example. Less stressful that way.

Ideally I'd like the one who wrote that migration (@SpartanCPA) to do it since they have more context, I'm happy with same migration at least. Wouldn't be instead of the PR, I'd like both if possible.

PR #22294, opened 2026-05-09
Update Invoke Rune effect to match Monster Core update

ammalagonc:
Critter no longer has Rock as an attack

### DIFF
--- a/packs/pf2e/bestiary-effects/effect-invoke-rune.json
+++ b/packs/pf2e/bestiary-effects/effect-invoke-rune.json
@@ -25,18 +25,45 @@
                 "choices": [
                     {
                         "label": "PF2E.NPCAbility.RuneGiant.InvokeRune.RuneOfDestruction.Label",
+                        "predicate": [
+                            {
+                                "not": "czaresia-invoke-rune"
+                            }
+                        ],
                         "value": "destruction"
                     },
                     {
                         "label": "PF2E.NPCAbility.RuneGiant.InvokeRune.RuneOfFlames.Label",
+                        "predicate": [
+                            {
+                                "not": "czaresia-invoke-rune"
+                            }
+                        ],
                         "value": "flames"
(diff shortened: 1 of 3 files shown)


simonward:
It looks like some of the variants have different attacks. Would something like this work instead?
```
{
  "choices": {
    "attacks": true,
    "predicate": [
      {
        "not": "item:category:unarmed"
      }
    ]
  },
  "flag": "weapon",
  "key": "ChoiceSet",
  "prompt": "PF2E.SpecificRule.Prompt.Weapon"
}
```

ammalagonc:
> It looks like some of the variants have different attacks. Would something like this work instead?
> 
> ```
> {
>   "choices": {
>     "attacks": true,
>     "predicate": [
>       {
>         "not": "item:category:unarmed"
>       }
>     ]
>   },
>   "flag": "weapon",
>   "key": "ChoiceSet",
>   "prompt": "PF2E.SpecificRule.Prompt.Weapon"
> }
> ```

This actually works!

Thank you!

packs/pf2e/bestiary-effects/effect-invoke-rune.json
simonward, line 68: Should add the Rune of Space.

packs/pf2e/bestiary-effects/effect-invoke-rune.json
```diff
@@ -79,8 +73,7 @@
                     "item:slug:{item|flags.system.rulesSelections.weapon}",
```
simonward, line 104: ~~Can probably boost the attack dice rather than having multiple specific deadly alterations?~~

Can we add deadly traits predicated on base die size?

ammalagonc: I don't see dice size as a RollOption for NPC...
Probably missing something.

PR #22298, opened 2026-05-10
Add weapon boost automation

Geliogabalus:
Add `boost` damage dice to the weapon damage rolls when `boost-*` trait is present. By default the extra dice are disabled. 

If `Effect: Boost` is present, the damage dice become enabled. When the effect is added via auxiliary actions buttom the effect is customized for the particular weapon.

The effect lasts till the end of the next turn and doesn't clear itself after the strike.

image

image


CarlosFdez:
Heyo, thanks for starting work on this. Talks of this have circulated and we kinda have an idea of what it should look like, so if you're patient, could you make these adjustments?

1) The equipment effect should be in a pf2e compendium and then added to duplicates.json. Otherwise it won't work for boost traits manually added by players in a pf2e world.

2) Use the traits config rather than parsing the trait. `weapon.system.traits.config.boost` should have the value if I correctly guessed its path. The value is a string unfortunately, so you'll have to parse that value. You can use something like `parseTermsFromSimpleFormula`, ensure there is only a single term, then add that.

3) It would be preferable for the damage die/modifier to be on the weapon always but disabled. That way if someone rolls damage, they can check on boost at that moment. The effect should instead enable it and otherwise do nothing (the modifier/die can include `self:effect:boost` in its predicate I think. Double check what I have the right roll option of course). Check out the `WeaponDamagePF2e#calculate()` function for where that occurs currently.

Geliogabalus:
I've started to do suggested updates and there is an incosistency with the rules in a way how boost traits are defined. Currently, there are following traits defined in the system: `boost-1`, `boost-1d4`, `boost-1d6`, `boost-1d8`, `boost-1d10`, `boost-1d12`, `boost-2d10`. Looking at the wording of the rules, it states that boost trait should has a die size (similar to fatal and deadly). The damage should scale with the quality of the weapon so this traits naming is not reflecting the rules. It seems there is an inconsistency in the rules where some weapons have `1d10` and some `d10` as a postfix, but no weapon specifies several dice in boost entry or fixed damage (`boost-1`). [Weapons with boost.](https://2e.aonsrd.com/traits/34-boost?trait=34&sort=name-asc&display=table&columns=trait)

I think, it makes sense to refactor the boost trait definition to make it consistent with `deadly` and `fatal` but it will need some migration I believe which I don't know how to do. 

In any case, I will process only `boost-1{dieSize}` in the current pr because two other traits don't make sense from the rules standpoint.

Geliogabalus:
> Heyo, thanks for starting work on this. Talks of this have circulated and we kinda have an idea of what it should look like, so if you're patient, could you make these adjustments?
> 
> 1. The equipment effect should be in a pf2e compendium and then added to duplicates.json. Otherwise it won't work for boost traits manually added by players in a pf2e world.
> 2. Use the traits config rather than parsing the trait. `weapon.system.traits.config.boost` should have the value if I correctly guessed its path. The value is a string unfortunately, so you'll have to parse that value. You can use something like `parseTermsFromSimpleFormula`, ensure there is only a single term, then add that.
> 3. It would be preferable for the damage die/modifier to be on the weapon always but disabled. That way if someone rolls damage, they can check on boost at that moment. The effect should instead enable it and otherwise do nothing (the modifier/die can include `self:effect:boost` in its predicate I think. Double check what I have the right roll option of course). Check out the `WeaponDamagePF2e#calculate()` function for where that occurs currently.

@CarlosFdez 
Updated the PR according to your notes.

- moved effect to the pf2e compendium with `duplicated.json` update
- updated damage calculation to include disabled boost entry by default. It is enabled when effect is present (effect has RollOption rule element, I wanted to restrict it for the particular weapon so I didn't rely on just `self:effect:boost` present). Still parsing damage die from the trait name as I am restricting extra dice to the valid traits only (see previous comment).

CarlosFdez:
@TikaelSol what's the reason we have `boost 1`? Is there a weapon with it?

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,24 @@ class WeaponDamagePF2e {
             modifiers.push(modifier);
         }
 
+        // Boost trait, ignore invalid boost traits
+        for (const trait of weaponTraits.filter((t) => t.startsWith("boost-1d"))) {
+            if (!weapon.isOfType("weapon")) continue;
+            const dieSize = trait.substring(trait.indexOf("-") + 2) as DamageDieSize;
+            const isBoosted = options.has(`item:${weapon.id}:boosted`) || options.has("item:weapon:boosted");
```
CarlosFdez: you can move this to the predicate field of the lower damage die object. `[{ or: [] }]` if you need both.

src/module/actor/character/helpers.ts
```diff
@@ -247,6 +247,23 @@ function getWeaponAuxiliaryActions(weapon: WeaponPF2e<CharacterPF2e>): WeaponAux
                     }
                 }
 
+                if (
+                    traitsArray.includes("boost-1d4") ||
+                    traitsArray.includes("boost-1d6") ||
+                    traitsArray.includes("boost-1d8") ||
+                    traitsArray.includes("boost-1d10") ||
+                    traitsArray.includes("boost-1d12")
+                ) {
```
CarlosFdez:
```suggestion
                if (weapon.system.traits.config.boost) {
```

src/module/actor/character/helpers.ts
```diff
@@ -247,6 +247,23 @@ function getWeaponAuxiliaryActions(weapon: WeaponPF2e<CharacterPF2e>): WeaponAux
...
+                    const isBoosted = actor.itemTypes.effect.some(
+                        (e) => e.slug === "effect-boost" && e.flags.sf2e.grantedBy?.id === weapon.id,
```
CarlosFdez:
```suggestion
                        (e) => e.slug === "effect-boost" && e.flags[SYSTEM_ID].grantedBy?.id === weapon.id,
```

src/module/actor/character/helpers.ts
```diff
@@ -247,6 +247,23 @@ function getWeaponAuxiliaryActions(weapon: WeaponPF2e<CharacterPF2e>): WeaponAux
...
+                    );
+                    if (!isBoosted) {
```
CarlosFdez: The aux action should always exist. You could make it no opt if already boost (perhaps disabled?) or be a toggle like modular. It shouldn't disappear once you boost the weapon.

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,24 @@ class WeaponDamagePF2e {
             modifiers.push(modifier);
         }
 
+        // Boost trait, ignore invalid boost traits
+        for (const trait of weaponTraits.filter((t) => t.startsWith("boost-1d"))) {
```
CarlosFdez: Use and parse `weapon.system.traits.config.boost` instead.  The current implementation of deadly/fatal just hasn't been refactored yet because we didn't have time.

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,24 @@ class WeaponDamagePF2e {
...
+            if (!weapon.isOfType("weapon")) continue;
```
CarlosFdez: This should work on npcs as well. Bo-Shek for example has 2d10 boost. You can more or less copy what deadly does for the diceNumber. While npcs don't support aux actions yet, the existance of the effect as well as the toggle in the damage roll should be handy.

src/module/system/damage/weapon.ts
CarlosFdez: Doing that will also allow a disabled boost row to exist in the damage roll as well in the same way deadly does on non-crits, allowing it to work with npcs as well if you remove the weapon check.

CarlosFdez:
Fair point on boost being per weapon. There might be a better way to handle boost being per weapon if we want to restrict that using a choice set, but I'm not certain about it so I'll come back to you with it later!

> I've started to do suggested updates and there is an incosistency with the rules in a way how boost traits are defined. Currently, there are following traits defined in the system: `boost-1`, `boost-1d4`, `boost-1d6`, `boost-1d8`, `boost-1d10`, `boost-1d12`, `boost-2d10`. Looking at the wording of the rules, it states that boost trait should has a die size (similar to fatal and deadly). The damage should scale with the quality of the weapon so this traits naming is not reflecting the rules. It seems there is an inconsistency in the rules where some weapons have `1d10` and some `d10` as a postfix, but no weapon specifies several dice in boost entry or fixed damage (`boost-1`). [Weapons with boost.](https://2e.aonsrd.com/traits/34-boost?trait=34&sort=name-asc&display=table&columns=trait)

Boost 1 looks like a holdover from playtest and we can probably remove it (depends on what tikael says). Boost-2d10 is used by an NPC (Bo-Shek). I'm thinking its like deadly where the number of dice are on the trait itself.

EDIT: We may have to move boost-2d10 to npc attack traits.

src/module/system/damage/weapon.ts
Geliogabalus: I don't understand exactly. It was showing boost toggle even when there were no effect present and now after removing weapon check it works for npc as well

src/module/actor/character/helpers.ts
Geliogabalus: Made no op on click when already boosted

src/module/system/damage/weapon.ts
Geliogabalus: Made it work for npc and tested it. It is working as intended when the effect is applied from the compendium and user otherwise can toggle it manually. We can make checkbox in the sheet since it is controlled by roll option right now.

src/module/system/damage/weapon.ts
CarlosFdez: No you're right, I misread. Preferable use predicate because its cleaner but not to fix any bug or anything.

CarlosFdez:
boost-1 can be removed, feel free to KO it from the traits file.

TikaelSol:
> boost-1 can be removed, feel free to KO it from the traits file.

For clarity, it just snuck in from the playtest. Nothing uses it to my knowledge.

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,29 @@ class WeaponDamagePF2e {
...
+        // Boost trait
+        if (weapon.system.traits.config.boost) {
+            const boostConfig = weapon.system.traits.config.boost;
+            const slug = `boost-${boostConfig}`;
+
+            const dieSize = boostConfig.substring(boostConfig.indexOf("d")) as DamageDieSize;
+            const baseNumber = Number(/(\d)d\d{1,2}$/.exec(boostConfig)?.at(1)) || 1;
+            const diceNumber = strikingDice > 0 ? baseNumber + strikingDice : baseNumber;
+
+            const isBoosted = options.has(`item:${weapon.id}:boosted`) || options.has("item:weapon:boosted");
```
TikaelSol: We can construct a choice set that works for both PCs and NPCs and provides the weapon flag in question. So we can use the effect to choose the weapon and just rely on a single roll option that will contain the ID, making the code itself simpler. This would auto set `item:boosted:{item|id}`.
```json
{
  "choices": {
    "ownedItems": true,
    "predicate": [
      {
        "or": [
          {
            "and": [
              "self:type:npc",
              "item:type:melee"
            ]
          },
          {
            "and": [
              "self:type:character",
              "item:type:weapon"
            ]
          }
        ]
      }
    ],
    "types": [
      "melee",
      "weapon"
    ]
  },
  "flag": "weapon",
  "rollOption": "item:boosted",
  "key": "ChoiceSet",
  "prompt": "PF2E.SpecificRule.Prompt.Weapon",
  "predicate": [
    {
      "or": [
        "self:type:npc",
        "self:type:character"
      ]
    }
  ]
}
```

CarlosFdez: might need an `item:trait:boost` in the predicate but yeah that might work better. Then the code would basically pre-set the value.

Geliogabalus: Refactored effect to use ChoiseSet rule element. Also removed `boost-1` trait and moved `boost-2d10` to npc traits.

src/module/actor/character/auxiliary.ts
```diff
@@ -157,6 +160,21 @@ class WeaponAuxiliaryAction {
                 selected: Number(selection),
             });
             if (!updated) return;
+        } else if (this.annotation === "boost") {
+            const isBoosted = actor.itemTypes.effect.some(
+                (e) => e.slug === "effect-boost" && e.flags.sf2e.grantedBy?.id === weapon.id,
```
CarlosFdez:
```suggestion
                (e) => e.slug === "effect-boost" && e.flags[SYSTEM_ID].grantedBy?.id === weapon.id,
```

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,29 @@ class WeaponDamagePF2e {
...
+            damageDice.push(
+                new DamageDicePF2e({
+                    selector: `${weapon.id}-damage`,
+                    slug,
+                    label: traitLabels[slug],
+                    diceNumber: diceNumber,
+                    dieSize,
+                    critical: false,
+                    enabled: isBoosted,
+                }),
```
CarlosFdez:
```suggestion
            damageDice.push(
                new DamageDicePF2e({
                    selector: `${weapon.id}-damage`,
                    slug,
                    label: traitLabels[slug],
                    diceNumber: diceNumber,
                    dieSize,
                    critical: false,
                    predicate: [`item:boosted:${weapon.id}`],
                }),
```

CarlosFdez:
Beyond the suggestions, effect boost might need a pack and extract to fix its indentation and strip out certain stuff, but it looks fine enough to me after that. Thanks for working on this.

src/module/system/damage/weapon.ts
Geliogabalus: Does it make die enabled when the predicate is true? I did try this setup and it is still disabled when rolling.

src/module/system/damage/weapon.ts
```diff
@@ -359,6 +359,28 @@ class WeaponDamagePF2e {
...
+
+            damageDice.push(
+                new DamageDicePF2e({
+                    selector: `${weapon.id}-damage`,
+                    slug,
+                    label: traitLabels[slug],
+                    diceNumber: diceNumber,
+                    dieSize,
+                    critical: false,
+                    enabled: options.has(`item:boosted:${weapon.id}`),
```
CarlosFdez, line 379: Is there a reason you can't use the predicate parameter here? Predicates (for damage dice) are resolved lazily, and would allow any sort of option inclusion from any source or phase to be resolved here, including from macros where someone might want to roll a "damage but boosted". Using enabled instead hard restricts it to the state it has during data preparation.

If there's an issue I understand, I just saw you're avoiding it for some reason.

CarlosFdez: Disregard this reply, I saw you responded elsewhere

src/module/system/damage/weapon.ts
CarlosFdez: It should, if it isn't something is up. I'm looking at the code and I see that `dice.test(options)` is used in the scenario where predicate is passed. That predicate is not being evaluated is a bit suspicious. I guess leave as is and I'll poke later.

CarlosFdez: Finished poking. The behavior I described is for statistics, damage modifiers and dice are reconstructed when the roll happens or the label is retrieved, not during data preparation. So what you did was right.

PR #22299, opened 2026-05-10
Remove inversion filter from light mode status effect hud icons

In3luki:
The filter is already disabled in dark mode by core CSS.
<br><br>
status-effects


stwlam:
It's not disabled in dark mode; it's _set_ in light mode. It's a core issue on account of the stock icons being white SVGs with no drop shadow.

stwlam:
~~Sent up a fix to core~~ core fix not happening, at least in V14

In3luki:
This will be fixed in core after all

PR #22302, opened 2026-05-10
Set cone angle to 90 when created from scene controls

stwlam:


CarlosFdez:
I find it kinda amusing that the icon has a 90 degree angle but the default cone isn't 90 degrees.

PR #22307, opened 2026-05-11
Check predicate for strike RE with replace-all

### LINKED ISSUE #17641: Have `Strike` RE `replaceAll` option be predicated
Right now, if i create a `Strike` RE with both a `predicate`, the strike only appears when the predicate is true but the `replaceAll` is always active, removing all the other strikes even when the one associated with the RE isn't present.

iDantar:
Closes https://github.com/foundryvtt/pf2e/issues/17641

### DIFF
--- a/src/module/rules/rule-element/strike.ts
+++ b/src/module/rules/rule-element/strike.ts
@@ -181,8 +181,7 @@ class StrikeRuleElement extends RuleElement<StrikeSchema> {
 
     /** Exclude other strikes if this rule element specifies that its strike replaces all others */
     override afterPrepareData(): void {
-        if (this.ignored || !this.actor.isOfType("character")) return;
-
+        if (this.ignored || !this.actor.isOfType("character") || !this.test()) return;
         if (this.replaceAll) {
             const systemData = this.actor.system;
             systemData.actions = systemData.actions


src/module/rules/rule-element/strike.ts
```diff
@@ -181,8 +181,7 @@ class StrikeRuleElement extends RuleElement<StrikeSchema> {
 
     /** Exclude other strikes if this rule element specifies that its strike replaces all others */
     override afterPrepareData(): void {
-        if (this.ignored || !this.actor.isOfType("character")) return;
-
+        if (this.ignored || !this.actor.isOfType("character") || !this.test()) return;
```
stwlam, line 184: A failed `test()` in `beforePrepareData` sets `ignored` to true as a side effect.

iDantar: @stwlam I just tested it on dev-14 and either that behavior has changed or I'm missing something - can you help me figure it out, please?
image
Rules used:
`{"category":"unarmed","damage":{"base":{"damageType":"piercing","dice":1,"die":"d4"}},"group":"brawling","img":"systems/pf2e/icons/features/feats/cobra-stance.webp","key":"Strike","label":"PF2E.SpecificRule.Stance.Attack.CobraFang","range":null,"replaceAll":true,"traits":["agile","deadly-d10","finesse","unarmed","venomous"],"predicate":["test-predicate"]}`

`{"key":"RollOption","option":"test-predicate","toggleable":true}`

On this branch, when toggle is off, other attacks are present:
image

PR #22319, opened 2026-05-13
Allow the didactic strike application to attack spells

### LINKED ISSUE #22308: Effect: Didactic Strike is not working with attack spells
The Didactic Strike says "each ally's first attack deals an extra 2d6 precision damage to the target if it hits". Currently, when you drag and drop the effect on an ally, the precision damage gets added only on strikes, not on attacks.

iDantar:
closes https://github.com/foundryvtt/pf2e/issues/22308

### DIFF
--- a/packs/pf2e/feat-effects/effect-didactic-strike.json
+++ b/packs/pf2e/feat-effects/effect-didactic-strike.json
@@ -26,7 +26,7 @@
                 "diceNumber": 2,
                 "dieSize": "d6",
                 "key": "DamageDice",
-                "selector": "strike-damage"
+                "selector": ["attack-damage"]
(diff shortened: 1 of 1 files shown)


packs/pf2e/feat-effects/effect-didactic-strike.json
```diff
@@ -26,7 +26,7 @@
                 "diceNumber": 2,
                 "dieSize": "d6",
                 "key": "DamageDice",
-                "selector": "strike-damage"
+                "selector": ["attack-damage"]
```
simonward:
```suggestion
                "selector": "attack-damage"
```

PR #22320, opened 2026-05-13
Fix greater Endless Quiver material choices

### LINKED ISSUE #22253: Effect: Convert Arrows does not allow all valid choices for Endless Quiver (Greater)
Endless Quiver (Greater) should have access to all 4 material type choices. The effect currently offers the 2 that Endless Quiver allows only for that item and not the Greater version. `"parent:origin:item:slug:endless-quiver-greater"` needs to be added to the predicate for cold iron and silver.

iDantar:
closes https://github.com/foundryvtt/pf2e/issues/22253

### DIFF
--- a/packs/pf2e/equipment-effects/effect-convert-arrows.json
+++ b/packs/pf2e/equipment-effects/effect-convert-arrows.json
@@ -43,12 +43,7 @@
                         "label": "PF2E.PreciousMaterialAdamantine",
                         "predicate": [
                             {
-                                "or": [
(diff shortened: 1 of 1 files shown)


simonward:
Would it be cleaner to just use the predicate "not": "parent:origin:item:slug:endless-quiver" on the higher quality materials, and remove the predicate altogether on the lesser materials?

iDantar:
Agree, this is much easier to digest. Applied, thanks!

PR #22327, opened 2026-05-14
Convert NPC Skill Editor to AppV2 and Svelte

jfn4th:
This also fixes processing of SCSS in svelte files. Up until now, they've been treated as standard CSS even when marked as SCSS. `item-summary.svelte` is the only other Svelte app I've found using a SCSS mixin, and its appearance doesn't appear to have changed (its tags are busted in dark mode, but that was the case before this PR and I'll address that separately if no one else gets around to it first).
Screenshot 2026-05-14 at 4 16 45 PM
Screenshot 2026-05-14 at 4 16 20 PM


CarlosFdez:
tags busted in dark mode for item-summary should be fixed recently now as of https://github.com/foundryvtt/pf2e/pull/22311

I recall adding that vite mixin adding errors where it doesn't allow defining `@use` at the top level. Are you sure you're not getting those errors? You can get the full svelte errors if you use `npm run build:system`

jfn4th:
@CarlosFdez I get no such error!

jfn4th:
Svelte errors if using `@use` anywhere other than the top level. Maybe that came up for you while you were making the item-summary?

PR #22332, opened 2026-05-15
Add {item|id}-{meleeOrRanged}-damage domain to attacks, to enable ammunition rules

### LINKED ISSUE #22330: Elemental Ammunition Damage not working
Elemental Ammunition no matter what type does no damage in V14. I believe this is because of the item selector which should be item-id-damage instead of item-id-ranged-damage

Image

iDantar:
- Closes #22330 

Kudos to @pedrogrullada for suggested solution
UPD: also updates Magnetic Shots's ItemAlteration rules to add Deadly trait directly to the weapon, not to ammunition.

### DIFF
--- a/packs/pf2e/equipment/antler-arrow.json
+++ b/packs/pf2e/equipment/antler-arrow.json
@@ -43,12 +43,7 @@
                 "outcome": [
                     "success"
                 ],
-                "predicate": [
-                    {
-                        "not": "item:ranged-usage:trait:combination"
-                    }
-                ],
-                "selector": "{item|_id}-damage",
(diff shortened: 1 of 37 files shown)


CarlosFdez:
I don't like the idea of globally mandating that all ammo rule elements handle edge cases of generic ammo usage.

It should either shut off ammo rules for combination weapon melee usage at the weapon level, or fix what likely actually broke and restore ranged-damage selectors as they likely broke when I implemented area fire.

iDantar:
Added the domain to handle the existing rules, thanks

PR #22333, opened 2026-05-15
Cleanup Persistent Damage Editor lifecycle

jfn4th:


src/module/item/condition/persistent-damage-editor.ts
```diff
@@ -57,13 +80,13 @@ class PersistentDamageEditor extends fa.api.HandlebarsApplicationMixin(fa.api.Ap
...
-            selectedItemId: this.selectedItemId,
+            selectedItemId: this.#selectedItemId,
             existing,
-            damageTypes: this.#prepareDamageTypes(),
+            damageTypes: PersistentDamageEditor.#prepareDamageTypes(),
         };
     }
 
-    #prepareDamageTypes(): DamageTypeData[] {
+    static #prepareDamageTypes(): DamageTypeData[] {
```
CarlosFdez: Why static? This isn't an event handler, this is basically a helper for prepareContext.

EDIT: If anything we'd remove the helper and merge it into the main method as its quite small.

jfn4th: Point taken. Largely it was just because it doesn't rely on `this`.

CarlosFdez:
Thanks for the PR

PR #22335, opened 2026-05-15
Add missing `replaceAll: true` for Talon and Claw stance effects

iDantar:
Just noticed that of all 'The only Strikes you can make are X' feats (excluding weapon ones) these are missing the replaceAll

### DIFF
--- a/packs/pf2e/feat-effects/stance-claw-stance.json
+++ b/packs/pf2e/feat-effects/stance-claw-stance.json
@@ -22,6 +22,7 @@
         },
         "rules": [
             {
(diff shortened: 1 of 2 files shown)


packs/pf2e/feat-effects/stance-claw-stance.json
simonward, line 26: Should this have `"baseType": "claw"`?

iDantar: Makes sense to me - added :)

PR #22340, opened 2026-05-16
Fix initialization of army actors in v14

### LINKED ISSUE #22338: Army schema throws error when calling `getInitialValue()` in v14
If a module tries to access the Actor data models, it fails on only the Army entry. Something about the Army schema is unable to get a clean initial data. The error seems to trigger on the AC initial value.

Can reproduce by running this command in the console:
```js
CONFIG["Actor"].dataModels.army.schema.getInitialValue();
```

Image

Practically, I found this trying to debug an error when opening the ItemPiles module's Settings menu, which tries to access the system's Actor types to build a list of all available Actor types to configure Item Piles for. I edited the module code to skip checking the Army actor type and the rest all work fine, so I've confirmed it's just Army. See https://github.com/fantasycalendar/FoundryVTT-ItemPiles/issues/837

CarlosFdez:
Closes https://github.com/foundryvtt/pf2e/issues/22338

### DIFF
--- a/src/module/actor/army/data.ts
+++ b/src/module/actor/army/data.ts
@@ -33,8 +33,8 @@ class ArmySystemData extends ActorSystemModel<ArmyPF2e, ArmySystemSchema> {
                     required: true,
                     nullable: false,
                     integer: true,
-                    initial: (d: DeepPartial<ArmySystemSource>) =>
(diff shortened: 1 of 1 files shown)


CarlosFdez:
This is actually due to a core bug. If rejected somehow I'll reopen.

PR #22343, opened 2026-05-16
Fix missing Tool Rat statistics

Geliogabalus:
Fix missing data for [Tool Rat](https://2e.aonsrd.com/creatures/111-tool-rat) creature entry.


TikaelSol:
Thanks for the PR. As a note PR titles should be in imperative case, finishing the sentence "If merged this PR will...". That way we can use the PR titles directly in the changelog.

Issue #22339, opened 2026-05-16
Include spell gems, chips, scrolls, and wands in Anachronism


CarlosFdez:
SF2e Anachronism does not allow creating spell chips nor spell gems, and PF2e anachronism does not allow creating scrolls nor wands. We can import the content system's spell items during the build and setup the startup script to add them to the config.

PR #22344, opened 2026-05-16
Include scrolls, wands, spell gems in Anachronism

jfn4th:
Implements #22339 

This doesn't touch spell chips, as the equipment items are already included in Anachronism and work the same as they do in the SF2e system, i.e., they do nothing.  I saw murmurings in the discord that full spell chip implementation may be waiting on changes to how spells get embedded in items rather than just using the same flow as wands. What's the current plan there? I'd be interested in taking a look at it.

### DIFF
--- a/build/anachronism/index.ts
+++ b/build/anachronism/index.ts
@@ -4,6 +4,7 @@ import { LANGUAGES_BY_RARITY } from "@actor/creature/values.ts";
 import { ItemSourcePF2e } from "@item/base/data/index.ts";
 import { itemIsOfType } from "@item/helpers.ts";
 import { ancestryTraits, classTraits } from "@scripts/config/traits.ts";
+import { SPELL_CONSUMABLES_BY_SYSTEM } from "@scripts/config/spell-consumables.ts";
 import { objectHasKey, recursiveReplaceString, sluggify } from "@util/misc.ts";
 import fs from "fs";
 import path from "path";
@@ -48,9 +49,13 @@ console.log(`Building Modules: ${contentSystems.map((c) => `${c}-anachronism`)}`
 /** Root module file contents of anachronism. This should be moved to a file somewhere. */
 const moduleSourceContents = String.raw`
 import UUID_REDIRECTS from "./uuid-redirects.json" with { type: "json" };
+import SPELL_CONSUMABLES from "./spell-consumables.json" with { type: "json" };
 for (const [from, to] of Object.entries(UUID_REDIRECTS)) {
     CONFIG.compendium.uuidRedirects[from] = to;
 }
+Hooks.once("init", () => {
+    Object.assign(CONFIG.PF2E.spellcastingItems, SPELL_CONSUMABLES);
+});
 `;
 
 console.log("Starting build, loading compendiums for both systems");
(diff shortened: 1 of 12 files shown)


CarlosFdez:
Try to imitate what we have for languages if possible.

build/anachronism/index.ts
```diff
@@ -48,9 +49,13 @@ console.log(`Building Modules: ${contentSystems.map((c) => `${c}-anachronism`)}`
 /** Root module file contents of anachronism. This should be moved to a file somewhere. */
 const moduleSourceContents = String.raw`
 import UUID_REDIRECTS from "./uuid-redirects.json" with { type: "json" };
+import SPELL_CONSUMABLES from "./spell-consumables.json" with { type: "json" };
 for (const [from, to] of Object.entries(UUID_REDIRECTS)) {
     CONFIG.compendium.uuidRedirects[from] = to;
 }
+Hooks.once("setup", () => {
```
CarlosFdez: looking at the code I think it can be done as early as init.

CarlosFdez:
For context, look at actor/creature/values.ts, at `LANGUAGES_BY_RARITY`. It is an object key'd by pf2e/sf2e. This would make tree shaking impossible but its a small object so no big deal. But it lets you key by it via SYSTEM_ID in the system and by contentSystem in anachronism. Then you can use recursiveReplaceString from anachronism to replace contentSystem with contentSystem-anachronism. When at all possible we should handle most of the remapping during the anachronism build process rather than adding that complexity to the system itself. This would avoid needing all those helper methods that accept a pack id.

jfn4th:
Thank you for the info, I can take another look tomorrow.

build/anachronism/index.ts
jfn4th: Yep, works fine!

CarlosFdez:
No problem! I already had an idea of how I wanted to do it, but I was preoccupied and putting it off and made the issue to remind myself. But I appreciate someone else handling it, so thank you!

jfn4th:
Always happy to help when I have the time, hence my string of recent AppV2/Svelte stuff (I've had a fair amount of Vue experience, so Svelte is fun for me)! I'd like to continue helping out there, especially on finally getting item and actor sheets closer to being ready to move. Like I mentioned in the description, if there are any ideas you guys already have on how we'd like spell chips to be handled I'd be interested in helping out there too.

CarlosFdez:
I just need to get off my arse with spell chips. I have an idea for allowing spell embedding in general using references rather than our current janky system, I just gotta do it.

I am 70% done with the loot sheet svelte rewrite, which should hopefully open up the rest. Glad to hear you're enjoying svelte.

jfn4th:
@CarlosFdez all updated. I also tightened up some spell types a bit. Kept that as a separate commit in case it you think it should be a separate PR.

jfn4th:
I noticed cantrip decks sat awkwardly outside the new per-system structure. Strictly speaking, a cantrip deck in SF2e without anachronism enabled is wrong. I've added another commit that moves cantrip decks into `SPELL_CONSUMABLES_BY_SYSTEM.pf2e` to match the rest of the consumables, plus some associated cleanup.

If that too should be a separate PR, or there isn't a desire to require anachronism, I'll drop it from here.

CarlosFdez:
putting cantripdeck 5 in all that is a good idea

CarlosFdez:
Took a look again. So I'm noticing that you changed some args to require OneToTen or SpellConsumableItemType but then cast the params rather than check and error. I think it should either remain strings/number (and clamp/error internally) or check using some early tests that error out, not simply cast an arbitrary string or number.

Its fine to cast at the same time you clamp, since that's more of a failure of the clamp method and the lack of numerical range types in typescript. But avoid doing so elsewhere.

The rest looks fine to me.

jfn4th:
Updated with checks and an error on invalid type. Didn't bother erroring on an `undefined` rank as `createConsumableFromSpell` accepts it and already handles throwing an error in that case. Coming from the UI, it really never should not fall within 1-10 anyway.

CarlosFdez:
Looks good, thanks for the PR!

PR #22355, opened 2026-05-17
Hide ActorsDeadAtZero setting from players

iDantar:
Got my break, thanks ;)


stwlam:
Death icon too (it's a world setting).

iDantar:
Updated, take another look please

PR #22366, opened 2026-05-19
Add back TakesNoDamage text on applyDamage card

### LINKED ISSUE #22364: Damage message for immunities in chat is coming up empty
Image

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22364

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1291,13 +1291,17 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
-        const persistentDamages = result.persistent
-            .map((instance) =>
-                instance.type === "bleed"
-                    ? `${instance.head.expression} bleed`
-                    : `${instance.head.expression} persistent ${instance.type} damage`,
-            )
-            .join(", ");
+        const persistentDamages = game.i18n
+            .getListFormatter({ style: "short", type: "conjunction" })
+            .format(
+                result.persistent.map((instance) => {
+                    const formula = instance.head.expression;
+                    const damageType = _loc(CONFIG.PF2E.damageTypes[instance.type]);
+                    return instance.type === "bleed"
+                        ? _loc(`${locPrefix}.PersistentEntry.bleed`, { formula, damageType })
+                        : _loc(`${locPrefix}.PersistentEntry.default`, { formula, damageType });
+                }),
+            );
         const persistentOnly =
             persistentCreated.length > 0 && (!hitPoints.max || finalDamage - damageAbsorbedByActor === 0);
 
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1291,7 +1291,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
-        const persistentDamages = game.i18n.getListFormatter({ style: "short", type: "conjunction" }).format(
+        const persistentDamages = game.i18n.getListFormatter({ style: "long", type: "conjunction" }).format(
             result.persistent.map((instance) => {
                 const formula = instance.head.expression;
                 const damageType = _loc(CONFIG.PF2E.damageTypes[instance.type]);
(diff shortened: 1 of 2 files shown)


src/module/actor/base.ts
```diff
@@ -1297,7 +1297,7 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
                 return hitPoints.value === hitPoints.max ? `${locPrefix}.AtFullHealth` : `${locPrefix}.HealedForN`;
             }
             if (!hitPoints.max || finalDamage - damageAbsorbedByActor === 0) {
-                return persistentCreated ? null : `${locPrefix}.TakesNoDamage`;
```
CarlosFdez, line 1300: this looks like the intent is to not show the "takes no damage" message if persistent damage was created. But persistentCreated is always truthy. This should clearly be persistentCreated.length instead of outright removal.

iDantar: I tried that, but imo it didn't look good
image

Maybe if we also move the undo button lower, into Persistent Damage block?

stwlam: Have it use the natural language that would otherwise be there ("Leng Spider starts taking 1d4 persistent damage.").

iDantar: Sorry, not sure if I got you right - something like this?
image

stwlam: Almost, but no big heading or bulleted list

"Leng Spider starts taking 1d4 bleed and 1d4 persistent acid damage."

iDantar: Pushed the changes, looks like this now:
image

CarlosFdez: Avoid the `:` and use whatever rendering the inline damage cards do.

`@Damage[{4[persistent,fire],5[persistent,cold]}]` turns into `4 persistent fire + 5 persistent cold`

iDantar: Just to make sure: this is a message about damage received - they usually don't render the inline damage rolls. Could you confirm that they are needed here, please?

iDantar: I removed the colon, but please confirm about the inline damage rendering @CarlosFdez - are you sure we need to add it here?

src/module/actor/base.ts
```diff
@@ -1291,13 +1291,23 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
+        const persistentDamages = result.persistent
+            .map((instance) =>
+                instance.type === "bleed"
+                    ? `${instance.head.expression} bleed`
```
Cuingamehtar: Please localize these as well.
And for the list, there is `game.i18n.getListFormatter({style:"narrow", type:"conjunction"})` (or short instead of "narrow", if you want "and")

iDantar: Good catch, thanks! Updated, now looks like this
image

src/module/actor/base.ts
```diff
@@ -1291,13 +1291,25 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             persistentDamage.length ? await this.createEmbeddedDocuments("Item", persistentDamage) : []
         ) as ConditionPF2e<this>[];
 
+        const persistentDamages = game.i18n.getListFormatter({ style: "short", type: "conjunction" }).format(
```
stwlam:
```suggestion
        const persistentDamages = game.i18n.getListFormatter({ style: "long", type: "conjunction" }).format(
```
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/ListFormat

src/module/actor/base.ts
stwlam: remove redundant "persistent"s, lowercase damage types

iDantar: Updated, looks like this now
image

iDantar: Gentle reminder, this is ready for review if you have a moment

stwlam: Some regret here; compare with published text containing multiple instances of persistent damage (I think it's "ndm persistent type1, ndm persistent type2, and ndm persistent type3").

iDantar: This is the best reference I was able to find: https://2e.aonprd.com/Conditions.aspx?ID=86#:~:text=X%20persistent%20%5Btype%5D%20damage
It matches your description, so I added 'persistent' back into message for all damage types except Bleed

PR #22368, opened 2026-05-19
Brush up persistent damage chat card

### LINKED ISSUE #22367: Damage taken card in chat is ignoring some styles for Persistent Damage
This is how message used to look like on Foundry v12
Image
This is how it looks now on v14
Image

The `border-bottom: 1px solid var(--color-underline-header);` is still in src\styles\ui\sidebar\chat\_damage.scss, but it's not applied for some reason

(Based on a conversation in ui-ux, kudos to websterguy and Rigo for bringing this up)

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22367

Looks like this after update:
image

### DIFF
--- a/src/styles/_colors.scss
+++ b/src/styles/_colors.scss
@@ -90,6 +90,7 @@ body,
 
     /* Borders */
     --color-border-divider: #baa991;
+    --color-underline-header: #782e22;
     --color-border-dark-input: #d3ccbc;
     --color-border-medium: gray;
 
--- a/src/styles/_colors.scss
+++ b/src/styles/_colors.scss
@@ -90,7 +90,6 @@ body,
 
     /* Borders */
     --color-border-divider: #baa991;
-    --color-underline-header: #782e22;
     --color-border-dark-input: #d3ccbc;
     --color-border-medium: gray;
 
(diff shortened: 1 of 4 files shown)


src/styles/_colors.scss
```diff
@@ -90,6 +90,7 @@ body,
 
     /* Borders */
     --color-border-divider: #baa991;
+    --color-underline-header: #782e22;
```
CarlosFdez: @stwlam override me here due to familiarity but this looks like an appv1 color, which I think we aren't supposed to reinstate. Might be good to find an appv2 equivalent (or remove the underline, I don't think we need it).

jfn4th: Chiming in as I've done a lot of core css var digging recently. It is an appv1 color, used as the default border under h1/h2/h3. Appv2 doesn't automatically underline headings anymore and doesn't have a direct equivalent.

iDantar: Personally, I like it a bit more with the line, especially if there are multiple instances of persistent, but I'm by no means a UI person
image

I removed the --color-underline-header and used --color-scrollbar for now, plus replaced that color's usage in all other places

stwlam: Without!

iDantar: Without it is. Also rolled back other color changes, to avoid unnecessary side-effects

iDantar: @stwlam Could you take another look please?

PR #22369, opened 2026-05-19
Apply materials from base damage to splash damage roll

### LINKED ISSUE #19669: Splash damage should inherit special material traits from originating weapon
The level 11 Bomber class feature allows an alchemist to infuse special materials into their bombs, however the splash damage from those bombs does not seem to inherit the material trait and trigger resistances and weaknesses.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/19669

### DIFF
--- a/src/module/system/damage/damage.ts
+++ b/src/module/system/damage/damage.ts
@@ -273,10 +273,11 @@ export class DamagePF2e {
                 .filter((s) => s.total > 0);
             const rolls: RollJSON[] = [];
             for (const splash of splashInstances) {
-                const formula = `(${splash.total}[splash])[${splash.damageType}]`;
-                const roll = await new DamageRoll(formula).evaluate();
-                roll.options.splashOnly = true;
-                rolls.push(roll.toJSON());
+                const flavor = [splash.damageType, ...roll.materials].join(",");
+                const formula = `(${splash.total}[splash])[${flavor}]`;
+                const splashRoll = await new DamageRoll(formula).evaluate();
+                splashRoll.options.splashOnly = true;
+                rolls.push(splashRoll.toJSON());
             }
 
             return rolls;
--- a/src/module/system/damage/damage.ts
+++ b/src/module/system/damage/damage.ts
@@ -275,9 +275,9 @@ export class DamagePF2e {
             for (const splash of splashInstances) {
                 const flavor = [splash.damageType, ...roll.materials].join(",");
                 const formula = `(${splash.total}[splash])[${flavor}]`;
-                const roll = await new DamageRoll(formula).evaluate();
-                roll.options.splashOnly = true;
-                rolls.push(roll.toJSON());
+const splashRoll = await new DamageRoll(formula).evaluate();
+splashRoll.options.splashOnly = true;
+rolls.push(splashRoll.toJSON());
             }
             return rolls;
         })();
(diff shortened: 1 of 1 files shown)


stwlam:
~~Materials can't be used as damage flavor tags.~~ except for the fact that I added it myself :melting_face:

iDantar:
Sorry, I'm not sure I follow what's meant by "can't be used" in this context.
If it's "this wouldn't work" - at least from my testing it looks to work
image
If you meant more like "this shouldn't be used because of unwanted side-effects" - could you please share more details on what this could affect?

CarlosFdez:
The only thing that brings me pause here is that we inject a lot of material options using stuff like `damage:material:silver` into roll options of inline damage. So there is a mismatch of expectations somewhere. I think this one is up to stwlam to resolve. So long as its possible I'm not fussy with the how.

src/module/system/damage/damage.ts
```diff
@@ -273,10 +273,11 @@ export class DamagePF2e {
...
             for (const splash of splashInstances) {
-                const formula = `(${splash.total}[splash])[${splash.damageType}]`;
-                const roll = await new DamageRoll(formula).evaluate();
-                roll.options.splashOnly = true;
-                rolls.push(roll.toJSON());
+                const flavor = [splash.damageType, ...roll.materials].join(",");
+                const formula = `(${splash.total}[splash])[${flavor}]`;
+                const splashRoll = await new DamageRoll(formula).evaluate();
+                splashRoll.options.splashOnly = true;
+                rolls.push(splashRoll.toJSON());
```
stwlam, line 280:
```suggestion
                const roll = await new DamageRoll(formula).evaluate();
                roll.options.splashOnly = true;
                rolls.push(roll.toJSON());
```

since we already know it's a splash roll

src/module/system/damage/damage.ts
```diff
@@ -273,10 +273,11 @@ export class DamagePF2e {
...
             }
```
stwlam:
```suggestion
```

src/module/system/damage/damage.ts
```diff
@@ -273,12 +273,12 @@ export class DamagePF2e {
                 .filter((s) => s.total > 0);
             const rolls: RollJSON[] = [];
             for (const splash of splashInstances) {
-                const formula = `(${splash.total}[splash])[${splash.damageType}]`;
+                const flavor = [splash.damageType, ...roll.materials].join(",");
+                const formula = `(${splash.total}[splash])[${flavor}]`;
                 const roll = await new DamageRoll(formula).evaluate();
                 roll.options.splashOnly = true;
                 rolls.push(roll.toJSON());
```
stwlam:
```suggestion
const splashRoll = await new DamageRoll(formula).evaluate();
splashRoll.options.splashOnly = true;
rolls.push(splashRoll.toJSON());
```

src/module/system/damage/damage.ts
```diff
@@ -273,12 +273,12 @@ export class DamagePF2e {
...
             for (const splash of splashInstances) {
-                const formula = `(${splash.total}[splash])[${splash.damageType}]`;
-                const roll = await new DamageRoll(formula).evaluate();
-                roll.options.splashOnly = true;
-                rolls.push(roll.toJSON());
+                const flavor = [splash.damageType, ...roll.materials].join(",");
+                const formula = `(${splash.total}[splash])[${flavor}]`;
+const splashRoll = await new DamageRoll(formula).evaluate();
+splashRoll.options.splashOnly = true;
+rolls.push(splashRoll.toJSON());
```
stwlam:
```suggestion
                const splashRoll = await new DamageRoll(formula).evaluate();
                splashRoll.options.splashOnly = true;
                rolls.push(splashRoll.toJSON());
```

stwlam:
This is going to need a followup to make sure it's only applying the material to splash damage if the damage type matches.

PR #22372, opened 2026-05-19
Update GrantItemRuleElement to allow implementation of Nimble Crawl

### LINKED ISSUE #7491: [Feature Request] Don't apply flatfooted from prone when you have Nimble Crawl and are Legendary in Acrobatic

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/7491

### DIFF
--- a/packs/pf2e/conditions/prone.json
+++ b/packs/pf2e/conditions/prone.json
@@ -35,6 +35,11 @@
             {
                 "inMemoryOnly": true,
                 "key": "GrantItem",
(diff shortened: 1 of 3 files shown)


stwlam:
We can't do this. Not all conditions can be correctly evaluated at that point.

PR #22386, opened 2026-05-22
Localize family ability glossary

iDantar:
Adds an ItemAlteration for name on each of the abilities in `packs\pf2e\bestiary-family-ability-glossary`, to replace it with a localized one, without parenthesis mentioning the family name. So `(Vampire, Jiang-Shi, Basic) One More Breath` becomes just `One More Breath`

To a kind reviewer: updates were done via script (included in PR for now, for convenience), so please don't be discouraged by change's size :)
One change post-script that I did was to Skymetal Metamorphosis abilities, as those already had entries in re-en

Kudos to @pedrogrullada for the idea


scripts/localize-family-glossary.mjs
```diff
@@ -0,0 +1,129 @@
+/**
+ * Apply bestiary-family-ability-glossary name localization for local testing.
+ *
+ * - Prepends an ItemAlteration rule that overrides each ability's display name
+ * - Adds PF2E.NPCAbility.<Family>.<Ability>.Name entries to static/lang/re-en.json
+ * - Ability keys use the pack name after the parenthetical prefix
+ *   (e.g. "(Vampire, Jiang-Shi, Minister) Dark Enlightenment" → DarkEnlightenment)
+ *
+ * Usage: node scripts/localize-family-glossary.mjs [--dry-run]
+ */
```
TikaelSol: Why the script?

iDantar: Thought this could be helpful for the review - not planning to include it in the final merge

iDantar: Removed it - it's still available in commit history, if needed

PR #22402, opened 2026-05-24
Add missing localization for Commander Fortitude Expertise

TikaelSol:


static/lang/re-en.json
```diff
@@ -3337,6 +3337,7 @@
                     "bard": {
                         "Description": "Your physique has grown hardy in your adventures. Your proficiency rank for Fortitude saves increases to expert."
                     },
+                    "commander": "Your physique has grown hardy in your adventures. Your proficiency rank for Fortitude saves increases to expert.",
```
Cuingamehtar: It needs to be under `{class}.Description` like other classes - the path is generated using the class slug.

PR #22413, opened 2026-05-26
Add toggleable Glitching Failure roll option

### LINKED ISSUE #22161: There is currently no way to activate the Glitching condition on DCs
The description of the Glitching condition says that if you have the Glitching condition and have failed the flat check, you should have an item penalty to all your checks and DCs equal to your glitching value or the glitching value of the item you're attempting to use, including saving throws, attack modifiers, and the DCs of abilities, effects, and spells you use.

The Glitching condition contains a Flat Modifier to all rolls equal to the value of the badge, but with a predicate ["glitching-failure"]
Currently there isn't any way to make that predicate true. The toggle can be turned on manualy when the Glitching creature makes a roll, but there is no such toggle when the Glitching creature is on the receiving side.
This means that, for example, if someone makes a Trip attempt against the Glitching creature, the Reflex DC should get the item penalty, but there is currently no way to apply it. Same for the AC of the Glitching creature.

Noisyink:
The Glitching condition's flat-check failure penalty (-value item to all checks and DCs) was gated behind a `glitching-failure` predicate that had no way to be activated when the glitching creature was the target of a roll, only when it was the roller. The penalty therefore could not apply to AC or to defender-side DCs (e.g. Reflex DC against Trip).

This adds a toggleable `RollOption` on the condition itself so the GM can mark the failed flat check. With the toggle on, the existing `FlatModifier { selector: "all" }` propagates the item penalty to both the creature's own checks and to its AC / save DCs when targeted (same propagation pattern as the (unpredicated) Frightened condition).

Can be made automatic on condition gain if wanted/needed.

Closes #22161

## Test plan
- [x] Apply Glitching 3 to an SF2e actor
- [x] Verify the new toggle appears under the actor's Actions panel
- [x] Tick the toggle, roll a Reflex save — modifier breakdown shows `-3 item (Glitching)`
- [x] Have another actor attack the glitching creature — AC shows the `-3 item` penalty
- [x] Tick off — penalty no longer applies

### DIFF
--- a/packs/sf2e/conditions/glitching.json
+++ b/packs/sf2e/conditions/glitching.json
@@ -32,9 +32,7 @@
         "rules": [
             {
                 "key": "FlatModifier",
-                "predicate": [
-                    "glitching-failure"
-                ],
+                "predicate": ["glitching-failure"],
                 "selector": "all",
                 "type": "item",
                 "value": "-@item.badge.value"
@@ -46,6 +44,12 @@
             {
                 "key": "RollOption",
                 "option": "self:condition:glitching:{item|badge.value}"
+            },
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.Glitching.FailureToggle",
+                "option": "glitching-failure",
+                "toggleable": true
             }
         ],
         "start": {
--- a/packs/sf2e/conditions/glitching.json
+++ b/packs/sf2e/conditions/glitching.json
@@ -32,7 +32,9 @@
         "rules": [
             {
                 "key": "FlatModifier",
-                "predicate": ["glitching-failure"],
+                "predicate": [
+                    "glitching-failure"
+                ],
                 "selector": "all",
                 "type": "item",
                 "value": "-@item.badge.value"
@@ -47,7 +49,6 @@
             },
             {
                 "key": "RollOption",
-                "label": "PF2E.SpecificRule.Glitching.FailureToggle",
                 "option": "glitching-failure",
                 "toggleable": true
             }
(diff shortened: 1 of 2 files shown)


packs/sf2e/conditions/glitching.json
```diff
@@ -32,9 +32,7 @@
         "rules": [
             {
                 "key": "FlatModifier",
-                "predicate": [
-                    "glitching-failure"
-                ],
+                "predicate": ["glitching-failure"],
```
simonward: It looks like this was manually entered rather than extracted.

Noisyink: Whoops, thanks for that - will note for future.

packs/sf2e/conditions/glitching.json
```diff
@@ -46,6 +44,12 @@
             {
                 "key": "RollOption",
                 "option": "self:condition:glitching:{item|badge.value}"
+            },
+            {
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.Glitching.FailureToggle",
```
simonward: This label seems a bit long. Could we get by with the default "Glitching"?

Noisyink: Yeah I would imagine we probably could, wanted to leave things open for any future specific rules without needing to rewrite. Happy to change this to leave it as glitching if we aren't expecting any other specific rules to be added.

simonward: We can always add a custom label later. Can't easily remove them from the strings file later.

Noisyink: Fair point! I'll make the change and test tonight, can't imagine it'll have any impact on functionality.

Noisyink: Done in f6a2902 - dropped the label and the strings entry, also restored the predicate to the multiline extracted format. Tested in a dev world, toggle row now reads "Glitching" via the item-name fallback.

PR #22414, opened 2026-05-26
Preserve token mirror state when applying TokenImage scale override

### LINKED ISSUE #22351: Tokens using a Token Image rule element can no longer be mirrored
Passing along an issue that was reported a while ago on my module: https://github.com/MAClavell/Token-Mirror-Button/issues/3

If a token is using a Token Image rule element to override scale, mirroring no long works. This happens even without my module when trying to set the mirror settings in the token's appearance tab.

I think the offending lines are here https://github.com/foundryvtt/pf2e/blob/ec5924284f733a0bbf1ce215acc4371650dcada3/src/module/scene/token-document/document.ts#L322

My repro:
1. Create a token
2. Create an action on the token
3. Add a Token Image rule element to the action
4. Set a scale override on the rule element
5. Attempt to mirror the token by going to the token's appearance tab and toggling either "Mirror Image" setting

Noisyink:
Fixes #22351.

The `TokenDocumentPF2e.prepareBaseData` override was blindly assigning `scaleX` / `scaleY` from a TokenImage rule element's `scale` field, overwriting any negative sign the user had set to mirror the token. Toggling Mirror Horizontally / Vertically in the token's Appearance tab appeared to do nothing on tokens with such a rule element.

Preserves the existing sign of `scaleX` / `scaleY` when applying the override, matching the pattern already used in `prepareScale` higher up in the same file (lines 254-257).

## Test plan
- [x] Created an effect with a TokenImage rule (`value: icons/svg/mystery-man.svg`, `scale: 1.5`), dropped on a blank actor (no existing token image) on the scene, confirmed the token reskinned and rescaled
- [x] Toggled Mirror Horizontally on that token via Configure Token, confirmed the mirror sticks across subsequent renders (this is the regression being fixed)
- [x] No regression on the simple path: the change is scoped inside the `if ("scaleX" in tokenOverrides.texture)` branch, so tokens without a TokenImage rule element never hit the new code

### DIFF
--- a/src/module/scene/token-document/document.ts
+++ b/src/module/scene/token-document/document.ts
@@ -319,8 +319,8 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         if (tokenOverrides.texture) {
             this.texture.src = tokenOverrides.texture.src;
             if ("scaleX" in tokenOverrides.texture) {
-                const mirrorX = this.texture.scaleX < 0 ? -1 : 1;
-                const mirrorY = this.texture.scaleY < 0 ? -1 : 1;
+                const mirrorX = Math.sign(this.texture.scaleX);
+                const mirrorY = Math.sign(this.texture.scaleY);
                 this.texture.scaleX = mirrorX * Math.abs(tokenOverrides.texture.scaleX);
                 this.texture.scaleY = mirrorY * Math.abs(tokenOverrides.texture.scaleY);
                 this.flags[SYSTEM_ID].autoscale = false;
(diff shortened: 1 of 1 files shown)


src/module/scene/token-document/document.ts
```diff
@@ -319,8 +319,10 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
         if (tokenOverrides.texture) {
             this.texture.src = tokenOverrides.texture.src;
             if ("scaleX" in tokenOverrides.texture) {
-                this.texture.scaleX = tokenOverrides.texture.scaleX;
-                this.texture.scaleY = tokenOverrides.texture.scaleY;
+                const mirrorX = this.texture.scaleX < 0 ? -1 : 1;
+                const mirrorY = this.texture.scaleY < 0 ? -1 : 1;
```
stwlam:
```suggestion
                const mirrorX = Math.sign(this.texture.scaleX);
                const mirrorY = Math.sign(this.texture.scaleY);
```

Noisyink:
Noted the style preference @stwlam. I'll submit PRs in line with changes.

PR #22427, opened 2026-05-27
Disambiguate FeatGroup slot ids when levels repeat

### LINKED ISSUE #22417: [Homebrew feature improvement] Allow multiple slots for the same level with campaignFeatSections
Currently you can only define unique levels per slot in the campaignFeatSections objects.

E.g. 
This is allowed: `slots: [1, 2, 3]`

But this isn't currently possible: `slots: [1, 1, 1]`

The latter does display correctly in the sheet, but in the dom each `<li>` slot gets the exact same data-slot-id, so the drag & drop behavior doesn't work correctly. (if a single slot is already taken, any new item dropped shifts to the Bonus Feats section.

---

Low prio feature request. I currently have a workaround to **_not_** set any `slots` at all. That way you can drag infinite amount of items on the feat category.

Noisyink:
Closes #22417.

Numeric `campaignFeatSections` like `slots: [1, 1, 1]` derived the same slot id three times, so only the last survived in `this.slots` and drops fell through to bonus feats. `R.unique` on the input collapses literal duplicates to one entry.

## Test plan
- [x] Synthetic `campaignFeatSection` with `slots: [1, 1, 1]`: renders as a single slot row, drops land in it
- [x] Workbench v7.2.0 dual class + ancestral paragon (slot arrays `[1,2,4,...,20]` and `[1,3,7,...,19]`): no duplicates, behave identically to pre-PR

### DIFF
--- a/src/module/actor/character/feats/group.ts
+++ b/src/module/actor/character/feats/group.ts
@@ -56,7 +56,7 @@ class FeatGroup<TActor extends ActorPF2e = ActorPF2e, TItem extends FeatLike = F
 
         if (data.slots) {
             this.slotted = true;
-            for (const slotData of data.slots) {
+            for (const slotData of R.unique(data.slots)) {
                 const slotObject =
                     typeof slotData === "object"
                         ? slotData


CarlosFdez:
Test workbench's dual class and ancestral paragon features. If it works, then we should be fine and dandy, but use `foundry.utils.logCompatibilityWarning`.

iDantar:
Also please take into account this PR, as it interacts with feat slot assignment as well https://github.com/foundryvtt/pf2e/pull/22249

Noisyink:
> Also please take into account this PR, as it interacts with feat slot assignment as well #22249

No worries, I'll be working through getting Carlos' comment tested on the weekend so I can have a look over that pull at the same time

Noisyink:
Pushed [a94a1cf](https://github.com/foundryvtt/pf2e/pull/22427/commits/a94a1cf) + [0edf87a](https://github.com/foundryvtt/pf2e/pull/22427/commits/0edf87a). Warning now routes through `foundry.utils.logCompatibilityWarning`, with a small module-scoped Set on top to keep it at one log line per group per session (Foundry's built-in `once` keys by stack, which fired ~5 times per actor lifecycle).

**Workbench checked on v7.2.0:** with dual class + ancestral paragon enabled on a level-20 PC, both sections render with the expected slot counts (11 and 6), slot IDs are unchanged from pre-PR (no `-N` suffixes, neither variant repeats levels), drops land in the right slot, and the warning fires once per group then stays quiet on resets.

**@iDantar** re #22249, no semantic overlap (this PR touches `evenLevels` in the free-archetype group and slot-id derivation; #22249 adds heritages to the `ancestry` group). Happy to rebase on top of yours if it lands first.

CarlosFdez:
I've been thinking about this more. For stuff like workbench, we could migrate data for now so that when finally removed, people who were playing were fine. But given that this is in a case of stagnancy (we can only really resume create a UI for it once homebrew settings are converted to svelte), I'm starting to wonder if instead just doing R.unique() on the slots so that duplicate slots just become one might not also solve the problem.

CarlosFdez:
Yeah im leaning on passing the slots through R.unique(). The problem is multiple slots with matching ids. It kinda reduces all this to a one liner but I think it'll be better for now.

Noisyink:
> Yeah im leaning on passing the slots through R.unique(). The problem is multiple slots with matching ids. It kinda reduces all this to a one liner but I think it'll be better for now.

Yeah easy enough, I'll cut it all down this arvo then 🙂 I'll keep the branch open incase there's a need for this later down the track for whatever reason.

Noisyink:
@CarlosFdez force pushed - down to a single line fix - validated it works using a synthetic "slots: [1, 1, 1]" rendered as one working slot. Ran a quick smoke test, no issues.

CarlosFdez:
That should be enough and handle the situation. Thanks for the PR and for looking into it.

PR #22430, opened 2026-05-28
Add missing SF physical item compendiumSources; remove non-compendium compendiumSources

Cuingamehtar:


packs/sf2e/starfinder-society-bestiary/season-1/1-17/corpse-fleet-infantry.json
```diff
@@ -275,9 +284,6 @@
         },
         {
             "_id": "9NHl8lYJ4iaZsysl",
-            "_stats": {
-                "compendiumSource": "Item.R6vSuJ9b5F3AUj6q"
```
CarlosFdez, line 279: Looks like this one might be a missing one?

Cuingamehtar: The only entry I removed manually just not to have to switch between worlds one more time. Thanks.

CarlosFdez:
Looks good to me! This looked annoying. Maybe it'd be worth adding some filter for world uuids in the extract if this is a common occurrence.

Cuingamehtar:
Actually, I just made two macros for it. One that strips, another that assigns.
Because those are just physical items, it was easy to search by name+description match.

An effective way to curate the field would be to replace the uuid with readable one on extracting (if possible). That way when converting those on build, bad uuids will cause errors.

I'm actually curious why so much ammunition from recent PF packs was missing sources. It seems like a systematic error.

Cuingamehtar:
This is probably not exhaustive, but If there are cases that I've missed, I can probably catch them when doing translation.

CarlosFdez:
I'm guessing what's happening is that people are making all the stuff in the world and then dragging it into the item, and there's nothing that validates it or warns about it.

Issue #22408, opened 2026-05-25
[Foundry 14] Feature suggestion: Charactersheet as detachable Window


vaka42:
Feature suggestion:
The character sheet should use the newly implemented pop-out feature. You can detach other windows, but not the character sheet. 

Played on:
Foundry 14 Build 361
Pathfinder 2e System v8.1.2

Noisyink:
The detach control is part of ApplicationV2's window header, and PF2e character sheets are still on V1. Migrating them is a sizeable refactor, effectively blocked on the broader V1 to V2 sheet migration.

In the meantime, [PopOut!](https://github.com/League-of-Foundry-Developers/fvtt-module-popout) works on Foundry V14 with PF2e's V1 sheets once the V13 max cap is removed (tracked at [League-of-Foundry-Developers/fvtt-module-popout#165](https://github.com/League-of-Foundry-Developers/fvtt-module-popout/issues/165)). Verified on PopOut! v2.23 + Foundry 14.363 + PF2e v8.1.2.

websterguy:
> In the meantime: the PopOut! (https://github.com/League-of-Foundry-Developers/fvtt-module-popout) module still works on V14 with PF2e's V1 sheets

Popout has never worked correctly with many of pf2e's sheet's features.

The PC sheets aren't going to be migrated to AppV2 any time soon.

Noisyink:
> > In the meantime: the PopOut! (https://github.com/League-of-Foundry-Developers/fvtt-module-popout) module still works on V14 with PF2e's V1 sheets
> 
> Popout has never worked correctly with many of pf2e's sheet's features.
> 
> The PC sheets aren't going to be migrated to AppV2 any time soon.

What in particular isn't working? I tested quite a number of things with PopOut! and it seemed to be working fine. I'm working on some local patching for PopOut! for my own games where needed, could potentially look to raise a couple PRs to fix compatibility.

Edit;: I'll start putting together a playwright spec to test this anyway - see what surfaces

Noisyink:
Following up on the PopOut! compatibility concern, ran a Playwright harness across a 27-feature matrix on PC and NPC sheets, popped-in vs popped-out, three runs each, on Foundry V14.363 / PF2e v8.1.2 / PopOut! v2.23 with the V14 cap removed.

[...]

PR #22434, opened 2026-05-29
Fix DOM helpers returning null cross-window in PopOut!

Noisyink:
Partial fix for #22408 (the visible-error symptoms on popped-out sheets, not the underlying ApplicationV1-to-V2 migration).

`instanceof Element` and `instanceof Document` return false when the node belongs to a different window's globals, which is what PopOut! does. Switched the guards to duck-typed property presence (`"querySelector"`, `"querySelectorAll"`, `"closest"`); prototype-agnostic but still rejects `EventTarget`s that lack the methods.

## Test plan
- [x] Foundry v14.363 + PopOut! v2.23, PF2e and SF2e v8.1.2 dev worlds
- [x] Pre-patch: clicking a roll-option toggle on a popped sheet throws `Sheet navigation not found` and the sheet stops rendering
- [x] Post-patch: toggle flips cleanly, sheet re-renders, no error
- [x] Repro: drag any feat with a togglable RollOption (Devise a Stratagem, Hunt Prey) onto a character, pop the sheet, click the toggle

### DIFF
--- a/src/util/dom.ts
+++ b/src/util/dom.ts
@@ -89,7 +89,7 @@ function htmlQuery<K extends keyof HTMLElementTagNameMap>(
 function htmlQuery(parent: MaybeHTML, selectors: string): HTMLElement | null;
 function htmlQuery<E extends HTMLElement = HTMLElement>(parent: MaybeHTML, selectors: string): E | null;
 function htmlQuery(parent: MaybeHTML, selectors: string): HTMLElement | null {
-    if (!(parent instanceof Element || parent instanceof Document)) return null;
+    if (!parent || !("querySelector" in parent)) return null;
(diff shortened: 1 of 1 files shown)


stwlam:
Please dispense with the AI-generated text dump in this and future PRs.

Noisyink:
> Please dispense with the AI-generated text dump in this and future PRs.

Easy enough, figured more was better but can cut it down to bare bones 🤙 just using a lot of the same pipelines I designed for work so it's probably more verbose than it needs to be for community projects

Have gone and retro-actively removed majority of the fluff from everything around issue diagnosis and issue replication - just leaving it simple. Have done this for all my active PRs.

PR #22435, opened 2026-05-29
Fix hiding roll opposer names

### LINKED ISSUE #22396: Bug - NPC/Creature Names Showing in Chat
I think this has started since the move to system version on Foundry 14. It also affects both PF2e and SF2e.

I have the option in 'limit metagame information' enabled for:

Image

But on attacks targeting NPCs/Creatures, the name of the creature is being revealed in chat:

Image

Tested with all modules disabled.

Image

jfn4th:
Closes #22396, opposer is the only thing I found not respecting the setting.


CarlosFdez:
Works, thanks for the PR!

PR #22438, opened 2026-05-29
Make Crossbow Crack Shot's bonus damage and backstabber upgrade mutually exclusive

### LINKED ISSUE #22377: Crossbow Crack Shot doesn't work as intended
1. It doesn't increase crossbows' range increment.
2. It applies both Crack Shot's precision damage and increased backstabber precision damage to backstabber crossbows.

pedrogrullada:
Also remove unnecessary domain from its Roll Option toggle.

One might also read that backstabber weapons don't get the range increment upgrade when their target is off-guard. Seems RAW but I couldn't get the automation for it to work anyway. Someone else might want to look into it.

- Closes #22377

### DIFF
--- a/packs/pf2e/feats/class/gunslinger/level-1/crossbow-crack-shot.json
+++ b/packs/pf2e/feats/class/gunslinger/level-1/crossbow-crack-shot.json
@@ -39,7 +39,7 @@
                     {
                         "nand": [
                             "item:trait:backstabber",
-                            "target:trait:off-guard"
+                            "target:condition:off-guard"
                         ]
                     }
                 ],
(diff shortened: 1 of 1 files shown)


packs/pf2e/feats/class/gunslinger/level-1/crossbow-crack-shot.json
```diff
@@ -36,7 +35,13 @@
                 "damageCategory": "precision",
                 "key": "FlatModifier",
                 "predicate": [
-                    "crossbow-crack-shot"
+                    "crossbow-crack-shot",
+                    {
+                        "nand": [
+                            "item:trait:backstabber",
+                            "target:trait:off-guard"
```
simonward:
```suggestion
                            "target:condition:off-guard"
```

PR #22440, opened 2026-05-29
Fix description in Gathered lore and Precise discipline

rectulo:


simonward:
Consider fixing the same issue in "Precise Discipline".

PR #22442, opened 2026-05-29
Add pf2e.castSpell hook

### LINKED ISSUE #20761: [Feature Request] New event: pf2e.castSpell
Hopefully this is a simple enough request to fulfill. But as a module developer wanting to make some automations I have ran into a roadblock related to how the sheet handles spell interactions. The Cast button is great, but there's no way for modules to differentiate between a spell being printed to chat through clicking the spell icon vs a spell being printed to chat from the Cast button.

So i propose a new event. `pf2e.castSpell`. When the player presses the Cast button on their sheet this event is fired. It really only needs to include the spell being cast, the rank it's being cast (also the spell's base level), then the uuids of the usual context stuff (actor uuid, item uuid for the spell, spellcasting entry uuid). This will make automating things around spells being cast a lot easier!

Noisyink:
Closes #20761. Thanks @MrVauxs for the design input on the pre-hook shape.

Adds two hooks fired from each `BaseSpellcastingEntry.cast()` (document, rituals, item-spellcasting, trick):

- `Hooks.call("pf2e.preCastSpell", spell, feature, options)` at the top of `cast()`. Handlers can mutate `options` (rank, consume, message) in place to alter the cast, or return `false` to cancel entirely. Motivated by Vauxs's use cases: Essence Casting needs to skip slot consume and advance its own resource; spellshapes need to mutate cast rank.
- `Hooks.callAll("pf2e.castSpell", spell, feature, { ...options, castRank })` right before `spell.toMessage()`. Fire-and-forget post-cast, distinguishes deliberate casts from spell-icon chat prints.

## Test plan
- [x] Post-hook fires with `(spell, feature, { ..., castRank })` on Cast button click (verified Caustic Blast @ rank 3 via Arcane Spontaneous)
- [x] Pre-hook mutating `opts.consume = false` produces a chat card but does not decrement the spell slot
- [x] Pre-hook returning `false` cancels the cast: no chat card, no consume, no error
- [x] No-hook regression: with all `pf2e.preCastSpell` / `pf2e.castSpell` listeners removed, Cast button decrements slots and renders the chat card as before

### DIFF
--- a/src/module/item/spellcasting-entry/document.ts
+++ b/src/module/item/spellcasting-entry/document.ts
@@ -264,6 +264,7 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
         if (message && valid) {
             spell.system.location.value ??= this.id;
             const castRank = spell.computeCastRank(rank);
+            Hooks.callAll("pf2e.castSpell", spell, this, { ...options, castRank });
             await spell.toMessage(null, { mode: options.messageMode, data: { castRank } });
         }
     }
--- a/src/module/item/spellcasting-entry/document.ts
+++ b/src/module/item/spellcasting-entry/document.ts
@@ -257,6 +257,7 @@ class SpellcastingEntryPF2e<TParent extends ActorPF2e | null = ActorPF2e | null>
 
     /** Cast the given spell as if it was part of this spellcasting entry. */
     async cast(spell: SpellPF2e<ActorPF2e>, options: CastOptions = {}): Promise<void> {
+        if (Hooks.call("pf2e.preCastSpell", spell, this, options) === false) return;
         const consume = options.consume ?? true;
         const message = options.message ?? true;
         const rank = options.rank ?? spell.rank;
--- a/src/module/item/spellcasting-entry/item-spellcasting.ts
+++ b/src/module/item/spellcasting-entry/item-spellcasting.ts
@@ -87,6 +87,7 @@ class ItemSpellcasting<TActor extends CreaturePF2e = CreaturePF2e> implements Sp
     }
 
     async cast(spell: SpellPF2e, options: CastOptions = {}): Promise<void> {
+        if (Hooks.call("pf2e.preCastSpell", spell, this, options) === false) return;
         const message = options.message ?? true;
         if (message && this.canCast(spell, { origin: spell.parentItem })) {
             spell.system.location.value = this.id;
--- a/src/module/item/spellcasting-entry/item-spellcasting.ts
+++ b/src/module/item/spellcasting-entry/item-spellcasting.ts
@@ -90,6 +90,7 @@ class ItemSpellcasting<TActor extends CreaturePF2e = CreaturePF2e> implements Sp
         const message = options.message ?? true;
         if (message && this.canCast(spell, { origin: spell.parentItem })) {
             spell.system.location.value = this.id;
+            Hooks.callAll("pf2e.castSpell", spell, this, { ...options, castRank: spell.rank });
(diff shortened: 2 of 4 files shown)


CarlosFdez:
Few things:
- "it builds" isn't really a "test plan". All checks passing in github is effectively that.
- Entry should be a parameter. It also shouldn't be called entry, it should perhaps be called feature. The intention is that a spellcasting entry is basically a feature that can take a spell and cast it. A spellcasting collection is basically just your spontaneous spells/signature spells. Its not properly split right now, and item casting's current implementation and what to do with focus spells does muck it up, but that was the "end goal". 
- I suspect people may want to modify the spell. You should poke Vauxs in the discord. He's done the most toying with spells.

Noisyink:
> Few things:
> 
>     * "it builds" isn't really a "test plan". All checks passing in github is effectively that.
> 
>     * Entry should be a parameter. It also shouldn't be called entry, it should perhaps be called feature. The intention is that a spellcasting entry is basically a feature that can take a spell and cast it. A spellcasting collection is basically just your spontaneous spells/signature spells. Its not properly split right now, and item casting's current implementation and what to do with focus spells does muck it up, but that was the "end goal".
> 
>     * I suspect people may want to modify the spell. You should poke Vauxs in the discord. He's done the most toying with spells.

1. Noted on tests - removed and won't bother including in future
2. Noted - Did you want an issue raised to investigate the split? Probably more than I could handle solo depending on the size of the split, but if I'm talking to Vauxs anyway I could potentially see what he thinks. I might have a glance myself to see how much time it'd be.
3. Easy - I'll reach out to him, I've probably taken too literal of an approach to the hook.

Noisyink:
@CarlosFdez had a quick chat with Vauxs - updates below
- Renamed the second hook arg to feature in the description and test plan. Left the BaseSpellcastingEntry.cast() references and the directory/class names alone
- Added pf2e.preCastSpell alongside the existing post-hook. Vauxs flagged a couple things that cause mutations so I've included Hooks.call so handlers can mutate options in place or return false to cancel.
- New tests added

CarlosFdez:
Preferably don't touch the split, that's a major system project and its mired with differing ideas on what everything means. It will likely not be touched until after the appv2 rewrite.

Few things now, that you should probably verify:
- What is the use case for castSpell now? It looks to be pre message. Would people prefer post message if pre message exists
- pre cast is done before computeCastRank(), which is the focus spell/cantrip auto-scale. I think that needs to be set in the options before pre cast, but verify of course..

Noisyink:
> Preferably don't touch the split, that's a major system project and its mired with differing ideas on what everything means. It will likely not be touched until after the appv2 rewrite.
> 
> Few things now, that you should probably verify:
> 
>     * What is the use case for castSpell now? It looks to be pre message. Would people prefer post message if pre message exists
> 
>     * pre cast is done before computeCastRank(), which is the focus spell/cantrip auto-scale. I think that needs to be set in the options before pre cast, but verify of course..

Yeah no objections on the split actions - Still learning the codebase so definitely don't want to touch anything with major break potential haha

1. Ahh yep makes sense. I'll rename castSpell to postCastSpell and move it after toMessage(). Keeps functionality for fire and forget listeners, happy for others to comment here though if there is a different direction needed. - I've flicked @MrVauxs  the thread if he wants to weigh in?
2. Ok, so just to clarify my interpretation there, it's probably required to have auto-rank calculate before any mutations otherwise they'll get ovewitten by computeCastRank() for things like cantrips and focus spells?

I'll make those changes and run some additional testing if you're happy with the above.

stwlam:
We're not doing this right now, sorry.

Noisyink:
> We're not doing this right now, sorry.

No worries!

PR #22443, opened 2026-05-30
Add text filter and slot-header toggle to PC spellcasting tab

Noisyink:
Closes #19937 (text portion).

Adds a name search and a per-user toggle to collapse the spell slot rank headers, both in the spellcasting tab header.

## Test plan
- [x] Type in the search: non-matching spell rows hide, clearing restores
- [x] Click the toggle: rank headers collapse and restore, persists across reload

Issue #19937, opened 2025-08-17
Feature Request: text filter for PC spell list


aedelor:
High level casters can have a _lot_ of spells. A text entry at the top of the spell tab which filters the list by entered text would be amazing. Maybe other kinds of filters would be nice, but I think even just text filtering would be a huge QOL improvement.

Noisyink:
Created two PRs - One for basic text filtering #22443, another for advanced filtering #22444.

PR #22444, opened 2026-05-30
Add advanced filters to PC spellcasting tab

Noisyink:
Additional for #19937 closed by #22443 - this is completely optional. Split into a second PR to make it easier to reject if it's too much of a code addition.

Depends on #22443 (text filter). This branch bundles #22443's changes; once it merges, this rebases down to just the advanced filter additions.

Chip filters for rank, tradition, action cost, save, damage type, and concentration. ANDs with text search. Tradition and damage type chips only show options the actor's spells actually use. Active-count badge on the toggle when the panel is closed.

## Test plan
- [x] Level-20 PC with ~80 spells across all ranks in a spontaneous Arcane entry
- [x] Each filter dimension individually and in combination, ANDed with text search
- [x] Clear button resets all chips, toggle badge reflects count, panel reflows on resize


Noisyink:
Moves with sheet resize - I'm honestly still a little unsure about the order of elements as it creates a bit of a "gappy" UI feature when at default size, but couldn't work out how to make it look better. I think it looks good but happy to take notes.

image

image

Cuingamehtar:
It's a minor point, but what's this "Concentration" toggle? If you meant "Sustained", then it should say so.

Noisyink:
> It's a minor point, but what's this "Concentration" toggle? If you meant "Sustained", then it should say so.

Right, sorry, had 5e on the brain, easy fix

Noisyink:
image

Straight up find and replace - all renamed to sustained

Noisyink:
**Rank search**
image

**Rank search + Save**

image

-----------

**Damage search** 

image

-----------

**Text search**
image

**Text search ANDed**
image

Noisyink:
image

Updates from PR #22443 included - this PR treats 22443 as a dependency.

PR #22445, opened 2026-05-30
Remove migrations for foundry versions up to and including 10

stwlam:


CarlosFdez (approved):
772's pre update's being missing might trip up remigrates after the flag fix but also maybe migration 845 might take care of it even if it is somehow a problem.

PR #22454, opened 2026-05-31
Convert deletion keys to forced-deletion operators in recent migrations

stwlam:


src/module/migration/migrations/950-ammo-consumable-to-ammo-ammo.ts
```diff
@@ -168,13 +167,10 @@ export class Migration950AmmoConsumableToAmmoAmmo extends MigrationBase {
         if ("game" in globalThis) {
             // Assign to both system and ==system. Future migrations can write to system and it "should" be fine, they're the same reference
             source.system = system as ConsumableSystemSource;
-            source["==system"] = system;
```
CarlosFdez, line 171: Does foundry still error if we change item type?  That was the reason for the ==

CarlosFdez (approved):

PR #22455, opened 2026-05-31
Restore rules accurate measured template coverage

### LINKED ISSUE #21958: V14 effect-area shapes are wrong
They're just not right.

jfn4th:
This only affects regions with coverage highlighting (`highlightMode: "coverage"`), the measured templates. Regions using the default shape highlighting aren't touched.

Circles, cones, and lines are grid counted again instead of using Foundry's geometric shapes. Also restores cursor aiming after placement and walls cutting off line of effect.

A few decisions worth a look:

- Blocked squares are shaded (v13 behavior), not dropped. Could exclude them entirely instead and would be simpler.
- Blocking uses `move` collision and could read the origin's traits later, like auras do for auditory and visual.
- Lines measure from the grid corner they fire from. Measuring from the center could consistently result in an extra 5ft being highlighted at certain angles. The outline box isn't re-snapped, so it can sit up to half a cell off the highlighted squares for off-corner placements.

Screenshot 2026-05-30 at 10 24 57 PM
Screenshot 2026-05-30 at 10 25 04 PM
Screenshot 2026-05-30 at 10 25 36 PM
Screenshot 2026-05-30 at 10 25 48 PM
Screenshot 2026-05-30 at 10 26 02 PM
Screenshot 2026-05-30 at 10 26 24 PM
Screenshot 2026-05-30 at 10 26 40 PM
Screenshot 2026-05-30 at 10 27 42 PM
Screenshot 2026-05-30 at 10 28 06 PM
Screenshot 2026-05-30 at 10 28 32 PM
Screenshot 2026-05-30 at 10 28 58 PM
Screenshot 2026-05-30 at 10 58 59 PM
Screenshot 2026-05-30 at 11 29 28 PM

### DIFF
--- a/src/module/canvas/layer/region.ts
+++ b/src/module/canvas/layer/region.ts
@@ -1,5 +1,11 @@
+import type { PlaceablesLayerPointerEvent } from "@client/canvas/layers/base/placeables-layer.d.mts";
+import type { ConeShapeData, LineShapeData } from "@common/data/data.d.mts";
 import { RegionDocumentPF2e } from "@scene";
 import { RegionPF2e } from "../index.ts";
+
+/** Cones fire along the eight grid directions; lines can be aimed at any cell, so they step much finer. */
+const directionSnap = (type: string): number => (type === "cone" ? 45 : 5);
+const isAimable = (type: string): boolean => type === "cone" || type === "line";
 
 export class RegionLayerPF2e extends fc.layers.RegionLayer<RegionPF2e> {
     override placeRegion(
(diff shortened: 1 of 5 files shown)


jfn4th:
Now that ring support is merged, I'm wondering if the logic this applied to circles should similarly be applied to rings? Logically I'd think yes, but examples only exist for bursts and emanations. To my knowledge rings like wall of fire are not explicitly defined as either and I know we try to avoid editorializing. @stwlam let me know your thoughts.

jfn4th:
Now also restores the Shift to place directly and Ctrl for free rotate placement modifiers the existing tooltip mentions.

PR #22390, opened 2026-05-23
Add support for ring templates

simonward:
Screenshot 2026-05-22 at 7 46 03 PM
Screenshot 2026-05-22 at 7 46 25 PM

Closes https://github.com/foundryvtt/pf2e/issues/22374

PR #22458, opened 2026-05-31
Add Area Damage roll Option to Wall of Fire's damage

ammalagonc:
Thank you @websterguy for the solution :)


TikaelSol:
If #22390 is merged wouldn't we be able to just add a proper area to this with variants?

CarlosFdez:
Thanks for reminding me of that PR Tikael 😄

PR #22459, opened 2026-05-31
Fix EphemeralEffects application when using TokenMark as predicate

### LINKED ISSUE #20594: Ephemeral Effect with Token Mark predicate does not behave as expected
If using a TokenMark for a predicate in EphemeralEffect, instead of `"target:mark:mark-slug"`, the Ephemeral triggers off `"self:mark:mark-slug"`

Examples used:
```json
{
  "key": "TokenMark",
  "slug": "feinted-against"
}
```
```json
{
  "key": "EphemeralEffect",
  "predicate": [
    "self:mark:feinted-against"
  ],
  "selectors": [
    "melee-attack-roll","melee-damage"
  ],
  "uuid": "Compendium.pf2e.conditionitems.Item.AJh5ex99aV6VTggg"
}
```
This works, but `target:mark:feinted-against` does not.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/20594

Collect markOptions separately for EphemeralEffects, so 'origin' and 'target' in mark predicate are independent of context

### Tests
These two were broken before, fixed now:

- Pointed Question
- Unbelievable Connection

This is regression testing, confirmed to still work:

- Aquatic Ambush
- Surprise Attack
- Deinonychu's bleed
- Bomber's Eye Elixir
- Tidewater Guard
- Extravagant Parry
- Cringe 
- Prey Mutagen
- Grandmother Spider's curse
- Disarm Success

### DIFF
--- a/src/module/actor/roll-context/base.ts
+++ b/src/module/actor/roll-context/base.ts
@@ -219,7 +219,7 @@ abstract class RollContext<
         const distanceOption = Number.isInteger(distance) ? `${opposingRole}:distance:${distance}` : null;
 
         // Extract origin and target marks
-        const markOptions = (() => {
+        const { markOptions, ephemeralMarkOptions } = (() => {
             const originActor = unresolved.origin?.actor;
             const originUuid = unresolved.origin?.token?.uuid;
             const targetActor = unresolved.target?.actor;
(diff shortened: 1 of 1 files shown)


stwlam:
What do you mean by "this is regression"? The PR breaks new things?

iDantar:
No, regression testing - confirmed not to break these cases. Updated the description

PR #22473, opened 2026-06-02
Override aeon trait for SF2e

### LINKED ISSUE #22462: Wrong tooltip for the Aeon trait
The Aeon trait on weapons currently uses the same tooltip as the Aeon trait on creatures.

Image

It should be

You can fit one aeon stone or spell gem into this weapon with a single Interact action. While you possess the weapon with an aeon stone in its indentation, you gain all the benefits of the aeon stone as if it were orbiting your head. While the weapon has a spell gem attached, you gain a number of bonus battery charges equal to the spell gem’s item level × 5. These charges regenerate during your daily preparations. In addition, you can use the spell gem while wielding the weapon, though using the gem in this manner still destroys it.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22462


iDantar:
Just for visibility: this is not the best solution, as it wouldn't work with Anachronism. Thinking on the better approach

PR #22483, opened 2026-06-04
Add inline roll options to Engulf

simonward:


packs/pf2e/rage-of-elements-bestiary/living-grove.json
```diff
@@ -178,7 +178,7 @@
...
+                    "value": "<p>@Check[reflex|dc:22|options:damaging-effect,inflicts:grabbed,inflicts:slowed], @Damage[5d8[bludgeoning]] damage, [[/act escape dc=20]], Rupture 10</p>\n<p>A creature Engulfed by the living grove must also attempt a @Check[fortitude|dc:22|options:inflicts:stunned] save as it's battered between the thin, tightly packed trunks.</p>\n<p><strong>Critical Success</strong> The creature is unaffected.</p>\n<p><strong>Success</strong> The creature is @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 1}.</p>\n<p><strong>Failure</strong> The creature is @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 2}.</p>\n<p><strong>Critical Failure</strong> The creature is @UUID[Compendium.pf2e.conditionitems.Item.Stunned]{Stunned 4}.</p>\n<hr />\n<p>@Localize[PF2E.NPC.Abilities.Glossary.Engulf]</p>"
```
simonward: Even though the source lists the second save as basic, there's no damage component so I believe it's an error.

TikaelSol: I think the basic is right, the ability is just weird. they are having the reflex be to avoid it and fortitude be to mitigate damage.

PR #22488, opened 2026-06-05
Fix attribute builder breaking after opening it for a second character

### LINKED ISSUE #22487: Attempting to edit attribute boosts on a character if another one is open throws errors in log and prevents window from opening
I was comparing two PCs in my campaign and noticed that weirdness happens if attempting to edit two characters at once.

specific steps:

1. open the character sheets for characters 'A' and 'B'
2. click the edit boosts button on 'A' and see the appv2 boost editor open
3. click the edit boosts button on 'B' expecting a second boost editor to open (like editing languages would)
4. watch the boost editor for 'A' become the boost editor for 'B'
5. click the edit boosts button on 'A' again
    > TypeError: Cannot read properties of null (reading 'offsetWidth')
    at AttributeBuilder2._updatePosition (foundry.mjs:30977:93)
    at AttributeBuilder2.setPosition (foundry.mjs:30953:26)
    at #render (foundry.mjs:30484:39)
    at async #try (foundry.mjs:7635:17)   

from this point on, one cannot  edit the boosts for 'A' until one reloads the webpage.

I was able to reproduce this with a fresh module-free world

jfn4th:
Closes #22487

The old attribute builder allowed multiple instances to be open, but when I rewrote it to AppV2 it was moved to a single instance with a static ID in review. Needs to be closed before another can be opened.

When a second instance renders with the same static ID, core never closes the first. It just evicts it from the instances registry and detaches its element. `_tearDown` never runs, leaving a zombie app in actor.apps that throws on any later render.

### DIFF
--- a/src/module/actor/character/apps/attribute-builder/app.ts
+++ b/src/module/actor/character/apps/attribute-builder/app.ts
@@ -61,8 +61,8 @@ class AttributeBuilder extends SvelteApplicationMixin<
 
     protected override _configureRenderOptions(options: DeepPartial<AttributeBuilderRenderOptions>): void {
         super._configureRenderOptions(options);
-        const actor = options.actor;
-        if (actor && actor !== this.#actor) {
+        const actor = options.actor ?? this.#actor;
+        if (actor !== this.#actor) {
             delete this.#actor.apps[this.id];
             this.#actor = actor;
             // Registration with the actor on first render is handled by `_onFirstRender`
--- a/src/module/actor/character/apps/attribute-builder/app.ts
+++ b/src/module/actor/character/apps/attribute-builder/app.ts
@@ -11,8 +11,15 @@ interface AttributeBuilderConfiguration extends fa.ApplicationConfiguration {
     actor: CharacterPF2e;
 }
 
+interface AttributeBuilderRenderOptions extends fa.ApplicationRenderOptions {
+    /** A character to which this application should switch its focus */
+    actor?: CharacterPF2e;
+}
+
 class AttributeBuilder extends SvelteApplicationMixin<
-    AbstractConstructorOf<fa.api.ApplicationV2> & { DEFAULT_OPTIONS: DeepPartial<AttributeBuilderConfiguration> }
+    AbstractConstructorOf<fa.api.ApplicationV2<fa.ApplicationConfiguration, AttributeBuilderRenderOptions>> & {
+        DEFAULT_OPTIONS: DeepPartial<AttributeBuilderConfiguration>;
+    }
 >(fa.api.ApplicationV2) {
     /** Number of attributes in the game (str, dex, con, int, wis, cha) */
     static ALL_ATTRIBUTES_COUNT = 6;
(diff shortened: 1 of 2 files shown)


CarlosFdez:
Are you sure this isn't a core bug? What prevents this being an issue for anything else with a shared id, like the spell prep dialog (beyond being appv1).

EDIT: Or ABCPicker, which is appv2

jfn4th:
I can look a little more and create an upstream issue, but weapon reloader and trade dialog both have similar guards currently. ABC doesn't use a static id and allows multiple instances.

CarlosFdez:
ABC does not allow multiple instances for the same actor and type. It has a uniqueId set in the initialization. I just saw that weapon reloader and trade dialog has such a guard.

I'll leave it to @stwlam to decide if this is a core issue since he has closed fixes for core issues before. But otherwise the PR itself looks fine to me.

jfn4th:
I should have been more specific, I was referring to it allowing multiple instances of the same type across different actors, which attribute builder doesn't allow.

ABC actually gets orphaned the same way (double-clicking the same actor's button evicts the first instance without closing it), but it never breaks because nothing re-renders the zombie. The handler always constructs a fresh instance, and ABCPicker doesn't register in actor.apps. The attribute builder crashed because the old handler retrieved the orphan from actor.apps and called render() on its detached element.

Mr-Dufus:
will this fix allow me to have two attribute builders open at once? ... the use-case that triggered the bug was me trying to rebuild a 10th level character from scratch as the original had gotten severely corrupted, and I needed to see the old characters boosts while editing the new character?

also wow at getting right on top of this one :)

src/module/actor/character/sheet.ts
```diff
@@ -826,11 +826,15 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
...
-            return builder.render({ force: true });
+        handlers["edit-attribute-boosts"] = async () => {
+            const existing = foundry.applications.instances.get("attribute-builder");
+            if (existing instanceof AttributeBuilder) {
+                // Re-rendering an existing instance with the same element ID for a different actor would detach the
+                // open builder's element without closing it, leaving it in a broken render state: close it first.
+                if (existing.options.actor === this.actor) return existing.render({ force: true });
+                await existing.close();
+            }
+            return new AttributeBuilder({ actor: this.actor }).render({ force: true });
```
stwlam: Instead of closing and reopening, allow for `actor` as an optional render option, which will change the actor being operated on.

jfn4th: Thanks. Since AppV2 freezes options, I moved the actor onto a mutable field and swap it through a render option. I'll probably go through and audit other AppV2s sometime this week.

jfn4th:
@Mr-Dufus sorry, only being able to have one open at a time is intentional. This only addresses the builder breaking when performing a swap.

src/module/actor/character/apps/attribute-builder/app.ts
```diff
@@ -39,25 +46,40 @@ class AttributeBuilder extends SvelteApplicationMixin<
...
 
     /** Re-evaluate ABP status on each access rather than caching at construction to avoid need for page reload */
     get #abpEnabled(): boolean {
         return game.pf2e.variantRules.AutomaticBonusProgression.isEnabled(this.#actor);
     }
 
+    protected override _configureRenderOptions(options: DeepPartial<AttributeBuilderRenderOptions>): void {
+        super._configureRenderOptions(options);
+        const actor = options.actor;
+        if (actor && actor !== this.#actor) {
```
stwlam:
```suggestion
        const actor = options.actor ?? this.#actor;
        if (actor !== this.#actor) {
```

stwlam (approved):

Mr-Dufus:
> @Mr-Dufus sorry, only being able to have one open at a time is intentional. This only addresses the builder breaking when performing a swap.

darn,

PR #22489, opened 2026-06-06
Correct spelling of 'Talamandor' to 'Talmandor' in For Talmandor! For…

rectulo:
… Freedom!


iDantar:
Looks like CI failed because the file name doesn't match with the updated action name

rectulo:
I don't know how to fix it in this case. I'm just a poor lonesome translator...

iDantar:
Let's see if I can help you :)

iDantar:
Looks like nothing would break if you just rename the file to match the spelling: `for-talmandor-for-freedom.json`

ammalagonc:
Data lead can chime here, but the correct order of operations:
- Make a new feat
- Do a redirect in pf2e.json
- Delete the old feat

Cuingamehtar:
No need to make a new feat - it's just a typo fix. All uuids will be correct still.
So yes, the only thing left to do is to change the file name to match the item slug.

rectulo:
Nothing is broken

simonward:
You would just need to rename it in Foundry, then extract. Since the id is the same, there's no issue.

PR #22490, opened 2026-06-06
Make clown car respect elevation and levels

### LINKED ISSUE #22321: The Party token always drops the party members at elevation 0.
_**The Bug**_
The clowns (or is that just my players) from the clown car always get dropped at elevation 0 which might not be the level where you intended to drop them.

_**How to Reproduce:**_
Take a scene with multiple levels (like the Restored Keep scene), 
put the party token down on a level that isnt at elevation 0,
deploy the clowns from the clown car,
and watch as they roll up/down to elevation 0.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22321

### DIFF
--- a/src/module/scene/token-document/clown-car.ts
+++ b/src/module/scene/token-document/clown-car.ts
@@ -73,7 +73,7 @@ class PartyClownCar {
             return {
                 _id: token._id ?? "",
                 ...(square
-                    ? { ...R.pick(square, ["x", "y"]), elevation: car.elevation, level: car.level }
+                    ? { ...R.pick(square, ["x", "y"]), ...R.pick(car, ["elevation", "level"]) }
                     : carPosition),
             };
         });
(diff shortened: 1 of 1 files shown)


src/module/scene/token-document/clown-car.ts
```diff
@@ -69,7 +70,12 @@ class PartyClownCar {
             const square = freeSpaces.findSplice(
                 (s) => Math.abs(s.x - token.x) >= widthPixels && Math.abs(s.y - token.y) >= widthPixels,
             );
-            return { _id: token._id ?? "", ...R.pick(square ?? car, ["x", "y"]) };
+            return {
+                _id: token._id ?? "",
+                ...(square
+                    ? { ...R.pick(square, ["x", "y"]), elevation: car.elevation, level: car.level }
```
CarlosFdez:
```suggestion
                    ? { ...R.pick(square, ["x", "y"]), ...R.pick(car, ["elevation", "level"]) }
```

Also test it to be sure I didn't mess up.

CarlosFdez:
There's one suggestion I have, but otherwise looks fine at first glance.

iDantar:
Thanks! Re-tested it after running linter fix, all good :)

CarlosFdez (approved):

PR #22491, opened 2026-06-06
Correct Malfunctioning Perfumery's Crushing Notes DC value

ammalagonc:


simonward:
Consider adding `inflicts` to the saving throws.

PR #22494, opened 2026-06-06
Add melee/ranged attack roll domains to spell attack rolls

### LINKED ISSUE #21592: Ranged spell attacks do not add the `ranged-attack-roll` option to their rolls, so automation fails for Curse of Creeping Ashes
Image

The automation for Curse of Creeping Ashes relies on `ranged-attack-roll` as a selector, and it works on ranged weapon attacks (e.g. shortbow) but fails on ranged spell attacks (e.g. Divine Lance, Ignition (ranged)).

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/21592

### DIFF
--- a/src/module/system/statistic/statistic.ts
+++ b/src/module/system/statistic/statistic.ts
@@ -411,10 +411,14 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
         })();
 
         const self = this.actor;
-        const domains = this.domains;
+        const item = args.item ?? null;
+        const domains = [...this.domains];
+        if (item?.isOfType("spell") && domains.includes("spell-attack-roll")) {
+            if (item.isMelee) domains.push("melee-attack-roll");
+            else if (item.isRanged) domains.push("ranged-attack-roll");
+        }
         const selfToken = args.token ?? self.getActiveTokens(true, true).shift() ?? null;
         const selfIsTarget = self === args.target || (!args.target && this.type === "saving-throw");
-        const item = args.item ?? null;
         const targetToken = selfIsTarget
             ? selfToken
             : (args.target?.getActiveTokens(true, true)?.find((t) => t.actor?.isOfType("army", "creature", "hazard")) ??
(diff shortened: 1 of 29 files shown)


src/module/system/statistic/statistic.ts
```diff
@@ -411,10 +411,14 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
         })();
 
         const self = this.actor;
-        const domains = this.domains;
+        const item = args.item ?? null;
+        const domains = [...this.domains];
+        if (item?.isOfType("spell") && domains.includes("spell-attack-roll")) {
```
stwlam: Check that this won't affect saving throws against spells.

iDantar: Tested - looks good to me: the "spell-attack-roll" check works

Image

stwlam:
The reason we passed over this before is because there are a lot of abilities that predicate on "ranged-attack-roll" but shouldn't operate on spells. A migration might be necessary.

src/module/system/statistic/statistic.ts
```diff
@@ -411,10 +411,14 @@ class StatisticCheck<TParent extends Statistic = Statistic> {
...
+            if (item.isMelee) domains.push("melee-attack-roll");
+            else if (item.isRanged) domains.push("ranged-attack-roll");
+        }
```
stwlam: This isn't an appropriate spot for this: `Statistic` is agnostic as to what it's rolling for. Check methods like `SpellPF2e#rolllAttack`.

iDantar: Moved this into the SpellPF2e#rolllAttack, thanks
Had to add a new argument for `StatisticRollParameters`. It seems in line with other arguments like `extraRollOptions`, but expanding an interface just for one use-case doesn't feel right, even if it's the cleanest way I see at the moment

iDantar:
Regarding predicates and selectors using 'ranged-attack-roll'. I vent through all of them, TL:DR: some would require migration, but most don't. 

They can be grouped like this:

**Migration required, should work only for a weapon strike:** Forge-Blessed Shot, Assisting Shot, Incredible Aim, Sniper's Aim, Triangle Shot, Effect: Aquatic Combat, Sentry's Aim, double-shot, triple-shot, multishot stance

**Fixed by this change, work same as Curse of Creeping Ashes from initial bugreport:** Spirits' Interference, 'Effect: -2 circumstance penalty to ranged attacks', 'Spell Effect: Curse of Recoil'

**Unaffected:**
Have other predicates: Legion's Aim, Far Throw, Impossible Volley
Kingmaker armies without spells
Affect only range increment, irrelevant for spells: hunt prey and it's variants

iDantar:
Updated affected items and added a migration

PR #22501, opened 2026-06-08
Update more foundry types

CarlosFdez:


types/foundry/client/canvas/placeables/sound.d.mts
```diff
@@ -73,6 +73,17 @@ export default class AmbientSound<
...
     updateSource(options?: { defer?: boolean; deleted?: boolean }): void;
 
+    /* -------------------------------------------- */
+    /*  Sound Source Management                     */
+    /* -------------------------------------------- */
+
+    /**
+     * Compute the field-of-vision for an object, determining its effective line-of-sight and field-of-vision polygons
+     * @param [options] Options which modify how the audio source is updated
+     * @param [options.deleted] Indicate that this SoundSource has been deleted.
```
stwlam:
```suggestion
     * @param options Options which modify how the audio source is updated
     * @param options.deleted Indicate that this SoundSource has been deleted.
```

types/foundry/client/documents/wall.d.mts
```diff
@@ -15,6 +17,37 @@ export default class WallDocument<TParent extends Scene | null = Scene | null> e
...
+     * @param options Options which modify how the edge is initialized
+     * @param [options.deleted] Delete the edge of this Wall document?
+     * @param [options.priorLevels] The IDs of prior Levels, if the levels this Wall document is included in has changed
+     * @param [options.changedTypes] The restriction types that either now affected or no longer affected by this Wall document
+     */
+    initializeEdge(options?: {
+        deleted?: boolean;
+        priorLevels?: string[] | Set<string>;
+        changedTypes?: Set<EdgeRestrictionType>;
+    }): void;
```
stwlam: These should go in the class definition, I think

types/foundry/client/global.d.mts
```diff
@@ -15,6 +15,12 @@ declare module "pixi.js" {
     export import LegacyGraphics = PixiJS.Graphics;
     export import smooth = PixiGraphicsSmooth;
     export import particles = PixiParticles;
+    enum UPDATE_PRIORITY {
+        PERCEPTION = 2,
+        PRIMARY = 3,
+        INTERFACE = 22,
+        OBJECTS = 23,
+    }
```
stwlam, line 23: No enums 🙂 

(core uses it but never in anything new, and it was basically used to mean "record")

CarlosFdez: The thing we are declaration merging is an enum. We can change the original though I guess.

CarlosFdez: Wait no we can't, the original is in PIXI types. They are injected by Foundry in client/canvas/board.mjs in `#activateTicker()`, but the original type we are merging into is an enum.

stwlam: Ah, gotcha.

PR #22505, opened 2026-06-08
Prevent icons in advanced alchemy items from wrapping

iDantar:
Kudos to @CarlosFdez for the suggestion

### DIFF
--- a/src/styles/actor/character/_crafting.scss
+++ b/src/styles/actor/character/_crafting.scss
@@ -278,7 +278,11 @@
             }
 
             .item-controls {
+                align-items: center;
+                display: flex;
+                flex-wrap: nowrap;
                 grid-area: controls;
+                justify-content: end;
                 justify-self: end;
                 font-size: var(--font-size-12);
                 margin-right: var(--space-4);


iDantar:
Before/after screenshot
image

src/styles/actor/character/_crafting.scss
```diff
@@ -278,7 +278,11 @@
             }
 
             .item-controls {
+                align-items: center;
+                display: flex;
+                flex-wrap: nowrap;
```
CarlosFdez, line 283: If the problem is that it doesn't have enough space, shouldn't that mean that control width needs to be wider?

iDantar: That was my first thought too! I started by increasing 
`.crafting-entry {
        --controls-width: 3rem;`
to 4rem, but after testing it with nowrap, that increase felt unnecessary:

Image

CarlosFdez: if there's no possible way for a mismatch in controls, then using min-content there is fine.

iDantar: Expended formula could cause a mismatch (this is with min-content)
image

PR #22511, opened 2026-06-10
Update Recall Knowledge breakdown and ABC picker on actor updates, brush up Item Transfer and Select Item popups

jfn4th:
Following up on #22488

- Moved item transfer, select item, update currency, distribute coins, scroll/wand creator, earn income, and item attacher from static to templated IDs (`earn-income-{id}`). These kept static IDs from appv1 where duplicates didn't matter, but can legitimately have multiple instances open. Item transfer and select item also resolve a promise on close, so refocusing or retargeting an existing instance would leave the new caller hanging.
- Actual singletons (check prompt, browser settings, license viewer, variant rules, world clock, ABC picker, persistent damage editor) get a `_canRender` guard that refocuses the open instance instead of rendering over it.
- Recall knowledge breakdown reuses the open instance and switches its data via render option, like the attribute builder.

Also registers the ABC picker with its actor's apps.

### DIFF
--- a/src/module/actor/sheet/popups/item-transfer-dialog.ts
+++ b/src/module/actor/sheet/popups/item-transfer-dialog.ts
@@ -97,6 +97,10 @@ class ItemTransferDialog extends fa.api.DialogV2<ItemTransferConfiguration> {
             buttons.push({ type: "submit", icon, label, action: mode, callback });
         }
 
+        // Close the open dialog first to avoid colliding on the shared id; this cancels the prior request.
+        // TODO: re-target in place (no close) once this moves from DialogV2 to a full ApplicationV2.
+        await foundry.applications.instances.get("item-transfer-dialog")?.close();
+
         return super.wait(Object.assign(options, { item, mode, newStack, lockStack, content, buttons }));
     }
 
--- a/src/module/actor/character/apps/abc-picker/app.ts
+++ b/src/module/actor/character/apps/abc-picker/app.ts
@@ -64,6 +64,19 @@ class ABCPicker extends SvelteApplicationMixin<
         return initialized;
     }
 
+    protected override async _onFirstRender(
+        context: ABCPickerContext,
+        options: fa.ApplicationRenderOptions,
+    ): Promise<void> {
+        await super._onFirstRender(context, options);
+        this.options.actor.apps[this.id] = this;
+    }
+
+    protected override _tearDown(options: fa.ApplicationClosingOptions): void {
+        delete this.options.actor.apps[this.id];
+        super._tearDown(options);
+    }
+
     /** Gather all items of the request type from the world and across all item compendiums. */
     async #gatherItems(): Promise<ABCItemRef[]> {
         const { actor, itemType } = this.options;
--- a/src/module/actor/npc/sheet.ts
+++ b/src/module/actor/npc/sheet.ts
@@ -400,7 +400,12 @@ class NPCSheetPF2e extends AbstractNPCSheet {
         };
 
         handlers["open-recall-breakdown"] = () => {
-            return new RecallKnowledgePopup({ identificationData: this.actor.identificationDCs }).render(true);
+            // Reuse an existing popup if one is open: constructing a second instance with the same element ID
+            // would detach the open popup's element without closing it, breaking its render state.
+            const actor = this.actor;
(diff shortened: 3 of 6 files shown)


src/module/actor/sheet/popups/item-transfer-dialog.ts
```diff
@@ -73,7 +74,7 @@ class ItemTransferDialog extends fa.api.DialogV2<ItemTransferConfiguration> {
             mode,
             newStack,
             lockStack,
-            rootId: "item-transfer-dialog",
+            rootId: `item-transfer-dialog-${fu.randomID()}`,
```
stwlam:
```suggestion
            rootId: this.id,
```

jfn4th: the `wait` method we're in here is static, `rootId` is set before instantiating the item transfer dialog. So `this.id` resolves to the class, not an instance, and lands as undefined in the label/input ids. Core hits the same "no stable id yet" case in `DocumentSheetV2#_prepareContext`, which I took this pattern from.

stwlam:
I'm not really following the rationale behind this:
- What is an "orphaned" application?
- Why do our singleton applications (these are all intended to be singletons) need a `_canRender` guard?

jfn4th:
By "orphaned" I mean these are intended singletons but nothing enforces that at runtime. Core's settings handler does `new menu.type(); app.render(true)` on every click (config.mjs:212), and DialogV2.wait/render constructs a fresh instance too. When a second instance renders with an id already in `foundry.applications.instances`, core evicts the first and swaps its element without closing it. The first is left in a rendered state with a detached element, `_onClose`/`_tearDown` never run. It leaks, and re-rendering it throws offsetWidth on a null parent.

Easy repro on the real Legal Notice viewer. Hit the button to open it twice without closing it in between. After the second click:

- `foundry.applications.instances.get("license-viewer")` is now the second instance
- the first is still `state === 2` (RENDERED) but `element.isConnected === false`
- only one #license-viewer is in the DOM, but the first is still in memory
- Calling `.render()` throws `Cannot read properties of null (reading 'offsetWidth')`

Every extra click strands another one. The `_canRender` guard fixes it by refocusing the open instance and cancelling the duplicate render, which is what actually makes them singletons.

Here's a minimal standalone version.

```js
class Demo extends foundry.applications.api.ApplicationV2 {
  static DEFAULT_OPTIONS = { id: "orphan-demo", window: { title: "Demo" } };
  async _renderHTML() { return "<p>content</p>"; }
  _replaceHTML(r, c) { c.innerHTML = r; }
}
const a = new Demo(); await a.render({ force: true });
const b = new Demo(); await b.render({ force: true }); // same id, e.g. a 2nd click
foundry.applications.instances.get("orphan-demo") === b; // true: a was evicted
a.state;                        // 2 (RENDERED), never closed
a.element.isConnected;          // false: detached
await a.render();               // throws: Cannot read properties of null (reading 'offsetWidth')
```

If this sounds like a core bug to you I'll kick this over there and trim this PR down to just my first and third bullets in the description.

stwlam:
> Easy repro on the real Legal Notice viewer. Hit the button to open it twice without closing it in between. After the second click:
> 
> * `foundry.applications.instances.get("license-viewer")` is now the second instance
> * the first is still `state === 2` (RENDERED) but `element.isConnected === false`
> * only one #license-viewer is in the DOM, but the first is still in memory

What's keeping the first in memory other than manually tracking it in the dev console?

jfn4th:
You're right. I was overzealous here. Retested and trimmed it way down.

The pure singletons get their orphan garbage collected with nothing re-rendering it, so it's a lifecycle oddity, not a leak. Dropped the guards from those, and put all the dialogs back to singletons.

The three changes this now makes

- ItemTransfer / SelectItem → still singletons, but they're await-based, so opening a second while the first was up left the first's promise unsettled. SelectItem now reuses the open dialog and re-targets it in place, ItemTransfer closes the open one first (with a TODO to re-target once it's off DialogV2 and a full AppV2). Either way the first request resolves cleanly now.
- ABCPicker → wasn't registered in actor.apps, so the heritage picker's list didn't react when the ancestry changed while it was open. Registering it fixes that and matches the sibling pickers.
- RecallKnowledge → holds the actor and registers now, so the breakdown picks up elite/weak changes instead of showing stale DCs. Follows the AttributeBuilder pattern.

Issue #20090, opened 2025-08-31
Include origin item roll options in effects dragged from a character sheet


pedrogrullada:
We currently automate some feats, spell, and equipment effects by having those effects pick up roll options from their origin item, such as their slug (in the case of item varities such as elixirs of life), or their tags (typically added via Item Alteration, such is the case with psychic amps, thaumaturge implements, etc). This works out as intended when these effects are dragged from chat, but when dragged from an actor's sheet, they lack these origin item roll options, which results in some awkward UX.

PR #22514, opened 2026-06-10
Include origin item roll options in effects dragged from a sheet

jfn4th:
Implements #20090

Before: Had to select item when applying effect

https://github.com/user-attachments/assets/d5e94b1d-53e2-4fa2-bedd-f252fe503c2c

After:

https://github.com/user-attachments/assets/9518ecdb-33b2-4263-aa2f-399407b9e3a4


stwlam:
- github is picky about linking issues: use "- Closes #[number]" or "- Fixes #[number]"
- go ahead and add labels in future PRs ("pr: bugfix" in this case)

jfn4th:
@stwlam I think I have to be added as a collaborator in order to add labels.

PR #22515, opened 2026-06-10
Update blurb for Ran-to from grappler to wrestler

rectulo:
according to the other versions of this NPC and to AON


simonward:
The source PDF does say grappler, as opposed to the other two.
Screen Shot 2026-06-10 at 1 28 58 PM
That said, it's probably an oversight?

rectulo:
I don't know what is best ? Do I need to change the two other versions of Ran-To (lvl 14 and lvl 20) from wrestler to grappler ?

SpartanCPA:
We use the PDF as the source of truth. If the stat blocks say different things, then they say different things.

PR #22517, opened 2026-06-10
Implement One-Inch Punch for Monastic Archer Stance

### LINKED ISSUE #18106: Monastic Archer Stance in Combination with One-Inch Punch, additional Dice are not added on the Bow Attack

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/18106

It's a slightly hacky way to do it, but doesn't require code changes

### DIFF
--- a/packs/pf2e/feat-effects/stance-monastic-archer-stance.json
+++ b/packs/pf2e/feat-effects/stance-monastic-archer-stance.json
@@ -20,7 +20,152 @@
             "remaster": true,
             "title": "Pathfinder Player Core 2"
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "attack-roll",
+                "key": "RollOption",
+                "option": "monastic-archer:half-increment",
+                "predicate": [
+                    "item:group:bow",
+                    "target:range-increment:1",
+                    {
+                        "or": [
+                            {
+                                "and": [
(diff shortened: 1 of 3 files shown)


packs/pf2e/feat-effects/stance-monastic-archer-stance.json
```diff
@@ -20,7 +20,152 @@
             "remaster": true,
             "title": "Pathfinder Player Core 2"
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "attack-roll",
```
TikaelSol, line 25: Why two rules? Just do one and omit the domain to make it all.

iDantar: This is a weird edge-case: if domain is not specified, the default is 'all'. It works fine for attack rolls, but in case of damage rolls, unless we specify the domain, beforeRoll  wouldn't fire - in this case the damage option for one-inch-punch would be disabled in damage roll window, even if it's checkbox is activated and the one-millimeter-punch would just not work
For this fix leaving only the damage roll would be enough, but it would be useful for implementing other features, like Metal Strikes, to work with the Monastic Archer Stance (once the code changes allow that)

PR #22520, opened 2026-06-11
Improve trade initiation: party-sheet member drops, request expiry with alerts on both sides, prevent sending to familiars

### LINKED ISSUE #20988: Dragging an item from PC's inventory to another character in the Party sheet's stash tab results in an error instead of opening the trade menu

jfn4th:
- Closes #20988

Moves trade initiation into `ActorSheetPF2e#moveItemBetweenActors` so party sheet drops request a trade instead of erroring. Auto-decline unanswered prompts after 30 seconds with alerts on both sides. Previously the request would time out on the sender's end while the receiver's prompt stayed open and, if accepted, opened the trade dialog for that user alone.

Also removes some redundant checks and restrict trade recipients to PCs and NPCs explicitly. `canTrade` only required that the recipient be a creature, so a familiar passed the sender-side check and the request was sent, but the receiver threw during validation (requires PCs/NPCs) before showing any prompt, and the sender was immediately shown a "no response after 30 seconds" timeout error.


iDantar:
One neat trick shark may appreciate during a review: 
If description line with linked issue is marked as list, github expands the link to show issue's summary, for example: `- Closes #20988` instead or `Closes #20988`

PR #22525, opened 2026-06-12
Prevent chat bubbles if message doesn't look like in-character text

### LINKED ISSUE #22522: Various interface buttons send chat message bubbles hovering over token when chat is in "In Context" mode
Image

This happens when an item is posted to chat, spell is cast, damage or check rolled and rerolled.

I think I figured out that this bubble pops up if ChatMessage options don't explicitly have `chatBubble` set to false, and the user is using IC chat mode.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/22522

Checks the message, if chatBubble is undefined, checks if the message has any marks it's not character speach (like damage dice, being posted from sheet, etc) and sets chatBubble to false.
Regular text, /ic, /emote still create a chat bubble

### DIFF
--- a/src/module/chat-message/document.ts
+++ b/src/module/chat-message/document.ts
@@ -377,6 +377,28 @@ class ChatMessagePF2e extends ChatMessage<UserPF2e | null> {
         if (canvas.ready) this.token?.object?.emitHoverOut(nativeEvent);
     }
 
+    protected override async _preCreate(
+        data: DeepPartial<this["_source"]>,
+        options: ChatMessageCreateCallbackOptions,
+        user: fd.BaseUser,
+    ): Promise<boolean | void> {
+        const pf2e = data.flags?.[SYSTEM_ID];
+        if (
+            options.chatBubble === undefined &&
+            (data.rolls?.length ||
+                data.flavor ||
+                data.style === CONST.CHAT_MESSAGE_STYLES.OTHER ||
+                pf2e?.origin ||
+                pf2e?.context ||
+                pf2e?.casting ||
+                pf2e?.modifiers?.length ||
+                pf2e?.dice?.length)
+        ) {
+            options.chatBubble = false;
+        }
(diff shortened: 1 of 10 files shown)


src/module/chat-message/document.ts
```diff
@@ -377,6 +377,28 @@ class ChatMessagePF2e extends ChatMessage<UserPF2e | null> {
...
+        if (
+            options.chatBubble === undefined &&
+            (data.rolls?.length ||
+                data.flavor ||
+                data.style === CONST.CHAT_MESSAGE_STYLES.OTHER ||
+                pf2e?.origin ||
+                pf2e?.context ||
+                pf2e?.casting ||
+                pf2e?.modifiers?.length ||
+                pf2e?.dice?.length)
```
jfn4th: Would `data.rolls?.length || data.flags?.[SYSTEM_ID]` suffice here? Every system message that sets flavor/origin/context/casting/etc. also carries system flags, so this should cover the same cases and would cover future message types. It also means a module's IC speech message doesn't lose its bubble just because it has flavor attached.

iDantar: Thanks! Yeah, that was an overkill on my side. 
I had to leave the flavor in, though: actions like 'raise shield' were popping up in bubbles othervise

jfn4th: Ah I see. I think auxiliary actions like that are the only system thing to set flavor. Perhaps they should just pass `chatBubble` as false.

iDantar: That makes sense. I added explicit chatBubble=False for actions like that, but a bit unsure if I got them all

stwlam:
Look for `CONST.CHAT_MESSAGE_STYLES.EMOTE` and kill any you find.

iDantar:
All targets eliminated, awaiting orders

PR #22529, opened 2026-06-12
Truncate effect link if it doesn't fit in a single line

iDantar:
Cuts effect's name if it wouldn't fit in a single line
Before/After
imageimage

### DIFF
--- a/src/styles/ui/sidebar/chat/_action.scss
+++ b/src/styles/ui/sidebar/chat/_action.scss
@@ -110,12 +110,8 @@
             padding-bottom: 0.1rem;
         }
 
-        .effect-applied > a.content-link {
-            flex: 0 1 auto;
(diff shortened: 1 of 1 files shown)


stwlam:
Just fix the wrapping

iDantar:
I minimized fix to only wrapping, looks like this now
image
Longer entries still look rather broken, but at least it doesn't explode vertically

stwlam:
But that's overflowing, not wrapping

iDantar:
Sorry, @stwlam I feel there is a miscommunication somewhere. I interpreted your first comment as 'only fix the wrapping'. But when testing I saw that it still looks bad, even if for a different reason. I think it would be better with hidden overflow, like this:
image
Could I return it?

iDantar:
@stwlam Gentle reminder

PR #22532, opened 2026-06-12
Craft prepared formulas at prepared quantity

iDantar:
Fixes bug: clicking 'Craft' on an item in Advanced Alchemy expends the formula and creates 1 item, no matter how many there actually are
image


stwlam:
There's a bug here, but it's that you can't continue crafting after using one of a higher number of allocated slots.

PR #22541, opened 2026-06-15
Recalculate encounter threat/XP when a combatant's level or alliance changes

### LINKED ISSUE #16460: Encounter metrics (threat, xp) doesn't take level adjustments into account
Applying Elite or Weak to NPCs in an encounter doesn't alter the calculated threat.  This is misleading.

jfn4th:
- Closes #16460

### DIFF
--- a/src/module/actor/base.ts
+++ b/src/module/actor/base.ts
@@ -1941,6 +1941,20 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
             }
         }
 
+        // Recalculate encounter metrics if this actor's level, adjustment, or alliance changed
+        const detailsChanged = changed.system?.details ?? {};
+        const attributesChanged = changed.system?.attributes ?? {};
+        if (
+            game.combats.size > 0 &&
+            ("level" in detailsChanged || "alliance" in detailsChanged || "adjustment" in attributesChanged)
+        ) {
+            const encounters = game.combats.filter((c) => c.combatants.some((combatant) => combatant.actor === this));
+            for (const encounter of encounters) {
+                encounter.metrics = encounter.analyze();
+            }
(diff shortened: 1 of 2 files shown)


src/module/actor/base.ts
```diff
@@ -1941,6 +1941,20 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
+        // Recalculate encounter metrics if this actor's level, adjustment, or alliance changed
+        const detailsChanged = changed.system?.details ?? {};
+        const attributesChanged = changed.system?.attributes ?? {};
+        if (
+            game.combats.size > 0 &&
+            ("level" in detailsChanged || "alliance" in detailsChanged || "adjustment" in attributesChanged)
+        ) {
+            const encounters = game.combats.filter((c) => c.combatants.some((combatant) => combatant.actor === this));
+            for (const encounter of encounters) {
+                encounter.reset();
```
stwlam:
```suggestion
                encounter.analyze();
```
Can we make do with just refreshing the metrics?

src/module/actor/base.ts
```diff
@@ -1941,6 +1941,20 @@ class ActorPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | n
...
+            }
+            if (encounters.length > 0) ui.combat.render();
```
stwlam:
```suggestion
            if (encounters.length > 0) ui.combat.render({ parts: ["metrics"] });
```
Similarly

src/module/actor/base.ts
jfn4th: yes, but `encounter.metrics = encounter.analyze();`

src/module/actor/base.ts
jfn4th: needs a type update, pushing now

PR #22542, opened 2026-06-15
Fix total price not updating on item transfer dialogs

### LINKED ISSUE #22533: Total Price value is not updated on item transfer dialogs
When transferring from a loot actor to a regular actor, there is normally supposed to be a total purchase price value that updates on the button itself. It currently looks like this.

Image

HOWEVER it is supposed to look this (I hacked it to work momentarily)

Image

It looks like there are two reasons this is failing. One of them is easy, the other I have no idea beyond rewriting it to no longer be a DialogV2.

1) The quantityInput in `#renderPurchasePrice()` no longer works. That can be changed to `range-picker[name=quantity]`. This will allow the price result to show up on the initial render.

2) The _onChangeForm() is no longer firing when the form updates. This prevents the price input from attempting to update when updating quantity.

This is currently giving me some difficulty with fixing https://github.com/foundryvtt/pf2e/issues/22518 entirely. While its easy enough to prevent credstick stacking, I want to add a toggle for transferring credits vs transferring credsticks, and I need to disable the quantity input when the checkbox changes, which requires me to add something to _onChangeForm().

jfn4th:
- Closes #22533

This is a minimal fix to unblock the credstick issue. I used the selector fix from the issue. For the `_onChangeForm` half, the reason it never fires is that DialogV2 only ever attaches a `submit` listener to its form and never a `change` one, so there's nothing to trigger it. Wired up a change listener manually in `_onRender` to get it going again.

imo this should eventually move off DialogV2 and into a full AppV2 + Svelte since it's really a live form rather than static content + buttons like DialogV2 is meant for, but I kept this minimal for now.

~~@CarlosFdez I'd mark you explicitly as reviewer, but I'm not a collaborator. Any chance I could get added so I can start tagging my things properly?~~ thanks shark!

### DIFF
--- a/src/module/actor/sheet/popups/item-transfer-dialog.ts
+++ b/src/module/actor/sheet/popups/item-transfer-dialog.ts
@@ -117,6 +117,11 @@ class ItemTransferDialog extends fa.api.DialogV2<ItemTransferConfiguration> {
         options: fa.api.HandlebarsRenderOptions,
     ): Promise<void> {
         await super._onRender(context, options);
+        // DialogV2 never attaches a change listener to its form, so wire one up manually to drive `_onChangeForm`.
+        this.element.querySelector("form")?.addEventListener("change", (event) => {
+            const formConfig = this.options.form;
+            if (formConfig) this._onChangeForm(formConfig, event);
+        });
         this.#renderPurchasePrice();
     }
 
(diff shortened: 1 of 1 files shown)


CarlosFdez (approved):
Looks good and tested well enough to me, though I agree that long term dialog isn't really the right choice for this form.

PR #22565, opened 2026-06-23
Fix runes added to handwraps via item alterations not applying to unarmed strikes

### LINKED ISSUE #20574: Fundamental runes, added to Handwraps via ItemAlteration RE, don't apply to unarmed strikes
`CharacterPF2e.prepareAttacks` uses `handwraps?._source.system.runes`, thus working with the item data, and not with RE-modified properties.

jfn4th:
- Closes #20574

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -47,7 +47,7 @@ import { ARMOR_CATEGORIES } from "@item/armor/values.ts";
 import { ActionCost } from "@item/base/data/index.ts";
 import { getPropertyRuneDegreeAdjustments, getPropertyRuneStrikeAdjustments } from "@item/physical/runes.ts";
 import type { EffectAreaShape, ItemType } from "@item/types.ts";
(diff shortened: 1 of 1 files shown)


src/module/actor/character/document.ts
```diff
@@ -1024,7 +1024,9 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
                 w.isEquipped &&
                 w.isInvested,
         );
-        const unarmedRunes = fu.deepClone(handwraps?._source.system.runes) ?? { potency: 0, striking: 0, property: [] };
+        const unarmedRunes: WeaponRuneSource = handwraps
+            ? R.pick(fu.deepClone(handwraps.system.runes), ["potency", "striking", "property"])
```
stwlam:
```suggestion
            ? R.pick(handwraps.system.runes, ["potency", "striking", "property"])
```
`pick` doesn't mutate the passed object

PR #22567, opened 2026-06-25
Replace core inline roll with Shove inline action in Bands of Force

pedrogrullada:
Is this Shove supposed to be affected by bonuses and penalties affecting the character? Who knows, but this at least maintains current functionality while making it a real Shove. This idea could probably be used in other similar items, if any.


TikaelSol:
I would lean to say that yes other modifiers should apply, at least frightened, etc. Let's talk about it in data entry.

PR #22570, opened 2026-06-26
Convert migration summary to AppV2 and Svelte 

jfn4th:
This port also displays a list of all unmigrated actors and items, along with the reason for failure on hover. A plaintext report can also be downloaded, which hopefully will make it easier to share info when troubleshooting. In these examples I am running a migration on a kingmaker world that fails anything beginning with the letter A.  Some small changes are also made to the migration runner in order to prevent the UI from locking up during longer migrations. The runner now yields periodically while processing documents and tokens, so the progress bar keeps advancing instead of freezing partway until the run completes. The runner also now retries a failed save batch one document at a time, so a single failed save no longer discards the entire batch.

Light and dark mode when failure reasons are recorded (after failed migration or remigration)
summary-light

summary-dark 
<p></p>

When we have unmigrated documents but no recorded reasons (migration fails, refresh page, haven't run remigration yet)
summary-no-info
<br>
Hovering over a failure with reason
Screenshot 2026-06-26 at 6 32 38 PM
<br>

And a succesful migration
happy-migration
<br>

Here's a video

https://github.com/user-attachments/assets/735b76a5-f9f5-4ba8-9e60-b95a0d8bf81b

<br>

and an example text file
[pf2e-migration-report.txt](https://github.com/user-attachments/files/29399207/pf2e-migration-report.txt)


src/module/apps/migration-summary/app.svelte
```diff
@@ -0,0 +1,286 @@
...
+    const data = $derived(getState());
+
+    let isRemigrating = $state(false);
+    let okButton = $state<HTMLButtonElement>();
+
+    const hasUnmigrated = $derived(data.unmigrated.actors.length + data.unmigrated.items.length > 0);
+    const hasFailureReasons = $derived(
+        data.unmigrated.actors.some((d) => d.reason) || data.unmigrated.items.some((d) => d.reason),
+    );
+    const showHelpResources = $derived(hasFailureReasons && !isRemigrating);
```
stwlam, line 18: Style note: in general, double line breaks should correspond with comments, and vice versa. E.g.,

```js
const a = 1;
const b = 2;
doStuff(a, b);
```

or

```js
const a = 1;
const b = 2;

// Time to do some stuff.
doStuff(a, b);
```

src/module/apps/migration-summary/app.svelte
```diff
@@ -0,0 +1,286 @@
...
+        </button>
+    </li>
+{/snippet}
+
+<div class="container" aria-busy={isRemigrating}>
+    <h3>
+        {_loc(data.troubleshoot ? "PF2E.Migrations.Summary.Troubleshoot.Hint" : "PF2E.Migrations.Finished", {
+            version: data.systemVersion,
+        })}
+    </h3>
```
stwlam: These heading are coming out over the top. They're a victim of core styling, but maybe we can try `h4`+.

src/module/apps/migration-summary/app.svelte
```diff
@@ -0,0 +1,286 @@
...
+                    {_loc(
+                        hasFailureReasons
+                            ? "PF2E.Migrations.Summary.UnmigratedDocumentsHint"
+                            : "PF2E.Migrations.Summary.UnmigratedDocumentsHintRemigrate",
+                    )}
+                </p>
+                <div class="document-list">
+                    {#if data.unmigrated.actors.length > 0}
+                        {#if data.unmigrated.items.length > 0}
+                            <h5>{_loc("PF2E.Actor.Plural")}</h5>
```
stwlam: `h6`? 🤷

src/module/apps/migration-summary/app.svelte
jfn4th: Default h4 is actually `1.5rem`, though a condensed font, and I've already set this h3 to `1rem`. Rather than picking and choosing each of core's headings just for visuals, I've updated to just follow correct h order (h2, then 3, etc), use default heading font, and size them as we'd like
Screenshot 2026-07-17 at 10 58 05 AM

PR #22581, opened 2026-06-29
Update Know-It-All to account for Archetypes

ammalagonc:


packs/pf2e/feats/class/shared-class-feats/level-8/know-it-all.json
simonward:
```suggestion
            "value": "<p><strong>Bard</strong> When you succeed at a check to @UUID[Compendium.pf2e.actionspf2e.Item.Recall Knowledge], you gain additional information or context. When you critically succeed at a check to Recall Knowledge, you get additional information or context or can ask an additional follow-up question (the GM chooses which).</p>\n<p><strong>Thaumaturge</strong> Having heard and scrutinized every rumor or story in the book, you know for a fact that if you've heard something at all, you've probably heard about it at great length. When you succeed or critically succeed at a check to Recall Knowledge, you can ask an additional follow-up question. At the GM's discretion, you might gain even more additional context than normal.</p>"
```

PR #22587, opened 2026-06-29
Fix two-handed grip toggle on strike rule elements

### LINKED ISSUE #18891: Two handed toggle on Strike RE is non-functional
If you give a two handed trait to a Strike you get the button to switch hands, but it is non-functional. This functionality is needed for Solarian's Solar Weapon, which was errata'd to allow for the two handed d10 trait as a selection. This could be done via toggle, but it would be cleaner to instead use the existing automation.

jfn4th:
- Closes #18891

This persists the grip state of rule-element strikes on the rule element itself (like the modular/versatile toggles already do). Also a little bit of numeric literal type cleanup.

### DIFF
--- a/src/module/actor/army/document.ts
+++ b/src/module/actor/army/document.ts
@@ -10,6 +10,7 @@ import type { CampaignFeaturePF2e } from "@item";
 import type { ItemSourcePF2e } from "@item/base/data/index.ts";
 import type { ItemType } from "@item/types.ts";
 import { ChatMessagePF2e } from "@module/chat-message/document.ts";
+import type { OneToTwo } from "@module/data.ts";
 import { extractDamageDice, extractModifierAdjustments, extractModifiers } from "@module/rules/helpers.ts";
 import { eventToRollParams } from "@module/sheet/helpers.ts";
 import type { TokenDocumentPF2e } from "@scene/index.ts";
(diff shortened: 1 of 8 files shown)


src/module/actor/army/document.ts
```diff
@@ -327,7 +328,7 @@ class ArmyPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nu
             type: "strike",
             glyph: "A",
             variants: [0, 1, 2].map((idx) => {
-                const mapModifier = idx === 0 ? null : createMapModifier(`map${idx as 1 | 2}`);
+                const mapModifier = idx === 0 ? null : createMapModifier(`map${idx as OneToTwo}`);
```
stwlam, line 331: a venerable system type 👍

src/module/rules/rule-element/strike.ts
```diff
@@ -292,15 +299,25 @@ class StrikeRuleElement extends RuleElement<StrikeSchema> {
         return weapon;
     }
 
-    /** Toggle the modular or versatile trait of this strike's weapon */
-    async toggleTrait({ trait, selected }: UpdateToggleParams): Promise<void> {
+    /** Persist changes to this rule element's own source on the parent item */
+    async #updateSource(changes: Partial<StrikeSource>): Promise<void> {
```
stwlam: Confusing name alongside the core synchronous `updateSource`

jfn4th: renamed to `persistSource`

PR #22596, opened 2026-07-01
Accept a leading "+" in modifier dialog inputs on Firefox

### LINKED ISSUE #22597: Adding modifiers in Check Modifiers and Damage Roll dialogs doesn't work well on Firefox
The modifier value input is set to `"type"="number"`, and firefox is stricter than chrome about this fact. The placeholder text shows `+1`, but firefox will treat any number with a + in front of it as just 1. People then come to troubleshoot and learn that they need to leave out the + on firefox. Perhaps the input could be changed so firefox doesn't screw it up (using text works fine for +<number> but I could see that would require more sanitation after the fact), or maybe just a different placeholder would alleviate confusion.

jfn4th:
- Closes #22597

While Chrome sanitizes the input and strips the character, Firefox more strictly adheres to the html spec and treats anything with a leading "+" as invalid, returning "".

### DIFF
--- a/src/module/system/check/dialog.ts
+++ b/src/module/system/check/dialog.ts
@@ -133,10 +133,23 @@ export class CheckModifiersDialog extends fav1.api.Application {
             });
         }
 
+        // Restrict manual entry to a signed integer: a leading +/- followed by digits.
+        const modifierValueInput = html.querySelector<HTMLInputElement>(".add-modifier-value");
+        modifierValueInput?.addEventListener("beforeinput", (event) => {
+            if (!(event instanceof InputEvent) || event.data === null) return;
+            const input = event.currentTarget as HTMLInputElement;
+            const newValue =
+                input.value.slice(0, input.selectionStart ?? 0) +
+                event.data +
+                input.value.slice(input.selectionEnd ?? input.value.length);
+            if (!/^[+-]?\d*$/.test(newValue)) event.preventDefault();
+        });
+
         const addModifierButton = html.querySelector("button.add-modifier");
         addModifierButton?.addEventListener("click", () => {
             const parent = addModifierButton.parentElement as HTMLDivElement;
-            const value = Number(parent.querySelector<HTMLInputElement>(".add-modifier-value")?.value || 1);
+            const valueInput = parent.querySelector<HTMLInputElement>(".add-modifier-value");
+            const value = Number(valueInput?.value || 1);
             const type = String(parent.querySelector<HTMLSelectElement>(".add-modifier-type")?.value);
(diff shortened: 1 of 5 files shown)


src/module/system/check/dialog.ts
```diff
@@ -133,14 +133,21 @@ export class CheckModifiersDialog extends fav1.api.Application {
...
+        modifierValueInput?.addEventListener("beforeinput", (event) => {
+            if (event instanceof InputEvent && event.data?.includes("+")) event.preventDefault();
+        });
+
         const addModifierButton = html.querySelector("button.add-modifier");
         addModifierButton?.addEventListener("click", () => {
             const parent = addModifierButton.parentElement as HTMLDivElement;
-            const value = Number(parent.querySelector<HTMLInputElement>(".add-modifier-value")?.value || 1);
+            const valueInput = parent.querySelector<HTMLInputElement>(".add-modifier-value");
+            const value = Number(valueInput?.value || 1);
```
stwlam, line 152:
```suggestion
            const value = valueInput?.valueAsNumber ?? NaN;
```
if keeping it as a number input: an empty-string `value` will have a `valueAsNumber` of `NaN`

src/module/system/check/dialog.ts
```diff
@@ -133,14 +133,21 @@ export class CheckModifiersDialog extends fav1.api.Application {
...
             const type = String(parent.querySelector<HTMLSelectElement>(".add-modifier-type")?.value);
             let name = String(parent.querySelector<HTMLInputElement>(".add-modifier-name")?.value);
             const errors: string[] = [];
-            if (Number.isNaN(value)) {
+            if (valueInput && !valueInput.checkValidity()) {
```
stwlam:
```suggestion
            if (Number.isNaN(value)) {
```
appropriateness conditional on the above

stwlam (changes requested):
I think we should probably still allow this so as to not upset user habits. Off the top of my head, some options might be:
- Simply fix the placeholder so we're not leading Firefox users astray.
- Switch to a text input and clean it up either as the user types or afterward.
- If it'll fit in the small available space, a range slider of like 1-50.

src/module/system/check/dialog.ts
stwlam: On second thought, it may have been intentional that we treat no input as 1.

jfn4th:
I'll go with a filtered text input, I think a range slider would feel too imprecise.

jfn4th:
Did a quick a11y pass now that it's a text input now and could be a little confusing. Placeholders aren't reliable on screen readers so added some `aria-label`s

stwlam (approved):

PR #22601, opened 2026-07-01
Fix formatting of some PF and SF entries

Cuingamehtar:


packs/pf2e/class-features/spellstrike.json
simonward, line 25: Changes to anything in SoM are frozen for now.

Cuingamehtar: I can drop it. It's likely going to be refreshed in the new book anyway.

PR #22620, opened 2026-07-03
Apply strike adjustments to area fire attacks

### LINKED ISSUE #22618: Cannot use `AdjustStrike`  to add traits to `grenade` base items
Because of the launchers giving `tracking` traits to grenades, i wanted to make a `RollOption` + `AdjustStrike` combo to handle it, but it appears that grenades do not get the traits no matter what i tried.

Simply add the following rule to an actor and see nothing changing for grenades while other weapons do:

```json
{"key":"AdjustStrike","mode":"add","property":"weapon-traits","value":"tracking-1"}
```

jfn4th:
- Closes #22618

Also some `private` -> `#` cleanup

### DIFF
--- a/src/module/actor/character/document.ts
+++ b/src/module/actor/character/document.ts
@@ -58,6 +58,7 @@ import {
     extractRollSubstitutions,
     extractRollTwice,
 } from "@module/rules/helpers.ts";
+import type { StrikeAdjustment } from "@module/rules/synthetics.ts";
 import { eventToRollParams } from "@module/sheet/helpers.ts";
 import { TokenDocumentPF2e } from "@scene/index.ts";
 import { Check, CheckCheckContext, CheckRoll } from "@system/check/index.ts";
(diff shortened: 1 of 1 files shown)


src/module/actor/character/document.ts
```diff
@@ -462,7 +463,7 @@ class CharacterPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e
         }
 
         super.prepareDataFromItems();
-        this.prepareBuildData();
+        this.#prepareBuildData();
```
stwlam, line 466: I don't think we can do this: try cloning a PC (`actor.clone()`).

jfn4th: Tried both `actor.clone()` and `getContextualClone`, both appear to work just fine. Core's clone calls `new this.constructor(...)`, not something like `Object.create()`, so `#` should be reachable.

stwlam: Try creating a new actor, then.

stwlam:
Technical js things aside:
- The AdjustStrike rule element does precisely that. It's not reasonable to expect to also support adjusting area fire, spellcasting, etc., actions.
- AdjustStrike is being phased out in favor of ItemAlteration, which the issue reporter notes works just fine.

PR #22623, opened 2026-07-03
Change Stellar Rush to give bonus to all speeds

Cuingamehtar:
Due to `traversal` trait.


simonward:
~~Not sure about this one. I read the traversal as applying to the Stride twice part, not the speed bonus.~~

PR #22633, opened 2026-07-05
Add header, traits, and publication options to feat embeds

Cuingamehtar:
Includes config options:
- header: when `true`, adds UUID link with feat level (no custom link label as that doesn't happen often)
- traits: when `true`, adds non-common rarity trait followed by alphabetically sorted list of traits
- publication: when `true`, adds publication source label
- trailingLink: when `false`, removes a link at the end of the description - intended for dedications that link to the current journal page.

Example
image

Leaving it as a draft for now.

### DIFF
--- a/src/module/item/feat/document.ts
+++ b/src/module/item/feat/document.ts
@@ -379,7 +379,7 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
                   (config.hr === false ? "" : "<hr>")
                 : "",
             config.trailingLink === false
-                ? this.description.replace(/<p>@UUID\[[^\]]+\](?:\{[^}]+\})<\/p>$/, "")
+                ? this.description.replace(/<p>@UUID\[[^\]]+\](?:\{[^}]+\})?<\/p>$/, "")
                 : this.description,
             publication,
         ].join("");
--- a/src/module/item/feat/document.ts
+++ b/src/module/item/feat/document.ts
@@ -343,14 +343,46 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
         return rollOptions;
     }
 
-    protected override embedHTMLString(config: DocumentHTMLEmbedConfig & { hr?: boolean }): string {
-        const list = this.system.prerequisites?.value?.map((item) => item.value).join(", ") ?? "";
-        return (
-            (list
-                ? `<p><strong>${_loc("PF2E.FeatPrereqLabel")}</strong> ${list}</p>` +
+    protected override embedHTMLString(
+        config: DocumentHTMLEmbedConfig & {
+            hr?: boolean;
+            traits?: boolean;
+            publication?: boolean;
+            header?: boolean;
+            trailingLink?: boolean;
+        },
+    ): string {
+        const header = config.header
+            ? `<h2>@UUID[${this.uuid}] <span style="float:right">${_loc("PF2E.Item.Feat.LevelN", { level: this.level })}</span></h2>`
+            : "";
+
+        const traits = config.traits
+            ? this.system.traits.value
+                  .map((t) => _loc(CONFIG.PF2E.featTraits[t]))
+                  .sort((a, b) => a.localeCompare(b))
+                  .reduce(
+                      (allTraits, t) => allTraits + `<li class="tag traits">${t}</li>`,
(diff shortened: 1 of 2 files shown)


Cuingamehtar:
I don't think there's much to add here.

`trailingLink` could have a better name - I'm open to suggestions.
An argument could be made to have prerequisites as an option, but it will affect existing embeds.

Cuingamehtar:
And they still show up in the journal navigation, which is nice.

src/module/item/feat/document.ts
```diff
@@ -343,13 +343,55 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+            ? this.system.traits.value
+                  .map((t) => _loc(CONFIG.PF2E.featTraits[t]))
+                  .sort((a, b) => a.localeCompare(b))
+                  .reduce(
+                      (allTraits, t) => allTraits + `<li class="tag traits">${t}</li>`,
+                      this.system.traits.rarity !== "common"
+                          ? `<li class="tag rarity ${this.system.traits.rarity}">${_loc(CONFIG.PF2E.rarityTraits[this.system.traits.rarity])}</li>`
+                          : "",
+                  )
+            : "";
```
jfn4th, line 378:
```suggestion
        const rarity = this.system.traits.rarity;
        const traits = config.traits
            ? [
                  rarity !== "common"
                      ? `<li class="tag rarity ${rarity}">${_loc(CONFIG.PF2E.rarityTraits[rarity])}</li>`
                      : "",
                  ...this.traitChatData().map((t) => `<li class="tag traits">${t.label}</li>`),
              ].join("")
            : "";
```
`traitChatData` already does the map and sort.

src/module/item/feat/document.ts
```diff
@@ -343,13 +343,55 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
 
-    protected override embedHTMLString(config: DocumentHTMLEmbedConfig & { hr?: boolean }): string {
-        const list = this.system.prerequisites?.value?.map((item) => item.value).join(", ") ?? "";
+    protected override embedHTMLString(
+        config: DocumentHTMLEmbedConfig & {
+            hr?: boolean;
+            traits?: boolean;
+            publication?: boolean;
+            header?: boolean;
+            trailingLink?: boolean;
```
jfn4th: let's call this `journalLink`

src/module/item/feat/document.ts
```diff
@@ -343,13 +343,55 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+        },
+    ): string {
+        // Add header with Item link and feat level
+        const header = config.header
+            ? `<h2>@UUID[${this.uuid}] <span style="float:right">${_loc("PF2E.Item.Feat.LevelN", { level: this.level })}</span></h2>`
```
jfn4th: Since this is code now, use a class on the heading `h2` and the source `p` you have later, rather than an inline style. Set styles in the `journal-styling` mixin within `_typography.scss`. Also, two embeds in a row are causing the first one's source to float up into the second one's h2. Instead of a float, you can flex the heading with space between and `text-align: right` on the source `p` to get what you want.

jfn4th (commented):
Just a few little things. This branch has also been sitting a little bit and could use a rebase.

Cuingamehtar:
image
Two consecutive embeds after the changes

src/module/item/feat/document.ts
```diff
@@ -349,13 +349,53 @@ class FeatPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Item
...
+            hr?: boolean;
+            traits?: boolean;
+            publication?: boolean;
+            header?: boolean;
+            journalLink?: boolean;
+        },
+    ): string {
+        // Add header with Item link and feat level
+        const header = config.header
+            ? `<h2 class="embed heading"><div>@UUID[${this.uuid}]</div><div> </div><div>${_loc("PF2E.Item.Feat.LevelN", { level: this.level })}</div></h2>`
```
jfn4th:
```suggestion
            ? `<h2 class="embed heading"><span>@UUID[${this.uuid}]</span> <span>${_loc("PF2E.Item.Feat.LevelN", { level: this.level })}</span></h2>`
```
While it works, `div` within a heading isn't supposed to be valid html.

PR #22635, opened 2026-07-05
Skip redundant modifier work in Statistic check/dc views

jfn4th:
The check/dc views on `Statistic` re-clone and re-stack all of the parent's modifiers on construction even when there's nothing view specific to add, which is the common case. Now the views adopt clones of the parent's already-resolved modifiers when no synthetics, adjustments, or inline modifiers target the view-only domains. Also stops `dc.value` from re-stacking on every read and makes `getTraceData` build only the side it reports.

In my testing this typically cuts ~10% of the time spent during an `actor.reset()` call/data prep, with the absolute time savings growing as statistics accumulate more modifiers at higher levels. I've measured up to ~30% for item-heavy characters outside of combat/without a bunch of effects on them.

### DIFF
--- a/src/module/system/statistic/statistic.ts
+++ b/src/module/system/statistic/statistic.ts
@@ -127,6 +127,20 @@ class Statistic<TActor extends ActorPF2e = ActorPF2e> extends BaseStatistic<TAct
         return this.check.mod;
     }
 
+    /**
+     * Whether a check/dc view's modifiers would equal this statistic's own: true when the view has no inline
+     * modifiers, no synthetics or adjustments target its view-only domains, and no modifier carries a predicate or
+     * attached adjustments (which the skipped view-level `StatisticModifier` run would retest and apply against the
+     * view's fuller roll-option set). When true the view can adopt clones of the already-resolved modifiers and skip
+     * the extract/stack work.
+     */
+    viewMatchesParent(viewDomains: readonly string[], inlineModifiers: readonly Modifier[] | undefined): boolean {
+        if ((inlineModifiers ?? []).length > 0) return false;
+        if (this.modifiers.some((m) => m.predicate.length > 0 || m.adjustments.length > 0)) return false;
+        const { modifiers, modifierAdjustments } = this.actor.synthetics;
+        return !viewDomains.some((d) => (modifiers[d]?.length ?? 0) > 0 || (modifierAdjustments[d]?.length ?? 0) > 0);
+    }
+
     override createRollOptions(domains = this.domains, args: RollOptionConfig = {}): Set<string> {
         const { item, extraRollOptions, target } = args;
         const origin = args.origin ?? (item?.actor && item.actor.uuid !== this.actor.uuid ? item.actor : null);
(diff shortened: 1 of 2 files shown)


src/module/system/statistic/statistic.ts
```diff
@@ -731,14 +764,9 @@ class StatisticDifficultyClass<TParent extends Statistic = Statistic> {
...
-            10 +
-            new StatisticModifier(
-                "",
-                this.modifiers.map((m) => m.clone()),
-                this.options,
-            ).totalModifier
-        );
+        // Sum currently-enabled modifiers. Avoids re-cloning and re-stacking on every read while remaining correct
+        // if a caller mutates `this.modifiers` after construction (as ArmorStatistic does to add armor bonuses).
+        return 10 + R.sumBy(this.modifiers, (m) => (m.enabled ? m.modifier : 0));
```
CarlosFdez, line 772: Out of curiousity, how well does this handle stacking modifiers and deduplication?

CarlosFdez: I'm seeing that there's also modifier adjustments that StatisticModifier does as well (which was probably the reason for the cloning).

jfn4th: I believe dedup and stacking are already handled by `BaseStatistic`'s own `StatisticModifier` run, and this adds nothing new on top, so those come out identical. But I'll have to double check `StatisticModifier` adjustments, thanks for mentioning.

jfn4th: @CarlosFdez okay, adjustments from the statistic's own domains are only applied by the view-level StatisticModifier pass and were being skipped. Also found predicated modifiers weren't re-tested against the fuller roll option set either.  Bail on both now. At a quick glance, average savings are closer to 8% now.

CarlosFdez: The other thing to worry about is deferred modifiers. Sometimes a modifier will not be available for extraction until later, or resolve into a different value later at different steps of preparation. I don't know if those are handled here, but I suspect if you want ultimate safety, the best way to do it might be to freeze extracted modifiers when data preparation has completed. 

But its been a long time since I touched this and I'm not 100% aware of what you changed, just something for you to keep in mind regarding why things are the way they are.

EDIT: I think most statistics are received post data prep, so you'd still find savings even if you only limit optimizations to anything done post-prep I think.

jfn4th: It looks like deferred modifiers are put in `synthetics.modifiers[selector]` before statistics are built. `viewMatchesParent` checks if there is anything registered for the view, so any deferred modifier, even resolving to `null`, fails and goes down the normal extraction path.

If I'm reading the freeze idea right, that fits a world where prepared statistics stick around after prep and later reads reuse them, frozen so nothing drifts underneath. This shortcut lives entirely inside a single prep, everything is still rebuilt from scratch on reset, so there's nothing that persists long enough to freeze. If statistics ever get memoized across resets, freezing the cached lists is the guard I'd go for.

On the edit, from what I can tell, a lot of the view construction actually happens during prep, at least for creatures. `prepareDerivedData` looks to store `getTraceData` results for perception, saves, skills, class DCs, AC, and initiative on both PCs and NPCs, so limiting the shortcut to post-prep might give up most of the savings. Unless I'm missing something, safety should be the same either way, since the guard runs at view construction, the same moment the extraction it skips would run, so both paths see identical synthetics state mid-prep or after.

CarlosFdez: What I think you're probably missing (if you are missing something) is that certain AEs or adjust modifiers or predicate results may not have finished yet because not everything runs at the same time. After data prep nothing is going to change, but during data prep the value of any individual statistic is up for possible changing.

Deferred modifiers were not added for no reason, they were added to solve a specific use case. But it has been many years, so I can't give you a breaky example off the top of my head.

CarlosFdez: I haven't dug through the code though and I'm operating on very old memory which is unreliable, so if deferred modifiers will lead to the same result even with data changes to other stuff, and this has nothing to do with the deferred modifier results, then carry on I suppose.

PR #22648, opened 2026-07-07
Fix Draddeth's Edge potency bonus

Longstrider:
Not sure if it's relevant though, since `dying-recovery-check` bonuses aren't working


packs/pf2e/equipment/draddeths-edge.json
```diff
@@ -57,7 +57,7 @@
                 "key": "FlatModifier",
                 "selector": "dying-recovery-check",
                 "type": "item",
-                "value": "{item|system.runes.potency.value}"
+                "value": "{item|system.runes.potency}"
```
simonward:
```suggestion
                "value": "@item.system.runes.potency"
```

PR #22666, opened 2026-07-09
Add universal Ancestry trait to feats

### LINKED ISSUE #22660: Cannot select `Unconventional Weaponry` feat using `Ancestral Paragon` due to lack opf trait
We cannot select `Unconventional Weaponry` as an ancestry feat because it lacks traits, while we can work around it when selecting an actual ancestry feat (by removing the ancestry trait in the filter), this completely prevents use to select it using `Ancestral Paragon`.

iDantar:
- Compliments https://github.com/foundryvtt/pf2e/pull/23227
- Together they close https://github.com/foundryvtt/pf2e/issues/22660

I added all but one feats mentioned here https://2e.aonsrd.com/rules/1290#:~:text=the%20novian%20trait.-,Standardized%20Ancestry%20Feats,-Feat 
[Internal Compartment](https://2e.aonsrd.com/feats/1053-internal-compartment) is taken from PF2, so not sure if it should receive the same treatment


src/module/apps/compendium-browser/tabs/feat.ts
```diff
@@ -101,11 +99,6 @@ export class CompendiumBrowserFeatTab extends CompendiumBrowserTab {
                     this.preparePublicationSource(pubSource, publications),
                 ];
 
-                // Tag ancestry items without an ancestry trait
-                if (category === "ancestry" && !traits.some((t) => t in this.#creatureTraits)) {
-                    options.push("trait:ancestry:universal");
-                }
```
TikaelSol, line 107: There's still some ancestry feats without the ancestry trait that are universal, the reincarnation feats for example.

static/lang/en.json
```diff
@@ -4513,6 +4513,7 @@
         "TraitAmphibious": "Amphibious",
         "TraitAnadi": "Anadi",
         "TraitAnalog": "Analog",
+        "TraitAncestry": "Ancestry",
```
TikaelSol, line 4546: Should add a trait description. From the book:

> You can select a standardized ancestry feat in place of an ancestry feat whenever you gain an ancestry feat if you meet its prerequisites and if it's listed among the standardized ancestry feat options for your ancestry. These feats have the uncommon trait as a reminder that they are only available if an ancestry's entry specifically lists the standardized ancestry feat as one that is available for that ancestry. The ancestry trait represents whatever ancestry your character is.

iDantar: Added, thanks

src/module/apps/compendium-browser/tabs/feat.ts
iDantar: Good call, returned and updated to match the new trait

iDantar:
I'm still not sure about the https://2e.aonsrd.com/feats/1053-internal-compartment - could you please tell if it also should have a standardized trait?

TikaelSol (approved):

stwlam:
There's no reason to combine these data and codebase updates into a single PR. Go ahead and split it up, leaving the data updates in this one.

PR #22669, opened 2026-07-10
Always use action image for unusable actions in character sheet

CarlosFdez:
If an action on the character sheet has no use button, it will now always show the action cost image in the character sheet. Before if someone changed the icon of an action, it'd be impossible to see the action cost.

## Before

image

## After

image


stwlam:
Hm, who's this someone? Do system compendiums have any examples?

CarlosFdez:
Nothing in system compendiums. More a thing that could happen if a player or gm tries to get smart.

PR #22694, opened 2026-07-15
Convert SpellPreparation to ApplicationV2

jfn4th:
- Rewrite the spell preparation sheet as a Svelte app, built on a new reusable spell-list component family (meant to later back spellcasting lists on actor sheets)
- Prepare/unprepare directly from the list: click to prepare (rank menu when several ranks are eligible), right-click or shift-click to unprepare, cantrips show a check instead of a count. Much more accessible and friendly if the window is popped out.
- Text search over names and traits, built on a new generic TextSearch class (same term processing as the compendium browser) intended to replace the ad-hoc searches in the formula picker, trade dialog, and elsewhere
- View-only column sorting: name, defense, range, and prepared count via the group usage indicator
- Drag reordering with an insertion indicator: drops from the compendium or other actors land at the drop position
- Keyboard and screen reader support: arrow keys walk the rows, rank groups are headings, menus are arrow-navigable, icon buttons and sort direction have accessible names
- Fix dropping a spell from a duplicated actor (or an unlinked token of the same actor) onto another's spellcasting entry silently doing nothing (changes in `collection.ts`)

New general components (`src/module/sheet/components/`):
- `PopoverMenu`: an anchored menu rendered in the top layer, so nothing can clip it (scrolling lists included). CSS anchor positioning with a flip fallback, arrow-key navigation, Escape/click-outside dismissal, and it keeps core keybindings (canvas pan etc.) out of the way while open. Useful for context menus, dropdowns, and action menus generally.
- `InlineIconButton`: a thin wrapper over core's `button.inline-control.icon`, giving sheets real focusable buttons with core's styling in place of the traditional `<a>` icon links, which keyboards can't reach at all

Also reworked `ItemSummary`: the description fetch is now keyed on the item's modification time, so summaries load lazily on first expand, re-expands don't refetch, and editing the item while its summary is open refreshes it live. A slow stale response can no longer overwrite a newer one, and a failed load renders nothing instead of rejecting unhandled. Gains an `id` prop so disclosure triggers can point `aria-controls` at it.

Dark:
dark

Light:
light

Rank menu:
menu

Prepared count breakdown:
badge-tooltip

Search:
search

Flexible collection:
flexible


stwlam:
The tips are a little confusing, I guess:



Maybe something like the inventory quantity adjuster would be more appropriate?

jfn4th:
> Maybe something like the inventory quantity adjuster would be more appropriate?

That does sound like a better idea! It would let me ditch the right click and shift click stuff too.

jfn4th:
Updated!
badge-tooltip

menu

PR #22700, opened 2026-07-17
Fix quantum reaver and reaction breacher stats

Longstrider:


packs/sf2e/equipment/weapons/flamethrower.json
```diff
@@ -57,7 +57,7 @@
         "quantity": 1,
         "range": 15,
         "reload": {
-            "value": "1"
+            "value": "2"
```
simonward: Do you have a source for this? My PDF shows reload 1, and I don't see it in the errata.

Longstrider: Sorry, you're right - it's my bad. Not sure where I got this, maybe confused it with a blockthrower 🤦‍♂️

PR #22706, opened 2026-07-18
Convert inline confirmation dialogs (sell-item, sell-all-treasure, delete-spell-variant) to DialogV2

jfn4th:


src/module/actor/sheet/base.ts
```diff
@@ -779,22 +779,14 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
 
                 const content = document.createElement("p");
                 content.innerText = _loc("PF2E.SellItemQuestion", { item: item.name });
```
stwlam:
```suggestion
                content.textContent = _loc("PF2E.SellItemQuestion", { item: item.name });
```
Outside the changes, of course, but this is something we should prefer generally.

src/module/actor/sheet/base.ts
```diff
@@ -1443,26 +1435,12 @@ abstract class ActorSheetPF2e<TActor extends ActorPF2e> extends fav1.sheets.Acto
...
-                    icon: fa.fields.createFontAwesomeIcon("times").outerHTML,
-                    label: "Cancel",
-                },
-            },
-            default: "cancel",
-        }).render(true);
+        const confirmed = await foundry.applications.api.DialogV2.confirm({
+            id: "sell-all-treasure-{id}",
+            window: { title: "PF2E.SellAllTreasureTitle" },
+            content: `<p class="note">${_loc("PF2E.SellAllTreasureQuestion")}?</p>`,
```
stwlam: Can that question mark move to the localization string?

PR #22718, opened 2026-07-20
Implement and improve Plant Form effects, brush up entu feats and add Bursting Surge effect

Longstrider:


packs/sf2e/feats/ancestry/entu-colony/level-13/fungal-evolution.json
Longstrider, line 1: It's probably not the most elegant solution and the feat's writing is quite ambiguous, so I'm open to suggestions.

PR #22740, opened 2026-07-25
Fix cost and duration for Connective Current

rectulo:


simonward:
Consider adding missing duration

packs/pf2e/spells/spells/rank-1/connective-current.json
simonward:
```suggestion
            "value": "1 minute"
```

PR #22742, opened 2026-07-25
Fix Compendium Browser sort by price

Cuingamehtar:
~~`priceInCopper` is not a proper field, and `price` should be an actual Coins object, but as the type system doesn't check it, I've added a null check in case it isn't there for some reason.~~

Added `priceInCopper` field instead so it doesn't have to recalculate values for every comparison

### DIFF
--- a/src/module/apps/compendium-browser/tabs/equipment.ts
+++ b/src/module/apps/compendium-browser/tabs/equipment.ts
@@ -106,7 +106,7 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
                         uuid: itemData.uuid,
                         level: itemData.system.level?.value ?? 0,
                         price: priceCoins,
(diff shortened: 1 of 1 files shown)


src/module/apps/compendium-browser/tabs/equipment.ts
```diff
@@ -106,6 +106,7 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
                         uuid: itemData.uuid,
                         level: itemData.system.level?.value ?? 0,
                         price: priceCoins,
+                        priceInCopper: priceCoins.copperValue,
```
jfn4th:
```suggestion
                        priceInCopper: coinValue,
```

looks like we already grab `priceCoins.copperValue` a few lines up.

PR #22766, opened 2026-07-28
fix attaching loudener to weapons

### LINKED ISSUE #22291: Loudener isn't able to be attached to a weapon
The Loudener weapon module shows in the item that it installs on a weapon like it should but unlike the other weapon modules, there isn't an option for it to be attached to a weapon once it's in the inventory. 

SF2 system:
Loudener (Commercial) Compendium.sf2e.equipment.Item.tMVTprjg28EEYqz9 
Loudener (Tactical) Compendium.sf2e.equipment.Item.OyqLJ1zQy8PpnSHR

SF2A nachronism module for PF2 system:
Loudener (Commercial) Compendium.sf2e-anachronism.equipment.Item.tMVTprjg28EEYqz9
Loudener (Tactical)  Compendium.sf2e-anachronism.equipment.Item.OyqLJ1zQy8PpnSHR

steve148:
Closes #22291

@CarlosFdez was right, the issue is the `installed-on` instead of `installed-in` for the usage value. Below are links to a before and after of this change. I didn't end up changing the `TraitInstalledOnAWeaponWithoutASilencer` trait name it doesn't have a functional impact, but can change that to avoid confusion later on (all installed-in values start with `TraitInstalledIn`).

Before 

https://github.com/user-attachments/assets/4bff9997-03d1-4155-8bd6-981a4dd15863

After

https://github.com/user-attachments/assets/e3dbb7fd-471d-45ae-a09f-0cfc5116b393

### DIFF
--- a/src/module/item/physical/usage.ts
+++ b/src/module/item/physical/usage.ts
@@ -67,8 +67,8 @@ function getUsageDetails(usage: string): UsageDetails {
         return { value: usage, type: "attached", where };
     }
 
-    if (usage.startsWith("installed-in") || usage.startsWith("installed-on")) {
-        const where = usage.replace(/^installed-in-/, "").replace(/^installed-on-/, "");
+    if (/^installed-(?:i|o)n-/.test(usage)) {
+        const where = usage.replace(/^installed-(?:i|o)n-/, "");
         return { value: usage, type: "installed", where };
     }
 
--- a/src/module/item/physical/usage.ts
+++ b/src/module/item/physical/usage.ts
@@ -67,8 +67,8 @@ function getUsageDetails(usage: string): UsageDetails {
         return { value: usage, type: "attached", where };
     }
 
-    if (usage.startsWith("installed-in")) {
-        const where = usage.replace(/^installed-in-/, "");
+    if (usage.startsWith("installed-in") || usage.startsWith("installed-on")) {
+        const where = usage.replace(/^installed-in-/, "").replace(/^installed-on-/, "");
         return { value: usage, type: "installed", where };
     }
 


src/scripts/config/usage.ts
```diff
@@ -7,7 +7,7 @@ const WEAPON_UPGRADES = {
     "installed-in-one-handed-weapon-grip": "PF2E.TraitInstalledInOneHandedWeaponGrip",
     "installed-in-ranged-weapon-without-a-loudener": "PF2E.TraitInstalledInRangedWeaponWithoutALoudener",
     "installed-in-two-handed-weapon": "PF2E.TraitInstalledInTwoHandedWeapon",
-    "installed-on-a-weapon-without-a-silencer": "PF2E.TraitInstalledOnAWeaponWithoutASilencer",
+    "installed-in-a-weapon-without-a-silencer": "PF2E.TraitInstalledOnAWeaponWithoutASilencer",
```
TikaelSol: Since we're changing the usage slug we'd need to migrate the slugs for existing equipment

steve148: That makes sense to me. I haven't added a migration here yet, but I'll take a peek at what's there and figure it out. Will try to get this done later today.

CarlosFdez: Could expand the match in the code to installed-in and installed-on? This is a thing sf2e does in the book and they're not gonna stop doing it.

TikaelSol: Would be easier to expand the match, yeah.

steve148: Commit amended, now it's just a widening of the usage condition.

src/module/item/physical/usage.ts
```diff
@@ -67,8 +67,8 @@ function getUsageDetails(usage: string): UsageDetails {
         return { value: usage, type: "attached", where };
     }
 
-    if (usage.startsWith("installed-in")) {
-        const where = usage.replace(/^installed-in-/, "");
+    if (usage.startsWith("installed-in") || usage.startsWith("installed-on")) {
+        const where = usage.replace(/^installed-in-/, "").replace(/^installed-on-/, "");
```
CarlosFdez, line 71:
```suggestion
    if (/^installed-(?:i|o)n-/.test(usage)) {
        const where = usage.replace(/^installed-(?:i|o)n-/, "");
```

Briefer is better

PR #22767, opened 2026-07-28
Add rule for Guerrilla Weaponry weapon die override

iDantar:
P.S. had to add a hacky rule for DamageDice, otherwise striking runes wouldn't apply


packs/pf2e/feats/archetype/guerrilla/level-4/guerrilla-weaponry.json
simonward, line 1: Consider `GrantItem` for the Guerrilla Assault action

iDantar: Good call, added

PR #22798, opened 2026-07-30
Add missing traditions from Sift the Sphere

ammalagonc:


packs/sf2e/spells/spells/rank-2/sift-the-sphere.json
simonward:
```suggestion
            "value": "10 minutes"
```

PR #22817, opened 2026-07-31
Add new Impossible Magic property runes

nythz:
basic additions for the property runes dropdown lists on items details
moved greaterQuenching to be in the correct alphabetical order position in the code


src/module/item/physical/runes.ts
```diff
@@ -490,6 +490,14 @@ export const ARMOR_PROPERTY_RUNES: { [T in ArmorPropertyRuneType]: ArmorProperty
         slug: "coldResistant",
         traits: ["magical"],
     },
+    convincing: {
+        name: "PF2E.ArmorPropertyRuneConvincing",
```
TikaelSol, line 494: We'll want to add the localizations to static/lang/en.json

nythz: Thanks for the pointer, On it.

PR #22826, opened 2026-07-31
Update publication source for Command a Thrall

nobbyfix:


packs/pf2e/actions/class/necromancer/command-a-thrall.json
simonward:
```suggestion
        "category": "interaction",
```

PR #22839, opened 2026-08-01
Condense Potion of Stable Form Effects

ammalagonc:


packs/pf2e/equipment-effects/effect-potion-of-stable-form.json
simonward: Consider linking to both potions and either removing the bonus number or listing both.

PR #22859, opened 2026-08-02
Remove class-specific Spell Repertoires and have the base description use localized strings

Cuingamehtar:
No need to redirect - they already are being redirected.

This just simplifies the Repertoire Item to reduce text duplication. The description is already being overridden by its RE anyway.


TikaelSol:
Were these linked anywhere on the classes? Probably need a redirect

Cuingamehtar:
As I've noted, those are already in the redirects file
I had to update the iconics because it didn't let me build otherwise

PR #22892, opened 2026-08-05
Add roll option to differentiate actual casting from send-to-chat

reonZ:
This PR adds an extra flag to spell messages indicating that the spell was actively `cast`, to differentiate it from the other mean of displaying a spell message: "send to chat".

We have had an eruption of automation modules in the last few years and a lot of them would benefit from such a small addition.

Notes:
- The `actualCast` is a placeholder as i didn't really know how it should be named.
- I wasn't sure if it was needed to add it to the DOM so i abstained


src/module/chat-message/data.ts
```diff
@@ -31,7 +31,7 @@ type ChatMessageFlagsPF2e = ChatMessageFlags & {
         damageRoll?: DamageRollFlag;
         context?: ChatContextFlag;
         origin?: ItemOriginFlag | null;
-        casting?: { id: string; tradition: MagicTradition; embeddedSpell?: SpellSource } | null;
+        casting?: { actualCast: boolean; id: string; tradition: MagicTradition; embeddedSpell?: SpellSource } | null;
```
jfn4th:
```suggestion
        casting?: { actualCast?: boolean; id: string; tradition: MagicTradition; embeddedSpell?: SpellSource } | null;
```

Existing messages won't have the flag. As for the name, maybe `spellCast`?

reonZ: Not making it optional was definitely an oversight.

As for the name,  `spellCast` doesn't feel right to me but it doesn't matter in the end, i will settle for anything.

jfn4th:
I talked with Shark on this one, we'd rather this be a roll option than a flag. The catch is that `flags.context` is only set when a spell goes against a defense, so this'll have to be added to `flags.origin.rollOptions` in `SpellPF2e#toMessage`. The roll option should be a constant defined in `item/spell/values.ts` as thus
```ts
const CAST_A_SPELL_OPTION = "origin:action:slug:cast-a-spell";
```

Also it looks like your last commit was pushed during the github actions outage and checks didn't run, should be good once new commits are pushed.

jfn4th (changes requested):

reonZ:
At least it solves the problem of having to find an appropriate variable name :3

src/module/item/spell/document.ts
```diff
@@ -793,6 +794,11 @@ class SpellPF2e<TParent extends ActorPF2e | null = ActorPF2e | null> extends Ite
                     messageMode: mode,
                 };
             }
+
+            // We actually cast the spell and not simply send it to chat
+            if (actualCast && flags.origin?.rollOptions) {
```
reonZ: Wasn't sure how to handle typescript telling me that `flags.origin.rollOptions` could be undefined so i went for the simple check.

reonZ:
> The catch is that `flags.context` is only set when a spell goes against a defense

Also, while i already did the roll-option, something is confusing here, the `actualCast` flag was in the `flags.casting` and not `flags.context` which is independent of having a defense. So should we still go the roll-option route or was it just a misunderstanding ?

src/module/item/spell/document.ts
jfn4th: Let's set this outside of `if (spellcasting?.statistic)` so it gets added to rituals too, in case those do something someday. Right after `const flags` should be good.

src/module/chat-message/listeners/cards.ts
```diff
@@ -143,13 +144,18 @@ class ChatCards {
                     return;
                 }
                 case "spell-variant": {
+                    const actualCast = message.flags[SYSTEM_ID].origin?.rollOptions?.includes(CAST_A_SPELL_OPTION);
```
jfn4th:
```suggestion
                    const actualCast = !!message.flags[SYSTEM_ID].origin?.rollOptions?.includes(CAST_A_SPELL_OPTION);
```

jfn4th (changes requested):
Sorry, I could have worded that better. Normally we'd have a roll option added to `flags.context.options`, as `self:action:slug:cast-a-spell`. That wouldn't work here because `flags.context.options` isn't set on spells that don't go against defense, hence adding to `origin.rollOptions`.

The original placement in `flags.casting` wasn't wrong per say, but roll options are a more standard system way of doing this, plus we can put them on rituals too since `flags.casting` only is added when there's a statistic.

Should be good to merge after these last tweaks

src/module/chat-message/listeners/cards.ts
reonZ: i thought about placing  it outside the if  statistic right after i pushed the change but wasn't sure ti would change anything.

There isn't really a need to cast `actualCast` as a boolean here since the property it is used for accepts undefined, but if this is something that needs to be done then fine.

jfn4th (approved):

PR #22894, opened 2026-08-05
Allow inline template distance to use resolve

### LINKED ISSUE #22376: Make the `distance` param resolvable in the `@Template`
There is at least one use case that i came up with that would require this, the `Rattle the Earth` action that has an emanation based on the radius of the kineticist aura. So it needs the following:

```
@Template[emanation|distance:resolve(@actor.flags.pf2e.kineticist.auraRadius)]
```

reonZ:
Make the `distance` argument of `@Template` be resolvable.

Closes #22376

### DIFF
--- a/src/module/system/text-editor.ts
+++ b/src/module/system/text-editor.ts
@@ -209,13 +209,13 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
         if (data.length < 4) return null;
         const rollData = resolveRollData(options.rollData);
         const item = rollData.item ?? null;
+        const actor = rollData.actor ?? item?.actor ?? null;
         const [_match, inlineType, paramString, inlineLabel] = data;
 
         switch (inlineType) {
             case "act":
                 return this.#createAction(data.groups?.slug ?? "", data.groups?.options ?? "", data.groups?.label);
             case "Check": {
-                const actor = rollData.actor ?? item?.actor ?? null;
                 return this.#createCheck({ paramString, inlineLabel, item, actor });
             }
             case "Damage":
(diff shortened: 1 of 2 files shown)


src/module/system/text-editor.ts
```diff
@@ -307,9 +306,14 @@ class TextEditorPF2e extends foundry.applications.ux.TextEditor {
         params.traits ||= item?.system.traits.value?.toString() ?? null;
         params.itemUuid ||= item?.uuid ?? null;
 
+        const distance =
+            params.distance.startsWith("resolve") && (item || actor)
+                ? resolveValue(actor, item, params.distance) || 5
+                : Number(params.distance) || 5;
```
jfn4th: This would always render as 5ft in the compendium. Existing actor resolves show 0. Also would be good to keep the old `!Number(params.distance)` guard for literal values.

```suggestion
        const isResolvable = params.distance.startsWith("resolve");
        if (!isResolvable && !Number(params.distance)) {
            ui.notifications.error("PF2E.InlineTemplateErrors.DistanceNoNumber", {
                format: { distance: params.distance },
            });
            return null;
        }
        const distance = isResolvable ? resolveValue(actor, item, params.distance) : Number(params.distance);
```

reonZ: Maybe we should change the error message then, because this would write `Error in @Template: dimension 0 is not a number` with distance of `0`.

reonZ: Or maybe `0` should actually be valid and the check should be for `isNan`

jfn4th: `Error in @Template: distance {distance} is not valid` sounds appropriate

PR #22895, opened 2026-08-05
Do not consume charges if trick-magic-item check fails or is canceled…

### LINKED ISSUE #17308: Make consumables not being consumed right away when using `Trick Magic Item`
Right now, scrolls and wands are used no matter what when clicking on the `Use` button, this cause issues for two situations:

- The user simply cancel the `Trick Magic Item` popup
- The user fails or critically fails the `Trick Magic Item` roll check (this prevents the use of the item and therefore shouldn't consume it)

For wands, we can just replenish its charge but for scrolls, if you happen to only have one of that type, you have to re-add it to your character.

reonZ:
Right now, using `Trick Magic Item` consumes a charge as soon as the button is pressed, even if the user cancel the check all together.

Also, as per the feat's description, you shouldn't be able to use the item at all if you fail the check, which means that is should not be consumed either.

closes #17308

### DIFF
--- a/src/module/actor/sheet/trick-magic-item-popup.ts
+++ b/src/module/actor/sheet/trick-magic-item-popup.ts
@@ -5,7 +5,7 @@ import { TRICK_MAGIC_SKILLS, TrickMagicItemEntry } from "@item/spellcasting-entr
 import { ErrorPF2e, localizer, tupleHasValue } from "@util";
 
 /** Prompt the user for a skill with which to trick a magic item, then roll the check and cast the embedded spell. */
-async function trickMagicItem(item: ConsumablePF2e): Promise<void> {
+async function trickMagicItem(item: ConsumablePF2e): Promise<boolean> {
     const localize = localizer("PF2E.TrickMagicItemPopup");
     if (!item.isOfType("consumable")) {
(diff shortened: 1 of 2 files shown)


src/module/actor/sheet/trick-magic-item-popup.ts
```diff
@@ -32,15 +32,17 @@ async function trickMagicItem(item: ConsumablePF2e): Promise<void> {
...
-    if (!tupleHasValue(TRICK_MAGIC_SKILLS, skill)) return;
+    if (!tupleHasValue(TRICK_MAGIC_SKILLS, skill)) return false;
 
-    actor.skills[skill].check.roll({
+    const check = await actor.skills[skill].check.roll({
         extraRollOptions: ["action:trick-magic-item"],
         dc: { value: checkDC[skill] ?? 0 },
         item: consumable,
     });
+    if (!check?.degreeOfSuccess || check.degreeOfSuccess < 2) return false;
```
jfn4th: Could you import `DEGREE_OF_SUCCESS` and try this?
```ts
if ((check?.degreeOfSuccess ?? 0) < DEGREE_OF_SUCCESS.SUCCESS) return false
```

PR #22905, opened 2026-08-06
Convert XP macro to `ApplicationV2`

jfn4th:
This is what lead me to replace Jest, I replaced the table with a metrics view matching that of the encounter tab and importing `TextEditorPf2e` to get at `convertXMLNode` lead to Svelte eventually being loaded and breaking Jest.

I also tweaked the text shadow on the threat level to make it easier to read in light mode.

xp-macro-light
xp-macro-dark


stwlam:
Not for this PR, but we ought to put together a dark-theme pf2e table style.

PR #22912, opened 2026-08-07
Implementing etched subitems?

reonZ:
This is more a "testing the waters" kind of thing than a proper PR, there is quite a lot of unknowns on how you guys actually planned to implement runes, if they stay as simple equipment, if they still use the hardcoded stuff in the system, etc.

So here is a preliminary (but working) draft.

https://github.com/user-attachments/assets/1272400f-81c8-4801-99ca-3af4b9d7a3a2


Cuingamehtar:
On principle, I'm for it. It will be consistent with how SF modules look.
However, I don't like how busy it makes the inventory. The changes are already reflected in the item name (assuming, of course, it's the same as the base name). Fully packed item will occupy 5 extra lines (6 with mythic). Add armor to it, a backup weapon, and some high level looted gear, and you'll have to scroll through the sheet to get to consumables.

reonZ:
> On principle, I'm for it. It will be consistent with how SF modules look. However, I don't like how busy it makes the inventory. The changes are already reflected in the item name (assuming, of course, it's the same as the base name). Fully packed item will occupy 5 extra lines (6 with mythic). Add armor to it, a backup weapon, and some high level looted gear, and you'll have to scroll through the sheet to get to consumables.

The system could decide to not display rune subitems in the inventory ofc, and only in the weapon sheet, but that is just a detail, it is not like i did any work to display them there, this is how the system works with subitems and how it was intended to be.

Allalinor:
Maybe a toggle in the inventory (potentially per category) that hides/shows subitems? Similar to how on the spellcasting tab we have one to hide/show unused spellslots.

Cuingamehtar:
Idle is correct here in that it's two separate issues. They are related, of course, but both don't have to be addressed at the same time.

As for the technical implementation, I noticed that the restriction on the number of property runes that can be attached at the same time is missing.

reonZ:
> As for the technical implementation, I noticed that the restriction on the number of property runes that can be attached at the same time is missing.

It is not limiting in the adding of the rune but it is then subtracted from the the system data. It was already done like that in the system, i didn't want to modify too much of the existing code in my draft. 

https://github.com/foundryvtt/pf2e/blob/a83ce6d84368eb1fd6b57eab2d438b0f363de02a/src/module/item/weapon/document.ts#L440

PR #22915, opened 2026-08-07
Add physical item usage filters to Compendium Browser

Cuingamehtar:
image
There is currently no good way to search for SF2e gear upgrades, and users have to refer to AoN instead of using the compendium browser.

Considering that the same usage types are being used to check if items can be attached, it seemed reasonable to use it as a filter.

Types of usages have been modified to tie them to the config object data, but somebody more knowledgeable of TS than me should check if its ok.

### DIFF
--- a/src/module/apps/compendium-browser/tabs/equipment.ts
+++ b/src/module/apps/compendium-browser/tabs/equipment.ts
@@ -6,6 +6,7 @@ import { CompendiumBrowser } from "../browser.svelte.ts";
 import { ContentTabName } from "../data.ts";
 import { CompendiumBrowserTab } from "./base.svelte.ts";
 import { CompendiumBrowserIndexData, EquipmentFilters, RangesInputData } from "./data.ts";
+import { getUsageDetails } from "@item/physical/usage.ts";
 
 export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
     tabName: ContentTabName = "equipment";
@@ -34,7 +35,8 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
         const physicalItemFields = [...baseFields, "system.level.value"];
         const runedItemFields = [...physicalItemFields, "system.runes"];
         const armorAndWeaponFields = [...runedItemFields, "system.category", "system.group"];
-        const indexFields = R.unique([...armorAndWeaponFields]).sort();
+        const equipmentFields = [...physicalItemFields, "system.usage.value"];
+        const indexFields = R.unique([...armorAndWeaponFields, "system.usage.value"]).sort();
         const publications = new Set<string>();
 
         for await (const { pack, index } of this.browser.packLoader.loadPacks(
@@ -57,6 +59,8 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
                                 return !this.hasAllIndexFields(itemData, baseFields);
                             case "shield":
                                 return !this.hasAllIndexFields(itemData, runedItemFields);
+                            case "equipment":
+                                return !this.hasAllIndexFields(itemData, equipmentFields);
                             default:
                                 return !this.hasAllIndexFields(itemData, physicalItemFields);
                         }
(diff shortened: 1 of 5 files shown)


src/module/apps/compendium-browser/tabs/equipment.ts
```diff
@@ -57,6 +59,8 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
                                 return !this.hasAllIndexFields(itemData, baseFields);
                             case "shield":
                                 return !this.hasAllIndexFields(itemData, runedItemFields);
+                            case "equipment":
+                                return !this.hasAllIndexFields(itemData, equipmentFields);
```
jfn4th, line 63:
```suggestion
```
All our equipment has `system.usage.value`, and you later fallback to "none". The only thing this does is hide module or homebrew content missing usage. You can also get rid of `equipmentFields`

src/module/apps/compendium-browser/tabs/equipment.ts
```diff
@@ -107,6 +115,7 @@ export class CompendiumBrowserEquipmentTab extends CompendiumBrowserTab {
                         level: itemData.system.level?.value ?? 0,
                         price: priceCoins,
                         rarity: itemData.system.traits.rarity,
+                        usage: usageType,
```
jfn4th, line 118:
```suggestion
```
The others are there because they're either shown or used in sorting. Filtering only reads `options`, so no need to add this.

jfn4th (changes requested):
This is pretty close, but I don't think `usage.type` should be the source. `getUsageDetails` only supports a few specific usages, many get thrown into Worn, and that won't change without actual support for affixing them. Using `USAGES` from `config/usage.ts`, you could make affixed/etched/applied/tattooed/other groups like there are for weapon and armor upgrades. Options should also come from the groups you actually hit while indexing like `publications` does, else you'll see ones with no results, like Installed in pf or Attached in sf.

Something like:

```ts
const AFFIXED = {
    "affixed-to-a-shield": "PF2E.TraitAffixedToAShield",
    // etc
};

const USAGES = { ...WEAPON_UPGRADES, ...ARMOR_UPGRADES, ...AFFIXED, ...ETCHED, /* etc */ };

const USAGE_GROUPS_BY_VALUE: Record<string, UsageGroup> = Object.fromEntries(
    Object.entries({ affixed: AFFIXED, etched: ETCHED, applied: APPLIED, tattooed: TATTOOED, other: OTHER }).flatMap(
        ([group, usages]) => Object.keys(usages).map((usage) => [usage, group as UsageGroup]),
    ),
);
```

Then in the equipment tab you can do a lookup on that, falling back to `getUsageDetails`. The `usageTypes` object you added can stay as the checkbox labels, just add affixed/etched/etc to it. It's not a `UsageType` mirror anymore, so `usage.ts` can be reverted.

Cuingamehtar:
Actually, I'm currently having second thoughts about this one.

The goal was to distinguish SF modules from other items, and while `installed` helps, it's not explicit enough in a "hey, this is a module and it's supposed to count in the item upgrade limit" way. I feel like it needs a different approach somehow.
Like it would have more correct to just add a tag to a SF module and filter by it but equipment doesn't allow other tags in its form.

Also, in light of the new development by Shark, I'd rather draft it or close it for now to see what happens, if you don't suggest I continue with the current approach.

> Options should also come from the groups you actually hit while indexing like publications does...

This is something I've noticed in other filters as well, and wanted to tackle it too, but didn't do it here so as it is a semirelated change.

jfn4th:
Sure, we can wait and see how Shark's rune rework leaves things, but I think this is about as close as it can get currently. Usage is the only thing to go off of for most of what you're going for.

Perhaps instead of naming the groups after the usage verb, name them after what they are. Upgrades, Augmentations, Affixed, Tattoos. That's explicit in the way you want, and it matches how `WEAPON_UPGRADES` and `ARMOR_UPGRADES` are already grouped in that file, by what they are rather than by prefix.

Issue #22808, opened 2026-07-31
Automaton Lesser Augmentation feat not letting allowing to choose enhanced augmentation for Hardened Chassis


harisbinali:
When I add Lesser Augmentation it just automatically applies the enhancement for Reinforced Chassis without letting me choose

<img width="428" height="165" alt="Image" src="https://github.com/user-attachments/assets/dc41c84a-ee29-4cc7-b088-55659e1ddb42" />

PR #22917, opened 2026-08-08
Add augmentation automation to IM Automaton feats (#22808)

sgennaoui:
The Automaton feat Lesser Augmentation allows a player to select one of
their 1st- or 5th-level automaton ancestry feats to gain its
enhancement benefits.

Lesser Augmentation already enables selection of some feats, such as
Arcane Eye or Resilient Chassis, but Hardened Chassis did not have the
active effect rules which enable it to be selected.

This commit allows Hardened Chassis and the remaining unselectable
1st- or 5th-level feats to be selected for either Lesser
Augmentation or Greater Augmentation.

### DIFF
--- a/packs/pf2e/feats/ancestry/automaton/level-1/expressive-face.json
+++ b/packs/pf2e/feats/ancestry/automaton/level-1/expressive-face.json
@@ -27,6 +27,34 @@
         },
         "rules": [
             {
+                "key": "ActiveEffectLike",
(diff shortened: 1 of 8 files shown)


TikaelSol:
Thanks for the PR, as a note we use imperative style for PR titles, so the titles can easily be used as the patch note. Basically write titles that complete the sentence "If merged this PR will..."

PR #22926, opened 2026-08-08
Add "area" support to `ItemAlteration` and use for Directed Channel

simonward:


src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -149,6 +150,47 @@ const ITEM_ALTERATION_HANDLERS = {
...
+                    data.alteration.value ?? "burst",
+                );
+                if (newValue instanceof validation.DataModelValidationFailure) {
+                    throw newValue.asError();
+                }
+                if (data.item.system.area) {
+                    data.item.system.area.type = newValue;
+                } else {
+                    data.item.system.area = { type: newValue, value: 5 };
+                }
```
stwlam:
```suggestion
                data.item.system.area ??= { type: newValue, value: 5 };
                data.item.system.area.type = newValue;
```

src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -149,6 +150,47 @@ const ITEM_ALTERATION_HANDLERS = {
...
+            const mode = data.alteration.mode;
+            if (mode === "override") {
+                const newValue = AELikeRuleElement.getNewValue(
+                    mode,
+                    data.item.system.area?.type ?? "burst",
+                    data.alteration.value ?? "burst",
+                );
+                if (newValue instanceof validation.DataModelValidationFailure) {
+                    throw newValue.asError();
+                }
```
stwlam:
```suggestion
                if (newValue instanceof validation.DataModelValidationFailure) throw newValue.asError();
```

src/module/rules/rule-element/item-alteration/handlers.ts
```diff
@@ -149,6 +150,47 @@ const ITEM_ALTERATION_HANDLERS = {
             data.item.system.area.value = Math.max(nearestFive, 5);
         },
     }),
+    "area-type": new ItemAlterationHandler({
+        fields: {
+            itemType: new fields.StringField({ required: true, choices: ["spell"] }),
+            mode: new fields.StringField({
+                required: true,
+                choices: ["override", "remove"],
+            }),
```
stwlam:
```suggestion
            mode: new fields.StringField({ required: true, choices: ["override", "remove"] }),
```
whitespace comes at a premium these days

stwlam:
Would prefer an "effect-area" alteration that can just reference the current value if need be. It seems changing both the type and value occurs at least as often as changing the type alone.

PR #22932, opened 2026-08-09
LoseHitPoints Rule Element: Test predicate during rule evaluation

### LINKED ISSUE #22931: LoseHitPoints Rule Element doesn't check predicate
I'm implementing automation for the [mutagenist's field vials](https://2e.aonprd.com/ResearchFields.aspx?ID=7) by adding predicates to the drawbacks of all mutagens.  This works well for everything except for [Quicksilver Mutagens](https://2e.aonprd.com/Equipment.aspx?ID=3319), which use the LoseHitPoints rule element for their drawback.
As it turns out, the LoseHitPoints Rule Element doesn't test its predicate, so its effect can't be disabled without removing the effect containing the element.

neonyuzu:
Also set actor HP to its effective max hp during the afterPrepareData phase if its current hp is greater than it.

Closes https://github.com/foundryvtt/pf2e/issues/22931.

### DIFF
--- a/src/module/rules/rule-element/lose-hit-points.ts
+++ b/src/module/rules/rule-element/lose-hit-points.ts
@@ -18,14 +18,14 @@ class LoseHitPointsRuleElement extends RuleElement<LoseHitPointsRuleSchema> {
     }
 
     override onCreate(actorUpdates: Record<string, unknown>): void {
-        if (this.ignored) return;
+        if (this.ignored || !this.test()) return;
         const value = Math.trunc(Math.abs(Number(this.resolveValue(this.value)) || 0));
         const currentHP = this.actor._source.system.attributes.hp.value;
         actorUpdates["system.attributes.hp.value"] = Math.max(currentHP - value, 0);
     }
 
     override beforePrepareData(): void {
-        if (this.ignored) return;
+        if (this.ignored || !this.test()) return;
 
         const { actor } = this;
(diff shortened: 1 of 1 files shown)


neonyuzu:
For ease of testing, I put a feat on a test actor with the following two rule elements:

`{"key":"RollOption","label":"Disable LoseHP","option":"disable-lose-hp","toggleable":true,"value":false}`
`{"key":"LoseHitPoints","recoverable":false,"value":5,"predicate":[{"not":"disable-lose-hp"}]}`

src/module/rules/rule-element/lose-hit-points.ts
```diff
@@ -34,8 +34,35 @@ class LoseHitPointsRuleElement extends RuleElement<LoseHitPointsRuleSchema> {
...
+        if (this.ignored || !this.test()) return;
+
+        const { actor } = this;
+        if (!this.recoverable) {
+            const maxHP = actor.hitPoints.max;
+            const unrecoverableHP = actor.system.attributes.hp.unrecoverable;
+            const effectiveMaxHP = maxHP - unrecoverableHP;
+            const currentHP = actor.system.attributes.hp.value;
+            if (currentHP > effectiveMaxHP) {
+                this.actor.update(
```
stwlam, line 53: No updates during data preparation.

PR #22934, opened 2026-08-09
Cap character sheet HP input to max less unrecoverable

### LINKED ISSUE #22933: Character Sheet current hp doesn't immediately update when setting hp higher than effective recoverable hp
If a character has unrecoverable HP (such as from the LoseHitPoints rule element from a Quicksilver Mutagen), and their current HP is equal to their effective max HP, and they then set their HP to a value higher than that through the character sheet, the character sheet appears to allow the change until something else causes the sheet to be updated.  

This is a display only bug.  The character's HP doesn't actually change.  In fact, the reason this happens is because the HP doesn't actually change, so the sheet is not re-rendered.  If the character's HP is less than their effective max HP and it is then set to a value higher than that, the sheet immediately updates to show the correct HP value.

neonyuzu:
Fixes https://github.com/foundryvtt/pf2e/issues/22933 by setting the data-max of the current hp input on the character sheet to the effective max hp, taking into account unrecoverable hp.  

Since we can't do math from the templates, I added a new property to the sheet data for the effectiveMaxHP and store the calculation into that.  This seemed safer than trying to put the property into the ActorHitPoints struct with the rest of the hit points data as I wasn't sure how best to keep the calculation updated, and anywhere else that needs the effectiveMaxHP can just do the calculation themselves.


src/module/actor/character/sheet.ts
```diff
@@ -120,6 +120,11 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
             if (tab) tab.initial = "biography";
         }
 
+        /** Max HP when taking into account unrecoverable HP.  Needed to set the data-max of the current hp
+         *  input since we can't do math in the template.
+         */
+        sheetData.effectiveMaxHP = this.actor.attributes.hp.max - this.actor.attributes.hp.unrecoverable;
```
jfn4th:
```suggestion
        sheetData.effectiveMaxHP = Math.max(actor.hitPoints.max - actor.hitPoints.unrecoverable, 0);
```
`unrecoverable`  could be greater than `max`,  it gets set before max hp is computed and isn't capped by it.

src/module/actor/character/sheet.ts
```diff
@@ -1649,6 +1654,10 @@ interface CharacterSheetData<TActor extends CharacterPF2e = CharacterPF2e> exten
     crafting: CraftingSheetData;
     data: CharacterSystemSheetData;
     deity: DeityPF2e<CharacterPF2e> | null;
+    /** Max HP when taking into account unrecoverable HP.  Needed to set the data-max of the current hp
+     *  input since we can't do math in the template.
+     */
+    effectiveMaxHP: number;
```
jfn4th:
```suggestion
    /** Max HP minus unrecoverable HP */
    effectiveMaxHP: number;
```

jfn4th (commented):
Thanks for the PR. Just a couple small suggestions.

PR #22937, opened 2026-08-09
Brush up Absalom Station items

Longstrider:


packs/sf2e/equipment/augmentations/magitech/limscan-lenses.json
```diff
@@ -35,7 +35,32 @@
...
+                    "saving-throw"
+                ],
+                "type": "item",
+                "value": 1
+            },
+            {
+                "key": "FlatModifier",
+                "predicate": [
+                    "magical-trap-perception"
+                ],
```
TikaelSol, line 57:
```suggestion
                "predicate": [
                    "target:trait:trap",
                    "target:trait:magical"
                ],
```
Just generally favoring real roll options when available, even if they are unlikely to have the thing actually targeted.

PR #22967, opened 2026-08-12
Expire the Greater Cover effect from tower shield

### LINKED ISSUE #22943: Greater Cover effect from raised tower shield isn't removed when shield is lowered
When a character has a raised tower shield and uses the inline Take Cover button, the system applies a "Greater Cover" effect (+4 circumstance bonus to AC), tagged with system.traits.otherTags: ["tower-shield"]. When the shield is later lowered (the effect-raise-a-shield effect is deleted), the cover effect is not cleaned up. It stays on the actor indefinitely, silently inflating their AC even after the shield is no longer raised.

To Reproduce

Equip a tower shield on a PC and raise it.
Click the inline Take Cover button to apply Greater Cover.
Lower the shield (remove/toggle off effect-raise-a-shield).
Check the actor's effects panel and AC.

Expected behavior

Lowering the shield should also remove the cover effect that was granted by raising it, since the rules state the AC bonus only applies "until the shield is no longer raised."

Actual behavior

The cover effect persists after the shield is lowered, and the actor keeps the +4 circumstance bonus to AC with no shield raised.

Environment

Foundry VTT v13+ (also confirmed on v14)
pf2e system v6+

Additional context

I put together a small module that patches around this by hooking item creation/deletion (flagging cover effects tagged tower-shield on creation, then removing them when effect-raise-a-shield is deleted) rather than touching a system compendium file: [https://github.com/InportantMageKing/skywall-tower-shield-fix](https://github.com/InportantMageKing/skywall-tower-shield-fix)

I haven't confirmed whether Fortress Shield's similar "raise and take cover" interaction has the same issue or uses a different tag, but it may be worth checking as part of a proper fix.

sgennaoui:
After raising a tower shield or fortress shield, a player can gain the Greater Cover effect by taking the Take Cover action. By default, the cover effect has an unlimited duration and no expiry.

With this change, when the cover effect for the tower shield is created, the duration is also set to expire after 1 round at start-of-turn.

Fixes #22943


sgennaoui:
I had considered doing it as a new cover effect (e.g. effect-greater-cover-for-1-round), but opted for the simple approach first.
Let me know if a new effect is preferred!

src/module/actor/character/auxiliary.ts
```diff
@@ -190,6 +190,12 @@ class WeaponAuxiliaryAction {
             if (effect instanceof EffectPF2e) {
                 const data = { ...effect.toObject(), _id: null };
                 data.system.traits.otherTags.push("tower-shield");
+                data.system.duration = {
+                    expiry: "turn-start",
+                    sustained: false,
+                    unit: "rounds",
+                    value: 1,
+                };
```
stwlam:
```suggestion
                data.system.duration = { expiry: "turn-start", unit: "rounds", value: 1 };
```

src/module/actor/character/auxiliary.ts
```diff
@@ -190,6 +190,7 @@ class WeaponAuxiliaryAction {
             if (effect instanceof EffectPF2e) {
                 const data = { ...effect.toObject(), _id: null };
                 data.system.traits.otherTags.push("tower-shield");
+                data.system.duration = { expiry: "turn-start", unit: "rounds", value: 1 };
```
stwlam:
```suggestion
                data.system.duration = { expiry: "turn-start", unit: "rounds", value: 1, sustained: false };
```

PR #22980, opened 2026-08-15
Add icon to Revealing Mist item

ammalagonc:


stwlam:
We're already using this one for Storm Breath: any good alternatives?

PR #23006, opened 2026-08-17
Fix display of Sanctum Guardian's immunity to starvation and thirst

simonward:
~~Ordinarily I would prefer to use a custom immunity for this, but it's already been shipped with these values~~

Also add bleed missing bleed immunity

Before:
Screenshot 2026-08-16 at 11 34 18 PM

After:
Screenshot 2026-08-16 at 11 32 53 PM


TikaelSol:
Since we already have the bad entries in we should probably just do a custom unless the current actor is just busted

PR #23007, opened 2026-08-17
Fix Medium Armor Expertise showing inventor text for every class

Cuingamehtar:


packs/pf2e/class-features/medium-armor-expertise.json
```diff
@@ -11,7 +11,7 @@
         },
         "category": "classfeature",
         "description": {
-            "value": "<p>You've learned to defend yourself better against attacks. Your proficiency ranks for light armor, medium armor, and unarmored defense increase to expert.</p>\n<p><strong>Inventor</strong> If you have a medium or heavy armor innovation, you gain access to the critical specialization effect with your armor innovation.</p>"
+            "value": "<p>@Localize[PF2E.SpecificRule.ClassFeatures.MediumArmorExpertise.GeneralDescription]</p>\n<p><strong>Inventor</strong> @Localize[PF2E.SpecificRule.ClassFeatures.MediumArmorExpertise.inventor.Description]</p>"
```
TikaelSol: I'm confused why we're doing the localizations here

Cuingamehtar: Purely because its content exists to be replaced. It will only be seen by a user in its compendium entry.

I guess that unlike Spell Repertoire this one is fairly small. If this is an undesired change, I'll revert it.

Cuingamehtar: Actually, that's not the sole reason - this way the contents are effectively guaranteed to be up to data, as if the localization data changes, the content will change as well.
In addition, if some future class provides additional effect to it, it will be instantly evident that the content need to be handled with the RE.

simonward:
Conflicts!

Cuingamehtar:
fixed

iDantar:
Looks like more conflicts happened, unfortunately

PR #23022, opened 2026-08-19
Remap default system action images to corresponding action symbols

SpartanCPA:


stwlam:
What's being redirected? And why the change?

ammalagonc:
The title is misleading - they're changing/swapping the `action.svg` for the corresponding action icons.

PR #23025, opened 2026-08-20
Fit emanations created from region tools to hovered tokens

FolkvangrForgent:
Allow emanations placed via region tools to auto fit to a token in measured template mode. This happens only if there is a single token where the emanation is being placed.

### DIFF
--- a/src/module/canvas/layer/region.ts
+++ b/src/module/canvas/layer/region.ts
@@ -94,7 +94,7 @@ export class RegionLayerPF2e extends fc.layers.RegionLayer<RegionPF2e> {
     /** Fit an emanation region shape base to underlying token while drawing with the region tools. */
     protected override _createDragShapeData(event: PlaceablesLayerEvent<RegionPF2e>): BaseShapeData {
         const shape = super._createDragShapeData(event) as EmanationShapeData;
-        if (!canvas?.scene || shape?.type !== "emanation" || shape?.base?.type !== "token") return shape;
+        if (!this.templateMode || shape.type !== "emanation" || shape.base.type !== "token") return shape;
         const { x, y } = event.interactionData.origin;
         const tokens = canvas.tokens.quadtree.getObjects(new PIXI.Rectangle(x, y, 0, 0));
         const token = tokens.values().next().value?.document;
--- a/src/module/canvas/layer/region.ts
+++ b/src/module/canvas/layer/region.ts
@@ -2,7 +2,13 @@ import type {
     PlaceablesLayerEvent,
     PlaceablesLayerPointerEvent,
 } from "@client/canvas/layers/base/placeables-layer.d.mts";
-import type { ConeShapeData, LineShapeData, EmanationShapeData, TokenShapeData } from "@common/data/data.d.mts";
+import type {
+    BaseShapeData,
+    ConeShapeData,
+    LineShapeData,
+    EmanationShapeData,
+    TokenShapeData,
+} from "@common/data/data.d.mts";
 import { RegionDocumentPF2e } from "@scene";
 import { RegionPF2e } from "../index.ts";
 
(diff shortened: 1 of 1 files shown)


src/module/canvas/layer/region.ts
```diff
@@ -81,4 +84,27 @@ export class RegionLayerPF2e extends fc.layers.RegionLayer<RegionPF2e> {
...
+        const shape = super._createDragShapeData(event) as EmanationShapeData;
+        if (canvas?.scene && shape?.type === "emanation" && shape?.base?.type === "token") {
+            const gridSize = canvas.grid.size;
+            const inBoundTokens = canvas.scene.tokens.filter((token) => {
+                const { x, y, width, height } = token;
+                return (
+                    event.interactionData.origin.x.between(x, x + width * gridSize) &&
+                    event.interactionData.origin.y.between(y, y + height * gridSize)
+                );
+            });
```
stwlam: This should be a quadtree retrieval

types/foundry/client/canvas/layers/base/placeables-layer.d.mts
```diff
@@ -398,6 +398,12 @@ export default class PlaceablesLayer<TObject extends PlaceableObject = Placeable
      */
     protected _confirmDeleteKey(documents: TObject["document"][]): Promise<boolean>;
 
+    /**
+     * Create the shape data from the drag start event.
+     * @param event  The PIXI InteractionEvent which wraps a PointerEvent
+     */
+    protected _createDragShapeData(event: PIXI.FederatedEvent): foundry.data.BaseShapeData;
```
stwlam: Import instead of referencing by global namespace

src/module/canvas/layer/region.ts
```diff
@@ -81,4 +84,27 @@ export class RegionLayerPF2e extends fc.layers.RegionLayer<RegionPF2e> {
             this._updateDragPreview(event);
         }
     }
+
+    /** Fit an emanation region shape base to underlying token while drawing with the region tools. */
+    protected override _createDragShapeData(event: PlaceablesLayerEvent<RegionPF2e>): foundry.data.BaseShapeData {
+        const shape = super._createDragShapeData(event) as EmanationShapeData;
+        if (canvas?.scene && shape?.type === "emanation" && shape?.base?.type === "token") {
```
stwlam: Return early instead of nesting the entire contents of the method. This should also be limited to measured-template mode.

stwlam (changes requested):
This is really more appropriate as an upstream feature request, but I'm open to it being a system feature if something comparable is not looking likely to end up in core any time soon.

FolkvangrForgent:
> This is really more appropriate as an upstream feature request, but I'm open to being a system feature if something comparable is not looking likely to end up in core any time soon.

The closest core issue I found was https://github.com/foundryvtt/foundryvtt/issues/13863 which was closed. I took that as a sign that this is not going to end up in core any time soon and figured it would be a nice addition to this system.

stwlam:
> The closest core issue I found was [foundryvtt/foundryvtt#13863](https://github.com/foundryvtt/foundryvtt/issues/13863) which was closed. I took that as a sign that this is not going to end up in core any time soon and figured it would be a nice addition to this system.

That was closed as a duplicate of foundryvtt/foundryvtt#13892, which was closed as completed: does it have any relevance here?

FolkvangrForgent:
> That was closed as a duplicate of [foundryvtt/foundryvtt#13892](https://github.com/foundryvtt/foundryvtt/issues/13892), which was closed as completed: does it have any relevance here?

No, unless I am missing something https://github.com/foundryvtt/foundryvtt/issues/13892 does not address the functionality https://github.com/foundryvtt/foundryvtt/issues/13863 was asking for: `dropping one ontop of a token on the canvas automatically resized the base of it to match the token dimensions` when using the region layer tools.

src/module/canvas/layer/region.ts
FolkvangrForgent: changed to a quadtree retrieval

types/foundry/client/canvas/layers/base/placeables-layer.d.mts
FolkvangrForgent: no longer referencing by global namespace

src/module/canvas/layer/region.ts
FolkvangrForgent: Return early instead of nesting method.

Why do you want this limited to measured-template mode? I don't think there isn't anything specifically that would imply emanations drawn from the region tool would or should only set the base token in measured-template mode.

stwlam: The system is hands off when not in measured-template mode. Regions may have subsumed measured templates, but they're still used for more than that.

FolkvangrForgent: Restricted behavior to measured template mode

stwlam:
> No, unless I am missing something [foundryvtt/foundryvtt#13892](https://github.com/foundryvtt/foundryvtt/issues/13892) does not address the functionality [foundryvtt/foundryvtt#13863](https://github.com/foundryvtt/foundryvtt/issues/13863) was asking for: `dropping one ontop of a token on the canvas automatically resized the base of it to match the token dimensions` when using the region layer tools.

I think what we instead got was an API means of doing a little more than what was requested: it makes an "aura" of sorts, attaching the region to the token, but also automatically resizes the hole to fit the token. While it's true that in pf2e an emanation isn't necessarily an aura, we've been able to determine via some paizo designer arm-twisting that all non-aura emanations resolve instantly and should never have a duration (making any such case an error requiring either the addition of the Aura trait or the removal of the duration). This would make an emanation region following a token around theoretically fine since moving such a token implies either a mistaken placement or that the region is no longer relevant and should be deleted.

stwlam (approved):

PR #23036, opened 2026-08-22
Remove empty filters and traits not associated with any items in compendium browser

Cuingamehtar:
This change will allow removing some compendium browser filters that would come up with no items in them, in a similar fashion to how publication selection works. As the behavior is identical, the publication source helper also reuses it (but could be removed completely to use the new method).

It will also implement this behavior for weapon and armor groups to remove empty groups, and for armor categories to remove barding in Starfinder. I didn't find it necessary to implement it other categories and tabs.

@jfn4th Pretty much the functionality you've asked for in the other PR.

### DIFF
--- a/src/module/apps/compendium-browser/components/filters.svelte
+++ b/src/module/apps/compendium-browser/components/filters.svelte
@@ -108,16 +108,18 @@
         <Traits bind:traits={filter.traits} />
     </FilterContainer>
     {#each Object.entries(filter.checkboxes) as [key, checkbox]}
-        <FilterContainer
-            isExpanded={checkbox.isExpanded}
-            clearButton={{
-                options: { visible: checkbox.selected.length > 0 },
-                clear: getClearFunction(checkbox),
-            }}
-            label={checkbox.label}
-        >
-            <Checkboxes bind:checkbox={filter.checkboxes[key as keyof BrowserFilter["checkboxes"]]} />
-        </FilterContainer>
+        {#if !R.isEmpty(checkbox.options)}
+            <FilterContainer
+                isExpanded={checkbox.isExpanded}
+                clearButton={{
+                    options: { visible: checkbox.selected.length > 0 },
+                    clear: getClearFunction(checkbox),
+                }}
+                label={checkbox.label}
+            >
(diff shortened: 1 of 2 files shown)


jfn4th:
Thank you. After thinking about it a bit, there's a simpler way to do this that can be done generically instead of opting each filter in one at a time and making more sets. Every entry already carries its options and the checkbox keys match them, so pruning after load would cover every group in every tab, including any added later.

```ts
#pruneEmptyFilterOptions(): void {
    if (!this.filterData || !("checkboxes" in this.filterData)) return;
    const present = new Set(this.indexData.flatMap((e) => [...(e.options ?? [])]));
    const checkboxes: Record<string, CheckboxData> = this.filterData.checkboxes;
    for (const [key, checkbox] of R.entries(checkboxes)) {
        const prefix = checkbox.optionPrefix ?? key;
        checkbox.options = R.pickBy(checkbox.options, (_v, k) => present.has(`${prefix}:${k}`));
    }
}
```

Added to `base.svelte.ts`, called In `init` right after `await this.loadData()`.`filters.svelte` will need to be updated skip empty ones.

Cuingamehtar:
Yep, this is certainly a cleaner solution, no doubt.

Also, I took this as an opportunity to clean up the list of traits as well, removing extra ones.

jfn4th:
Great! Looks good but you still need an empty check in `filters.svelte`, in the `{#each ...}` block you can wrap `<FilterContainer>` in an `{#if !R.isEmpty(checkbox.options)}`

Cuingamehtar:
Of note, `options` are all Sets, but currently undeclared as such in the type. Could be reduced with union, I think, but would require creating an intermediary type specifying `options: Set<string>`.

I'm not sure I like double checking for property, but its either that or lazily populating the set of options.

Cuingamehtar:
I guess traits selection can also be removed if empty.
But honestly, likelihood that the checkbox options object or the trait list will be completely empty is nil.

jfn4th:
Ah, you're right the types could use some tightening up, but I can handle that separately. Thank again.

PR #23037, opened 2026-08-22
Update some NPCs to use `AdjustStrike` to add runes

simonward:
Also add automation for Caydserris Arudora's Charg's Final Gift


pedrogrullada:
These look good to me! Would you consider changing the PR title to be more ambiguous for the sake of spoilers for the AP? (or let's make note to obfuscate it for the next patch notes)

TikaelSol (approved):

PR #23041, opened 2026-08-22
Make `_validateType` throw errors instead of returning them

oWave:
v13 allowed returning `DataModelValidationFailure` in validators (`DataField.validate` in `common/data/field.mjs`), but v14 only checks if the returned value is `true` or `false`.
Returned errors get ignored, so invalid REs like `{"key":"RollOption","label":"Test","option":"test","suboptions":"invalid","toggleable":true}` get processed and cause exciting problems. 
Having just that RollOption RE on a feat just causes the feat to not get added to the sheet, but more complex setups can break actor data prep (https://discord.com/channels/170995199584108546/697845559435853865/1540673817754603610)

### DIFF
--- a/src/module/system/schema-data-fields.ts
+++ b/src/module/system/schema-data-fields.ts
@@ -341,10 +341,7 @@ class DataUnionField<
         return super.clean(value, options, _state) as MaybeProp;
     }
 
-    protected override _validateType(
-        value: unknown,
-        options?: DataFieldValidationOptions | undefined,
-    ): boolean | void | validation.DataModelValidationFailure {
+    protected override _validateType(value: unknown, options?: DataFieldValidationOptions | undefined): boolean | void {
         const errors: { field: TField; result: validation.DataModelValidationFailure }[] = [];
         for (const field of this.fields) {
             const result = field.validate(value, options);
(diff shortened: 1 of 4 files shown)


src/module/system/schema-data-fields.ts
```diff
@@ -619,12 +616,10 @@ class RecordField<
...
 
-    protected override _validateType(
-        values: unknown,
-        options?: DataFieldValidationOptions,
-    ): boolean | validation.DataModelValidationFailure | void {
+    protected override _validateType(values: unknown, options?: DataFieldValidationOptions): boolean | void {
         super._validateType(values, options);
-        return this.validateValues(values, options);
+        const failure = this.validateValues(values, options);
+        if (failure) throw failure;
```
stwlam, line 622:
```suggestion
        if (failure) throw failure.asError();
```

oWave: Core `DataField.validate` turns any thrown error that isn't a DataModelValidationFailure back into one. Wouldn't throwing via `asError()` discard the original failure data?
The core DataFields like SchemaField also throw DataModelValidationFailure, not DataModelValidationError

stwlam: Bah. Core is having a little too much fun with the fact that JS lets you throw anything.

PR #23042, opened 2026-08-22
Pin pnpm to 10.34.5

jfn4th:
pnpm will auto switch to the version set on the `packageManager` field in `package.json`. Ideally with pnpm, everyone should be using the exact same pnpm version, though that doesn't matter as much for us until we move package installation to it. But this makes life easier if you have to juggle multiple pnpm versions for other things like I do.

`branches` also does nothing on `workflow_dispatch` triggers, so cleared those out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,6 @@ on:
   push:
     branches: [v13-dev, v14-dev]
   workflow_dispatch:
-    branches: [v13-dev, v14-dev]
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -6,7 +6,6 @@ on:
   push:
     branches: [v13-dev, v14-dev]
   workflow_dispatch:
-    branches: [v13-dev, v14-dev]
```
stwlam, line 9: Syntax error, ain't it?

jfn4th: valid, just yaml being yaml https://github.com/orgs/community/discussions/26098

PR #23050, opened 2026-08-23
Add Force Fling spell effect

Longstrider:


packs/pf2e/spell-effects/effect-force-fling.json
```diff
@@ -0,0 +1,60 @@
...
+            {
+                "key": "DamageAlteration",
+                "mode": "override",
+                "predicate": [
+                    "item:melee",
+                    {
+                        "not": "item:trait:unarmed"
+                    },
+                    {
+                        "or": [
```
simonward: Can probably use `"item:damage:category:physical"` instead.

packs/pf2e/spell-effects/effect-force-fling.json
simonward, line 1: Might be possible to use `MultipleAttackPenalty` RE here.

PR #23069, opened 2026-08-25
Add pf2e.aurasChanged hook

MrVauxs:
A second go at this, the hook is a very minimalistic notification for modules that a given tokens auras have changed.

This is primarily for animations modules to synchronize their animation sizes with that of the system. All of the relevant data is already available, just not **when** it updates.

```js
Hooks.on("pf2e.aurasChanged", (token) => {
    for (const [slug, aura] of token.auras) {
        console.log(`The aura "${slug}" size is now ${aura.radius} feet.`);
    }
});
```


stwlam:
Not accepting hook PRs, sorry.

Issue #22380, opened 2026-05-21
Add extra flag to spell `cast` and action `use` to differentiate from a simple send-to-chat


reonZ:
Adding an extra flag indicating that a spell/action was actually cast/used (to differentiate it from a regular send-to-chat) would not only be a small change but greatly benefit all the automation modules (which there are a tons of by now for pf2e) out there.

For actions, it would require the extra step of adding the `Use` button to all of them instead of restricting it to those that have charges or have a self-applied effect. Adding the button by itself is already a plus for users "i want to use my action, i press the `Use` button", instead of having to mix between some action requiring to click on the image.

reonZ:
I reframed the request

MrVauxs:
Adding the ability to distinguish actions being posted vs. used would also mean a step towards automating frequency much like spells are.

reonZ:
- The spell part was added #22892
- The action part has been rejected #23071

PR #23071, opened 2026-08-26
Make all actions usable

reonZ:
This makes all actions have the `Use` button instead of limiting it to the those that have uses, have a self-applied effect, or are part of the crafting feature.

The reasons for such change are threefold:
- Consistency between all actions, instead of having to sometimes click on the `Use` button and other times the icon
- Being able to use custom icons for any action while still showing the action cost (it is in the use btn)
- Being able to add a rollOption to differentiate an actual use of the action from a plain send-to-chat (which would come as another PR later if this is accepted)

related #22380

### DIFF
--- a/src/module/actor/character/sheet.ts
+++ b/src/module/actor/character/sheet.ts
@@ -430,8 +430,6 @@ class CharacterSheetPF2e<TActor extends CharacterPF2e> extends CreatureSheetPF2e
                 ...baseData,
                 img: ((): ImageFilePath => {
                     const actionIcon = getActionIcon(item.actionCost);
-                    if (!baseData.usable) return actionIcon;
-
                     const defaultIcon = ItemPF2e.getDefaultArtwork(item._source).img;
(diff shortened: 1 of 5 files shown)


stwlam:
It's useful to know whether or not an action does something.

reonZ:
> It's useful to know whether or not an action does something.

I agree, my original code was to convert the `AbilityViewData.usable` into `AbilityViewData.hasSpecialUse`, i removed it because it ended up not being used anywhere in the code but i can definitely re-add it. Though not sure what to do with it from there.

reonZ:
image

What about something like that?

stwlam:
Mitigating the downside is kind of question-begging: why do this in the first place? The entire point of the Use button is that the action does something. The problems it seeks to solve aren't really related.

MrVauxs:
Frequency automation would also require this sort of button to distinguish real usage and posting to chat, just like spells.

PR #23075, opened 2026-08-26
Update Ash Form publication source and traits

### LINKED ISSUE #23073: Ash Form spell has only half been updated to remaster
The Ash Form currently in the system is partly legacy partly remaster. I noticed because it has the Cursebound trait but shouldn't.
It's also listed as OGL and from A Taste of Ashes (Blood Lords), but it's now ORC and Divine Mysteries.

pf2e 8.4.1

ammalagonc:
- Closes #23073

+some lint

### DIFF
--- a/packs/pf2e/spell-effects/spell-effect-ash-form.json
+++ b/packs/pf2e/spell-effects/spell-effect-ash-form.json
@@ -4,7 +4,7 @@
     "name": "Spell Effect: Ash Form",
     "system": {
         "description": {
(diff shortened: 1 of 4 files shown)


packs/pf2e/spell-effects/spell-effect-ash-form.json
simonward, line 1: Should we add a custom falling damage immunity? We could predicate it on `fall-damage`.

PR #23082, opened 2026-08-27
Prevent modifier duplication on derived speeds

### LINKED ISSUE #10887: Boots of Speed and Quick Climb resulting in incorrect climb speed
The Quick Climb feat grants a climb speed to a PC who is legendary in Athletics, and sets this equal to the PC's land speed. The Boots of Speed item gives an item bonus to the wearer's land, climb, and swim speed. When both of these are in effect on someone who otherwise has just a 25 foot land speed, it currently results in them having a 30 foot land speed and 35 foot climb speed; when instead, both of these should be 30 feet. It appears that the boots' item bonus is applying twice to result in the too-high climb speed.

### LINKED ISSUE #19011: [Starfinder 2e Data] Fly speed equal to swim speed
Some feats in Starfinder 2e grant a fly speed equal to some other speed like swim speed, rather than base land speed. This isn't possible with current rule elements.

### LINKED ISSUE #21145: Minimum land speed is not enforced.
To reproduce: create a merfolk player character with less than +3 strength and equip chain mail.
The land speed is listed as - (0). It should be 5.
Per page 271 of the Player Core: "While wearing a suit of armor, you take the penalty listed in this entry to your Speed, as well as to any other movement types you have, such as a climb Speed or swim Speed, to a minimum Speed of 5 feet."

### LINKED ISSUE #21570: Pushing Wind Duplicated Bonus
If you have Pushing Wind effect:
> So long as you're holding a fan, you and allies who start their turn in a 30- foot aura emanating around you gain a +5-foot circumstance bonus to land Speed for 1 round; you also gain this bonus to your fly Speed if you possess one, but it can't grant you the ability to fly if you wouldn't otherwise be able to.
And an item/feat that gives you a flight speed based on your land speed, it stacks with itself, granting a total of +10 to flight speed:

Image

Image

### LINKED ISSUE #22642: Base Speed Rule Element should be able to forcibly reduce existing movement speed
The Base Speed rule element currently only works when increasing base speeds. Any Base Speed rule elements that would decrease a character's speed are completely ignored.

This is easily seen with the Werecreature (Wereshark) dedication, where the hybrid and animal forms are supposed to reduce or remove most characters' land speeds (and according to the rule elements, they do!) but the character sheet will still show the higher base value.

Additionally testing shows that applying two Base Speed rule element effects to the same character will always set their base speed to the higher of the two values, regardless of what the priority values of each effect are set to.

iDantar:
Changes how the speeds are calculated to properly account for some speeds being derived from others. Fixes bonus duplication, like Pushing Wind + Fly speed derived from land.

Main features:
- Speeds are prepared in order of dependency - so that derived speeds have fully calculated 'parent' speed
- Added a way to discern if one speed is derived from other (for cases like max(50, land-speed)) and the kind of that relationship (direct or scaled, like climb speed being 0.5*land-speed)
- Added 'force' flag for BaseSpeed, to allow reducing base speed

Other changes:
- Allowed speeds to derive from any other speed, not only land
- Fixed lowest speed after penalties to 5 (but 'force' allows to set speed to 0)
- Rules that matched multiple selectors to same speed (for example `all-speeds` and `land-speed` for Land) would no longer show twice on the breakdown tooltip

Should fix all of the following issues
- closes https://github.com/foundryvtt/pf2e/issues/21570
- closes https://github.com/foundryvtt/pf2e/issues/10887
- closes https://github.com/foundryvtt/pf2e/issues/22642
- closes https://github.com/foundryvtt/pf2e/issues/19011
- closes https://github.com/foundryvtt/pf2e/issues/21145


iDantar:
I'll list what cases I tested just in case (no pun intended). But of course I may missed some, so suggestions are very appreciated :)
- Pushing wind + Fly spell (both cases of land speed)
- Quick Climb, Quick Swim + Boots of Speed
- Derived speeds + bonuses with all-speeds selector
- BaseSpeed RE to decrease land speed
- Land speed derived from Swim speed (potential implementation for Supramarine Chair)
- Stacking penalties totaling more than base speed (stays at 5)
- Multiple combinations of Unburdened Iron, Hellknight Mobility, Fortress Shield, Heavy Armor, Hindering trait and Encumbered
- Ring of Climbing + armor combinations from prior point

simonward:
Will it be able to handle [Feet to Fins](https://2e.aonprd.com/Spells.aspx?ID=1525)?

iDantar:
Oh, that kind of effect I didn't plan for, great catch! 
I was able to make a hacky RE for it, but it's not very clean: tooltip of land speed is not updated and the land speed is calculated with all bonuses and penalties before the spell's effect - not sure if that's RAW for 'normal land Speed'

src/module/actor/creature/document.ts
```diff
@@ -722,52 +723,156 @@ abstract class CreaturePF2e<
...
+                (synthetics[type] ?? []).map((entry) => entry.deferred({ test: baseSpeedOptions })),
+                R.isNonNull,
+            );
+            const candidates: SpeedBaseCandidate[] = fromSynthetics.map((s) => ({
+                value: s.value,
+                source: s.source,
+                force: s.force,
+                formula: s.formula,
+                dependsOn: s.dependsOn,
+            }));
```
stwlam: Use `R.pick` if you're just getting a subset of properties.

src/module/rules/rule-element/base-speed.ts
jfn4th: For a formula like `max(20, @actor.system.movement.speeds.land.value)`, this compares the resolved value against the literal after land has already taken its penalties, so how penalized land is ends up deciding whether the speed is treated as `independent` or `equal`. For example, under this a character with Spell Effect: Fly and Effect: Web on a 25ft land speed gets a 10ft fly speed, but the same two effects on a 30ft land speed gives 20ft fly.

Deleting this makes it consistent and puts it back to the current behavior. Whether 20ft is actually the right answer is worth its own discussion, but it shouldn't be picking between the two based on the penalty in the meantime. 

Longer term I think the actual fix is not having to work this out after the fact at all. The relationship gets inferred from the resolved number plus a regex over the formula, which is why the penalty ends up deciding the branch in the first place. If the RE declared what it derives from and carried the floor or cap as its own fields, the floor could apply after modifiers and nothing would need guessing. That would need migration for the min/max items though, so something to be addressed separately.

```suggestion
    if (resolved === parentValue) return "equal";
    return "scaled";
```

src/module/rules/rule-element/base-speed.ts
```diff
@@ -30,6 +80,7 @@ class BaseSpeedRuleElement extends RuleElement<BaseSpeedRuleSchema> {
             ...super.defineSchema(),
             selector: new fields.StringField({ required: true, blank: false, initial: undefined }),
             value: new ResolvableValueField({ required: true, nullable: false, initial: undefined }),
+            force: new fields.BooleanField({ required: false, nullable: false, initial: false }),
```
jfn4th, line 45: `force` will be resolved while picking the base speed, so anything deriving off of that speed sees the overridden value rather than the original. In the case of Feet to Fins, swim ends up at 5 alongside land instead of keeping the pre-transformation land speed. Supporting that properly would mean overrides running as their own pass once everything has derived, rather than being folded into base speed selection.

src/module/system/statistic/speed.ts
```diff
@@ -1,33 +1,47 @@
...
         this.rollOptions = this.createRollOptions(domains);
-        const additionalModifiers = (options.modifiers ??= []);
+        const additionalModifiers = options.modifiers ?? [];
         const modifierAdjustments = actor.synthetics.modifierAdjustments;
         for (const modifier of additionalModifiers) {
             modifier.adjustments = extractModifierAdjustments(modifierAdjustments, domains, modifier.slug);
         }
         const syntheticModifiers = extractModifiers(actor.synthetics, domains, { test: this.rollOptions });
-        this.modifiers = [...syntheticModifiers, ...additionalModifiers];
+        this.modifiers = dedupeModifiersByRule([...syntheticModifiers, ...additionalModifiers]);
```
jfn4th, line 79: This changes non derived speeds too, since a rule with both `all-speeds` and `land-speed` selectors would only apply once to land where it used to apply twice. Should probably be noted in the PR description.

src/module/system/statistic/speed.ts
```diff
@@ -40,32 +54,77 @@ class SpeedStatistic<TActor extends ActorPF2e, TType extends MovementType | "tra
...
         const localize = localizer("PF2E.Actor.Speed");
+        const parent = this.#parent;
+        if (parent) {
+            const parentRules = new Set(parent.modifiers.flatMap((m) => (m.rule ? [m.rule] : [])));
+            const parentSlugs = new Set(parent.modifiers.map((m) => m.slug));
+            const fromParent = (m: Modifier): boolean => (m.rule ? parentRules.has(m.rule) : parentSlugs.has(m.slug));
+            const shared = this.modifiers.filter((m) => m.enabled && fromParent(m));
+            const childOnly = this.modifiers.filter((m) => m.enabled && m.value !== 0 && !fromParent(m));
+            const displayBase = this.base + shared.reduce((sum, m) => sum + m.value, 0);
+            const typeLabel = localize(`Type.${parent.type.capitalize()}`);
```
jfn4th: A derived speed shows the parent's type in its breakdown. This would cause a climb speed tooltip to label it a Land speed.

```suggestion
            const typeLabel = localize(`Type.${this.type.capitalize()}`);
```

Travel speed derives through here too, and it looks like we're missing a `Type.Travel` localization key.

src/module/system/statistic/speed.ts
```diff
@@ -40,32 +54,77 @@ class SpeedStatistic<TActor extends ActorPF2e, TType extends MovementType | "tra
...
+
     /** The "total modifier" of this speed, even though it isn't a check or DC statistic */
     get value(): number {
-        this.#value ??= Math.max(
-            0,
-            this.base + new StatisticModifier("", this.modifiers, this.rollOptions).totalModifier,
-        );
+        if (this.#value === null) {
+            const total = this.base + new StatisticModifier("", this.modifiers, this.rollOptions).totalModifier;
+            this.#value = this.base > 0 ? Math.max(5, total) : Math.max(0, total);
```
jfn4th: RAW speeds are floored at 5 ft unless otherwise stated, but after some searching

src/module/actor/creature/document.ts
```diff
@@ -722,52 +723,156 @@ abstract class CreaturePF2e<
...
+            this.system.movement.speeds[type] = statistic.getTraceData() as CreatureMovementData["speeds"][K];
+        };
+
+        const pending = new Set<MovementType>(MOVEMENT_TYPES);
+        // prepare movement types in order of dependencies
+        for (const _Round of R.range(0, MOVEMENT_TYPES.length)) {
+            const ready = MOVEMENT_TYPES.filter(
+                (type) =>
+                    pending.has(type) &&
+                    (synthetics[type] ?? []).every((entry) => entry.dependsOn.every((dep) => !pending.has(dep))),
```
jfn4th: `dependsOn` is set before the predicate is tested, so a `BaseSpeed` that never applies still counts as a dependency. Two pointing at each other gives up on the ordering and falls through to the loop below, which can build a speed before its parent exists. For example, climb deriving off swim with a swim only bonus gives a 40ft climb speed, but adding a swim from climb rule whose predicate never passes drops it to 30ft.

Should skip synthetics whose predicate doesn't pass. Adding the test to `MovementTypeSynthetic` would let the ordering check it without resolving the value.

src/module/rules/rule-element/base-speed.ts
```diff
@@ -68,6 +126,7 @@ interface BaseSpeedRuleElement extends RuleElement<BaseSpeedRuleSchema>, ModelPr
 type BaseSpeedRuleSchema = RuleElementSchema & {
     selector: fields.StringField<string, string, true, false, false>;
     value: ResolvableValueField<true, false, true>;
+    force: fields.BooleanField<boolean, boolean, false, false, true>;
 };
 
-export { BaseSpeedRuleElement };
+export { BaseSpeedRuleElement, classifySpeedDeriveKind, getDeriveParentType };
```
jfn4th: `classifySpeedDeriveKind` and `getDeriveParentType` aren't used in here at all and are only concerned with how one speed statistic derives from another rather than anything to do with the RE itself. Better housed in `system/statistic/speed.ts`

src/module/rules/rule-element/base-speed.ts
```diff
@@ -7,6 +7,56 @@ import { RuleElement, RuleElementOptions } from "./base.ts";
...
+): SpeedDeriveKind {
+    if (typeof formula !== "string" || parentValue === undefined || !SPEED_VALUE_PATTERN.test(formula)) {
+        return "independent";
+    }
+    if (resolved === parentValue) return "equal";
+    if (/\b(?:min|max)\s*\(/i.test(formula)) {
+        const literals = [...formula.matchAll(/(?<![.\w])(\d+)(?![.\w])/g)].map((m) => Number(m[1]));
+        if (literals.includes(resolved)) return "independent";
+    }
+    return "scaled";
```
jfn4th: For a formula like `max(20, @actor.system.movement.speeds.land.value)`, this compares the resolved value against the literal after land has already taken its penalties, so how penalized land is ends up deciding whether the speed is treated as `independent` or `equal`. For example, under this a character with Spell Effect: Fly and Effect: Web on a 25ft land speed gets a 10ft fly speed, but the same two effects on a 30ft land speed gives 20ft fly. Deleting this makes it consistent and puts it back to the current behavior.

Longer term I think the actual fix is not having to work this out after the fact at all. The relationship gets inferred from the resolved number plus a regex over the formula, which is why the penalty ends up deciding the branch in the first place. If the RE declared what it derives from and carried the floor or cap as its own fields, the floor could apply after modifiers and nothing would need guessing. That would need migration for the min/max items though, so something to be addressed separately.


```suggestion
    if (resolved === parentValue) return "equal";
    return "scaled";
```

src/module/rules/rule-element/base-speed.ts
```diff
@@ -40,23 +91,30 @@ class BaseSpeedRuleElement extends RuleElement<BaseSpeedRuleSchema> {
...
 
     #createMovementType(type: MovementType): DeferredMovementType {
+        const dependsOn = getSpeedFormulaDependsOn(this.value, type);
         return (options: { test?: string[] | Set<string> } = {}): BaseSpeedSynthetic | null => {
             if (!this.test(options.test ?? [])) return null;
             const value = Math.trunc(Number(this.resolveValue(this.value)));
             if (!(value > 0)) {
                 if (!Number.isInteger(value)) this.failValidation("Failed to resolve value");
                 return null;
             }
```
jfn4th: We should allow a speed of 0 if `force` is set

src/module/rules/rule-element/base-speed.ts
iDantar: Force was mainly meant for implementing things like Werecreature Dedication, where overriding base speed and allowing modifiers to apply later looks fitting. Order of operations in Feet to Fins is a rare outlier, though I'd be happy to find a way to properly handle both, if possible - for now it could be hacked with a couple of REs like this: ``{"key":"ActiveEffectLike","phase":"afterDerived","path":"system.movement.speeds.land.value","mode":"override","value":5}``, overriding land value, base land and travel speed (but that doesn't update the land's tooltip)

src/module/actor/creature/document.ts
iDantar: Added test - great catch, thanks!

src/module/system/statistic/speed.ts
iDantar: I thought that would make where the speed is derived from more clear, but you are right.
Also added a label for Travel speed, but I assume it was missing because Travel speed is only shown on Party Sheet, and there is no tooltip for speed breakdown there (a potential separate PR, maybe?)

src/module/rules/rule-element/base-speed.ts
iDantar: Good call, moved

src/module/rules/rule-element/base-speed.ts
iDantar: Right, that would allow us to handle Battleforms better - thanks!

src/module/system/statistic/speed.ts
iDantar: I believe that they didn't apply twice before, because of same-type bonus rule, but from a quick test, they did show twice in the tooltip - I'll add that to description, thanks

src/module/rules/rule-element/base-speed.ts
iDantar: I agree that there is a weakness in current design, but if possible, I'd like to keep it until the solution you proposed is ready (and I'd be happy to implement it), as it fixes more than it breaks. Without it, speeds falsely marked as `scaled` would skip applying armor penalties altogether, for example

iDantar (commented):
Huge thanks for such a thorough review!

src/module/rules/rule-element/base-speed.ts
jfn4th: I think it's fine as is, no need for another pass just to support this one edge case. This was moreso me thinking aloud.

src/module/rules/rule-element/base-speed.ts
jfn4th: I don't think penalties get skipped for scaled speeds, at least generally speaking. On a 30ft land speed, a half speed climb is 15, and 10 with Web on. An equal climb goes from 30 to 20. That's the same on release and this branch, since the penalty comes through `land.value` either way.

The only case where the penalty would have no effect is flooring, where max picks the literal and throws away the penalized number. Again, that's already the current behavior, so we're not losing anything there.

Setting aside which answer is right, the penalty doesn't apply consistently if we keep this min|max. On a 25ft land speed, Web takes land to 15, so the spell's max(20, ...) floors at 20 and the penalty comes off that literal, giving 10. On a 30ft land speed, Web takes land to 20, which matches the floor exactly, so equal takes over and the penalty comes off land's base of 30 instead, giving 20. Whether floored speeds should take the penalty or not, it ought to be one or the other. Right now the system says no, they don't, and I'm not about to change that without there first being discussion (maybe this has already been decided long ago, I don't know!). So, such a change is outside the scope of this PR.

If we do decide penalties should apply there, I think the floor would need to apply to the base with the penalty coming after, which is part of my idea above.

src/module/system/statistic/speed.ts
jfn4th: Yeah separate PR for that. Label is worth having regardless.

src/module/rules/rule-element/base-speed.ts
iDantar: Yep, I was talking about your example from second paragraph - I was sure that's a big mistake in current calculation.
I don't believe there was a discussion, so it does make sense to have team's consensus before changing the behavior.
Removed that if block, thanks

jfn4th (approved):

PR #23090, opened 2026-08-29
Automate Smithing Weapons, Seek the Hidden Glyphs, Rune Ward, Fortifying Knock

iDantar:


packs/pf2e/feat-effects/effect-rune-ward.json
```diff
@@ -0,0 +1,48 @@
...
+            "value": 1
+        },
+        "publication": {
+            "license": "ORC",
+            "remaster": true,
+            "title": "Pathfinder Impossible Magic"
+        },
+        "rules": [
+            {
+                "key": "FlatModifier",
```
simonward:
```suggestion
                "key": "FlatModifier",
                "predicate": [
                    "item:type:spell"
  ],
```

packs/pf2e/feat-effects/effect-rune-ward.json
simonward, line 1: For defensive reactions we often add the automation directly to the feat via a toggleable roll option.
For example, see "Hit the Dirt!".

packs/pf2e/feat-effects/effect-rune-ward.json
iDantar: Included, thanks

packs/pf2e/feat-effects/effect-rune-ward.json
iDantar: Changed to that, thanks. Though personally, losing `removeAfterRoll` hurts - people tend to forget those checkboxes on. Maybe adding similar logic to toggles could be a useful feature?

packs/pf2e/feats/class/runesmith/level-1/rune-ward.json
simonward, line 1: Might be possible to sanity check the skill as well?
For example:
```
"item:trait:arcane"
{
	"gte": [
		"skill:arcana:rank",
		1
	]
}
```

iDantar: Yes, great catch!

packs/pf2e/feats/class/runesmith/level-1/smithing-weapons.json
```diff
@@ -25,7 +25,31 @@
             "remaster": true,
             "title": "Pathfinder Impossible Magic"
         },
-        "rules": [],
+        "rules": [
+            {
+                "domain": "damage",
+                "key": "RollOption",
+                "label": "PF2E.SpecificRule.Runesmith.SmithingWeapons.TargetBearsRune",
```
simonward: Not sure we need the custom label?

iDantar: I thought localizing each label was a common practice, but maybe I'm misreading this: should I remove the localization, or just make the label more generic, like 'Target bears your rune'?

simonward: We generally prefer to use the implicit label where possible to minimize the addition of localized strings. If it's ambiguous or if it's unclear, then a custom label makes sense.
If you prefer a custom label, then making it reusable in other contexts is often a good idea.

iDantar: Gotcha, thanks for clarifying!
I do think there are more runesmith features that check for target being a runebearer, so I'll update it to be more general

PR #23092, opened 2026-08-29
Automate Tracing Trance, Song of Glorious Invocation, Rune Singer and it's follow-ups

iDantar:
Unfortunately we can't remove the frequency from items, so Generational Rune Singer only replaces the note


packs/pf2e/actions/class/runesmith/trace-rune-song.json
```diff
@@ -0,0 +1,36 @@
+{
+    "_id": "p5QjF9uR0iK4oL7x",
+    "folder": "s8ZpeVG9b0MCbope",
+    "img": "systems/pf2e/icons/actions/OneAction.webp",
+    "name": "Trace Rune (Song)",
```
TikaelSol: This is not an action in the book. We don't add modified versions of actions like this.

packs/pf2e/feat-effects/effect-tracing-trance.json
```diff
@@ -0,0 +1,43 @@
+{
+    "_id": "dR7tU1vX5yZ9bC3q",
+    "img": "icons/magic/symbols/runes-triangle-blue.webp",
+    "name": "Effect: Tracing Trance",
+    "system": {
+        "description": {
+            "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Tracing Trance]</p>\n<p>You are @UUID[Compendium.pf2e.conditionitems.Item.Quickened]. You can use the extra action only to @UUID[Compendium.pf2e.actionspf2e.Item.Trace Rune]{Trace a Rune} (including to supply 1 action if using the 2-action version of Trace Rune). You can't use invocation actions this turn.</p>"
```
TikaelSol:
```suggestion
            "value": "<p>Granted by @UUID[Compendium.pf2e.feats-srd.Item.Tracing Trance]</p>\n<p>You are @UUID[Compendium.pf2e.conditionitems.Item.Quickened]. You can't use invocation actions this turn.</p>"
```
We only add what the effect actually does, and there's no way to limit the actions spent. For the invocations use an ItemAlteration to add to the description of invocation actions.

packs/pf2e/actions/class/runesmith/trace-rune-song.json
iDantar: Allright, but what would be a good way to keep track of activation frequency then? Adding frequency to the main action would probably be wrong, but nothing else comes to mind

TikaelSol: We don't do anything for it for now. At most we would just do a description addition

iDantar: Added description updates, thanks

packs/pf2e/feat-effects/effect-tracing-trance.json
iDantar: Added rules for Invocations, thanks

PR #23100, opened 2026-08-31
Add Icons to select consumables

ammalagonc:


packs/pf2e/bestiary-effects/effect-asp-coil.json
```diff
@@ -0,0 +1,55 @@
+{
+    "_id": "7SFfnWTAUIjnbMfR",
+    "img": "icons/creatures/reptiles/snake-poised-white.webp",
+    "name": "Effect: Asp Coil",
```
TikaelSol: Rename to Asp Stance (Irabel)

PR #23109, opened 2026-09-01
Add Combination Weapon usage toggle

### LINKED ISSUE #21870: Feature Request - Combination weapon configurement button
Was hoping that since you all just redesigned the ranged combat in the system if there would be a way to add a button to weapons with the combination tag so that players and GM's can keep track of which grip the character is in on their character sheet.

iDantar:
- Closes https://github.com/foundryvtt/pf2e/issues/21870

Adds toggle similar to Modular, hides attacks that are unavailable in selected mode (not 100% sure about last decision, would appreciate other's input on it)

2us64EwL3O

### DIFF
--- a/src/module/actor/character/auxiliary.ts
+++ b/src/module/actor/character/auxiliary.ts
@@ -14,7 +14,7 @@ import { ChoiceSetSource } from "@module/rules/rule-element/choice-set/data.ts";
 interface AuxiliaryInteractParams {
     weapon: WeaponPF2e<CharacterPF2e>;
     action: "interact";
-    annotation: "draw" | "grip" | "modular" | "pick-up" | "retrieve" | "sheathe" | "boost";
+    annotation: "combination" | "draw" | "grip" | "modular" | "pick-up" | "retrieve" | "sheathe" | "boost";
     hands?: ZeroToTwo;
 }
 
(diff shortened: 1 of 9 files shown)


stwlam:
We would have added a toggle long ago were the problem as simple as making it work like modular. How are you handling attacking after a success melee strike or the critical fusion trait?

iDantar:
That's why I wasn't sure about hiding the attacks for inactive mod - currently you'll have to switch to ranged mode, make an attack and switch back, which isn't ideal
Still, main request was to have an option to keep track of which mode is in use - even if we revert the part that hides inactive mod's buttons, I believe this would be an improvement

stwlam:
I'm thinking if we bother to add a toggle, it should do something. If we take the revisions as they currently are in this PR and remove the restriction against attacking with the inactive usage, we're back to square one except with a toggle that does nothing.

iDantar:
I agree that it would be cool for the toggle to have some mechanical impact, in addition to just being a reminder.
How about this: we would only hide ranged in melee if weapon is not loaded, and keep hiding melee in range mode?

stwlam:
I'm not saying it would be cool but that if we do anything at all, we should do that. What little was originally done made it possible to use combination weapons in any capacity; what follows needs to be a substantial step towards feature completion.

iDantar:
I hear you. What do you think about the idea I mentioned earlier, then?
Hiding melee while in range mode doesn't seem to have drawbacks. Hiding the unloaded ranged attack while in melee can be a bit confusing at first, though.
Since it sounds like you've thought about this issue before, I'd be interested to hear what approaches you've considered.

PR #23114, opened 2026-09-02
Fix adding and removing entries in `RecordField` updates

jfn4th:


src/module/system/schema-data-fields.ts
```diff
@@ -591,6 +591,8 @@ class RecordField<
         for (const [key, value] of fu.iterateEntries(values)) {
             const keyFailure = this.keyField.validate(key, options);
             if (keyFailure) failures.elements.push({ id: key, failure: keyFailure });
+            // A deletion has no value to validate
+            if (value instanceof foundry.data.operators.ForcedDeletion) continue;
```
stwlam: Also anticipate the possibility of forced replacement
```suggestion
            if (value instanceof foundry.data.operators.ForcedDeletion) continue;
            value = foundry.data.operators.ForcedReplacement.get(value);
```

PR #23120, opened 2026-09-02
Force base speeds for Werecreature Dedication

iDantar:
Apply new Force flag to make sure speeds from Werecreature dedication override base speeds
Also fix land speed for weresharks and remove unused effect


packs/pf2e/feat-effects/effect-change-shape-werecreature.json
```diff
@@ -1,372 +0,0 @@
-{
-    "_id": "Q8pD29NcGGuRxWwh",
-    "img": "icons/creatures/mammals/wolf-howl-moon-forest-blue.webp",
-    "name": "Effect: Change Shape (Werecreature)",
```
TikaelSol, line 4: Unless we are redirecting or migrating we should leave the effect in the system since we'd be making broken links otherwise. We have a tracking issue for condensing effects (which, honestly we can probably do a sweep though) #4144

iDantar: Got the effect back.
I could do a sweep - cleaning up feels good

Issue #23091, opened 2026-08-29
Feats with `archetype`, `general` and `skill` traits assigned during data preparation are not recognized as having those by compendium browser filters


Cuingamehtar:
For example, Seasoned Command feat.
It has "auditory", "concentrate", "general", "linguistic", "mental" and "skill" traits, but its data stores only "auditory", "concentrate", "linguistic" and "mental"

The rest are assigned in `FeatPF2e.prepareBaseData` and show on the feat sheet. 

However, as Compendium Browser works with object's data, it's not aware that this feat is supposed to have those traits.

This also applies to the "archetype" trait that's added in the same method too, but the only feat that has this issue regarding that trait is Sanguimancer dedication.

PR #23123, opened 2026-09-02
Add missing "archetype", "general" and "skill" traits to feat data and move their automatic assignment from data preparation to preUpdate

Cuingamehtar:
Attempts to handle https://github.com/foundryvtt/pf2e/issues/23091

As noted in the issue, traits that are automatically added in `prepareBaseData` due to the feat being of either skill or general category don't get assigned to the source data, and as such are invisible to the compendium browser filter.

It seemed fairly logical to move the check to `_preUpdate` so that this behavior would be set in the data. However, as it is, this behavior causes issues with current behavior of traits tagify field on the item sheet. As these traits are now in data, they are no longer considered readonly by `createTagifyTraits` and can be removed from the list. This will result in a visual disappearance of the trait until some other change happens that affects the trait list (such as a change of the feat category).

Currently I see several approaches on how to handle the issue so that it doesn't happen in the future
1. Remove all the automatically assigned traits, and require them all to be held in the source data. This will remove any surprises, but might result in longer time needed for the addition of data from new publications.
2. Duplicate the logic to the compendium browser (as it doesn't work with feat objects, just with data), and let it assign appropriate filter options based on other filter options
3. Have some data on the feat object that stores information about which traits are actually readonly, instead of `createTagifyTags` just checking which of them are or aren't in the source data (but this will probably need to be handled in multiple places, ex. ItemAlteration).

### DIFF
--- a/packs/pf2e/feats/archetype/artillerist/level-4/named-artillery.json
+++ b/packs/pf2e/feats/archetype/artillerist/level-4/named-artillery.json
@@ -36,7 +36,8 @@
         "traits": {
             "rarity": "common",
             "value": [
-                "archetype"
+                "archetype",
+                "skill"
             ]
         }
     },
--- a/packs/pf2e/feats/archetype/sanguimancer/sanguimancer-dedication.json
+++ b/packs/pf2e/feats/archetype/sanguimancer/sanguimancer-dedication.json
@@ -29,6 +29,7 @@
         "traits": {
             "rarity": "rare",
             "value": [
+                "archetype",
                 "dedication"
             ]
         }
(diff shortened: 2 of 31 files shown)


jfn4th:
Off the top of my head option 2 is probably best. I think the browser already does the same/similar for ancestry feats? You can probably move the trait logic in `prepareBaseData` to a helper and use it in both that and the browser feat tab.

Cuingamehtar:
Technically, with the adjustments to the feats handled in the compendium data as well, correcting item sources is no longer necessary.

jfn4th:
Looks good, but separate the data updates into their own pr for the data leads.

stwlam:
We refrain from actually adding these to the source because changing the category would yield a theoretically invalid result.

Cuingamehtar:
Understood. Thankfully, there was no need to change the feat category in data at the time

Issue #21654, opened 2026-02-28
Polymorph Spells like Dragon Form, Animal Form, Dinosaur Form, etc Don't Conform to new Errata re: Speeds


GorgoPrimus:
`Page 301: The polymorph trait has been updated to clarify how Speeds work. In the second paragraph, before the final sentence, add: “You lose your Speeds and gain those of the battle form.”`

I know the Polymorph description in System has been updated to the errata, but none of the spells sets non-specified form speeds to 0 to reflect this mechanically yet.

PR #23137, opened 2026-09-03
Make BattleForm remove unmentioned speeds (Errata)

iDantar:
- Based on https://github.com/foundryvtt/pf2e/issues/21654

The plan is to apply explicit 0 to all speeds not listed on battleforms, now that the code for it is in place. If this PR passes, next step is to update each instance of BattleForm.
Made dragons a separate PR because their RE is the most complicated and to make sure there's no objections to the approach
Also fixed hightened version's speed bonus


stwlam:
Given that a speed of zero theoretically can't exist, we should instead be removing speeds.

iDantar:
Essentially, this is what PR does. Here's a test character with all speeds, and below are his speeds after application of Adamantine dragon form
image

stwlam:
> Essentially, this is what PR does.

Let's go a step further than essentially and make it `null`.

packs/pf2e/spell-effects/spell-effect-dragon-form.json
```diff
@@ -28,399 +28,519 @@
                         "label": "PF2E.Dragon.Adamantine",
                         "value": {
                             "breathShape": "cone",
+                            "burrowSpeed": 20,
+                            "climbSpeed": 0,
```
stwlam: Place these in a single object instead of repeating "xSpeed".

packs/sf2e/spell-effects/spell-effect-dragon-form.json
```diff
@@ -180,8 +192,11 @@
                         }
                     },
                     "speeds": {
+                        "burrow": "{item|flags.system.rulesSelections.dragonForm.burrowSpeed}",
+                        "climb": "{item|flags.system.rulesSelections.dragonForm.climbSpeed}",
```
stwlam: These should be numeric, using `@` expressions

iDantar: From what I can see, `@` doesn't handle `null` values - would it be ok to use `@` expressions and leave speeds at 0 in data, but treat those cases as `null` when resolving speeds in BattleForms?
Or we can use explicit `null` in data, but without `@` expressions - whichever works best

iDantar: I did 0's in data and used `@` expressions for now

iDantar:
Thanks for the review - now we'll remove the non-land speeds if they are set to 0 on the battleform. For land, I'd like to keep this PR's changes narrow if possible, so I did a local workaround instead of making land nullable

packs/pf2e/spell-effects/spell-effect-dragon-form.json
```diff
@@ -31,6 +31,11 @@
                             "damageType": "bludgeoning",
                             "dragonType": "adamantine",
                             "saveType": "reflex",
+                            "speeds": {
+                                "burrow": 20,
+                                "climb": 0,
+                                "swim": 0
+                            },
```
stwlam, line 38: Nix any zero values. A movement type unmentioned in a given battle form is one the character doesn't have while under it.

iDantar: Updated the BattleForm to handle that and dropped zeroes - as it is now, changes would also affect all other battleforms without the need to update data

src/module/rules/rule-element/battle-form/rule-element.ts
```diff
@@ -144,6 +144,11 @@ class BattleFormRuleElement extends RuleElement<BattleFormRuleSchema> {
         this.modifierLabel = this.getReducedLabel();
         const bracket = this.brackets.findLast((b) => b.start <= (this.item.system.level?.value ?? 0));
         if (bracket) this.overrides = fu.mergeObject(this.overrides, bracket.value);
+        for (const selection of Object.values(this.item.flags[SYSTEM_ID].rulesSelections)) {
+            if (!R.isPlainObject(selection) || !R.isPlainObject(selection.speeds)) continue;
+            const extra = R.pickBy(selection.speeds, (value) => typeof value === "number" && value > 0);
+            if (!R.isEmpty(extra)) fu.mergeObject(this.overrides.speeds, extra);
+        }
```
stwlam: RE code doesn't know what RE-generated flags there are.

packs/pf2e/spell-effects/spell-effect-dragon-form.json
stwlam: I'd make sure everything has at least one speed defined for it.

iDantar: There's a single BattleForm that doesn't list any speeds - [Spirit Guide](https://2e.aonprd.com/Feats.aspx?ID=8131)
For that reason I added an exception - if battle form doesn't list speeds, we don't modify speeds at all

src/module/rules/rule-element/battle-form/rule-element.ts
iDantar: This was a workaround for dragon form's speeds extracted from choice set - if we just remove the speeds, `@`-s fail to resolve for unmentioned speeds and the whole rule is ignored. Would it be ok to get the zeroes back into choice sets for Dragon Form if this is a no-go?

iDantar: I saw the PR, thanks!
Updated the data and removed the workaround

packs/pf2e/spell-effects/spell-effect-dragon-form.json
stwlam: That's not a battle form.

iDantar: Well, it uses BattleForm RE, at least. But all other uses have a speed in it

iDantar:
Put the missing speeds as nulls, but had to switch away from `@`s

stwlam:
> Put the missing speeds as nulls, but had to switch away from `@`s

ಠ_ಠ

iDantar:
I was sure you didn't support them in your PR, apologies 😅

iDantar:
Fixed and tested on a fresh dev

PR #23147, opened 2026-09-05
Only count tokens inside an effect area's highlighted squares

### LINKED ISSUE #23044: System template regions incorrectly contain tokens outside of the highlighted cells
Image

While the system brought template regions back to how they are supposed to look in pf2e, the region itself still behave as if it wasn't changed. `RegionDocumentPF2e#tokens` still contains the list of tokens based on core foundry logic and we end up with tokens that are outside the highlighted cells still being counted as inside (as shown in the screenshot above).

https://github.com/user-attachments/assets/d670ad73-deab-43ef-b213-7632660e182e

This also means that any region behavior will incorrectly trigger as shown in the video here.

jfn4th:
- Closes #23044

Half of this is just pulling the coverage math from #22455 out of `RegionPF2e` and into `region-document/coverage.ts` so both the placeable and `testInsideRegion` can use it.

Coverage is cached per level since walls are, and `WallDocumentPF2e` exists to clear that cache. Walls don't affect region membership in core so nothing rechecks it when one changes, but they do affect ours through line of effect.

Separately, I'm finding core splits movement at the region boundary when there's an enter/exit behavior, which leaves the movement readout showing odd fractions like `50 ft (+8.9)`. It splits on the region's polygon rather than our covered squares, so those splits can happen on squares we don't count. The fractions issue predates my v14 region work though (probably was the case in v13 but I haven't checked) and overriding `segmentizeRegionMovementPath` isn't trivial, so not touching that here.

### DIFF
--- a/src/module/scene/region-document/coverage.ts
+++ b/src/module/scene/region-document/coverage.ts
@@ -1,0 +1,108 @@
+import type { Point } from "@common/_types.d.mts";
+import type { CircleShapeData, ConeShapeData, LineShapeData } from "@common/data/data.d.mts";
+import { measureDistance } from "@module/canvas/helpers.ts";
+
+/** Per-shape geometry for a template's grid coverage. */
+interface AreaCoverage {
+    /** Point line of effect is measured from */
+    origin: Point;
+    /** Center of the cell-search box, in pixels */
+    searchCenter: Point;
+    /** Half-extent of the search box, in grid squares */
+    reach: number;
+    /** Whether a cell center lies within the area, before line of effect is checked */
+    contains: (destination: Point) => boolean;
+}
+
+/** Resolve a template shape into its line-of-effect origin, a cell-search box, and a coverage test. */
+function areaCoverage(shape: CircleShapeData | ConeShapeData | LineShapeData): AreaCoverage {
+    const { size, distance } = canvas.dimensions;
+
+    if (shape.type === "circle") {
+        const circle = shape as CircleShapeData;
+        const origin = { x: circle.x, y: circle.y };
+        const radius = (circle.radius / size) * distance;
+        return {
+            origin,
+            searchCenter: origin,
+            reach: circle.radius / size,
+            contains: (destination) => measureDistance(destination, origin) <= radius,
+        };
+    }
+
+    if (shape.type === "cone") {
+        const cone = shape as ConeShapeData;
+        const radius = (cone.radius / size) * distance;
+        const { origin, contains } = coneCoverage(cone, { x: cone.x, y: cone.y });
+        return {
+            origin,
+            searchCenter: origin,
+            reach: cone.radius / size,
+            contains: (destination) => contains(destination) && measureDistance(destination, origin) <= radius,
+        };
+    }
+
+    // A line is a rectangle `length` long and `width` wide running from the origin along its rotation
+    const line = shape as LineShapeData;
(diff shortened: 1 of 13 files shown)


src/module/scene/wall-document/document.ts
```diff
@@ -0,0 +1,49 @@
...
+    protected override _onDelete(options: DatabaseDeleteCallbackOptions, userId: string): void {
+        super._onDelete(options, userId);
+        this.onEdgeChange();
+    }
+
+    /**
+     * Line of effect is part of an effect area's coverage, so a wall can move tokens in or out of one without the
+     * region changing. Core only rechecks containment when a region or token changes.
+     */
+    private onEdgeChange(): void {
```
stwlam: Same

src/module/scene/region-document/coverage.ts
```diff
@@ -0,0 +1,173 @@
...
+    // Bursts and cylinders are circles; only circles, cones, and lines are template shapes we count ourselves
+    return document.shapes.length === 1 && !!shape && ["circle", "cone", "line"].includes(shape.type);
+}
+
+/**
+ * The grid spaces a burst, cone, or line effect area reaches, counted by PF2e's grid rules rather than by intersection
+ * with the region's polygon. Walls are per-level, so line of effect is measured against `level`. Returns null for
+ * regions core should measure itself.
+ */
+function effectAreaOffsets(document: RegionDocumentPF2e, level: Level | null): EffectAreaOffsets | null {
```
stwlam: Any reason not to make these methods of `RegionDocumentPF2e`?

src/module/scene/token-document/document.ts
```diff
@@ -291,6 +292,38 @@ class TokenDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ext
...
+        const depth = data.depth ?? this._source.depth;
+        if (topInclusive ? elevation > top : elevation !== bottom && elevation >= top) return false;
+        const head = elevation + depth * (this.parent?.grid.distance ?? 0);
+        if (depth === 0 ? head < bottom : head <= bottom) return false;
+
+        return this.occupiedOffsets(data).some((o) => coverage.covers(o));
+    }
+
+    /** The grid spaces this token sits on, taken from square centers so unsnapped tokens count the right ones */
+    private occupiedOffsets(data: Partial<TokenPosition>): GridOffset2D[] {
```
stwlam: It should be safe to make this hard private so long as it's not hit during data preparation.

src/module/scene/region-document/document.ts
```diff
@@ -26,6 +40,91 @@ class RegionDocumentPF2e<TParent extends ScenePF2e | null = ScenePF2e | null> ex
...
+        // Bursts and cylinders are circles; only circles, cones, and lines are template shapes we count ourselves
+        return this.shapes.length === 1 && !!shape && ["circle", "cone", "line"].includes(shape.type);
+    }
+
+    /**
+     * The grid spaces this area covers by PF2e's rules, which differ from the region's polygon in that spaces are
+     * counted whole and cut off by walls. Null if core should measure the region itself.
+     * @param level The level whose walls block line of effect, defaulting to the viewed one
+     */
+    coverage(level: Level | null = canvas.level): EffectAreaOffsets | null {
```
stwlam: Sorry, sorry, wee style nit: we usually have methods in verb form (`getCoverage` or the like).

PR #23163, opened 2026-09-06
Allow `RuleElement#resolveValue` to return `null` from string inputs

stwlam:


src/module/rules/rule-element/base.ts
```diff
@@ -311,33 +311,31 @@ abstract class RuleElement<TSchema extends RuleElementSchema = RuleElementSchema
                             warn &&
                             !unresolveds.every((u) => u.startsWith("@target.") || u.startsWith("@actor.conditions."));
                         this.ignored = true;
```
iDantar, line 319: Would it be possible to also not set .ignored to true?

jfn4th:
I'm finding this breaks a number of things, including

- CreatureSize size strings fail `Roll.safeEval` so they default to medium
- AELike/ItemAlterations throw errors on string values
- FastHealing roll formulas like `"1d6"` are a SyntaxError in `new Function`

src/module/rules/rule-element/base.ts
stwlam: It wouldn't

jfn4th:
Those all seem fixed now. But with the #23137 data (absent speed keys) the @ still comes back unresolved, so the BattleForm gets ignored and Dragon Form stops granting its temp HP. The "null" check only ever fires for a literal "null" string in the flag.

stwlam:
> Those all seem fixed now. But with the #23137 data (absent speed keys) the @ still comes back unresolved, so the BattleForm gets ignored and Dragon Form stops granting its temp HP. The "null" check only ever fires for a literal "null" string in the flag.

The idea is those entries will get put back but with `null` instead of zero.

jfn4th:
> The idea is those entries will get put back but with null instead of zero.

That's what I'm getting at, `"null"` works but not literal `null` like you seem to intend. Core's `replaceFormulaData` treats `null` the same as `undefined` so it never resolves and the RE ends up ignored.

Screenshot 2026-09-07 at 11 02 25 AM


If you want literal `null` I think you need to either resolve the @ terms yourself before `replaceFormulaData` so you can distinguish `null`, or pass `{ missing: "null" }` and accept the loss of typo warnings.

iDantar:
Sorry for jumping in, but I pulled these changes locally and they seem to work fine with syntax like `{item|flags.system.rulesSelections.dragonForm.speeds.burrow}`

PR #23171, opened 2026-09-07
Fix TokenImage rule overrides on world load

### LINKED ISSUE #23169: Effect with TokenImage rule no longer gets applied when re-loading Game World
Since upgrading to the Foundry14 version of pf2e, monsters with an Effect with a TokenImage rule have reverted to their original token. New applications of Effects with TokenImage rules also reset upon reloading the Game World. 

**Replication:** I tested this in a new empty pf2e Game World with no modules. I imported a Goblin Dog from the compendium, assigned an image, and enabled the token ring. I dragged the Goblin Dog into the map. I created a new Effect, added an TokenImage rule, and populated its ImagePath. I applied the new Effect to the Goblin Dog on the map, the token transformed as expected. I reloaded the Game World, and the Goblin Dog had reverted to its original image.

jfn4th:
- Closes #23169

`game.ready` is false when `canvasInit` first fires, so synthetic actors never got constructed before textures loaded.

Also stops `prepareBaseData` writing the rule element's ring subject texture into token source data, which stuck around after the effect was removed.

### DIFF
--- a/src/scripts/hooks/canvas-init.ts
+++ b/src/scripts/hooks/canvas-init.ts
@@ -1,8 +1,7 @@
 /** Ensure all synthetic actors are constructed in a to-be-viewed scene. */
 export const CanvasInit = {
     listen: (): void => {
-        Hooks.once("canvasInit", () => {
-            if (!game.ready) return;
+        Hooks.on("canvasInit", () => {
             for (const token of canvas.scene?.tokens ?? []) {
                 if (!token.hasConstructedActor && token.baseActor) token.reset();
             }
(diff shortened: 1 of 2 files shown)


src/scripts/hooks/canvas-init.ts
```diff
@@ -1,8 +1,7 @@
 /** Ensure all synthetic actors are constructed in a to-be-viewed scene. */
 export const CanvasInit = {
     listen: (): void => {
-        Hooks.once("canvasInit", () => {
-            if (!game.ready) return;
+        Hooks.on("canvasInit", () => {
```
stwlam, line 4: Does this need to fire every time rather than getting moved to later hook?

jfn4th: Yeah it does. canvasInit is the last hook before textures load, next one is canvasReady and that's too late. It has to be every time since tokens in non viewed scenes aren't constructed. With `once` the first scene is fine but switching to a second one still shows the base art.

PR #23176, opened 2026-09-07
Migrate remaining thaumaturge weapon crit spec predicates.

griffirpf2e:
The initiate benefit gated crit spec on `feature:thaumaturge-weapon-expertise`, it seems that slug no longer exists after Weapon Expertise was unified, so the effect never applied. This seemed like a problem on v13 but fixed on v14.

However, it seems we never did the migration which causes existing chars to keep pointing to the old `weapon-expertise`. Hopefully this patch will fix old chars.

### DIFF
--- a/packs/pf2e/iconics/mios/mios-level-5.json
+++ b/packs/pf2e/iconics/mios/mios-level-5.json
@@ -3597,7 +3597,7 @@
                         "key": "CriticalSpecialization",
                         "predicate": [
                             "item:id:{actor|flags.system.thaumaturge.weaponImplement}",
-                            "feature:thaumaturge-weapon-expertise"
+                            "feature:weapon-expertise"
                         ]
                     }
                 ],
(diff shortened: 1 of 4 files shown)


TikaelSol:
I already closed your other PRs, given that I am not AI and have persistence of memory I'm fully aware that you're the same person trying to merge the same code I've already rejected twice. I reported you as a spammer. We will not be merging any of your code, even if it did work, which this won't (let's see if you can figure out why the migration would break older actors!).

griffirpf2e:
> I already closed your other PRs, given that I am not AI and have persistence of memory I'm fully aware that you're the same person trying to merge the same code I've already rejected twice. I reported you as a spammer. We will not be merging any of your code, even if it did work, which this won't (let's see if you can figure out why the migration would break older actors!).

Hey I am sorry for that truly, it was not my intention to spam. I was trying to fix that on my own fork. I apologize for the error.

PR #23179, opened 2026-09-07
Add shared spell tracking for troops

### LINKED ISSUE #22866: Troop tokens do not share a spell list
When placing a troop actor on the map the system automatically generates 4 tokens.  Cool!

However if the troop is a  caster, each token keeps its own spell list that is not shared. This means if a player casts a spell by clicking on one token , then next turn selects a different one the spell is still shown as available.

https://github.com/user-attachments/assets/18173f4c-2dd7-4cd6-bc45-7eb760f58f5f

sgennaoui:
Troops are represented by up to four sibling tokens whose actor data is partially synchronized. The existing synchronization handles:
* HP
* Elite/weak adjustments

This change extends actor synchronization to include focus points and spell slots.

Fixes #22866

### DIFF
--- a/src/module/actor/creature/sheet.ts
+++ b/src/module/actor/creature/sheet.ts
@@ -243,19 +243,27 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
         handlers["reset-spell-slots"] = (event): Promise<unknown> | void => {
             const row = htmlClosest(event.target, "[data-item-id]");
             const itemId = row?.dataset.itemId;
-            const item = actor.items.get(itemId, { strict: true });
+            const actors = actor.isOfType("npc") ? [actor, ...(actor.otherSegments ?? [])] : [actor];
+            const items = actors.flatMap((actor) => actor.items.contents.filter((item) => item.id === itemId));
 
-            if (item.isOfType("spellcastingEntry")) {
-                const { system } = item.toObject();
-                if (!system.slots) return;
+            if (items[0]?.isOfType("spellcastingEntry")) {
                 const groupNumber = spellSlotGroupIdToNumber(row?.dataset.groupId) || 0;
                 const propertyKey = goesToEleven(groupNumber) ? (`slot${groupNumber}` as const) : "slot0";
-                system.slots[propertyKey].value = system.slots[propertyKey].max;
-                return item.update({ system });
-            } else if (item.isOfType("spell")) {
-                const max = item.system.location.uses?.max;
-                if (!max) return;
-                return item.update({ "system.location.uses.value": max });
+                return Promise.all(
+                    items.map((item) => {
+                        if (!item.isOfType("spellcastingEntry") || !item.system.slots) return;
+                        const max = item.system.slots[propertyKey].max;
+                        return item.update({ [`system.slots.${propertyKey}.value`]: max });
+                    }),
+                );
+            } else if (items[0]?.isOfType("spell")) {
+                return Promise.all(
+                    items.map((item) => {
+                        if (!item.isOfType("spell")) return;
+                        const max = item.system.location.uses?.max;
+                        return max ? item.update({ "system.location.uses.value": max }) : undefined;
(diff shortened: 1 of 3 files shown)


src/module/actor/npc/document.ts
```diff
@@ -579,6 +584,48 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
...
+            for (const actor of this.otherSegments) {
+                const siblingEntry = actor.items.get(document.id);
+                if (siblingEntry?.isOfType("spellcastingEntry")) {
+                    actor.updateEmbeddedDocuments(
+                        "Item",
+                        [{ _id: document.id, system: { slots: fu.deepClone(slots) } }],
+                        { fromTroop: true } as Partial<DatabaseUpdateOperation<NPCPF2e>> & {
+                            fromTroop?: boolean;
+                        },
+                    );
```
stwlam: Really not crazy about item updates that trigger item updates. Could this be done as a batch operation?

sgennaoui: I was following the pattern from the actor data sync. I'll see if I can tackle it as a batch operation instead. Thanks!

src/module/actor/npc/document.ts
```diff
@@ -579,6 +584,48 @@ class NPCPF2e<TParent extends TokenDocumentPF2e | null = TokenDocumentPF2e | nul
...
+                }
+            }
+        }
+    }
+
```
sgennaoui:
```suggestion

```

src/module/actor/creature/sheet.ts
```diff
@@ -243,19 +243,30 @@ abstract class CreatureSheetPF2e<TActor extends CreaturePF2e> extends ActorSheet
...
                 const propertyKey = goesToEleven(groupNumber) ? (`slot${groupNumber}` as const) : "slot0";
-                system.slots[propertyKey].value = system.slots[propertyKey].max;
-                return item.update({ system });
-            } else if (item.isOfType("spell")) {
-                const max = item.system.location.uses?.max;
-                if (!max) return;
-                return item.update({ "system.location.uses.value": max });
+                return Promise.all(
+                    items.map(async (item) => {
+                        if (!item.isOfType("spellcastingEntry") || !item.system.slots) return;
```
stwlam, line 257: Would spellcasting entries not always have `slots`?

stwlam (commented):
By batch operation I mean the API endpoint. Try something like this in sheet.ts (untested):
```ts
handlers["reset-spell-slots"] = (event): Promise<unknown> | void => {
    const row = htmlClosest(event.target, "[data-item-id]");
    const itemId = row?.dataset.itemId ?? "";
    const targetItem = actor.items.get(itemId);
    if (!targetItem) return;
    const actors: ActorPF2e[] = actor.isOfType("npc") ? [actor, ...(actor.otherSegments ?? [])] : [actor];

    // Prepare batch update in case of multiple troop actors
    const operations = actors
        .map((actor): DatabaseUpdateOperation<ActorPF2e> | null => {
            const item = actor.items.get(itemId);
            const updates: Record<string, unknown>[] = [];
            if (item?.isOfType("spellcastingEntry") && item.system.slots) {
                const groupNumber = spellSlotGroupIdToNumber(row?.dataset.groupId) || 0;
                const slotKey = goesToEleven(groupNumber) ? (`slot${groupNumber}` as const) : "slot0";
                const max = item.system.slots[slotKey].max;
                updates.push({ _id: item.id, [`system.slots.${slotKey}.value`]: max });
            } else if (item?.isOfType("spell") && item.system.location.uses?.max) {
                updates.push({ _id: item.id, "system.location.uses.value": item.system.location.uses.max });
            } else {
                return null;
            }
            return { action: "update", documentName: "Item", parent: actor, updates };
        })
        .filter(R.isNonNull);
    if (operations.length === 1) return targetItem.update(operations[0].updates[0]);
    else return foundry.documents.modifyBatch(operations);
};
```

PR #23180, opened 2026-09-08
Correct slug to Astradaemon's Essence Drain ability

ammalagonc:


packs/pf2e/pathfinder-monster-core/astradaemon.json
simonward:
```suggestion
                    "value": "<p>When an astradaemon hits with their claw, jaws, or tail, they drain the target's spiritual and vital essences.</p>\n<p>The target takes 2d10 void damage and the astradaemon regains an equal number of Hit Points. The target must succeed at a @Check[fortitude|dc:37|options:inflicts:drained,inflicts:doomed] save or become @UUID[Compendium.pf2e.conditionitems.Item.Doomed]{Doomed 1} and @UUID[Compendium.pf2e.conditionitems.Item.Drained]{Drained 1}. If the target was already drained or doomed, it instead increases both conditions' value by 1, to a maximum of 4.</p>"
```

packs/pf2e/pathfinder-monster-core/astradaemon.json
```diff
@@ -924,7 +924,7 @@
                         "diceNumber": 2,
                         "dieSize": "d10",
                         "key": "DamageDice",
-                        "selector": "damage"
+                        "selector": "melee-damage"
```
simonward: Should we use the three specific selectors instead?

TikaelSol: We should. We should also make sure the strikes have stable slugs set when we do.

PR #23185, opened 2026-09-09
Add Raise a Shield button to NPC sheet when shield is held

### LINKED ISSUE #23001: NPC actors do not inherently get the Raise a Shield action when a shield is equipped
As the header implies, currently in the system, an NPC who has a shield equipped does not gain access to the Raise a Shield action like a PC does.

jfn4th:
- Closes #23001
Screenshot 2026-09-08 at 8 31 58 PM
Screenshot 2026-09-08 at 8 32 05 PM

Screenshot 2026-09-08 at 8 32 12 PM
Screenshot 2026-09-08 at 8 32 17 PM

### DIFF
--- a/src/styles/actor/npc/_sidebar.scss
+++ b/src/styles/actor/npc/_sidebar.scss
@@ -89,6 +89,35 @@
     }
 
     &.shield {
+        .raise-shield {
+            background: none;
+            border: none;
+            box-shadow: none;
+            color: var(--text-dark);
+            flex: 0 0 1.5rem;
+            height: 1.5rem;
+            line-height: 1;
+            opacity: 0.6;
+            outline: none;
+            padding: 0;
+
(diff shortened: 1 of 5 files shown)


src/styles/actor/npc/_sidebar.scss
```diff
@@ -89,6 +89,35 @@
...
+            padding: 0;
+
+            &:hover:not(:disabled),
+            &:focus-visible {
+                opacity: 1;
+                text-shadow: 0 0 8px var(--color-shadow-primary);
+            }
+
+            &.raised {
+                color: var(--primary);
```
stwlam:
```suggestion
                color: var(--color-pf-primary);
```
`--primary`, `--secondary`, etc. are legacy color variables slowly getting purged

PR #23192, opened 2026-09-10
Add label for Force field of BaseSpeed RE

iDantar:
Adds a localized label for Force

Kudos to Tickael for noticing the missing label 
image

### DIFF
--- a/static/lang/re-en.json
+++ b/static/lang/re-en.json
@@ -3751,6 +3751,14 @@
             }
         },
         "RULES": {
+            "BaseSpeed": {
+                "FIELDS": {
+                    "force": {
+                        "hint": "Overrides speeds from other sources",
+                        "label": "Force"
+                    }
+                }
+            },
             "Common": {
                 "FIELDS": {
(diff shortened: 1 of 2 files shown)


static/lang/re-en.json
```diff
@@ -3792,6 +3791,13 @@
                     }
                 }
             },
+            "BaseSpeed": {
```
jfn4th: Should be alphabetical, before `"Common"`. Add a hint as well.

iDantar:
Fixed, thanks!

static/lang/re-en.json
```diff
@@ -3752,6 +3751,14 @@
             }
         },
         "RULES": {
+            "BaseSpeed": {
+                "FIELDS": {
+                    "force": {
+                        "hint": "Overrides speeds from other sources",
```
jfn4th:
```suggestion
                        "hint": "Whether this speed overrides others instead of only applying if higher: allows lower values, including 0.",
```

PR #23207, opened 2026-09-12
Restore missing localization for alignment traits

ammalagonc:
Divine Vessel is still in the system - has not been reprinted/replaced yet.


stwlam:
Well, that ain't happening. Have alternatives in mind? Switch to holy/unholy? Drop that part of the spell? Drop the spell?

PR #23212, opened 2026-09-13
Convert `SpellcastingCreateAndEditDialog` to `ApplicationV2` and Svelte

jfn4th:
spellcasting-entry-dialog-dark
spellcasting-entry-dialog-light
spellcasting-entry-dialog-npc-dark


src/module/item/spellcasting-entry/apps/spellcasting-entry-dialog/app.ts
```diff
@@ -0,0 +1,281 @@
...
+            const statistic = new SpellcastingEntryPF2e(source, { parent: actor }).buildStatistic();
+            return statistic ? { attack: statistic.check.mod, dc: statistic.dc.value } : null;
+        } catch (error) {
+            // A preview shouldn't take the form down with it
+            console.error(error);
+            return null;
+        }
+    }
+
+    async confirm(draft: SpellcastingDraft): Promise<void> {
```
stwlam, line 192: Any reason to prefer this over having a proper form and a submission handler?

jfn4th: It's the same pattern as the other svelte apps, where the component calls into the app. Here it also sidesteps that disabled `<select>`s don't submit, which the old dialog had to work around too. Could switch to a form handler with hidden inputs for those if you'd rather.

PR #23217, opened 2026-09-13
Change recharge roll syntax for Eternal Eruption items to not be blind rolls

ammalagonc:
These should not be blind rolls.


simonward:
Since they’re equipment, consider making the saving throws showDC:all

PR #23224, opened 2026-09-14
Re-render world clock and sync darkness on all scenes upon setting change

stwlam:


jfn4th:
Could you filter to just scenes that are syncing darkness and batch it?

stwlam:
> Could you filter to just scenes that are syncing darkness and batch it?

Probably with the exception of the viewed scene: that's done with multiple updates if time is advanced across an entire 24-hour period for multiple phases of animation.

PR #23249, opened 2026-09-19
Clean up some equipment effects

Cuingamehtar:
Add missing backlinks and icons. Simplify descriptions.

Also, coalesce Energy Shield Tunic effects into one, and remove REs from the base Energy Shield Tunic.


Cuingamehtar:
Forgot to add a redirect. Will do it tomorrow.

PR #23252, opened 2026-09-19
Add temp HP roll options

### LINKED ISSUE #19067: Temp HP Roll Options
We have `hp-percent` and `hp-remaining` roll options but no `temp-hp` related roll options.

Implementation of such would allow for effects that last as long as the Temp HP last (`gt > 0` the temp hp)

jfn4th:
- Closes #19067 

Adds `hp-temp:N` alongside `hp-remaining` and `hp-percent`. Also adds `hp-temp:source:<slug>` when the temp HP came from a TempHP rule element and that item is still on the actor, since the effects that care (Avenge in Glory, Hellbreaker's Resolve, Muscle Barrier) all say "these temporary Hit Points". Manually entering a new temp HP value or zeroing it now clears the recorded source, while damage keeps it. Turn-start renewals record the source too. Also found the familiar schema was stripping `tempsource` entirely and existing cleanup wasn't working there.


stwlam:
Since the `tempsource` property was getting stripped, let's get it in normal style as `tempSource`.

PR #23253, opened 2026-09-19
Add area damage to Clockwork Dragon's Spearing Tail

ammalagonc:


simonward:
I think this is an addition to the tail damage instead?

